PR #24, opened 2020-12-12
Make withOnyx forwardRef

roryabraham:
In order for refs to work with a component wrapped in a HOC, the HOC must forward the ref to the wrapped component.

This PR is required so that https://github.com/Expensify/ReactNativeChat/pull/939 will work.


tgolen:
I think this one can be closed as well, but reopen if that's not correct!

PR #36, opened 2021-01-12
Move mocks to __mocks__. Fix discovered broken merge behavior.

marcaaron:
Came across an interesting behavior while testing Pusher stuff in `Expensify.cash` also took that opportunity to fix up these tests.

As far as I can tell, nothing is breaking from this (except the tests I'm trying to write), but it does seem incorrect and may lead to problems in the future.

**Here's what happens...**

- Do not have a default value for a key in storage 
- Queue up back to back `Onyx.merge()` updates
- We will only apply the _last update_ since the initial value is `null`. 
- The place where I saw this happen is with setting `myPersonalDetails` for the first time - but it's not reproducible in production when dealing with actual API requests that don't resolve immediately like in the testing environment - most likely because each will get merged without anything else waiting to be merged.

Regardless, there still may be times where we want to call `Onyx.merge()` several times in succession without worrying about whether we have a default value. We can do it when we have a value, but not when there is no value. Which doesn't really make sense. Calling `Onyx.set()` before `Onyx.merge()` shouldn't be a requirement to take advantage of our ability to queue and apply updates.


marcaaron:

tgolen:
Really nice catch on finding this! I was a little confused by the PR description without an example but seeing the tests that you added highlight the issue perfectly. There are places where we do use multiple merges in a row like that, so I'm glad this bug hasn't bitten us yet.

pecanoro (approved):
It looks good, but since I am not really sure what this is used for, can someone more knowledgeable take a look?

tgolen (approved):
@pecanoro Yep! I don't mind approving and merging it. Thanks for taking a look at it.

PR #37, opened 2021-01-13
Remove version action and reset version back to 1.0.0

AndrewGable:
Since we are not deploying `react-native-onyx` anywhere and we are just using git hashes to download via `npm`, we do not need to run the version or auto merge actions. I've reset the version in the `package.json` to be `1.0.0` since we have not deployed the code anywhere.

cc @Julesssss @tgolen


tgolen:
> since we have not deployed the code anywhere.

To clarify this a little... we have deployed it, but I think you meant that it hasn't been "published" anywhere. Is that correct? Should we also make sure that we also delete any version tags other than 1.0.0? (or maybe we should just delete them all until we want to publish this somewhere like NPM)

AndrewGable:
> To clarify this a little... we have deployed it, but I think you meant that it hasn't been "published" anywhere. 

Yes, correct. I've also deleted all the tags 👍

marcaaron (approved):
LGTM

PR #42, opened 2021-01-20
Shallowly copy a collection before updating it and setting it to withOnyx state

marcaaron:
Fixes https://github.com/Expensify/Expensify/issues/151770

**To reproduce the bug and test this fix we can:**

1. Navigate to a report
2. Add this diff to a react component subscribing to all reports via Onyx
```js
componentDidUpdate(prevProps) {
    console.log(prevProps.reports);
    console.log(this.props.reports);
}
```
3. Toggle a report from pinned to unpinned or vice versa
4. Verify that the `isPinned` property on the active report is the _different_ for both the `prevProps` and `this.props`


marcaaron:
cc @tgolen found another pretty weird one while testing some other things out

tgolen (approved):

PR #43, opened 2021-01-23
Code for mergeCollection

chiragsalian:
Helps with,
https://github.com/Expensify/Expensify/issues/152029

Rather than notifying the subscriber each time for every item within a collection now we can just notify the subscriber of the updated data once. This results in drastically less re-renders.

### DIFF
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -164,8 +164,8 @@ function keysChanged(collectionKey, collection) {
             && isCollectionKey(subscriber.key)
         ) {
             if (_.isFunction(subscriber.callback)) {
-                _.each(collection, (data) => {
-                    subscriber.callback(data, collectionKey);
+                _.each(collection, (data, dataKey) => {
+                    subscriber.callback(data, dataKey);
                 });
             } else if (subscriber.withOnyxInstance) {
                 subscriber.withOnyxInstance.setState((prevState) => {
@@ -529,10 +529,11 @@ function mergeCollection(collectionKey, collection) {
             ]), []);
 
             // New keys will be added via multiSet while existing keys will be updated using multiMerge
+            // This is because setting a key that doesn't exist yet with multiMerge will throw errors
             const existingCollectionPromise = AsyncStorage.multiMerge(keyValuePairsForExistingCollection);
             const newCollectionPromise = AsyncStorage.multiSet(keyValuePairsForNewCollection);
 
-            Promise.all([existingCollectionPromise, newCollectionPromise])
+            return Promise.all([existingCollectionPromise, newCollectionPromise])
                 .then(() => keysChanged(collectionKey, collection))
                 .catch(error => evictStorageAndRetry(error, mergeCollection, collection));
         });
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -517,7 +517,7 @@ function mergeCollection(collectionKey, collection) {
                 } else {
                     newCollection[key] = data;
                 }
-            })
+            });
 
             const keyValuePairsForExistingCollection = _.reduce(existingKeyCollection, (finalArray, val, key) => ([
                 ...finalArray,
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -505,7 +505,7 @@ function mergeCollection(parentKey, collection) {
     // Confirm all the collection keys belong to the same parent
     _.each(collection, (data, dataKey) => {
         if (!isKeyMatch(parentKey, dataKey)) {
-            throw `Provided collection does not have all its data belonging to the same parent. ParentKey: ${parentKey}, DataKey: ${dataKey}`;
+            throw new Error(`Provided collection does not have all its data belonging to the same parent. ParentKey: ${parentKey}, DataKey: ${dataKey}`);
         }
     });
 
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -151,6 +151,41 @@ function addAllSafeEvictionKeysToRecentlyAccessedList() {
 }
 
 /**
+ * When a collection of keys change, search for any callbacks matching the collection key and trigger those callbacks
+ *
+ * @param {string} parentKey
+ * @param {Object} collection
+ */
+function keysChanged(parentKey, collection) {
+    // Add or remove this parentKey from the recentlyAccessedKeys lists
+    if (!_.isNull(collection)) {
+        addLastAccessedKey(parentKey);
+    } else {
+        removeLastAccessedKey(parentKey);
+    }
+
+    // Find all subscribers that were added with connect() and trigger the callback or setState() with the new data
+    _.each(callbackToStateMapping, (subscriber) => {
+        if (subscriber && isKeyMatch(subscriber.key, parentKey) && isCollectionKey(subscriber.key) && subscriber.withOnyxInstance) {
+            subscriber.withOnyxInstance.setState((prevState) => {
+                const finalCollection = _.clone(prevState[subscriber.statePropertyName] || {});
+                _.each(collection, (data, dataKey) => {
+                    if (finalCollection[dataKey]) {
+                        lodashMerge(finalCollection[dataKey], data);
+                    } else {
+                        finalCollection[dataKey] = data;
+                    }
+                });
+
+                return {
+                    [subscriber.statePropertyName]: finalCollection,
+                };
+            });
+        }
+    });
+}
+
+/**
  * When a key change happens, search for any callbacks matching the key or collection key and trigger those callbacks
  *
  * @param {string} key
@@ -456,6 +491,28 @@ function merge(key, val) {
 }
 
 /**
+ * Merges a collection based on their keys
+ *
+ * @param {string} parentKey
+ * @param {object} collection
+ * @returns {Promise}
+ */
+function multiMerge(parentKey, collection) {
+    // AsyncStorage expenses the data in an array like:
+    // [["@MyApp_user", "value_1"], ["@MyApp_key", "value_2"]]
+    // This method will transform the params from a better JSON format like:
+    // {'@MyApp_user': 'myUserValue', '@MyApp_key': 'myKeyValue'}
+    const keyValuePairs = _.reduce(collection, (finalArray, val, parentKey) => ([
+        ...finalArray,
+        [parentKey, JSON.stringify(val)],
+    ]), []);
+
+    return AsyncStorage.multiMerge(keyValuePairs)
+        .then(() => keysChanged(parentKey, collection))
+        .catch(error => evictStorageAndRetry(error, multiMerge, collection));
+}
+
+/**
  * Initialize the store with actions and listening for storage events
  *
  * @param {Object} [options]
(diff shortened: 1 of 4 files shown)


chiragsalian:

lib/Onyx.js
```diff
@@ -150,6 +150,45 @@ function addAllSafeEvictionKeysToRecentlyAccessedList() {
         });
 }
 
+/**
+ * When a collection of keys change, search for any callbacks matching the collection key and trigger those callbacks
+ *
+ * @param {String} parentKey
+ * @param {Object} collection
+ */
+function keysChanged(parentKey, collection) {
```
marcaaron, line 159: Rather than call this a `parentKey` could we call it a `collectionKey` ? I think it is more in line with the terminology we are using in Onyx so far.

lib/Onyx.js
```diff
@@ -150,6 +150,45 @@ function addAllSafeEvictionKeysToRecentlyAccessedList() {
...
+    // Add or remove this parentKey from the recentlyAccessedKeys lists
+    if (!_.isNull(collection)) {
+        addLastAccessedKey(parentKey);
+    } else {
+        removeLastAccessedKey(parentKey);
+    }
```
marcaaron, line 165: I think we don't need these at all since last accessed should never be a collection key and `addLastAccessedKey()` will just return immediately anyway.

lib/Onyx.js
```diff
@@ -150,6 +150,45 @@ function addAllSafeEvictionKeysToRecentlyAccessedList() {
...
+
+    // Find all subscribers that were added with connect() and trigger the callback or setState() with the new data
```
marcaaron, line 167: >trigger the callback

We are not triggering any callback here only `setState()`. However, we probably need to?

Each time we use `Onyx.connect()` with a collection key like `ONYXKEYS.COLLECTION.REPORTS` that callback will expect to see receive an update for each key updated.

It makes sense that we might not want this to happen for a React component since it would re-render n report count times which is bad for performance.

But for the subscribers not using `withOnyx()` I think we should still send updates to them? Otherwise a call to `mergeCollection()` will fail to update it's non React subscribers.

marcaaron: e.g. the `handleReportChanged()` callback here will now not be updated when we merge a collection

https://github.com/Expensify/Expensify.cash/blob/adfdb884d1ccc726893c73dc1d331631b477dd42/src/libs/actions/Report.js#L645-L663

marcaaron (changes requested):
Looks like a good start and really excited for this improvement. I think we're gonna unlock some nice perf gains from this. But there's a couple of things to update. And also wanted to float the idea of writing some tests for this. I think they would go something like this...

1. Test that merging a collection only updates a `withOnyx` subscriber once
2. Test that merging a collection updates all `Onyx.connect()` subscribers once per record in that collection
3. Test that merging a collection with keys that do not match the collection key will throw

lib/Onyx.js
chiragsalian: We are triggering a setState. We're not doing a callback though.

Ah you're right about the handleReportChanged. I was just testing places that connected to the collection via withOnyx. Yeah i think with the current code it does not send updates to connections via onyx. Yup, i think we should send updates to them as well. I'll update and test it out. I might have to change their format from report object to reports array but I'll check it out.

chiragsalian:
Cool i'll add the tests as well. I haven't tried the tests before so might hit you up on it if i have any questions.

lib/Onyx.js
chiragsalian: Updated.

chiragsalian:
Updated code and tests and retested as well. I couldn't figure out how to track the number of renders and am using the help of jest console spyOn and clearing it at the end of the test as well. Let me know your thoughts on the changes.

lib/Onyx.js
```diff
@@ -150,6 +150,43 @@ function addAllSafeEvictionKeysToRecentlyAccessedList() {
...
+function keysChanged(collectionKey, collection) {
+    // Find all subscribers that were added with connect() and trigger the callback or setState() with the new data
+    _.each(callbackToStateMapping, (subscriber) => {
+        if (subscriber
+            && isKeyMatch(subscriber.key, collectionKey)
+            && isCollectionKey(subscriber.key)
+        ) {
+            if (_.isFunction(subscriber.callback)) {
+                _.each(collection, (data) => {
+                    subscriber.callback(data, collectionKey);
```
marcaaron, line 168: Probably the key in `subscriber.callback()` here should be the individual key not the collection key? That is what happens in `keyChanged()` here and we should consider keeping this behavior consistent.

https://github.com/Expensify/react-native-onyx/blob/4e810bfd30cbb02aaf663ba69456fd233346a774/lib/Onyx.js#L170-L172

Having said that, I don't recall whether anything is actually _using_ the `key` parameter in the callback (but there's basically no other way to tell which record in the collection you are seeing pop up in the callback so it makes sense for it to be there).

lib/Onyx.js
```diff
@@ -455,6 +492,52 @@ function merge(key, val) {
...
+            const keyValuePairsForExistingCollection = _.reduce(existingKeyCollection, (finalArray, val, key) => ([
+                ...finalArray,
+                [key, JSON.stringify(val)],
+            ]), []);
+            const keyValuePairsForNewCollection = _.reduce(newCollection, (finalArray, val, key) => ([
+                ...finalArray,
+                [key, JSON.stringify(val)],
+            ]), []);
+
+            // New keys will be added via multiSet while existing keys will be updated using multiMerge
```
marcaaron, line 531: Maybe this would be a good spot to add some "why" context? Probably there is no harm in adding 

>because setting a key that doesn't exist yet with `multiMerge` will throw errors

tests/components/ViewWithCollections.js
```diff
@@ -0,0 +1,31 @@
...
+    })),
+};
+
+const defaultProps = {
+    collections: {},
+};
+
+const ViewWithCollections = (props) => {
+    /* eslint-disable no-console */
+    console.log('rendering ViewWithCollections');
```
marcaaron, line 18: Really creative! Just thought of an alternative that might be more obvious. We should be able to pass a [`jest.fn()`](https://jestjs.io/docs/en/mock-functions) as a prop and call it in place of this `console.log` the test would then looks something like 

```js
const onRender = jest.fn();
render(<TestComponentWithOnyx onRender={onRender} />); 
expect(onRender.mock.calls.length).toBe(5);
```

Also kind of surprised there isn't a better way to do this with the test renderer.

tests/unit/onyxTest.js
```diff
@@ -145,18 +147,88 @@ describe('Onyx', () => {
...
+            },
+            test_3: {
+                ID: 345,
+                value: 'three'
+            }
+        })
+            .then(() => {
+                // 2 key values to update and 2 new keys to add.
+                // MergeCollection will perform a mix of multiSet and multiMerge
+                Onyx.mergeCollection(ONYX_KEYS.COLLECTION.TEST_KEY, {
```
marcaaron, line 196: Might be able to return this back to the `.then()` instead of returning `waitForPromisesToResolve()`?

tests/unit/onyxTest.js
```diff
@@ -145,18 +147,88 @@ describe('Onyx', () => {
...
+
+    it('should properly set and merge when using mergeCollection', () => {
+        const valuesReceived = {};
+        let numberOfCallbacks = 0;
+
+        connectionID = Onyx.connect({
+            key: ONYX_KEYS.COLLECTION.TEST_KEY,
+            initWithStoredValues: false,
+            callback: (data) => {
+                ++numberOfCallbacks;
```
marcaaron, line 173: This is also fine but tossing out that `jest.fn()` can maybe can help here? We can look at the end to see how many calls were made and what each one was made with:

```js
const mockCallback = jest.fn()
connectionID = Onyx.connect({
            key: ONYX_KEYS.COLLECTION.TEST_KEY,
            initWithStoredValues: false,
            callback: mockCallback,
    });
    ...
    expect(mockCallback.mock.calls.length).toBe(7);
    expect(mockCallback.mock.calls[0][0]).toEqual({ID: 123, value: 'one'});
```

marcaaron (changes requested):
Looks great just have a few more comments mostly minor stuff and suggestions about the tests.

lib/Onyx.js
chiragsalian: Yup, you're correct. Shall update.

tests/unit/onyxTest.js
chiragsalian: I'll try that out

tests/components/ViewWithCollections.js
chiragsalian: shall try that out

tests/unit/onyxTest.js
chiragsalian: Tried it. Doesn't work when I return this since it fails here `expect(numberOfCallbacks).toEqual(7);` thinking the value is 3. I think it takes time for the callback to finish so `waitForPromisesToResolve` is maybe waiting for the callback to also finish which is why I get the correct values in expect when I use `waitForPromisesToResolve`

tests/unit/onyxTest.js
chiragsalian: oh damn, mock functions are pretty cool. Didn't know i can check number of times it was called with calls.length. Cool stuff. I do want to see if i can save the values somewhere to confirm the merge worked correctly than a set so I'll check out how to configure my mock function.

marcaaron: yes! Can always save the value like you are doing and just call the mock function. So that example would become:

```js
    Onyx.connect({
            key: ONYX_KEYS.COLLECTION.TEST_KEY,
            initWithStoredValues: false,
            callback: (val) => {
                // save the value
               mockCallback();
            },
    });
```

tests/unit/onyxTest.js
marcaaron: >maybe waiting for the callback to also finish which is why I get the correct values

There is no async work happening in `keysChanged()` or the callbacks so I think it's something else...
I did notice we are returning the `Promise` created by `AsyncStorage.getAllKeys()` but not returning the inner promise here...

https://github.com/Expensify/react-native-onyx/blob/727c1347743ed35aac30393dca8bb530b15947bd/lib/Onyx.js#L512-L535

tests/unit/onyxTest.js
chiragsalian: thats cool, i was actually thinking of setting the value in the mock function like so,
```
const mockCallback = jest.fn(data => valuesReceived[data.ID] = data.value);
```
That should be fine right? Or is it recommended to have the mock function empty?

chiragsalian:
Updated code and ready for review

tests/unit/onyxTest.js
marcaaron: Nice! Can't think of any reason why we shouldn't do it as you've done.

marcaaron (approved):
LGTM 👍

PR #45, opened 2021-01-28
Add cache eviction test. Clean up some other tests.

marcaaron:
cc @chiragsalian @madmax330 might be interested in the conversation [here](https://expensify.slack.com/archives/C01GTK53T8Q/p1609359194296700)

Doesn't seem like there is a very clear consensus yet on whether we **need** to fix the existing cache eviction strategy. But the concerns in the Slack thread above make a lot of sense. I have also started to see storage errors popping up more and more frequently while testing Expensify.cash via the web proxy instead of local development. I expect that other engineers will see the same if they switch to testing via the proxy.

For now, this PR just adds tests for the existing cache eviction system since none were written when it was conceived. Eventually if/when we re-work cache eviction it will be good to have a test set up to build off of. Additionally, it seems like a good idea to get some of this stuff under test.

### DIFF
--- a/lib/Logger.js
+++ b/lib/Logger.js
@@ -1,5 +1,5 @@
 // Logging callback
-let logger;
+let logger = () => {};
 
 /**
  * Register the logging callback
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -240,14 +240,15 @@ function keyChanged(key, data) {
  * @param {string} [config.statePropertyName]
  * @param {function} [config.callback]
  * @param {*|null} val
+ * @param {String} key
  */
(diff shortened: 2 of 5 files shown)


robertjchen (commented):
This looks great, good idea to have this test in place in any case 👍

robertjchen (approved):

AndrewGable:
Not sure why but this PR didn't have GH actions run on it, meaning there was some lint errors that snuck in 🤔

marcaaron:
When did we add GH actions for this repo?

marcaaron:
I can't recall ever seeing them run, but maybe they were at one point.

PR #53, opened 2021-03-19
[Onyx] - Updated lodash dependency to fix vulnerability

Luke9389:
cc @tgolen 

### Details
<Explanation of the change or anything fishy that is going on>

Rather than using individual packages for each lodash method, we are using the main lodash package and selectively loading the methods we need (using this handy syntax: `lodash/merge`)

### Fixed Issues
<Please replace GH_LINK with the link to the GitHub issue this Pull Request is fixing>
Fixes https://github.com/Expensify/Expensify/issues/157899


bondydaa (approved):
Can you add a QA section that states regular regressions are all that are needed, no special tests.

PR #59, opened 2021-04-02
Delay withOnyx render when prop dependent key changes so new data appears at once

marcaaron:
### Details
This refactor addresses some issues with `withOnyx` handling of dynamic keys based on props. Previously if a prop dependent key changed we would:

1. Re-render the wrapped component first
2. Re-subscribe to the new keys that have changed
3. Then re-render the wrapped component once again per key that changes (if there are multiple keys depending on the same prop)

This creates some problems around tracking props changes. 

e.g. if you have a `report_${props.reportID}` and `reportActions_${props.reportID}` key and want to respond to the change in props by comparing the previous `reportID` to the current `reportID` then you will only be able to react to the `reportID` change and not changes to any Onyx derived props that use this value to generate a key.

Basically after this change if we subscribe to a key using a dynamic prop then it should be 100% reliable to look at the key change and then know that anything dependent on it should be updated in the same cycle.

This PR also removes the `initWithStoredValues` property since it's no longer relevant and made things a little complicated. There is no current situation where we need to subscribe to an Onyx key, but not use the stored values. If there was, we should treat that as an edge case and either use `Onyx.connect()` in `componentDidMount()` to access "new" values.

### Related Issues
https://github.com/Expensify/Expensify/issues/159220

### Automated Tests
There is an automated test added that validates a component only renders once when a prop updates that changes a key.

### Linked PRs
https://github.com/Expensify/Expensify.cash/pull/2220


marcaaron:
I'm going to close this for now but leave the original issue open in case we decide to address it at some point. Probably this implementation is out of date now and needs to be updated for consistency with the caching changes added.

PR #60, opened 2021-04-06
Create a test case for mergecollection value check

tugbadogan:
Related issue: https://github.com/Expensify/Expensify.cash/issues/2225

Created a new test case for mergeCollection function to see if it merges values as expected. This test will fail.

### DIFF
--- a/tests/unit/withOnyxTest.js
+++ b/tests/unit/withOnyxTest.js
@@ -108,11 +108,11 @@ describe('withOnyx', () => {
         const onRender = jest.fn();
         render(<TestComponentWithOnyx onRender={onRender} />);
         Onyx.mergeCollection(ONYX_KEYS.COLLECTION.TEST_KEY, {test_4: {ID: 456}, test_5: {ID: 567}});
-        Onyx.mergeCollection(ONYX_KEYS.COLLECTION.TEST_KEY, {test_4: {Name: "Test4"}, test_5: {Name: "Test5"}, test_6: {ID: 678, Name: "Test6"}});
+        Onyx.mergeCollection(ONYX_KEYS.COLLECTION.TEST_KEY, {test_4: {Name: 'Test4'}, test_5: {Name: 'Test5'}, test_6: {ID: 678, Name: 'Test6'}});
         return waitForPromisesToResolve()
             .then(() => {
                 expect(onRender.mock.calls.length).toBe(3);
-                expect(onRender.mock.instances[2].text).toEqual({ID: 456, Name: "Test4"});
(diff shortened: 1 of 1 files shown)


tugbadogan:

tests/unit/withOnyxTest.js
```diff
@@ -96,3 +96,23 @@ describe('withOnyx', () => {
...
+        const collectionItemID = 4;
+        const TestComponentWithOnyx = withOnyx({
+            text: {
+                key: `${ONYX_KEYS.COLLECTION.TEST_KEY}${collectionItemID}`,
+            },
+        })(ViewWithCollections);
+        const onRender = jest.fn();
+        render(<TestComponentWithOnyx onRender={onRender} />);
+        Onyx.mergeCollection(ONYX_KEYS.COLLECTION.TEST_KEY, {test_4: {ID: 456}, test_5: {ID: 567}});
+        Onyx.mergeCollection(ONYX_KEYS.COLLECTION.TEST_KEY, {test_4: {Name: 'Test4'}, test_5: {Name: 'Test5'}, test_6: {ID: 678, Name: 'Test6'}});
```
marcaaron, line 111: I see these two calls happening [here](https://github.com/Expensify/Expensify.cash/blob/0921feec98fa5d1ce14ee51071bebe9e148bccc8/src/libs/actions/PersonalDetails.js#L195) and [here](https://github.com/Expensify/Expensify.cash/blob/0921feec98fa5d1ce14ee51071bebe9e148bccc8/src/libs/actions/Report.js#L317).

This makes sense to me since `mergeCollection()` does not work the same way as `merge` so the assumption to treat the data the same was incorrect on my part. However, I'm unsure what would be the best way forward since `Onyx.merge()` does not require that we use `prevState` and the mechanism is a bit different.

See here: 

https://github.com/Expensify/react-native-onyx/blob/aa36bdb7b5cd2319c34b9ee1b4c73cb73e89f4a5/lib/Onyx.js#L452-L513

The merge is batched and uses the data in `localStorage` as the source of truth so that when we call `updateIndividualSubscriber()` we will merge them together and then set the whole thing to state.

Probably the answer is to treat `mergeCollection()` similar to `merge()`? Another idea is that we can only use `prevState` when using `mergeCollection()`. Open to suggestions on the best way forward.

marcaaron:
Hey @tugbadogan we're on somewhat of a tight deadline with this so I'm going to create the PR we will get you a bonus for catching this + helping write the test (which I'll include in the PR I'm drafting now). Thanks again!

PR #64, opened 2021-04-16
Add test case for passing props from one connected component to another

marcaaron:
Unrelated to any specific issue just adds a new test case.

### DIFF
--- a/lib/compose.js
+++ b/lib/compose.js
@@ -16,7 +16,7 @@
  *
  * @returns {Function}
  */
- export default function compose(...funcs) {
+export default function compose(...funcs) {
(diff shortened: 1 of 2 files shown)


Julesssss (approved):
Thanks, will use this example to test my problem. 

I wonder if this test would be better suited for Expensify.cash, where our compose function is liable to change. No harm in including it here too though!

marcaaron:
>I wonder if this test would be better suited for Expensify.cash, where our compose function is liable to change. No harm in including it here too though!

Ah we've just stolen that function from Redux so hopefully it won't ever change 😅

PR #66, opened 2021-04-22
Fix mergeCollection bug

tgolen:
### Details
This fixes a bug found in mergeCollection

### Related Issues
https://github.com/Expensify/Expensify/issues/161384

### Automated Tests
There is a test added which properly tests the behavior

### Linked PRs
https://github.com/Expensify/Expensify.cash/pull/2543

### DIFF
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -195,7 +195,10 @@ function keysChanged(collectionKey, collection) {
                     newData = lodashMerge(newData, prevState[subscriber.statePropertyName]);
                 }
                 return {
-                    [subscriber.statePropertyName]: newData,
+                    // If `newData` happens to not exist, then set the state back to the previous value because
+                    // if the state is set to an undefined value, then the item reverts to whatever is in
+                    // defaultProps
+                    [subscriber.statePropertyName]: newData || prevState[subscriber.statePropertyName],
                 };
             });
         }
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -199,7 +199,10 @@ function keysChanged(collectionKey, collection) {
 
             subscriber.withOnyxInstance.setState(prevState => ({
                 [subscriber.statePropertyName]: _.isObject(dataFromCollection)
-                    ? lodashMerge(dataFromCollection, prevState[subscriber.statePropertyName])
+                    ? {
+                        ...prevState[subscriber.statePropertyName],
+                        ...dataFromCollection,
+                    }
                     : dataFromCollection,
             }));
         }
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -189,20 +189,19 @@ function keysChanged(collectionKey, collection) {
                 });
             }
         } else if (isSubscribedToCollectionMemberKey) {
-            subscriber.withOnyxInstance.setState((prevState) => {
-                let newData = collection[subscriber.key];
-                if (_.isObject(newData)) {
-                    newData = lodashMerge(newData, prevState[subscriber.statePropertyName]);
-                }
-                return {
-                    // If `newData` happens to not exist, then set the state back to the previous value because
-                    // if the state is set to an undefined value, then the item reverts to whatever is in
-                    // defaultProps
-                    [subscriber.statePropertyName]: !_.isUndefined(newData)
-                        ? newData
-                        : prevState[subscriber.statePropertyName],
-                };
-            });
+            const dataFromCollection = collection[subscriber.key];
+
+            // If `dataFromCollection` happens to not exist, then return early so that there are no unnecessary
+            // re-renderings of the component
+            if (_.isUndefined(dataFromCollection)) {
+                return;
+            }
+
+            subscriber.withOnyxInstance.setState((prevState) => ({
+                [subscriber.statePropertyName]: _.isObject(dataFromCollection)
+                    ? lodashMerge(dataFromCollection, prevState[subscriber.statePropertyName])
+                    : dataFromCollection,
+            }));
         }
     });
 }
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -198,7 +198,7 @@ function keysChanged(collectionKey, collection) {
                     // If `newData` happens to not exist, then set the state back to the previous value because
                     // if the state is set to an undefined value, then the item reverts to whatever is in
                     // defaultProps
-                    [subscriber.statePropertyName]: newData || prevState[subscriber.statePropertyName],
+                    [subscriber.statePropertyName]: typeof newData !== 'undefined' ? newData : prevState[subscriber.statePropertyName],
                 };
             });
         }
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -197,7 +197,7 @@ function keysChanged(collectionKey, collection) {
                 return;
             }
 
-            subscriber.withOnyxInstance.setState((prevState) => ({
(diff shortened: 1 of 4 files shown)


tgolen:

lib/Onyx.js
```diff
@@ -195,7 +195,10 @@ function keysChanged(collectionKey, collection) {
                     newData = lodashMerge(newData, prevState[subscriber.statePropertyName]);
                 }
                 return {
-                    [subscriber.statePropertyName]: newData,
+                    // If `newData` happens to not exist, then set the state back to the previous value because
+                    // if the state is set to an undefined value, then the item reverts to whatever is in
+                    // defaultProps
+                    [subscriber.statePropertyName]: typeof newData !== 'undefined' ? newData : prevState[subscriber.statePropertyName],
```
marcaaron: Should we use `_.isUndefined()` ?

marcaaron (commented):
Looks great and nice catch!

lib/Onyx.js
tgolen: Ah, yeah... that's a good call. I guess we do have underscore here.

tgolen:
Updated to use underscore

lib/Onyx.js
```diff
@@ -195,7 +195,12 @@ function keysChanged(collectionKey, collection) {
                     newData = lodashMerge(newData, prevState[subscriber.statePropertyName]);
                 }
                 return {
-                    [subscriber.statePropertyName]: newData,
+                    // If `newData` happens to not exist, then set the state back to the previous value because
+                    // if the state is set to an undefined value, then the item reverts to whatever is in
+                    // defaultProps
+                    [subscriber.statePropertyName]: !_.isUndefined(newData)
+                        ? newData
+                        : prevState[subscriber.statePropertyName],
```
marcaaron, line 203: Thinking more on this... this could maybe be filed under "unrelated" or "not a problem", but I find it interesting (or rather incorrect/inefficient) that if a subscriber is connected to a single key, but it's data hasn't changed at all ... that we are still calling `setState()` with previous data?

That's not super great because it means that we will still trigger re-renders on those connected components and for no reason at all. So, probably if we have `undefined` then we should do less than set the previous state again and should do... well... nothing?

tests/unit/withOnyxTest.js
```diff
@@ -153,3 +153,70 @@ describe('withOnyx', () => {
...
+                // THEN the component subscribed to the modified item should have the new version of the item
+                // and all other components should be unchanged.
+                // Note: each component is rendered twice. Once when it is initially rendered, and then again
+                // when the collection is updated. That's why there are two checks here for each component.
+                expect(onRender1.mock.calls[0][0].testObject).toStrictEqual({ID: 1});
+                expect(onRender1.mock.calls[1][0].testObject).toStrictEqual({ID: 1, newProperty: 'yay'});
+                expect(onRender2.mock.calls[0][0].testObject).toStrictEqual({ID: 2});
+                expect(onRender2.mock.calls[1][0].testObject).toStrictEqual({ID: 2});
+                expect(onRender3.mock.calls[0][0].testObject).toStrictEqual({ID: 3});
+                expect(onRender3.mock.calls[1][0].testObject).toStrictEqual({ID: 3});
```
marcaaron, line 219: Like my expectation would be... not that this gets called again with same data but not called at all ? Why should it re-render since nothing has changed?

lib/Onyx.js
tgolen: Yeah, I think you make a good point. While it's not a problem currently, it's not a great design either. The reason I put it inside `setState` is because that's where it is getting `newData`. But now that I look at it closely, `newData` should be able to be retrieved before calling `setState()` and in the case where it is undefined, we can just skip calling `setState()`. How's that sound?

marcaaron: Yep, sounds good to me. 👍

tgolen:
OK, updated. Now with less re-renderings!

marcaaron (approved):

chiragsalian (dismissed):
LGTM and i like the changes 👍

tgolen:
OK, before merging this, I want to give this a little more testing because I haven't been able to test the most recent changes yet. Once I do that testing, I'll let you know when this has a green-light to merge.

stitesExpensify (approved):

tests/unit/withOnyxTest.js
```diff
@@ -153,3 +153,73 @@ describe('withOnyx', () => {
...
     });
 });
+
+describe('withOnyx', () => {
+    it('using mergeCollection to modify one item should only effect one component', () => {
+        const onRender1 = jest.fn();
+        const onRender2 = jest.fn();
+        const onRender3 = jest.fn();
+
+        // GIVEN there is a collection with three simple items in it
```
deetergp, line 163: NAB: I REALLY appreciate the `GIVEN/WHEN/THEN` docs here! It makes it so much easier to follow what's going on and what we're testing for. `:bigfan:`

deetergp (dismissed):
Looks good to me

tests/unit/withOnyxTest.js
tgolen: Yeah, I love that format for testing comments. @nkuoch was the one that originally showed it to me!

tgolen:
OK, I think I found another bug in my testing, but I want to write a test case to confirm that theory. Will continue this on Monday.

tgolen:
All right, I am taking off the hold here. When I tested this on Friday I found another bug with how the new collection data was being merged into the state. I have added a test case to cover that, fixed the bug (using spread operators instead of `lodash.merge()`), and I've tested this in the IOU PR to ensure it works the way it's supposed to.

chiragsalian (approved):
LGTM

lib/Onyx.js
```diff
@@ -189,15 +189,22 @@ function keysChanged(collectionKey, collection) {
...
+            // If `dataFromCollection` happens to not exist, then return early so that there are no unnecessary
+            // re-renderings of the component
+            if (_.isUndefined(dataFromCollection)) {
+                return;
+            }
+
+            subscriber.withOnyxInstance.setState(prevState => ({
+                [subscriber.statePropertyName]: _.isObject(dataFromCollection)
+                    ? {
+                        ...prevState[subscriber.statePropertyName],
```
marcaaron, line 203: This seems potentially dangerous as it assumes we will always have at least an empty object for `prevState[subscriber.statePropertyName]` if we are dealing with object shaped data. If we try to spread something like `undefined` or `null` then we get 💥.

Maybe some combination of `lodash.merge()` could still be helpful here to help guard against that?

marcaaron: Maybe there's no problem with that actually. I could just have made that up in my head.

marcaaron: I think spreading `undefined` in objects works OK but arrays doesn't.

deetergp (approved):

lib/Onyx.js
marcaaron: I guess the small difference between this and using `lodashMerge()` is that the spread is not going to recursively merge the data. So depending on how complicated that data is we could get some weird results. But we can address that if it becomes a problem.

marcaaron (approved):

lib/Onyx.js
tgolen: Cool, thanks Marc. I think it'll be OK for now. Seeing as how everyone has approved this, I'm going to go ahead and merge it.

Issue #65, opened 2021-04-16
Benchmark Onyx


kidroca:
### Problem
Measure the performance of Onyx

#### Execution time
- How much time it takes from start to finish for a public method exposed by the API 
- E.g. how much time passes between the initial call to `connect` and resolving data to the caller. Same for other methods like `merge`, `set`, `get` ...

#### Method call counts and keys used
- Is `connect` called often to retrieve the same key?
- How many times?
- When does that happen? Are there many calls during init - timestamps of the calls

#### Onyx memory footprint
- How much of the data provided by Onyx is in memory at a given time
- Have a way to check which of the Objects returned by Onyx are still in memory, and the amount of memory they take

### Solution
TBD

kidroca:
You don't have to patch timer logic inside Onyx code 

You can decorates existing methods like this:

```js
import Onyx from './lib/Onyx';
import withOnyx from './lib/withOnyx';

if (process.env.ONYX_INSTRUMENTATION == 'true') {
  const decorator = require('./lib/benchmarkDecorator');
  decorator.decorate(Onyx);
}

export default Onyx;
export {withOnyx};
```

And the decorator would do something like this

```jsx
function decorate(Onyx) {
  const promiseReturningMethods = ['merge', 'set', 'get', ...];

  promiseReturningMethods.forEach(methodName => {
     const original = Onyx[methodName];

     const decorated = (...args) => {
        const logId = Log.timeStart(methodName, args);
        
        return original.call(Onyx, ...args)
           .finally(() => Log.timeEnd(logId))
     };

     Onyx[methodName] = decorated;
  });
}
```

If Onyx had a prototype (was a class) you would've been able to also track internal usage by this decorating pattern

Something similar can be done for methods like `connect` that can't return a promise because they return `connectionID` 

This could be further extended to track call counts and memory usage. 
 - Data returned by Onyx can be tagged and you'd be able to use dev tools to see how much of the data is in memory, what's the actual content, how much memory does it take etc..

quinthar:
Oh, this is cool!  Some questions to help me understand it:

1. Are you suggesting that this be implemented as some kind of "profile mode", where when enabled it begins tracking this data?  Or are you suggesting this is something "always on"?  (I would vote an explicit "profile mode")

2. Are you suggesting that the profiling data be harvested centrally via some kind of server, or just perhaps output in the debug console?  (I would vote just logged into the debug console, so we'd use this internally for testing specific builds.)

[...]

PR #70, opened 2021-05-13
Feature: Benchmark Onyx

kidroca:
@marcaaron 

### Details
Moved some methods to `Onyx.internal.js` - e.g. to test private methods like `get` through the decorator pattern
Created `decorateWithMetrics` lib that serves to decorate Onyx methods similarly to how HOCs are used to enhance components
Updated some methods so they return the promises they work with so that they can be metered
Added `captureMetrics` flag to Onyx.init. Initially the intent was to use `process.env.BENCHMARK_ONYX` but `process.env` has only `NODE_ENV` so far - this can be configured after the draft is finalized

Created a sample usage that will print statistics for each navigation state change (only in Chrome dev tools - debug mode). Don't forget to set the `BENCHMARK_ONYX=true` in `.env`:  https://github.com/kidroca/Expensify.cash/tree/kidroca/sample-onyx-benchmark

Some methods cannot be tested as easy like the `connect` method, they would need some changes to incorporate the decorator - e.g. don't measure the whole `connect` but move the promise based logic to a separate method and measure that method - this way connect can still return a `connectionID` or a `removeListener` callback

### Related Issues

#65 
https://github.com/Expensify/Expensify.cash/issues/2762

### Automated Tests
Added basic covering tests TBD


kidroca:
Metering xlsx sample:

https://docs.google.com/spreadsheets/d/1R4YPeG6rHbTiY3gcKINV-kLFsW37ITnfhV_MYnWmJzs/edit?usp=sharing

(index) | method | startTime | endTime | duration | args | result
-- | -- | -- | -- | -- | -- | --
  |   |   |   |   |   |  
0 | "Onyx.merge" | 1350810.00ms | 1350827.00ms | 18.00ms | "["modal",{"isVisible":false}]" | undefined
1 | "Onyx.merge" | 1350810.00ms | 1350827.00ms | 17.00ms | "["modal",{"isVisible":false}]" | undefined
2 | "Onyx.merge" | 1350810.00ms | 1350827.00ms | 17.00ms | "["modal",{"isVisible":false}]" | undefined
3 | "Onyx.merge" | 1350810.00ms | 1350827.00ms | 17.00ms | "["modal",{"isVisible":false}]" | undefined
4 | "Onyx.merge" | 1350810.00ms | 1350827.00ms | 17.00ms | "["modal",{"isVisible":false}]" | undefined
5 | "Onyx.merge" | 1350810.00ms | 1350827.00ms | 17.00ms | "["modal",{"isVisible":false}]" | undefined
6 | "Onyx.merge" | 1350811.00ms | 1350827.00ms | 17.00ms | "["modal",{"isVisible":false}]" | undefined
7 | "Onyx.merge" | 1350811.00ms | 1350827.00ms | 16.00ms | "["modal",{"isVisible":false}]" | undefined
8 | "Onyx.merge" | 1350811.00ms | 1350827.00ms | 16.00ms | "["modal",{"isVisible":false}]" | undefined
9 | "Onyx.merge" | 1350811.00ms | 1350827.00ms | 16.00ms | "["modal",{"isVisible":false}]" | undefined
10 | "Onyx.merge" | 1350811.00ms | 1350827.00ms | 16.00ms | "["modal",{"isVisible":false}]" | undefined
65 | "Onyx.get" | 1350808.00ms | 1350850.00ms | 42.00ms | "["modal"]" | "{"isVisible":false}"
66 | "Onyx.get" | 1350820.00ms | 1350871.00ms | 50.00ms | "["currentlyViewedReportID"]" | ""71955477""
67 | "Onyx.get" | 1350859.00ms | 1350884.00ms | 25.00ms | "["report_71402263"]" | "{"reportID":71402263,"reportName":"Concierge","unr…cierge_2021.png"],"optimisticReportActionIDs":[]}"
68 | "Onyx.get" | 1350861.00ms | 1350891.00ms | 30.00ms | "["personalDetails"]" | "{"expensify3@kidroca.com":{"login":"expensify3@kid…":{"automatic":true,"selected":"Europe/Athens"}}}"
69 | "Onyx.get" | 1350865.00ms | 1350899.00ms | 34.00ms | "["session"]" | "{"authToken":"0DACCD10C10A007A181A76FE3D529FD286D4…m","loading":false,"shouldShowComposeInput":true}"
70 | "Onyx.get" | 1350926.00ms | 1350942.00ms | 16.00ms | "["report_71402263"]" | "{"reportID":71402263,"reportName":"Concierge","unr…cierge_2021.png"],"optimisticReportActionIDs":[]}"
71 | "Onyx.get" | 1350928.00ms | 1351045.00ms | 117.00ms | "["reportActions_71402263"]" | "{"0":{"actionName":"CREATED","created":"Mar 19 202…true,"clientID":"","reportActionID":"755784774"}}"
72 | "Onyx.get" | 1350931.00ms | 1351056.00ms | 125.00ms | "["session"]" | "{"authToken":"0DACCD10C10A007A181A76FE3D529FD286D4…m","loading":false,"shouldShowComposeInput":true}"
73 | "Onyx.get" | 1350935.00ms | 1351062.00ms | 126.00ms | "["modal"]" | "{"isVisible":false}"
74 | "Onyx.get" | 1350937.00ms | 1351065.00ms | 128.00ms | "["network"]" | "{"isOffline":false}"
75 | "Onyx.get" | 1350939.00ms | 1351074.00ms | 135.00ms | "["report_71402263"]" | "{"reportID":71402263,"reportName":"Concierge","unr…cierge_2021.png"],"optimisticReportActionIDs":[]}"
76 | "Onyx.get" | 1351055.00ms | 1351270.00ms | 215.00ms | "["report_71402263"]" | "{"reportID":71402263,"reportName":"Concierge","unr…cierge_2021.png"],"optimisticReportActionIDs":[]}"
77 | "Onyx.get" | 1351273.00ms | 1351510.00ms | 237.00ms | "["session"]" | "{"authToken":"0DACCD10C10A007A181A76FE3D529FD286D4…m","loading":false,"shouldShowComposeInput":true}"
78 | "Onyx.get" | 1351283.00ms | 1351522.00ms | 238.00ms | "["personalDetails"]" | "{"expensify3@kidroca.com":{"login":"expensify3@kid…":{"automatic":true,"selected":"Europe/Athens"}}}"
79 | "Onyx.get" | 1351293.00ms | 1351535.00ms | 243.00ms | "["personalDetails"]" | "{"expensify3@kidroca.com":{"login":"expensify3@kid…":{"automatic":true,"selected":"Europe/Athens"}}}"
80 | "Onyx.get" | 1351346.00ms | 1351595.00ms | 249.00ms | "["personalDetails"]" | "{"expensify3@kidroca.com":{"login":"expensify3@kid…":{"automatic":true,"selected":"Europe/Athens"}}}"
425 | "Onyx.getAllKeys" | 1353100.00ms | 1353521.00ms | 421.00ms | "[]" | "["account","betas","countryCode","credentials","cu…ort_73013597","report_73164379","session","user"]"
426 | "Onyx.getAllKeys" | 1353113.00ms | 1353535.00ms | 421.00ms | "[]" | "["account","betas","countryCode","credentials","cu…ort_73013597","report_73164379","session","user"]"
427 | "Onyx.getAllKeys" | 1353126.00ms | 1353551.00ms | 425.00ms | "[]" | "["account","betas","countryCode","credentials","cu…ort_73013597","report_73164379","session","user"]"
428 | "Onyx.getAllKeys" | 1353136.00ms | 1353566.00ms | 429.00ms | "[]" | "["account","betas","countryCode","credentials","cu…ort_73013597","report_73164379","session","user"]"
429 | "Onyx.getAllKeys" | 1353148.00ms | 1353578.00ms | 430.00ms | "[]" | "["account","betas","countryCode","credentials","cu…ort_73013597","report_73164379","session","user"]"
430 | "Onyx.getAllKeys" | 1353209.00ms | 1353605.00ms | 396.00ms | "[]" | "["account","betas","countryCode","credentials","cu…ort_73013597","report_73164379","session","user"]"
431 | "Onyx.getAllKeys" | 1353228.00ms | 1353618.00ms | 390.00ms | "[]" | "["account","betas","countryCode","credentials","cu…ort_73013597","report_73164379","session","user"]"
432 | "Onyx.getAllKeys" | 1353365.00ms | 1353740.00ms | 375.00ms | "[]" | "["account","betas","countryCode","credentials","cu…ort_73013597","report_73164379","session","user"]"
433 | "Onyx.getAllKeys" | 1353379.00ms | 1353743.00ms | 364.00ms | "[]" | "["account","betas","countryCode","credentials","cu…ort_73013597","report_73164379","session","user"]"
434 | "Onyx.getAllKeys" | 1353500.00ms | 1353810.00ms | 310.00ms | "[]" | "["account","betas","countryCode","credentials","cu…ort_73013597","report_73164379","session","user"]"
435 | "Onyx.getAllKeys" | 1353551.00ms | 1353837.00ms | 286.00ms | "[]" | "["account","betas","countryCode","credentials","cu…ort_73013597","report_73164379","session","user"]"
436 | "Onyx.getAllKeys" | 1353763.00ms | 1353971.00ms | 208.00ms | "[]" | "["account","betas","countryCode","credentials","cu…ort_73013597","report_73164379","session","user"]"
437 | "Onyx.getAllKeys" | 1353774.00ms | 1353974.00ms | 201.00ms | "[]" | "["account","betas","countryCode","credentials","cu…ort_73013597","report_73164379","session","user"]"

kidroca:

kidroca:
recheck

kidroca:
Now that I look at the data `merge` does not seem to be measured very accurately

```js
    if (mergeQueue[key]) {
        mergeQueue[key].push(val);
        return Promise.resolve();
    }
```

It returns early when it's filling a queue
So there are a lot of 7-10ms calls 
And then one big call ranging between 100ms and 900ms

lib/Onyx.js
```diff
@@ -405,7 +395,7 @@ function evictStorageAndRetry(error, ionMethod, ...args) {
  */
 function set(key, val) {
     // Write the thing to persistent storage, which will trigger a storage event for any other tabs open on this domain
-    return AsyncStorage.setItem(key, JSON.stringify(val))
+    return OnyxInternal.set(key, val)
         .then(() => keyChanged(key, val))
         .catch(error => evictStorageAndRetry(error, set, key, val));
 }
```
kidroca, line 401: Initially I wanted to move this whole function to `Onyx.internals` as well, but when I did I had to also move `keyChanged` and `evictStorageAndRetry` as well, as I moved those 2, I needed to also move, more and more functions and static variables, so I decided to just move the async part.

lib/Onyx.js
```diff
@@ -579,8 +570,15 @@ function init({
...
-    registerStorageEventListener
+    registerStorageEventListener,
+    captureMetrics = false,
 }) {
+    if (captureMetrics) {
+        // The code here is only bundled and applied when the captureMetrics is set
+        // eslint-disable-next-line no-use-before-define
+        applyDecorators();
+    }
+
```
kidroca, line 581: The initial idea was to use `process.env.BENCHMARK_ONYX` ENV variable directly here, but it seems it's not as straightforward as to include it in `.env` or provide it form a shell

Variables in `.env` are exposed through `react-native-config` but since Onyx does not have it as dependency I've decided to use the `captureMetrics` flag

I'm settings `captureMetrics` in E.cash by reading env variables using `react-native-config` 

```js
import Config from 'react-native-config';
import Onyx, {withOnyx} from 'react-native-onyx';

Onyx.init({
    keys: ONYXKEYS,
    safeEvictionKeys: [ONYXKEYS.COLLECTION.REPORT_ACTIONS],
    captureMetrics: Config.BENCHMARK_ONYX,
    /* ... */
});
```

lib/decorateWithMetrics.js
```diff
@@ -0,0 +1,120 @@
...
+/**
+ * Class instances are easily tracked by Dev tools memory snapshots
+ * Wrapping a result with such instance allows us to track it without
+ * incurring extra weight
+ */
+class TaggedItem {
+    constructor(data) {
+        this.ref = data;
+    }
+}
```
kidroca, line 16: So far this seems to only work on web - when I capture profile samples in react-native / debug / chrome / memory - the objects do not appear there, while in web / chrome / memory I see this 

image

lib/decorateWithMetrics.js
```diff
@@ -0,0 +1,120 @@
...
+    // eslint-disable-next-line no-param-reassign
+    instance[methodName] = (...args) => {
+        const startTime = performance.now();
+        let result;
+        let error;
+
+        const promise = originalMethod.apply(instance, args);
+
+        // Then handlers added here are not affecting the returned promise as they aren't attached to it
+        promise.then(data => data && (result = new TaggedItem(data)))
```
marcaaron, line 42: Would it be bad to create a `TaggedItem` even if `data` had no value? Just curious why we are preventing this since some `result` will be `undefined`.

lib/decorateWithMetrics.js
```diff
@@ -0,0 +1,120 @@
...
+ * The list would be empty if nothing is captured yet or the method is not metered
+ * @param {String} [methodName]
+ * @returns {Array<CallMeta>}
+ */
+function getMetrics(methodName) {
+    if (methodName) {
+        return stats.get(methodName) || [];
+    }
+
+    return Array.from(stats.values()).flat();
```
marcaaron, line 87: Not to stifle use of built-in array smoosh, but I think we would have some more consistency with the rest of our code if the `Map` was a plain `Object` and this returned something like...

```js
return _.chain(stats)
    .values()
    .flatten()
    .value();
```

marcaaron (commented):
This all looks pretty straightforward to me, but going to tag in @tgolen to give his thoughts.

marcaaron:
>Now that I look at the data merge does not seem to be measured very accurately

Regarding this comment. The only calls to `merge()` that we are interested in measuring are the ones that write to disk. I don't have a perfect answer on how to get around that. But one complicated idea would be...

Make `merge()` privately and unreliably return a `Promise` that can be measured _if_ one has been returned (e.g. check return value of calling `_merge()` to see if `instanceof Promise` and then have the public method always return `void` so `.then()` cannot be called on it). Not sure if the decorator pattern you are going for makes that easy or not. 

A simpler alternative would be to just manually wrap all the promises we want to measure and tag them. But you'd have to unwind the decorator pattern stuff.

marcaaron:
>Some methods cannot be tested as easy like the connect method, they would need some changes to incorporate the decorator - e.g. don't measure the whole connect but move the promise based logic to a separate method and measure that method - this way connect can still return a connectionID or a removeListener callback

Sounds like this could be another reason why benchmarking inline with a higher order function that wraps any `Promise` could be useful. This way we would not need to have methods that return `Promise`, but could track any `Promise` whether returned by a function or not. But I'll let you decide.

lib/decorateWithMetrics.js
```diff
@@ -0,0 +1,120 @@
...
+ */
+class TaggedItem {
+    constructor(data) {
+        this.ref = data;
+    }
+}
+
+/**
+ * Decorates an instance method with metrics capturing functionality
+ * The method needs to be async or to return a promise
```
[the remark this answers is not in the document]
kidroca: Hmm this is probably not documented well it should just say "The method needs to return a promise" 
For example `merge` was one of the methods that is async but did not return the promise it worked with
Also some people don't know that `async/await` methods return a promise, not that we use this syntax though

lib/decorateWithMetrics.js
kidroca: Some methods are `void` they don't return anything 
With `TaggedItem` we're trying to identify only actual data that uses memory

The check is a bit crude and should be updated to check only against `null` and `undefined` 

For example let say:
- you're in a chat with cleared stats (`resetMetrics` was just called)
- You switch to a different chat 
- You capture a snapshot in dev tools

The `TaggedItems` there would be pretty much how much data Onyx consumes for the 2nd chat - only Onyx data from storage, not rendered content and DOM stuff. We don't want to capture `TaggedItems` for void returns, though it would barely increase memory it would list more `TaggedItems` than the actual results returned from Onyx

lib/decorateWithMetrics.js
kidroca: Oki if you prefer that I can update it

kidroca:
> Sounds like this could be another reason why benchmarking inline with a higher order function that wraps any Promise could be useful. This way we would not need to have methods that return Promise, but could track any Promise whether returned by a function or not.

Benchmarking inline would not only include the benchmarking code in the source, but would also require you to change each place/method you want to track

The whole point for the decorator pattern was to avoid interfering with original logic (as much as possible) and to easily exclude the code from production builds. You can apply it to anything as long as it returns a promise with no changes to source. You can also easily define what you want to decorate - just `['get']` or perhaps `['get', 'set']`.

It seems you want Onyx methods to always be used like sync functions (even though originally there are a few that do return Promise). This can still be achieved by another decorator that can discard returned promises. This way you have the benefits of tracking and not introducing logic changes, while you still hide the returned promise

---

I've extract `Onyx.internal` to demonstrate how the pattern works, and what's needed to decorate private methods like `get` and `set`. 
For me the best approach is to convert `Onyx` to class - this way you can decorate any method without having to move it to a separate file. And you can still hide private methods from the public API

```js
const instance = new Onyx();

if (captureMetrics) {
  applyMetricsDecorator(instance);
}

applyPromiseHidingDecorator(instance);

const publicApi = _.pick(instance, [
    'connect',
    'disconnect',
    'set',
    'multiSet',
    'merge',
    'mergeCollection',
    'clear',
    'init',
    'registerLogger',
    'addToEvictionBlockList',
    'removeFromEvictionBlockList',
    'isSafeEvictionKey',
};

export default publicApi;
```

lib/decorateWithMetrics.js
```diff
@@ -0,0 +1,120 @@
+/**
+ * Each key is a method name and the value is an array of calls metadata
+ * @type {Map<String, Array<CallMeta>>}
+ */
+const stats = new Map();
```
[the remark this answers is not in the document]
marcaaron: I appreciate the preference (I wrote a blog post once about this data type 😂 ), but we can put personal bias aside to leave things uncomplicated and consistent for everyone else who has to work on this code. For context, this is the first usage out of all the JS our company has ever written.

lib/decorateWithMetrics.js
marcaaron: Thanks!

marcaaron:
>Benchmarking inline would not only include the benchmarking code in the source 

That's true, but my gut feeling is that this isn't a particularly important problem to solve, we're adding a very small amount of code to the source.

>but would also require you to change each place/method you want to track

That's also true, but we've had to change several things anyway to achieve the decorator pattern and we are still left with an incomplete picture as some methods do not return a `Promise`.

I think at the end of the day for me it comes down to which implementation would be easiest to understand and sometimes a less creative or tedious approach is better for that. But will let others share their thoughts.

lib/decorateWithMetrics.js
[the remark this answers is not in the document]
kidroca: Ok sure, thanks for the context

kidroca:
> > but would also require you to change each place/method you want to track
> 
> That's also true, but we've had to change several things anyway to achieve the decorator pattern and we are still left with an incomplete picture as some methods do not return a `Promise`.

These are just some draft changes to prove the concept. Yes I've moved code around, but in the end the code is pretty much the same - you can't say whether it uses benchmarking or not - how Onyx work is independent from benchmarking

The approach captures a full method execution - arguments, timings, return value. You have a clear idea of everything that is decorated because it's extracted in a single place - `applyDecorators`. Most of the essential sync methods are part of async flows so they are captured and any potential bottlenecks from them can be identified. 
Other valuable to meter methods like `connect` can be refactored to work by extracting async logic out

- Inlining benchmark logic adds code unrelated to the actual method purpose. 
- how many methods will have to be modified
- can you provide a sample of how that pattern looks/works
  - will it capture original method call parameters, or only what's used on the promise?
  - would it be possible to turn it off so that metrics aren't collected?

Even then methods like `connect` would have to be updated so that they return the inner promises they work with:

https://github.com/Expensify/react-native-onyx/blob/586c76e7b90dbbde051d7ec7adbc4d53b2d51cd1/lib/Onyx.js#L310-L338

- `Promise.all` won't be waited / timed correctly if it's not changed to `return Promise.all`
- `_.each(matchingKeys` won't be waited as well 
- I would extract this whole block to a separate function and update it to return a promise - then apply a decorator to it
- Generally not returning the inner promises is a bit of an antipattern - it won't make the code faster, but it can cover errors - I'm sure you've came across the error: "Unhandled promise rejection" - it makes tracking/debugging/testing promises harded

---

We need to identify which methods should be benchmarked. I thought the current decorated methods are just about it - we only need to add `connect`

tgolen (commented):
Hola!

I just wanted to pop in and check out the progress on this! I had seen that it was just a draft, so I didn't want to do too thorough of a review. I had some thoughts on this as well.

It seems like one of the original goals here was to use the decorator pattern so that there wouldn't need to be any changes to Onyx, but it looks like there are quite a few changes that need to be made to Onyx to get the decorator pattern to work, so I am not really seeing the value.

I think part of the idea was to not include some of the code in production also, but why is that a problem? 

I would like to see this simplified more and really pared down to the bare minimum that we need for benchmarking, even if it tightly couples the benchmarking code to Onyx.

This last one probably goes without saying (but I didn't see anything to the contrary) and that is that we should be sure to do the things listed in https://github.com/Expensify/Expensify.cash/issues/2762 one PR at a time. This could definitely be what you are already planning, and if so, this can serve as a good confirmation that you're on the right track!

kidroca:
> It seems like one of the original goals here was to use the decorator pattern so that there wouldn't need to be any changes to Onyx, but it looks like there are quite a few changes that need to be made to Onyx to get the decorator pattern to work

The only change made to Onyx are that 2 methods were updated to return a promise. The rest is just code being moved to a separate file and usages update e.g. calls to `get()` became `OnyxInternal.get()`.

What I'm suggesting is to move code and update usage - it does not change how it works, but allows enough flexibility to apply the decorator pattern

The rest of the changes in this PR are documentation, benchmark code and tests - you'll have them no matter what solution is accepted

> I think part of the idea was to not include some of the code in production also, but why is that a problem?

Including it is not that big of a problem - it will just be evaluated, stay in memory and never used.
Capturing metrics should be disabled in prod - I guess this can be applied for both solutions

> This last one probably goes without saying (but I didn't see anything to the contrary) and that is that we should be sure to do the things listed in Expensify/Expensify.cash#2762 one PR at a time.

Absolutely. 
We should use this draft to pick a benchmarking pattern and apply it. A separate PR would be opened (after), to apply the optimization. Do you need a more granular approach?

lib/Onyx.js
```diff
@@ -496,22 +486,23 @@ function applyMerge(key, data) {
  *
  * @param {string} key
  * @param {*} val
+ * @returns {Promise}
```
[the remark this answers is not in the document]
kidroca: On the other hand, the worst part for **not** returning a promise is the client of Onyx won't have a clue whether `merge` succeed or failed. This might be intended for E.cash, but I as a user would like to be free to handle the promise

kidroca (commented):
Updated and reverted most of the changes to Onyx 
Onyx.internal is not needed as I found a way to apply the decorator directly on Onyx original functions. It's a bit hacky but works

To capture `merge` "correctly" sans the sync calls I still need to extract the async logic to e separate function or maybe you can point me to track another function instead like `applyMerge`.

To capture `connect` I still need to extract the async logic I commented which in my previous comment.

The difference is now I can extract the logic in the same file

---

Here's how memory tracking works

The benchmarking code keeps method return values in a `WeakMap`
The key used is the original return value. This way when the return value
is no longer referenced the key and the value are garbage collected from the map
The value is our `TaggedItem` - this serves to easily find what's still retained 
It shallow copies the original item and stores a ref to it as well. This way we see actual memory retained by the item, and have a ref to the original so we can track what is causing it to be retained. 
The `TaggedItem` is referencing the original item directly, but since nothing is referencing the `TaggedItem` - we never expose it - it does not prevent the WeakMap from garbage collecting the original data and the `TaggedItem` that captured it

Here's a brief video 
 - I had to speed it up so I can put it here
 - The sizes are in bytes, so not much

https://user-images.githubusercontent.com/12156624/118322897-83c4a300-b508-11eb-8ffd-271ba641db75.mp4

About `WeakMap`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Keyed_collections#weakmap_object

lib/decorateWithMetrics.js
kidroca: I've posted more info here (2nd part): https://github.com/Expensify/react-native-onyx/pull/70#pullrequestreview-660153513
If you decide to keep this functionality I can add more details in the `README`

lib/Onyx.js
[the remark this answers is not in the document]
tgolen: > as a library user would like to be free to handle the promise

I think in this case, we would rather be opinionated about how Onyx is utilized, rather than just allowing for a lot of "use it however you want" scenarios. It doesn't feel like this is the right PR to be focused on how a consumer other than e.cash should behave. So, just to be clear, I'm not saying that we shouldn't have that convo, I just don't think this is the right place to have it.

If there is a way to get around not needing to return this Promise, I think we should do it. Though, if it's absolutely required, then I guess I would rather return a promise to have the benchmarking as opposed to nothing.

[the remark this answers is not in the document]
tgolen: Also, I see that we return promises from everything except `merge()`, so that weakens my argument tremendously already, and I don't think returning a promise here will have any significant impact.

lib/decorateWithMetrics.js
```diff
@@ -0,0 +1,132 @@
...
+            .catch(error => (result = error))
+            .finally(() => {
+                const endTime = performance.now();
+
+                if (!_.has(stats, alias)) {
+                    stats[alias] = [];
+                }
+
+                const methodStats = stats[alias];
+                methodStats.push({
```
tgolen, line 72: Looking at some of the output, I only found `methodName, args, duration` valuable. Also... what happened to duration?

tests/unit/decorateWithMetrics.js
```diff
@@ -0,0 +1,235 @@
...
+    resetMetrics
+} from '../../lib/decorateWithMetrics';
+import waitForPromisesToResolve from '../utils/waitForPromisesToResolve';
+
+describe('decorateWithMetrics', () => {
+    beforeEach(() => {
+        resetMetrics();
+    });
+
+    it('Should collect metrics for a single method, single call', () => {
```
tgolen, line 13: For unit tests, we like to have them contain comments in the following format:

```
// GIVEN <some initial condition>
// WHEN <an action happens>
// THEN <a result is expected>
```

I know some of the tests in this repo aren't very consistent with that, but we have found that providing comments in this style is extremely helpful to convey context and improve maintainability.

lib/decorateWithMetrics.js
```diff
@@ -0,0 +1,132 @@
...
+
+/**
+ * Each key is a method name and the value is an array of calls metadata
+ * @type {Record<String, Array<CallMeta>>}
+ */
+let stats = {};
+
+/**
+ * Class instances are easily tracked by Dev tools memory snapshots
+ * Wrapping a result with such instance allows us to track it without
```
tgolen, line 11: I'm not quite sure what this comment means or what it implies. What is the "extra weight"?

tgolen (changes requested):
I really like the cleanup of Onyx so that very little has changed there.

I'm still learning a lot about this. Weak maps and the decorator pattern are new to me. I think `decorateWithMetrics.js` is pretty clean, so I don't know what changes to suggest.

lib/Onyx.js
```diff
@@ -610,4 +618,25 @@ const Onyx = {
...
+    clear = decorate.decorateWithMetrics(get, 'Onyx:clear');
+    merge = decorate.decorateWithMetrics(get, 'Onyx:merge');
+    mergeCollection = decorate.decorateWithMetrics(get, 'Onyx:mergeCollection');
+    AsyncStorage.getAllKeys = decorate.decorateWithMetrics(AsyncStorage.getAllKeys, 'AsyncStorage:getAllKeys');
+    /* eslint-enable */
+
+    // Expose stats methods on Onyx
+    Onyx.getMetrics = decorate.getMetrics;
+    Onyx.resetMetrics = decorate.resetMetrics;
+    Onyx.isCapturingMetrics = true;
```
marcaaron, line 639: What does `isCapturingMetrics` do for us? If we have set `captureMetrics` to `true` then we should know that we are capturing metrics?

lib/decorateWithMetrics.js
```diff
@@ -0,0 +1,132 @@
...
+}
+
+/**
+ * Capture tags in a weak map
+ * When the original object (key) is no longer used anywhere in app - the data (TaggedItem)
+ * is free for garbage collection and removed
+ * Using dev tools we see which TaggedItems are still retained and what's using them - preventing GC
+ * @type {WeakMap<object, TaggedItem>}
+ */
+const taggedItems = new WeakMap();
```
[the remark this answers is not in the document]
marcaaron: Just to get clear on this, we are using `WeakMap` because it will _allow garbage collection_. And therefore reveal to us which `TaggedItem` no longer have references. Why does this matter?

lib/decorateWithMetrics.js
```diff
@@ -0,0 +1,132 @@
...
+
+    function decorated(...args) {
+        const startTime = performance.now();
+        let result;
+
+        const originalReturnValue = func.apply(this, args);
+
+        /*
+        * Promisify the original return value in case the decorated method does not return a promise
+        * Then handlers added here are not affecting the returned promise as they aren't attached to it
```
marcaaron, line 51: I think it would be more straightforward to skip timing here since there's nothing to time.

marcaaron: Another option would be to throw when `originalReturnValue` is not returning a `Promise` and enforce that the decorated function return one.

lib/decorateWithMetrics.js
marcaaron: +1

Except `result` seems valuable since if there are outliers then you might want to consider the data being returned when investigating why something is taking longer than expected.

lib/decorateWithMetrics.js
```diff
@@ -0,0 +1,132 @@
...
+ * By default clears all collected metrics, pass a methodName to only clear the metrics
+ * for the given method
+ * @param {String} [methodName]
+ */
+function resetMetrics(methodName) {
+    if (methodName) {
+        delete stats[methodName];
+    } else {
+        stats = {};
+    }
```
marcaaron, line 116:
```suggestion
    if (methodName) {
        delete stats[methodName];
        return;
    }
    
    stats = {};
```

marcaaron: We use early returns a lot in our code.

marcaaron (changes requested):
This is looking cleaner to me, thanks! Had a few more questions, comments, suggestions.

lib/Onyx.js
marcaaron: * Not a blocker

lib/decorateWithMetrics.js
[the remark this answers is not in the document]
kidroca: It's actually the opposite - it would reveal to us which `TaggedItems` are still retained and allow us to see what is retaining them

A `TaggedItem` would only be retained if the `ref` - the original data is still used somewhere

We can find retained `TaggedItems` in a heap snapshot and then on the **Retainers** section we'll can track what is retaining a specific `TaggedItem` all the way to the component that is using the original (tagged) data

Let's say we see a `TaggedItem` that is retaining a huge chunk of memory - we can track what's causing this

lib/decorateWithMetrics.js
kidroca: Oki, I just wanted to show that it's possible to make this more universal and work for sync methods as well

lib/decorateWithMetrics.js
kidroca: I didn't want to include any computation that can be avoided here
We're just capturing the raw data and then when we actually need to display it it's mapped to take a better shape

On my sample benchmark I use a method like this:

```js
    printMetrics() {
        const data = Onyx.getMetrics().map(call => ({
            method: call.methodName,
            startTime: Math.round(call.startTime),
            endTime: Math.round(call.endTime),
            duration: Math.round(call.endTime - call.startTime),
            args: JSON.stringify(call.args),
            result: JSON.stringify(call.result),
        }));

        // eslint-disable-next-line no-console
        console.table(data);

        Onyx.resetMetrics();
    }
```

https://github.com/kidroca/Expensify.cash/blob/3baa89133035429fdc13424ca2d8a5d67ab435c9/src/libs/Navigation/NavigationRoot.js#L40-L54

---

The times actually have nanosecond precision and they are lengthy float (double) point numbers like `3003.0000141` the value before the dot are milliseconds.

tests/unit/decorateWithMetrics.js
kidroca: I think I can apply it like this:

// GIVEN a mock function that returns a mock object
// WHEN it is decorated and executed
// THEN calling `getMetrics` should return data about the particular call

Or 

// GIVEN a mock function, returning a mock object that is decorated
// WHEN it is called with a parameter
// THEN  calling `getMetrics` should return data about the particular call

kidroca (commented):
I've actually hooked up to this to E.cash - it prints benchmark data on each route change, but have added a wrong link in the PR description:

It's on my fork here: https://github.com/kidroca/Expensify.cash/tree/kidroca/sample-onyx-benchmark - you can see stats each time you switch a chat
To see the `console.table` data you need to run in `debug` so that the Chrome/Safari debugger is launched

lib/decorateWithMetrics.js
kidroca: It also covers the case where `merge` would not always return a promise - in case the Promise handling changes in `merge` get reverted

lib/Onyx.js
marcaaron: >This way we don't have to import and test for the ENV variable that sets this flag again

seems reasonable to use an env variable

lib/decorateWithMetrics.js
marcaaron: >Oki, I just wanted to show that it's possible to make this more universal and work for sync methods as well

But we are not measuring any sync methods

>It also covers the case where merge would not always return a promise - in case the Promise handling changes in merge get reverted

If we are worried about behavior changing then we can write a test.

lib/Onyx.js
marcaaron: This is a NAB in any case, but perhaps we should initialize Onyx so that `Onyx.isCapturingMetrics` returns `false`.

lib/decorateWithMetrics.js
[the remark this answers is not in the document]
marcaaron: >it would reveal to us which TaggedItems are still retained and allow us to see what is retaining them

Sorry, can you add all of this context in the code as a comment or possibly a guide somewhere? I'm not sure how or why any of this works, why we want it, etc and if we agree it's useful then it should be properly documented. 

>Let's say we see a TaggedItem that is retaining a huge chunk of memory - we can track what's causing this

That's interesting. But this isn't the problem that we are trying to solve.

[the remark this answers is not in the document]
kidroca: > > it would reveal to us which TaggedItems are still retained and allow us to see what is retaining them
> 
> Sorry, can you add all of this context in the code as a comment or possibly a guide somewhere? I'm not sure how or why any of this works, why we want it, etc and if we agree it's useful then it should be properly documented.

I've made a demo and explained how that works in the top level comment when I pushed the changes: https://github.com/Expensify/react-native-onyx/pull/70#pullrequestreview-660153513



> > Let's say we see a TaggedItem that is retaining a huge chunk of memory - we can track what's causing this
> 
> That's interesting. But this isn't the problem that we are trying to solve.

Ok then there's no point documenting and spending more time over this, I'll just remove memory tracking

lib/decorateWithMetrics.js
kidroca: > > Oki, I just wanted to show that it's possible to make this more universal and work for sync methods as well
> 
> But we are not measuring any sync methods?

We're not measuring any sync methods because:
1. Before this change it wasn't possible
2. You haven't said explicitly what you need tested

Now any sync method can be added and tested to see its effects on the bigger picture, if a sync method is taking more than 16ms it's a tell for a performance issue

kidroca: > > It also covers the case where merge would not always return a promise - in case the Promise handling changes in merge get reverted
> 
> If we are worried about behavior changing then we can write a test.

I was talking specifically for my current changes to `merge` if I revert them and the method is sometimes `void` sometimes returns a promise it won't be able to work without the above handling.
It's still not clear whether my changes to `merge` would stay or I need to find a different way to measure `merge` or not measure it at all

lib/Onyx.js
kidroca: > > This way we don't have to import and test for the ENV variable that sets this flag again
> 
> seems reasonable to use an env variable

It's not very easy to find out during debugging - you can't check `process.env.ONYX_BENCHMARK` - it's not set there
It's set at the end of the decoration - you know it passed successfully when this is set to `true` 



> This is a NAB in any case, but perhaps we should initialize Onyx so that `Onyx.isCapturingMetrics` returns `false`.

I don't know what you mean by that, we do initialize to to `false` by default: `captureMetrics = false,`
It's only `true` when we want benchmarking

marcaaron: I'm suggesting we explicitly set the value to `false`, but again, it's not a blocker. I get that this `Onyx.isCapturingMetrics` will return `undefined` which is falsy, but maybe more clear if it was always a bool.

marcaaron: >It's not very easy to find out during debugging

Again not a blocker, but we import env variables to the `Config` object so it is very easy (at least for us) to find out 😂

lib/decorateWithMetrics.js
marcaaron: >You haven't said explicitly what needs benchmarking

I'm just reviewing this PR 🤷‍♂️ 

But my assumption is that we were mainly concerned with benchmarking the performance of `AsyncStorage` so yeah, I'm confused by the discussion about async/sync methods.

lib/decorateWithMetrics.js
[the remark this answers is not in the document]
marcaaron: Ok, I might be confused, but to clarify, is this something we agreed to do or not? If it is, then I'd suggest including documentation so people will know how to use it. If we don't need/want it, then yes, I'd remove it. You can always ask in the Slack channel for feedback.

lib/decorateWithMetrics.js
kidroca: Ok, so we don't know exactly what needs benchmarking but this change allows to benchmark any method, how's that a bad thing?

kidroca:
Updated

Extracted the async part of  `connect` to a separate method so it can be tracked by the benchmark and addressed most of the threads

For the rest we need to clear out some things 
- what Onyx methods should be benchmarked? e.g. `get, set, merge, connect`
- should I revert my changes to `merge`, `connect` (Then we'll benchmark just the base `get/set` methods)?
- should we track memory usage for data returned by Onyx? I've tried to explain how that works here: https://github.com/Expensify/react-native-onyx/pull/70#pullrequestreview-660153513. I can add this information to the README, or remove memory tracking
- anything else needs changing?

marcaaron:
After syncing with @tgolen today I'm thinking we should agree to:

1. Figure out exactly what needs benchmarking
2. Implement those things _specifically_ no more no less
3. Move onto the caching stuff

We took a stab at the two major reasons why benchmarking is valuable for us today and came up with...

**Provide a way to test/prove that implementing an in-memory cache will decrease the overall time spent reading/writing to Onyx by a significant amount**

(and I think could be used again if we choose to make improvements to `AsyncStorage`)

This PR seems to do that, but maybe we can simplify it by just focusing on a few methods and trim down some of the extra info for now. The timing stuff is the main thing we care about and the biggies should tell us most of what we need to know. Those would probably be:

`get(), set(), multiSet(), merge(), mergeCollection()`

**Provide a way to test/prove that implementing an in-memory cache that passes references (as opposed to copies) will decrease the total memory footprint of data sent to all subscribers by a significant amount**

I'm not sure we've really landed on a good solution for this yet (or I might still not be understanding what you're going for here). I appreciate all the extra info about `WeakMap`, retained objects, etc. It just seems like it could be helpful for other reasons (e.g. detecting a memory leak related to Onyx). But not sure if I'm getting how it will help us look at how much memory all of the subscribers are using.

Would it be possible to perhaps focus on the memory tracking for subscribers and create a simple Onyx method that dumps the total amount of memory all subscribers are currently using? Something like that sounds pretty valuable.

lib/decorateWithMetrics.js
marcaaron: Left a comment, but I'd rather discuss how to move forward

kidroca:
> **Provide a way to test/prove that implementing an in-memory cache will decrease the overall time spent reading/writing to Onyx by a significant amount**
> 
> This PR seems to do that, but maybe we can simplify it by just focusing on a few methods and trim down some of the extra info for now. The timing stuff is the main thing we care about and the biggies should tell us most of what we need to know. Those would probably be:
> 
> `get(), set(), multiSet(), merge(), mergeCollection()`

Ok, so I should probably revert my changes to `connect` and this method time can be somewhat represented by the separate `get` calls that happen

We should probably include `getAllKeys` as well as it turned out it's something that we've missed
I would like to extract a private `getAllKeys` method for Onyx that just delegates to `AsyncStorage.getAllKeys()`
We'll have to do that anyway if we want to apply cache to that method as well

> Provide a way to test/prove that implementing an in-memory cache that passes references (as opposed to copies) will decrease the total memory footprint of data sent to all subscribers by a significant amount

It will decrease it significantly but not noticeably
For example the raw Onyx data with my biggest chat open is about 18kb updating `get` to use cache instead of creating a new object from string would let's say cut this by 50%. This doesn't matter much to the view components that will use the data and render visual content e.g. rendering an image would still need let's say a MB of memory
The performance gains are mostly from reduced CPU time and reduced latency

> Would it be possible to perhaps focus on the memory tracking for subscribers and create a simple Onyx method that dumps the total amount of memory all subscribers are currently using? Something like that sounds pretty valuable.

The current solution displays how much raw Onyx data is still in memory (that all subscribers are currently using), and you can directly trace it to where it is used. You can see the total amount in dev tools but there's no way to print it

There are libraries that can approximate how much memory an object is using by traversing it - passing `callbackToStateMapping` to such library should be able to print the size on the console. There are no native JS APIs to tell you how much memory an object like `callbackToStateMapping` is using.

Printing the size on the console, can raise attention to a memory problem, but won't be able to pinpoint it

kidroca:
> Provide a way to test/prove that implementing an in-memory cache that passes references (as opposed to copies) will decrease the total memory footprint of data sent to all subscribers by a significant amount

Another possibility is to calculate memory after the cache solution is implemented.
We will have a `cache` object - we can print it's size (using a library)

To verify objects are coming from cache and not copied you can again trace them from
dev tools, but without having to keep a `WeakMap` and `TaggedItem`

tgolen:
> We should probably include getAllKeys as well

Yeah, I like including this in the list too.

> The performance gains are mostly from reduced CPU time and reduced latency

OK, that's a good perspective. So, in that case, I would change the  requirement to be:

**Provide a way to test/prove that implementing an in-memory cache that passes references (as opposed to copies) will decrease the CPU usage and latency of data sent to all subscribers by a significant amount**

kidroca:
> OK, that's a good perspective. So, in that case, I would change the requirement to be:
> 
> **Provide a way to test/prove that implementing an in-memory cache that passes references (as opposed to copies) will decrease the CPU usage and latency of data sent to all subscribers by a significant amount**

I don't think there's a way to capture CPU load/usage during method execution beside profiling the app using dev tools
Especially when a big part of the execution happens in native
We can only measure the time it took to complete the call, like we already do
With cache we'll be able to see that the time is significantly less for repeated calls as the round trip to native will be skipped

---

In order to proceed I need the items in my previous comment addressed:
- [ ] what Onyx methods should be benchmarked - I think this is almost cleared up: get, set, multiSet, merge, mergeCollection, getAllKeys  - it's down to whether we should measure `connect` - keep my changes there, or not measure it - drop the changes
- [ ] should we track memory usage - also kind of addressed by the above, but no one explicitly said to remove memory tracking
- [ ] anything else - like there's a thread about whether we should capture `result` or not: https://github.com/Expensify/react-native-onyx/pull/70/files#r633730659

tgolen:
>  it's down to whether we should measure connect

Let's not worry about connect for now.

> should we track memory usage 

Let's not focus on this for now either, since we are lacking a clear problem statement.

> capturing `result`

I personally don't think we need to capture this to begin with, I wasn't quite convinced with the argument for keeping it, but I don't feel strongly.

> anything else

I think one small thing I'd like to see in addition to all this is to have the method for interpreting the data be very clear and straightforward. Right now, the solution provides kind of "building blocks" and then you had a personal code snippet that you were running from the JS console. I'd actually like to see that code snippet put right into the code because 1) it's a good snippet and 2) it makes it easy for anyone else to come along and use it.

I think that covers everything? Might let @marcaaron weigh in on this too.

kidroca:
Updated

Addressed all the points in @marcaaron's comment 
Added the snippet requested by @tgolen  as `Onyx.printMetrics` + human friendly duration

This is an example output:

```
Onyx Benchmark
  Total:  135.9min
  Average:  45.3min

Onyx:get
  total: 51.5min
  max: 6.20sec
  min: 26.650ms
  avg: 4.24sec 
  calls: 728

Onyx:getAllKeys
  total: 84.3min
  max: 6.20sec
  min: 8.210ms
  avg: 3.78sec 
  calls: 1336

Onyx:set
  total: 10.09sec
  max: 5.01sec
  min: 43.125ms
  avg: 2.02sec 
  calls: 5
```

_Total time is lot `135.9min` because there are a lot of calls happening in parallel_ 

https://user-images.githubusercontent.com/12156624/119202098-ce11cb00-ba98-11eb-80b8-86b51d0edcf7.mp4

@tgolen 
> and then you had a personal code snippet that you were running from the JS console

I didn't run it through the console, I've added some code to E.cash to print statistics each time navigation state changed:

#### NavigationRoot.js
```js
    parseAndStoreRoute(state) {
        if (!state) {
            return;
        }

        const path = getPathFromState(state, linkingConfig.config);
        setCurrentURL(path);

        // Added this (previously it was a method on this component)
        if (Onyx.printMetrics) {
            Onyx.printMetrics();
            Onyx.resetMetrics();
        }
    }
```

@marcaaron I've removed `Onyx.isCapturingMetrics` as you can see above, if Onyx is capturing metrics this method will be defined

kidroca:
Updated

Resolved merge conflicts

marcaaron (commented):
Code looking good, thanks for the changes! 

Hooked this up on the web side and doesn't seem to be working with `mergeCollection()` any ideas why?

kidroca:
> Hooked this up on the web side and doesn't seem to be working with `mergeCollection()` any ideas why?

Yes I've just added a fix:

The original non decorated methods are instantly added to `Onyx` here:

```js
const Onyx = {
    connect,
    disconnect,
    set,
    multiSet,
    merge,
    mergeCollection,
    clear,
    init,
    registerLogger,
    addToEvictionBlockList,
    removeFromEvictionBlockList,
    isSafeEvictionKey,
};
```

This happens before `init` is called and the decorator applied

So we need to re-expose them like so

#### In applyDecorators()
```js
    // Re-expose decorated methods
    Onyx.set = set;
    Onyx.multiSet = multiSet;
    Onyx.clear = clear;
    Onyx.merge = merge;
    Onyx.mergeCollection = mergeCollection;
```

lib/Onyx.js
```diff
@@ -40,6 +39,14 @@ function get(key) {
         .catch(err => logInfo(`Unable to get item from persistent storage. Key: ${key} Error: ${err}`));
 }
 
+/**
+ * Returns current key names stored in persisted storage
+ * @returns {Array<String>}
```
tgolen, line 44: I think this returns a promise, right?

lib/decorateWithMetrics.js
```diff
@@ -0,0 +1,177 @@
...
+
+/**
+ * @typedef {Object} StatSummary
+ * @property {string} methodName
+ * @property {number} minTime - the quickest execution for this method
+ * @property {number} maxTime - the longest execution for this method
+ * @property {number} averageTime - average execution time for the method
+ * @property {number} totalTime - sum of the duration for all the calls
+ * @property {Array<CallMeta>} calls - a collection of all the calls for this method
+ */
```
tgolen, line 170: What are all these docs down here for? I think I would just remove them.

tgolen (changes requested):

lib/decorateWithMetrics.js
kidroca: I needed these while I was writing tests, I don't need them any more, will remove

---

These help with the editor's intellisense:

#### Hovering summaries
image

#### Aftering clicking StatSummary
image

#### And after clicking CallMeta
image

```js
/**
 * Returns total, average time and all captured stats mapped under
 * summaries.methodName -> StatSummary
 * @returns {{averageTime: number, summaries: Record<string, StatSummary>, totalTime: number}}
 */
```

The same can be achieved by making a separate file defining the public interface, libraries tend to have them
Here's the one for AsyncStorage: https://github.com/react-native-async-storage/async-storage/blob/master/types/index.d.ts

kidroca:
Updated with requested changes

lib/decorateWithMetrics.js
tgolen: Ah, OK. I learned something new!

tgolen (approved):

lib/decorateWithMetrics.js
```diff
@@ -0,0 +1,158 @@
...
+
+    function decorated(...args) {
+        const startTime = performance.now();
+
+        const originalPromise = func.apply(this, args);
+
+        /*
+        * Then handlers added here are not affecting the original promise
+        * They crate a separate chain that's not exposed (returned) to the original caller
+        * */
```
marcaaron, line 30:
```suggestion
        // The handlers added here are not affecting the original promise
        // They create a separate chain that's not exposed (returned) to the original caller
```

marcaaron (approved):
LGTM just had one NAB for a typo all you @Gonals

Gonals (approved):
This looks good!

PR #76, opened 2021-05-25
Feature: Onyx Cache

### LINKED ISSUE #63: Improvements for Onyx connect
Right now Onyx works fine on iOS
On Android though not so much, we can blame it on AsyncStorage, but there are
ways to improve the usage of AsyncStorage

## Querying `AsyncStorage`
- `withOnyx` receives a key -> val mapping
- each key is connected separately to Onyx - `this.connectMappingToOnyx(mapping, propertyName)` -> new connection `Onyx.connect` 
- each key mapping can result in one or more `matchingKeys`
- each key in `matchingKeys` makes a separate call to `AsyncStorage.getItem` `get(key) -> AsyncStorage.getItem(key)`
  - BTW any exception in `AsyncStorage.getItem` is swallowed and end users of the Onyx library won't be able to intercept and handle errors themselves.

Now all of the above happens for a single `withOnyx` usage e.g. One component wrapped with Onyx. 
In the simplest scenario each prop requested from Onyx would make a separate call to `AsyncStorage.getItem`

So let's say 10 components use data from Onyx and each reads 2 props from Onyx. That's a total of 20 `getItem` calls.
In practice the calls would be a lot more though and this happens during startup

### What happens under the hood 
**every `getItem` call would make a new SQL query**

*The rest is just proving that statement - you can skip to Improvements if you trust me 😉*

Async Storage for Android uses SQLite internally
- [AsyncStorage -> resolves to android native module](https://github.com/react-native-async-storage/async-storage/blob/aca9af97cc63dc258426aa2d223474f59aac3920/src/RCTAsyncStorage.js#L6)
- [AsyncStorage | android -> uses ReactDatabaseSupplier](https://github.com/react-native-async-storage/async-storage/blob/75c571b7b1265e7d5599e0b9e38e19acb6004181/android/src/main/java/com/reactnativecommunity/asyncstorage/AsyncStorageModule.java#L46)
- [ReactDatabaseSupplier -> uses SQLiteOpenHelper](https://github.com/facebook/react-native/blob/6e6443afd04a847ef23fb6254a84e48c70b45896/ReactAndroid/src/main/java/com/facebook/react/modules/storage/ReactDatabaseSupplier.java#L22)
- [SQLiteOpenHelper -> provides an api over SQLite Databases](https://developer.android.com/reference/android/database/sqlite/SQLiteOpenHelper)
	
Tracing this down you can see that every `getItem` call would make a new SQL query
- [getItem -> RTCAsyncStorage.multiGet](https://github.com/react-native-async-storage/async-storage/blob/a6957ba70d80b1610ad8ce940ab9427b7cfe4534/src/AsyncStorage.native.js#L119)
- [multiGet -> mReactDatabaseSupplier.get().query(...)](https://github.com/react-native-async-storage/async-storage/blob/75c571b7b1265e7d5599e0b9e38e19acb6004181/android/src/main/java/com/reactnativecommunity/asyncstorage/AsyncStorageModule.java#L132)
	
### Using `multiGet`
The easiest improvement would be to use the optimization provider from `AsyncStorage` - [AsyncStorage.multiGet](https://react-native-async-storage.github.io/async-storage/docs/api#multiget)

At least for the code here:
https://github.com/Expensify/react-native-onyx/blob/121c521fe72f37c2b1fb79c39755814558eca27c/lib/Onyx.js#L320-L324
- instead of creating separate promises fetch all keys using `multiGet`

### Batch connect calls keys
The next best thing:
Don't call AsyncStorage as soon as `Onyx.connect` is called. Wait for a brief time or debounce the execution of AsyncStorage to run after all the calls to `Onyx.connect`. Make a batch and each time `Onyx.connect` is called add to that batch, after the debounce period expires (lets say 5ms) fetch all the keys using `multiGet` and respond to the listeners.

When all components are initially mounting (App Startup) this should batch all the necessary keys and make just one query to AsyncStorage. It would also remove duplicate queries when multiple components use the same Onyx KEY

### In Memory Proxy
The next next best thing:
Don't always read from file storage, make an "in memory proxy" and provide recently accessed keys
Let's say there are 5 components subscribing to `CURRENTLY_VIEWED_REPORTID` each of them would retrieve the value
from file storage. Instead only the initial call should read from file and return the value to the rest

## Note on Onyx.set
Onyx.set does not resolve optimistically, it would first wait to perist to file
so Onyx set would not imeddiately update listeners, but only after write, if it's successful
 - `AsyncStorage.setItem` -> `keyChanged(key, val)` 

 You could make the change immediately. If it happens to fail you can retry in the background or handle it some other way
 E.g. the value of the `CURRENTLY_VIEWED_REPORTID` updates let listeners know that right away no matter what

kidroca:
@marcaaron, @tgolen 

### Details

Added a new file and defined `OnyxCache` class to encapsulate cache related methods and have some reuse
Update Onyx methods to use cache
Updated Onyx methods to deal with concurrent method calls while data is still being read
Added tests

### Related Issues

GH_LINK
Fixes #63
https://github.com/Expensify/Expensify.cash/issues/2762

### Automated Tests

Added test annotated with GIVEN, WHEN, THEN. They should be pretty self explanatory
Covering `OnyxCache` and the integration in Onyx and AsyncStorage calls


tests/unit/onyxMetricsDecorationTest.js
```diff
@@ -0,0 +1,80 @@
+import waitForPromisesToResolve from '../utils/waitForPromisesToResolve';
+
+describe('Onyx', () => {
+    describe('Metrics Capturing Decoration', () => {
```
kidroca, line 4: Moved these tests to separate file

tests/unit/withOnyxTest.js
```diff
@@ -25,7 +25,7 @@ describe('withOnyx', () => {
     it('should render with the test data when using withOnyx', () => {
         let result;
 
-        Onyx.set(ONYX_KEYS.TEST_KEY, 'test1')
+        return Onyx.set(ONYX_KEYS.TEST_KEY, 'test1')
```
kidroca, line 28: There are some tests here that need to be double checked
This one didn't return the promise so it always passed successfully

tests/unit/withOnyxTest.js
```diff
@@ -38,12 +38,12 @@ describe('withOnyx', () => {
             })
             .then(() => {
                 const textComponent = result.getByText('test1');
-                expect(textComponent).toBeTruthy();
-                expect(result).toHaveBeenCalledTimes(999);
+                expect(textComponent).not.toBeNull();
```
kidroca, line 41: And then started to fail after it was fixed. I guess `toHaveBeenCalledTimes(999)` was some leftover...

lib/Onyx.js
```diff
@@ -34,17 +35,42 @@ let defaultKeyStates = {};
...
-        .then(val => JSON.parse(val))
-        .catch(err => logInfo(`Unable to get item from persistent storage. Key: ${key} Error: ${err}`));
+    // When we already have the value in cache - resolve right away
+    if (cache.hasCacheForKey(key)) {
+        return Promise.resolve(cache.getValue(key));
+    }
+
+    /* Otherwise setup a task to retrieve it.
+       This ensures concurrent usages would not start the same task */
+    return cache.resolveTask(`get:${key}`, () => AsyncStorage.getItem(key)
```
kidroca, line 45: I guess this method and the description above can have a better name

The idea is the task (2nd parameter) would only be invoked if it needs to be started
E.g. the first call to get `session` would trigger the task and capture it (`get:session`)
While this `session` task is still running other calls arrive, they aren't resolved from cache
as the cache isn't filled yet, but they won't start another task for the same thing
they'll just be hooked to the existing call

I think the method documentation of `cache.resolveTask` describe this well, it's just that
the method name should changed to something like `startTaskIfNecessary` 
And the comment here is a bit confusing
`This ensures concurrent usages would not start the same task`
vs
`This ensures concurrent calls would not start the same task over`

lib/Onyx.js
```diff
@@ -414,12 +445,28 @@ function evictStorageAndRetry(error, ionMethod, ...args) {
...
+ * AsyncStorage expects array like: [["@MyApp_user", "value_1"], ["@MyApp_key", "value_2"]]
+ * This method transforms an object like {'@MyApp_user': 'myUserValue', '@MyApp_key': 'myKeyValue'}
+ * to an array of key-value pairs in the above format
+ * @param {Record<string, *>} data
+ * @return {Array<[string, string]>} an array of stringified key - value pairs
+ */
+function prepareKeyValuePairsForStorage(data) {
+    return _.keys(data).map(key => [key, JSON.stringify(data[key])]);
+}
+
```
kidroca, line 469: Extracted this from `multiSet` for reuse (in `mergeCollection`)

lib/Onyx.js
```diff
@@ -551,34 +594,31 @@ function mergeCollection(collectionKey, collection) {
...
-    const newCollection = {};
     return getAllKeys()
-        .then((keys) => {
-            _.each(collection, (data, dataKey) => {
-                if (keys.includes(dataKey)) {
-                    existingKeyCollection[dataKey] = data;
-                } else {
-                    newCollection[dataKey] = data;
-                }
-            });
```
kidroca, line 564: Refactored this part a bit.

lib/Onyx.js
```diff
@@ -551,34 +594,31 @@ function mergeCollection(collectionKey, collection) {
...
 
+            // Capture (non-stringified) changes to cache
+            cache.merge(_.pairs(collection));
+
+            // Optimistically inform collection subscribers
+            keysChanged(collectionKey, collection);
+
             return Promise.all([existingCollectionPromise, newCollectionPromise])
-                .then(() => keysChanged(collectionKey, collection))
                 .catch(error => evictStorageAndRetry(error, mergeCollection, collection));
```
kidroca, line 622: Shouldn't we also call `keysChanged` for the individual collection items

E.g. collection is just the collection name 
image

It seems `mergeCollection` was called to merge new content to all these collections but no subscriber would be notified for that
image

Isn't it possible to subscribe to a particular collection item like `report_71402263` ?
I think currently it doesn't matter because when a collection is updated we're usually already have the updated data rendered, but this might not always be the case

kidroca (commented):
There are still some things to complete 
Like clearing cached keys when there are no subscribers for a given key
Some existing tests started to fail after the update, I'll take a look tomorrow

marcaaron (commented):
Just leaving some initial comments since it seems like you are still working through stuff.

lib/Onyx.js
```diff
@@ -428,17 +475,15 @@ function set(key, val) {
...
+
+    // Capture (non-stringified) changes to cache
+    cache.merge(_.pairs(data));
+
+    // Optimistically inform subscribers
+    _.each(data, (val, key) => keyChanged(key, val));
 
     return AsyncStorage.multiSet(keyValuePairs)
-        .then(() => _.each(data, (val, key) => keyChanged(key, val)))
         .catch(error => evictStorageAndRetry(error, multiSet, data));
```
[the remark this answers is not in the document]
kidroca: > Not sure how we want to handle this. But if `evictStorageAndRetry()` is called then once it's done removing stuff it will call `multiSet()` again and also update the cache + call `keyChanged()` again.

Calling subscribers with the same data again isn't a big problem if any, at most they will re-render

But I see a bigger issue: This means that there is an existing concurrency issue even before my changes
1. multiSet is called to write some data
2. multiSet fails and schedules a retry - evictStorageAndRetry
3. another call to multiSet arrives before the retry is performed. It sets a new value for some of the keys
4. retry runs and overwrites the new value with the old value

The above is true for every method that currently uses `evictStorageAndRetry` to set up a retry

I've talked about a write queue - It would be best to implement one, but that's a separate issue

Methods that persist data to file should not write directly but push a task to a queue and ask the "queue manager" to flush the queue - if a flush is not running it will start. Flush will use the queue to write data to file - FIFO and handle any necessary retires. In the mean cache would always provide the most up to date data since it doesn't have to wait for the thing to be written to disk 

Furthermore having such a queue would allow you to group and set multiple keys at once (there is still some bridge flooding due to many writes running at the same time)
It would also allow you to skip outdated writes. Let's say something is updated very often in short intervals of time e.g. key "A" value "V" then again key "A" value "Val" and finally key "A" value "Value" - you don't need to first persist "V" then "Val" you can directly persist the latest available value - "Value"

lib/Onyx.js
kidroca: > > Shouldn't we also call keysChanged for the individual collection items
> > E.g. collectionKey is just the collection name
> 
> Correct, but we loop over all of the possible subscribers to find which ones need to be updated.

I see that now, I didn't know that `keysChanged` does that internally

lib/OnyxCache.js
```diff
@@ -0,0 +1,137 @@
...
+        return this.storageMap[key];
+    }
+
+    /**
+     * Check whether cache has data for the given key
+     * @param {string} key
+     * @returns {boolean}
+     */
+    hasCacheForKey(key) {
+        return key in this.storageMap;
```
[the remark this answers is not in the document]
kidroca: `null` is considered a valid value, e.g. if something is deliberately set to `null` - we'll cache that
`undefined` is not expected though, so I'll update this check to `isDefined(this.storageMap[key])` 

E.g. the difference is subtle:
- `undefined` the key was never set
- `null` the key was deliberately set to `null`

lib/OnyxCache.js
```diff
@@ -0,0 +1,137 @@
...
+     * Update or add data to cache
+     * @param {string} key
+     * @param {*} [value]
+     */
+    update(key, value) {
+        // Update all storage keys
+        this.storageKeys.add(key);
+
+        // When value is provided update general cache as well
+        if (isDefined(value)) {
```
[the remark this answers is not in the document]
kidroca: The 2nd parameter to `update(key, value)` is optional
This way the method works for just setting some available keys, without their values

When the 2nd parameter is not provided (`undefined`) we'll just capture a storage key - this happens in `Onyx.getAllKeys`
When the 2nd parameter is anything else will capture the value as well

lib/OnyxCache.js
```diff
@@ -0,0 +1,137 @@
...
+        delete this.storageMap[key];
+    }
+
+    /**
+     * Merge data to cache, any non existing keys will be crated
+     * @param {Array<[string, Object]>} pairs - array of key value pairs
+     */
+    merge(pairs) {
+        _.forEach(pairs, ([key, value]) => {
+            const currentValue = _.result(this.storageMap, key, {});
```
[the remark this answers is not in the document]
kidroca: Good points

I can add a check and do different handling:
- if the new value is different type from the old value - the new value overwrites the old
- for primitives like number and string - the new value overwrites the old
- for arrays - old and new are concatenated
- for objects - old and new are merged

lib/OnyxCache.js
```diff
@@ -0,0 +1,137 @@
...
+
+            if (_.isObject(value) || _.isArray(value)) {
+                const merged = {...currentValue, ...value};
```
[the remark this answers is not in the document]
kidroca: This will be changed added a comment on the thread above

lib/OnyxCache.js
```diff
@@ -0,0 +1,137 @@
...
+                this.update(key, merged);
+                return;
+            }
+
+            throw new Error(`The provided merge value is invalid: ${value}`);
```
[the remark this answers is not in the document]
kidroca: This will be changed added a comment on the thread above

kidroca (commented):
Thanks for the review @marcaaron 
Added some replies

kidroca:
Updated 
I've left some threads open for potential discussions 
The only thing remaining otherwise is a 26 of the test I've written for the cache use `async/await` syntax I find the tests more readable that way, but can convert them to promise chains if you want to

marcaaron:
>the test I've written for the cache use async/await syntax I find the tests more readable that way, but can convert them to promise chains if you want to

I believe that using async/await in tests would also be more readable, but this is more of a style guide issue than a personal preference. We've collectively agreed to not use it for now so I'd suggest either

1. Propose we begin using this syntax in the Slack channel and if everyone agrees feel free to use it + update the style guide

OR

2. Take the path of least resistance and change remove the usages

marcaaron:
That said, I often write tests using async/await and then convert them after so it's probably a good indicator 😄

lib/Onyx.js
```diff
@@ -367,7 +371,14 @@ function disconnect(connectionID, keyToRemoveFromEvictionBlocklist) {
         removeFromEvictionBlockList(keyToRemoveFromEvictionBlocklist, connectionID);
     }
 
+    const key = callbackToStateMapping[connectionID].key;
     delete callbackToStateMapping[connectionID];
+
+    // When the last subscriber disconnects, drop cache as well
+    const remainingConnectionsForKey = _.filter(callbackToStateMapping, {key});
+    if (remainingConnectionsForKey.length === 0) {
+        cache.remove(key);
```
marcaaron, line 380: Separately, doesn't this sort of invalidate at least some of the usefulness of having the cache. It seems like we are only optimizing for the case of multiple subscribers.

e.g. if it takes some non trivial amount of time to retrieve something like a large report from `AsyncStorage` if we remove it the cache once we are no longer subscribed that means we will have to incur the same cost again when connecting again.

Another option would be to clear the cache key based on some other criteria like no subscribers and also least recently subscribed. But not sure if we need this optimization at all. What is the worst that can happen if the cache entry is not removed at all and the cache is allowed to grow to the size of storage?

lib/Onyx.js
```diff
@@ -414,12 +445,28 @@ function evictStorageAndRetry(error, ionMethod, ...args) {
  * @returns {Promise}
  */
 function set(key, val) {
+    // Adds the key to cache when it's not available
+    cache.update(key, val);
```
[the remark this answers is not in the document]
marcaaron: When I see "update" I think of "modify" a value not "overwrite" the value. So, I think this should be `set()` based on what the method is doing...

```js
this.storageMap[key] = value;
```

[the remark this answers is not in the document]
marcaaron: also just generally keeps this consistent with the method calling it which seems reasonable.

lib/Onyx.js
```diff
@@ -428,17 +460,15 @@ function set(key, val) {
...
-    // This method will transform the params from a better JSON format like:
-    // {'@MyApp_user': 'myUserValue', '@MyApp_key': 'myKeyValue'}
-    const keyValuePairs = _.reduce(data, (finalArray, val, key) => ([
-        ...finalArray,
-        [key, JSON.stringify(val)],
-    ]), []);
+    const keyValuePairs = prepareKeyValuePairsForStorage(data);
+
+    // Capture (non-stringified) changes to cache
+    cache.merge(_.pairs(data));
```
marcaaron, line 466: Isn't this basically doing `multiMerge` not `multiSet()`?

```js
// previous value {loading: true}
Onyx.multiSet(['someKey', {data: 'test'}]); // -> {loading: true, data: 'test'}
```

and the value should be `{data: 'test'}` ?

lib/Onyx.js
```diff
@@ -428,17 +460,15 @@ function set(key, val) {
...
+
+    // Optimistically inform subscribers
+    _.each(data, (val, key) => keyChanged(key, val));
```
marcaaron, line 469: This comment might change based on the resolution to the comment above... but I think in this case since we've used a `merge()` that we are in some cases sending one copy of the `val` (object or array) to `keyChanged()` and another to the cache.

Wouldn't it be more correct to update these subscribers with the value _from the cache_ so future subscribers are referencing that same value instead of some referencing the value passed to `multiSet()` and others the cached value (retrieved via `get()`)?

lib/Onyx.js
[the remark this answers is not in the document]
marcaaron: >Calling subscribers with the same data again isn't a big problem if any, at most they will re-render

Cool. I'm not sure if it's a problem just noting the behavior change. It seems like if we could avoid it then it would be nice to do so. One alternative I'd propose is that we make the "retryable" part of [`evictStorageAndRetry`](https://github.com/Expensify/react-native-onyx/blob/0cdf133867dccbd0b99090850130fef3442e826f/lib/Onyx.js#L394-L407
) only operate on the data in storage and not manipulate the cache at all. That would solve the problem in another way.

>But I see a bigger issue: This means that there is an existing concurrency issue even before my changes

Sounds like an interesting future improvement/observation, but I'm not sure this is a problem we are experiencing atm. Usually what causes the storage to throw errors is that a very large `reportActions_*` set is processed and merged. Those are usually merged in and not written with `multiSet()` or `set()`.

lib/Onyx.js
```diff
@@ -551,34 +579,31 @@ function mergeCollection(collectionKey, collection) {
...
             // New keys will be added via multiSet while existing keys will be updated using multiMerge
             // This is because setting a key that doesn't exist yet with multiMerge will throw errors
             const existingCollectionPromise = AsyncStorage.multiMerge(keyValuePairsForExistingCollection);
             const newCollectionPromise = AsyncStorage.multiSet(keyValuePairsForNewCollection);
 
+            // Capture (non-stringified) changes to cache
+            cache.merge(_.pairs(collection));
+
+            // Optimistically inform collection subscribers
+            keysChanged(collectionKey, collection);
```
marcaaron, line 604: I think this maybe has the same problem I called out in another comment. The collection here is not referencing the cached value (which is a merged copy), but the value passed `mergeCollection()`?

lib/OnyxCache.js
```diff
@@ -0,0 +1,183 @@
...
+    /**
+     * Merge data to cache, any non existing keys will be crated
+     * Array items are concatenated
+     * Object items are merged
+     * If the existing value in cache and the new value are not the same type
+     * the existing value is replaced by the new value
+     * In case the merge value for a key is undefined - the key is removed from cache
+     * @param {Array<[string, *]>} pairs - array of key value pairs
+     */
+    merge(pairs) {
```
marcaaron, line 128: I wonder if passing `pairs` is the way to go here. Seems like it could be cleaner to just work with one `key/value` at a time and would also allow us to return a single value that we can call `keyChanged()` with after.

lib/OnyxCache.js
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    /**
+     * Get all the storage keys
+     * Provide a promise returning fallback that will be executed when
+     * keys aren't available in cache
+     * @param {function(): Promise<string[]>} fallback - a callback to retrieve keys from storage
+     * @returns {Promise<string[]>}
+     */
+    getAllKeys(fallback) {
```
marcaaron, line 46: NAB, I still think passing in `fallback` here and for `get()` is unnecessary. We are trying to avoid the cache having a dependency on `AsyncStorage`, but I'm not convinced there's any reason to do so.

marcaaron (changes requested):

kidroca (commented):
I think I'll be able to complete this tomorrow (or come close), so if you have anything regarding the threads where we don't know what to do exactly would be great

kidroca (commented):
Posted some comments back

Regarding the top level items:
- the fallback function abstraction - I don't think we'll gain much by moving the code to `OnyxCache`. It makes the util less flexible and harder to maintain. I like the initial version where it was clear what happens. 
- disconnecting / removing cache keys should consider collection key subscribers -> I've posted a comment about it - would add clean up code
- cache.merge() / mergeCollection() should be sending complete values to subscribers -> I'm not convinced we should change what `mergeCollection` sends to `keysChanged` in this PR. Not making a change means at least the current functionality is preserved, otherwise we could get stuck trying to fix things in the current PR

lib/OnyxCache.js
```diff
@@ -0,0 +1,183 @@
...
+
+            if (_.isObject(valueToMerge) && _.isObject(existingValue)) {
+                newValue = {...existingValue, ...valueToMerge};
+            }
+
+            if (_.isArray(valueToMerge) && _.isArray(existingValue)) {
+                newValue = [...existingValue, ...valueToMerge];
+            }
+
+            this.update(key, newValue);
```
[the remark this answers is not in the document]
kidroca: Here's the same for `AsyncStorage`:

image

So it looks like the end result of both `AsyncStorage.merge` and `lodashMerge` is identically and we should use `lodashMerge` in cache

lib/Onyx.js
```diff
@@ -367,7 +371,14 @@ function disconnect(connectionID, keyToRemoveFromEvictionBlocklist) {
         removeFromEvictionBlockList(keyToRemoveFromEvictionBlocklist, connectionID);
     }
 
+    const key = callbackToStateMapping[connectionID].key;
     delete callbackToStateMapping[connectionID];
+
+    // When the last subscriber disconnects, drop cache as well
+    const remainingConnectionsForKey = _.filter(callbackToStateMapping, {key});
```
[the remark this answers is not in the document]
marcaaron: >maybe we should leave cache to fill indefinitely ATM

I'm fine with this, but we should clarify the expectations of others and see if they agree. If you want to go that route, please bring it up on the original issue or in Slack and get a consensus.

Alternatively, we can just fix this so it works correctly. I'd prefer we do this option 😄

lib/Onyx.js
marcaaron: Alright, I appreciate all of the context here about references and I don't think we need to do anything yet with the `keysChanged()` stuff (my opinion). Although, you might want to clarify how things will work on the original issue to make sure the expectations are met with regards to the "reduced memory footprint" that some are expecting to see.

However... we must address how `mergeCollection()` works in this PR WRT to the cache merging because before the change we did:

- `AsyncStorage.multiMerge()` sets data in one way
- Subscribers are notified (with some receiving partial data atm)
- The data in storage matches what the next subscriber will get when they ask for the key

I will not recommend merging this PR if `cache.merge()` affects data differently than `AsyncStorage.multiMerge()`. The cache should reflect what is on disk... more on this in another comment.

lib/OnyxCache.js
marcaaron: Thanks for your thoughts. Let's go ahead and make this change.

lib/OnyxCache.js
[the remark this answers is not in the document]
marcaaron: Yes the cache should be performing the same logic as `multiMerge()` ?

https://github.com/react-native-async-storage/async-storage/blob/864c64bd0f6414c8959548ec237637a82b7faec3/src/AsyncStorage.js#L13-L19

Since you want to avoid changing too much behavior from what we have now the path of least resistance would be to rename this to `cache.multiMerge()` instead of `cache.merge()`.

This is mainly to avoid confusion with `Onyx.merge()` which actually batches merges + handles arrays properly + calls `set()`.

Sound good?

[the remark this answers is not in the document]
kidroca: I was thinking of just make it a regular merge that will get a single object and merge it to cache

We have only one use case here:

https://github.com/Expensify/react-native-onyx/blob/ee177e27eeae0208d3aeeef451c4eedf8badfe3b/lib/Onyx.js#L647-L648

That will be updated to 

```js
cache.merge(collection);
```

Since `loadashMerge` would update nested objects the delta from the `collection` would be merged to cache and the cache keys for the collection items updated

The other usage here is invalid

https://github.com/Expensify/react-native-onyx/blob/ee177e27eeae0208d3aeeef451c4eedf8badfe3b/lib/Onyx.js#L647-L648

And will be removed and the code will do this instead:

```js
    _.each(data, (val, key) => {
        // Update cache and optimistically inform subscribers
        cache.set(key, val);
        keyChanged(key, val);
    });
```

lib/Onyx.js
kidroca: > However... we must address how `mergeCollection()` works in this PR WRT to the cache merging

Yep, I'll make sure cache reflects what's stored on disk for the merged collection

lib/Onyx.js
kidroca: Updated to 

```js
    _.each(data, (val, key) => {
        // Update cache and optimistically inform subscribers
        cache.set(key, val);
        keyChanged(key, val);
    });
```

So there's no need for a test that reference matches in this particular case

marcaaron:
@kidroca Just a heads up I'm going out of office for a bit. If you would prefer to move forward without my review feel free to tag in a random reviewer with PullerBear and request for another reviewer in the Slack channel. Thanks!

lib/Onyx.js
kidroca: `cache.merge(collection);` merges each collection item to cache and cache reflects what's stored on disk

lib/Onyx.js
```diff
@@ -34,8 +35,11 @@ let defaultKeyStates = {};
  * @returns {Promise<*>}
  */
 function get(key) {
-    return AsyncStorage.getItem(key)
-        .then(val => JSON.parse(val))
+    return cache.getValue(key, () => AsyncStorage.getItem(key)
+        .then((val) => {
+            const parsed = JSON.parse(val);
```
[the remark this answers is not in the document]
kidroca: Updated how `get` and `getAllKeys` work and it should be clear how data retrieving works and when cache is involved

lib/OnyxCache.js
kidroca: Merge works with an object now

lib/OnyxCache.js
kidroca: Updated how get and getAllKeys work and it should be clear how data retrieving works and when cache is involved

lib/OnyxCache.js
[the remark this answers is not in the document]
kidroca: Updated `merge` to work with an object and to use `lodashMerge` internally

AsyncStorage `multiSet`/`multiMerge` uses key value pairs and for each pair `merges` (or sets) the data for the key
`cache.merge` uses a single object with cache keys and values (collection) - it merges data to those keys 
The end result is the same as the stored result

kidroca:
Updated
Addressed all threads that were unresolved
Most added lines in this PR are from tests 680+ and a tests moved to a separate file 80+
The actual code changes are around ~350 lines

lib/Onyx.js
```diff
@@ -350,6 +403,43 @@ function connect(mapping) {
...
+ */
+function cleanCache(key) {
+    const hasRemainingConnections = _.some(callbackToStateMapping, {key});
+
+    // When the key is still used in other places don't remove it from cache
+    if (hasRemainingConnections) {
+        return;
+    }
+
+    // When this is a collection - also recursively removed any unused individual items
```
roryabraham, line 418:
```suggestion
    // When this is a collection - also recursively remove any unused individual items
```

roryabraham (commented):
This is looking great, really excited for this. Most of my feedback / questions were pretty minor (sorry my review wasn't as substantive as @marcaaron's 😅). I didn't have time to test this locally today, so I'm not approving right now. I can spend some more time reviewing this tomorrow.

lib/OnyxCache.js
```diff
@@ -0,0 +1,150 @@
...
+    captureTask(taskName, promise) {
+        this.pendingPromises[taskName] = promise.finally(() => {
+            delete this.pendingPromises[taskName];
+        });
+
+        return this.pendingPromises[taskName];
+    }
+}
+
+const instance = new OnyxCache();
```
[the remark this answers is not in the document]
kidroca: This works for `js`

```js
const instance = new OnyxCache();
export default instance;
```

It's very similar to how Onyx exports an object with the difference that in Onyx the object is not instantiated with `new`

https://github.com/Expensify/react-native-onyx/blob/ee177e27eeae0208d3aeeef451c4eedf8badfe3b/lib/Onyx.js#L701-L714

https://github.com/Expensify/react-native-onyx/blob/ee177e27eeae0208d3aeeef451c4eedf8badfe3b/lib/Onyx.js#L747

`return OnyxCache.instance;` - this (returning) won't work in a class constructor

lib/OnyxCache.js
```diff
@@ -0,0 +1,150 @@
...
+         * @type {Set<string>}
+         */
+        this.storageKeys = new Set();
+
+        /**
+         * @private
+         * A map of cached values
+         * @type {Record<string, *>}
+         */
+        this.storageMap = {};
```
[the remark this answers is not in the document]
kidroca: Actually `Map` can be made to work with `lodashMerge` 
I intentionally avoided using a `Map` as I've done so previously (other PR) and was told if I don't have a specific reason to use Map over plain object to prefer a plain object. My specific reason was that Map IMO provides a better interface for mapped data, but as a personal preference it was disregarded 

On the other hand we're using `Set` as it will ignore adding duplicate keys so that field has a specific reason to use modern features

lib/OnyxCache.js
```diff
@@ -0,0 +1,150 @@
...
+    hasCacheForKey(key) {
+        return isDefined(this.storageMap[key]);
+    }
+
+    /**
+     * Saves a key in the storage keys list
+     * Serves to keep the result of `getAllKeys` up to date
+     * @param {string} key
+     */
+    addKey(key) {
```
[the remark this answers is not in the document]
kidroca: This was requested by Marc, and I agree with him that it's better now that we've extracted a separate handler

lib/OnyxCache.js
```diff
@@ -0,0 +1,150 @@
...
+        this.storageKeys.add(key);
+    }
+
+    /**
+     * Set's a key value in cache
+     * Adds the keys to the storage keys list as well
+     * Providing value is optional. Skipping value would just capture the key so that
```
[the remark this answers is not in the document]
kidroca: Initially `set` (was `update` at the time) and worked with an optional 2nd parameter
But the method was trying to be a bit too clever and we decided to drop this - e.g. usages in Onyx where you call `cache.update('myKey')` or `cache.set('myKey')` without providing a value are confusing, while `cache.addKey('myKey')` is more in line

You can see the old code with the now resolved threads about it here:
https://github.com/Expensify/react-native-onyx/pull/76/files/16d42ac061c23d8502280a1c5bb2e3e91a3f1cc5#diff-ae2f16183ba08951cb9b1a4bb48bd5b6c62e71f908ef121aa7e59524c901454fR70-R76

The documentation here should be updated to reflect that the 2nd parameter is not optional

lib/OnyxCache.js
```diff
@@ -0,0 +1,150 @@
...
+     * it's available in `getAllKeys`
+     * @param {string} key
+     * @param {*} value
```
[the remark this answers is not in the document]
kidroca: As per my other comment this is not intended to be optional, but a leftover documentation

lib/OnyxCache.js
```diff
@@ -0,0 +1,150 @@
...
+     * @returns {*} value - returns the cache value
+     */
+    set(key, value) {
```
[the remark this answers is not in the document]
kidroca: As per my other comment this is not intended to be optional, but a leftover documentation

You can still pass `undefined` or `null` but the handling is different

`undefined` would remove a value from cache - `set('myKey', undefined)`. E.g. if some iterative logic needs to update or remove values (like `Onyx.multiSet`). Otherwise prefer `remove('myKey')` 

`null` would set the cached value as `null` - it's a valid value and `hasCacheForKey` would return `true` in such case. For example `AsyncStorage.getItem` can return `null` if there's no data for a given `key` - in that case we still want to capture the result and prevent other callers from calling `AsyncStorage.getItem` again. We'll just tell them the result is `null` right away.

lib/OnyxCache.js
```diff
@@ -0,0 +1,150 @@
...
+     */
+    addKey(key) {
+        this.storageKeys.add(key);
+    }
+
+    /**
+     * Set's a key value in cache
+     * Adds the keys to the storage keys list as well
+     * Providing value is optional. Skipping value would just capture the key so that
+     * it's available in `getAllKeys`
```
kidroca, line 81:
```suggestion
     * Adds the key to the storage keys list as well
```

kidroca (commented):
Thanks, for the review 

I think there's one more place I need to bring to attention, perhaps this conversation should stay unresolved so it's visible: https://github.com/Expensify/react-native-onyx/pull/76/files#r641400108

I'll open a separate ticket about it but:

Basically it seems there's an existing problem with `Onyx.mergeCollection(key, collection)` where `collection` is supposed to be (or might be) just the delta for the collection items
but then `keysChanged` is called with the delta - `keysChanged(key, collection)`. Reacts components work it out as they merge to existing state, while non react component usage might have a problem receiving only the delta

kidroca:
Regarding testing the cache in E.cash I have a branch where I tested with and without cache: https://github.com/kidroca/Expensify.cash/tree/kidroca/sample-onyx-benchmark

If you want to use `Onyx.printStatistics` in the console you need to update you `.env` file (and rebuild android/ios) to have `BENCHMARK_ONYX=true` 

See the changes here:

https://github.com/kidroca/Expensify.cash/blob/85765d44fb8aa7443f365f887c571849ef4922f3/src/Expensify.js#L22-L40

lib/Onyx.js
```diff
@@ -522,17 +635,15 @@ function initializeWithDefaultKeyStates() {
...
+        .then((keys) => {
+            _.each(keys, (key) => {
                 keyChanged(key, null);
+                cache.remove(key);
             });
-
-            initializeWithDefaultKeyStates();
-        });
+        })
+        .then(() => AsyncStorage.clear())
```
roryabraham, line 645: NAB:
```suggestion
        .then(AsyncStorage.clear)
```

lib/OnyxCache.js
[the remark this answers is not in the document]
roryabraham: When in doubt, let's stick with the status quo.

lib/OnyxCache.js
```diff
@@ -0,0 +1,148 @@
...
+            'hasPendingTask', 'getTaskPromise', 'captureTask',
+        );
+    }
+
+    /**
+     * Get all the storage keys
+     * @returns {string[]}
+     */
+    getAllKeys() {
+        return [...this.storageKeys];
```
roryabraham, line 47: Any reason we can't just return `this.storageKeys` directly?

lib/OnyxCache.js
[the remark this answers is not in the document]
roryabraham: Okay, thanks.

lib/OnyxCache.js
kidroca: This actually converts the keys to array
It can be changed to something like `Array.from(this.storageKeys)` to be more explicit about what's happening

We can return `this.storageKeys` but then all usages have to be checked if they still work with a `Set` as currently they expect an `Array` 

https://github.com/Expensify/react-native-onyx/blob/27c0ba8ba1dee20171062ff249d03f2696d078a8/lib/Onyx.js#L373-L382

E.g. I'm not sure if this _.filter will still work, and don't want to break something

Jag96 (commented):
I did some testing on my physical iPhone and I can confirm this speeds up the app quite a bit! I'm noticing a couple of differences between this branch and master, so I want to confirm they aren't regressions:

1. On main I notice that after tapping a chat, I'm brought into the chat immediately and then there is a loading spinner. On this branch, after tapping the chat it doesn't immediately take me into the chat, and then after half a second or so I'm taken to the chat without a loading spinner
2. On main I notice that after logging in I see the E.cash logo and then all chats appear at the same time. On this branch, I notice that after I log in I see the E.cash logo and then I see one chat appear, and then after a couple of seconds the rest of them appear

To test this, I just updated the `react-native-onyx` hash in package.json for E.cash and built it on my phone. @kidroca have you tested this on a physical device at all? Just wondering if anybody else sees these differences as well.

As a note, when testing this on the simulator (both dev and production) I don't see the differences mentioned above, it is as if these changes aren't getting applied. I'm not sure why yet.

kidroca:
> I did some testing on my physical iPhone and I can confirm this speeds up the app quite a bit! I'm noticing a couple of differences between this branch and master, so I want to confirm they aren't regressions:
> 
> 1. On main I notice that after tapping a chat, I'm brought into the chat immediately and then there is a loading spinner. On this branch, after tapping the chat it doesn't immediately take me into the chat, and then after half a second or so I'm taken to the chat without a loading spinner
> 2. On main I notice that after logging in I see the E.cash logo and then all chats appear at the same time. On this branch, I notice that after I log in I see the E.cash logo and then I see one chat appear, and then after a couple of seconds the rest of them appear

Sorry I'm a bit confused about branches. At the start you mention "this branch and master" so I guess you talk about the Onyx project and the branch used in this PR, but then you speak about "main" and "this branch" and so this is the E.cash project, but I don't know what "this branch" is? Is it the branch I've shared: https://github.com/kidroca/Expensify.cash/tree/kidroca/sample-onyx-benchmark or do you mean that there are issues when you use the changes from this PR in E.cash?

> To test this, I just updated the react-native-onyx hash in package.json for E.cash and built it on my phone. @kidroca have you tested this on a physical device at all? Just wondering if anybody else sees these differences as well.

Yes, I have on both iPhone and Android. I have posted Android videos here: https://github.com/Expensify/Expensify.cash/issues/2762#issuecomment-848318315 I think I can see this right after 27sec. of the videos (both): 
_"I notice that after tapping a chat, I'm brought into the chat immediately and then there is a loading spinner. On this branch, after tapping the chat it doesn't immediately take me into the chat, and then after half a second or so I'm taken to the chat without a loading spinner"_ 
- On the before video - chat opens almost immediately, there's a spinner but the initial content is blank and slowly starts to appear
- On the after video there's a delay after the tap but then the fully loaded chat opens with images and all content rendered

I think the problem is the sheer amount of items that are pre-rendered when you select a chat - could be more than 200 chat messages.
- Before: the information would come with a delay so the chat is opened almost instantly, you see the blank page, and then as data for each of the 200 items is received they start to render and pre-render in the background
- After: with cache - the data comes instantly and the FlatList is choked to render 200 items at once as you're transitioning to the chat at the same time 

This can be addressed by updating the Virtualization Configuration not to render 200 items at once when you won't see more than 20 initially - https://github.com/Expensify/Expensify.cash/issues/2629

If that's not enough - update the logic in the `ReportScreen` so the loading screen is showed initially when you tap on a chat, and only after the transition is complete the `ReportView` is mounted which in turn mounts the FlatList and the chat. Key point - `ReportView` stays unmounted during the transition
Currently there's loading but at the same time the new chat is mounted 

> As a note, when testing this on the simulator (both dev and production) I don't see the differences mentioned above, it is as if these changes aren't getting applied. I'm not sure why yet.

Happened to me as well. I even went to `node_modules` to check if the `react-native-onyx` package is the one with cache - and it was. 

This fixed it for me:
```sh
npm run start -- --reset-cache
```
You should see
> warning: the transform cache was reset.

But even then I had to reload the app a few times until it start to work as expected

It's seems libraries from `node_modules` are cached so the react native server doesn't have to evaluate them each time it starts
In the end it works for me on a sim/emulator, just make sure to reset cache when you switch between onyx versions

lib/OnyxCache.js
roryabraham: I like the idea of switching this  to `Array.from`, it's a bit more explicit and should hopefully prevent someone from potentially making the mistake I almost made here.

Jag96:
> Sorry I'm a bit confused about branches. At the start you mention "this branch and master" so I guess you talk about the Onyx project and the branch used in this PR, but then you speak about "main" and "this branch" and so this is the E.cash project, but I don't know what "this branch" is?

Sorry, the only thing I was changing was the `react-native-onyx` branch between `kidroca:kidroca/onyx-cache` (this PR branch) and `master`, I kept E.cash on main the entire time. When typing out my notes initially I used `main` but didn't realize `react-native-onyx` was still `master`.

> npm run start -- --reset-cache

Ah looks like that did it, thanks!

> I think the problem is the sheer amount of items that are pre-rendered when you select a chat - could be more than 200 chat messages.

Ok, that makes sense, I agree this is already a big improvement so those 2 items aren't blockers, esp if they'll be reduced with the virtualization improvements.

Jag96 (commented):
Other than [this outstanding comment](https://github.com/Expensify/react-native-onyx/pull/76/files#r644962070) this PR looks good to me and tests well. @roryabraham @Julesssss any other things you'd like addressed? Ideally, we'd get this merged this week to have everyone test before sending N5

roryabraham:
👍 from me

lib/OnyxCache.js
kidroca:
```suggestion
        return Array.from(this.storageKeys);
```

kidroca:
Addressed the outstanding comment by explicitly converting the set to array

Jag96 (approved):
LGTM, will leave open to let @Julesssss get his review in

Jag96 (changes requested):
Re-tested this today on the latest E.cash main on an account that has an IOU and I'm getting this error on both web and desktop after signing in (iOS/Android work fine):

```
TypeError: Currency code is required with currency style.
    at new NumberFormat (<anonymous>)
    at oO (app-6bf1371e025ae4112ce5.bundle.js:78)
    at Object.numberFormat (app-6bf1371e025ae4112ce5.bundle.js:78)
    at IOUBadge (app-6bf1371e025ae4112ce5.bundle.js:78)
    at Yi (app-6bf1371e025ae4112ce5.bundle.js:50)
    at gs (app-6bf1371e025ae4112ce5.bundle.js:50)
    at lc (app-6bf1371e025ae4112ce5.bundle.js:50)
    at sc (app-6bf1371e025ae4112ce5.bundle.js:50)
    at Zs (app-6bf1371e025ae4112ce5.bundle.js:50)
    at app-6bf1371e025ae4112ce5.bundle.js:50

Uncaught (in promise) TypeError: Currency code is required with currency style.
    at new NumberFormat (<anonymous>)
    at oO (app-6bf1371e025ae4112ce5.bundle.js:78)
    at Object.numberFormat (app-6bf1371e025ae4112ce5.bundle.js:78)
    at IOUBadge (app-6bf1371e025ae4112ce5.bundle.js:78)
    at Yi (app-6bf1371e025ae4112ce5.bundle.js:50)
    at gs (app-6bf1371e025ae4112ce5.bundle.js:50)
    at lc (app-6bf1371e025ae4112ce5.bundle.js:50)
    at sc (app-6bf1371e025ae4112ce5.bundle.js:50)
    at Zs (app-6bf1371e025ae4112ce5.bundle.js:50)
    at app-6bf1371e025ae4112ce5.bundle.js:50
```

This results in a white screen on Web and Desktop. I tested this on dev after running `npx react-native-clean-project` and wiping the node_modules folder, and I also confirmed it happens when running `npm run desktop-build` as well. Here's the steps I did to test this, just to make sure I'm not missing anything:

1. Run `npx react-native-clean-project` and wipe node_modules folder
2. Update package.json `react-native-onyx` commit hash to `77b07e10dfd10831380c197a132a032e812b31f9`
3. Run npm install then npm run web and npm run desktop and npm run desktop-build
4. Sign into an account that has an IOU
5. See the above errors in the console and a white screen
6. Clear changes in package.json so the `react-native-onyx` hash is back to `d0a5d12c08f2f029e3038fb3516a93a9403e9746`
7. Run `npx react-native-clean-project` and wipe node_modules folder
8. Run npm install then npm run web and npm run desktop and npm run desktop-build
9. Confirm there is no white screen and the error doesn't appear after signing in

I chatted w/ @roryabraham and he is also going to test this to confirm

Jag96:
It looks like setting a default value for currency [here](https://github.com/Expensify/Expensify.cash/blob/027ac6cb8e366f835281ac1f75c9b7297f71b1cb/src/components/IOUBadge.js#L29) fixes the issue, going to create an E.cash PR.

Jag96:
PR merged, going to test again on the main branch to confirm everything is working

quinthar:
Woo!!!!

roryabraham (approved):
Spent a while testing this on iOS, web, desktop, and I couldn't find any regressions or obvious errors. Seems to work well.

Jag96 (approved):
This looks good to me as well, tested on all platforms and it's working great! I'll let @marcaaron have a final review/merge for this one

Julesssss (approved):
Sorry for the late review! I ran this locally and while it was tricky to do a visual comparison between dev and prod builds the improvement is noticeable. 

I had some of the same concerns as @Jag96 mentioned [here](https://github.com/Expensify/react-native-onyx/pull/76#pullrequestreview-674760326), but as discussed there are further optimizations that might take care of this. The main one is that even pages without huge lists _feel_ slightly slower to display. For example, the initial delay when tapping an option in the setting page feels like input lag (I know this isn't actually the case, but this might seem like input lag to a user). 

Anyway, thanks for leading this huge improvement. I look forward to testing on a prod build!

tests/unit/onyxCacheTest.js
```diff
@@ -0,0 +1,680 @@
...
+
+            // GIVEN a component connected to Onyx
+            const TestComponentWithOnyx = withOnyx({
+                text: {
+                    key: ONYX_KEYS.TEST_KEY,
+                },
+            })(ViewWithText);
+
+            // GIVEN some string value for that key exists in storage
+            AsyncStorageMock.getItem.mockResolvedValue('"mockValue"');
```
marcaaron, line 426: Is there some significance to doing a single and double quote here e.g. `'"mockValue"'` vs. just `'mockValue'`?

tests/unit/withOnyxTest.js
```diff
@@ -95,6 +97,7 @@ describe('withOnyx', () => {
     });
 
     it('should update withOnyx subscribing to individual key with merged value if mergeCollection is used', () => {
+        // Todo: ViewWithCollections does not expect a `text` prop
```
marcaaron, line 100: Who is this todo for / are we planning to clean this up? not sure what this comment means.

tests/unit/onyxCacheTest.js
```diff
@@ -0,0 +1,680 @@
+import React from 'react';
+import {render} from '@testing-library/react-native';
+
+import waitForPromisesToResolve from '../utils/waitForPromisesToResolve';
+import ViewWithText from '../components/ViewWithText';
+import ViewWithCollections from '../components/ViewWithCollections';
+
+describe('Onyx', () => {
+    describe('Cache Service', () => {
```
marcaaron, line 9: NAB, appreciate the thoroughness here, but rather than test the individual cache methods we could have just tested the behavior that we'd expect when Onyx is actually using the cache vs. using the cache in isolation.

tests/unit/onyxCacheTest.js
```diff
@@ -0,0 +1,680 @@
...
+        /** @type OnyxCache */
+        let cache;
+
+        // Always use a "fresh" instance
```
marcaaron, line 13: NAB, Is that what requiring again would do? When would this not be creating a "fresh" instance? Why should we use one instead of just clearing out the cache?

marcaaron (approved):
Left a few minor comments (mostly about tests and not blockers - but should be addressed). Still think we should move usages of `captureTask` and `AsyncStorage` into the cache itself, but it's not a hill I will die on so I'll let it go for now.

marcaaron:
Next we will just need to create a `Expensify/Expensify.cash` PR referencing this commit `4f341cfca83bcc4f9a8d89b593d81ea78903f04c` for `react-native-onyx`.

tests/unit/onyxCacheTest.js
kidroca: > NAB, appreciate the thoroughness here, but rather than test the individual cache methods we could have just tested the behavior that we'd expect when Onyx is actually using the cache

Actually we've tested both in isolation here and the behavior that we'd expect when Onyx is using cache (in the next section)

It's much harder to test all the expectations written here in Onyx, you have to 
- setup component(s) 
- configure async blocks and flush promises at specific times
- not as straightforward if you want to debug a test

The test for Onyx validate complete flows covering multiple cache methods - e.g. a key is not retrieved from storage when a rendered component uses it

tests/unit/onyxCacheTest.js
kidroca: > NAB, Is that what requiring again would do? When would this not be creating a "fresh" instance? Why should we use one instead of just clearing out the cache?

Simply requiring again would not do the trick. It would return the instance that was already created from the first time the script was required somewhere

There's no function that completely clears the cache. You could implemented something here or in the cache file, but if it's only for the test you could just tell jest you want a clean env

tests/unit/onyxCacheTest.js
```diff
@@ -0,0 +1,680 @@
...
+
+        function initOnyx() {
+            const OnyxModule = require('../../index');
+            Onyx = OnyxModule.default;
+            withOnyx = OnyxModule.withOnyx;
+            cache = require('../../lib/OnyxCache').default;
+
+            Onyx.init({
+                keys: ONYX_KEYS,
+                registerStorageEventListener: jest.fn(),
```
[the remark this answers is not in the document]
kidroca: > Not sure if we need to do anything about this. But seeing `registerStorageEventListener` got me wondering whether we should be handling cache values in that listener.

Yes, `cache.set` should probably be used here as well

- When a given tab changes storage (saves something) all listeners (other tabs) are notified of the changes (same as before)
- Then if you switch to some previous tab data is rendered correctly (same as before)
- Now if some other component needs to use the same key that the storage event listener fired with
  - if we have cache for that key -> bug it's stale
  - if we don't have cache it's going to be read from storage

It's rare, but could happen

tests/unit/withOnyxTest.js
kidroca: Not sure who are they for,  it's just something that I've spotted that's out of this PR's scope

kidroca (commented):
> Still think we should move usages of captureTask and AsyncStorage into the cache itself, but it's not a hill I will die on so I'll let it go for now.

Sorry I thought that's what you meant by https://github.com/Expensify/react-native-onyx/pull/76#discussion_r641772230
> Thanks for your thoughts. Let's go ahead and make this change.

I thought it was a reply to 
> Adding Onyx specifics here makes this util less flexible it should just provide some low level cache handling heplers like hasCacheForKey, get, set, merge and the resolveTask helper for concurrency calls and leave the caller the flexibility to use it however needed. Then it's transparent how items are retrieved in Onyx.js and the cache handling is on the same abstraction level

And so I switched to how Cache was set up initially 

Right now we have separation - cache logic in Cache. Onyx logic and flows in Onyx. It's transparent how Onyx uses and updates cache

It would be confusing to me if I read the high level of the code for `Onyx.get` and see that it just does `return cache.get(key)`. "What happens if there's no cache?" is my first question, then I have to open `OnyxCache` and have another question "Why does cache knows how to retrieve values from storage?" 

From the current code it's clear that it doesn't have to know and works

If it's just a Proxy - the "startTask" implementation where it doesn't know how the task would provide the value is OK, but it made it harder to test and it was confusing for everybody but me

kidroca:
> Next we will just need to create a `Expensify/Expensify.cash` PR referencing this commit `4f341cfca83bcc4f9a8d89b593d81ea78903f04c` for `react-native-onyx`.

Should I create this?
I guess yes, and I'll need to provide samples of my tests for all the platforms 
I can do this tomorrow

marcaaron:
>Sorry I thought that's what you meant by #76 (comment)

No. I was thanking you for your thoughts and requesting that you take my suggestion anyway.

tests/unit/withOnyxTest.js
marcaaron: Please create a new issue for them then. We don't leave random `todos` throughout our codebase for others to wonder how or if they should be actioned.

lib/OnyxCache.js
```diff
@@ -0,0 +1,148 @@
...
+     * Remove data from cache
+     * @param {string} key
+     */
+    remove(key) {
+        this.storageKeys.delete(key);
+        delete this.storageMap[key];
+    }
+
+    /**
+     * Deep merge data to cache, any non existing keys will be crated
```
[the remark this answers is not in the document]
kidroca: Addressed in #79

tests/unit/onyxCacheTest.js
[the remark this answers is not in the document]
kidroca: Addressed in #79

lib/OnyxCache.js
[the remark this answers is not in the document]
kidroca: Addressed in #79

tests/unit/onyxCacheTest.js
```diff
@@ -0,0 +1,680 @@
...
+        }
+
+        // Always use a "fresh" instance
+        beforeEach(() => {
+            jest.resetModules();
+            return initOnyx();
+        });
+
+        it('Expect a single call to getItem when multiple components use the same key', () => {
+            const AsyncStorageMock = require('@react-native-community/async-storage/jest/async-storage-mock');
```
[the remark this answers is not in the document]
kidroca: Addressed in #79

tests/unit/withOnyxTest.js
kidroca: Addressed in #79

Julesssss:
Moving this comment here, as [the linked issue](https://github.com/Expensify/Expensify.cash/pull/1785#issuecomment-857600896) was locked. 

______

This morning @trjExpensify and I noticed a potential regression with the Onyx changes:

The `iou` loading state values [that we initialize here](https://github.com/Expensify/Expensify.cash/blob/fe2355cc3b6b57d1ef19624e3756f30a712e549b/src/Expensify.js#L32) are [returned null here](https://github.com/Expensify/Expensify.cash/blob/fe2355cc3b6b57d1ef19624e3756f30a712e549b/src/pages/iou/IOUModal.js#L136) when opening the IOUModal. The data is valid within local storage and this is only reproducible about 1 in 4 attempts.

While the simple solution would be to add a conditional check to confirm the iou object is valid, we need to confirm whether this requires a fix, or is now expected behavior.

CC @kidroca 

https://user-images.githubusercontent.com/10736861/121342044-2e1dc380-c919-11eb-9670-fac9b03046e2.mov

Julesssss:
It looks like the cleanup PR supposedly [resolved this issue](https://github.com/Expensify/react-native-onyx/pull/79#issuecomment-856833531). But as far as I can tell, these commits are also on the staging build. So it seems to me that the issue is still occurring:

- [Tim's PR](https://github.com/Expensify/Expensify.cash/pull/3446) [bumps the Onyx commit](https://github.com/Expensify/Expensify.cash/pull/3446/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L85-R85) to `4ff94b6c068582a75f50c6e159d3ff69fa123138`
- Commit `4ff94b6` [is more recent ](https://github.com/Expensify/react-native-onyx/commits/master)than the onyx cache fix
- Tim's PR was[ deployed to staging](https://github.com/Expensify/Expensify.cash/pull/3446#issuecomment-857283974)

kidroca:
@Julesssss 
> This morning @trjExpensify and I noticed a potential regression with the Onyx changes

I'll investigate

https://github.com/Expensify/react-native-onyx/blob/9b03d95a851fadf4c0a5ef0c942aba1a3e5bd413/lib/Onyx.js#L723-L724

Are we supposed to put these `initialKeyStates` in `cache` like so?

```js
    // Set our default key states to use when initializing and clearing Onyx data
    defaultKeyStates = initialKeyStates;
    cache.merge(initialKeyStates);
```

I think they automatically go there:

https://github.com/Expensify/react-native-onyx/blob/9b03d95a851fadf4c0a5ef0c942aba1a3e5bd413/lib/Onyx.js#L730-L731

This function would internally call `merge` which would call `set` and this will update the values in cache

Maybe, this happens after enough delay that the initial value is `null` because it's not yet `set` ?

Anyway, I'll see what can I find

trjExpensify:
Thanks, @kidroca! We're imminently sending a newsletter out to all our users to announce the IOU features and welcome them into E.cash to test out what we’ve been built so far. We shipped the latest prod build off to the app stores last night for review, which was the last item on the list before hitting send. 

If this regression makes it to production, it delays our timeline for the announcement, as the IOU feature will be broken on Web/Desktop/mWeb. With that, it would be awesome if we could get a fix up ASAP to minimise the delay. Thanks again!

Julesssss:
Thanks! BTW I believe I'm seeing the same issue with the `betas` Onyx key, but I can only reproduce this on dev currently:
- Send an IOU to a non-expensify account
- Sign into the non-expensify account
- Try to open the IOU Details Modal to pay an IOU
- `Permissions.js:34 Uncaught (in promise) ReferenceError: betas is not defined`

_____

> Are we supposed to put these initialKeyStates in cache like so?

Yeah, I think that would make sense. Perhaps we should just expect these to be initially null... but I think having the defaults added to cache would be preferable.

kidroca:
I traced what is causing the issue. It's our cache cleanup logic and a bug in `cache.remove` 

image

- When you close the IOU modal, since it's the last and only connection that uses the `iou` key
- on disconnect it tells cache to remove the `iou` from cache - because there are no more usages on screen
- but `cache.remove` would also remove it from the `storageKeys` (the bug) 
- the next time someone tries to connect to the `iou` key (You open the IOU modal)
- `getAllKeys` is not returning the `iou` because it was removed
- since `getAllKeys` does not return it `connect` short-circuits the value to `null`

https://github.com/Expensify/react-native-onyx/blob/9b03d95a851fadf4c0a5ef0c942aba1a3e5bd413/lib/Onyx.js#L373-L382

`cache.remove` should not remove the key from `storageKeys` so that `getAllKeys` continues to return correct result

https://github.com/Expensify/react-native-onyx/blob/9b03d95a851fadf4c0a5ef0c942aba1a3e5bd413/lib/OnyxCache.js#L91-L98

It should just drop the cached `value` for that key

---

@Julesssss I'll open a PR shortly

kidroca:
@Julesssss pushed a fix here #81 

And confirmed it against the `main` branch of E.cash 

https://user-images.githubusercontent.com/12156624/121382610-735ce800-c94f-11eb-9bc7-5e08ac67c52d.mp4

PR #79, opened 2021-06-08
Kidroca/onyx cache cleanup

kidroca:
@marcaaron @Jag96 @roryabraham

### Details

Address some NAB leftovers from #76

### Related Issues

#63, Expensify/Expensify.cash#2762

### Automated Tests

Cache tests were updated to require AsyncStorage in the before each block

### Linked PRs

https://github.com/Expensify/Expensify.cash/pull/3423

### DIFF
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -731,7 +731,10 @@ function init({
     initializeWithDefaultKeyStates();
 
     // Update any key whose value changes in storage
-    registerStorageEventListener((key, newValue) => keyChanged(key, newValue));
+    registerStorageEventListener((key, newValue) => {
+        cache.set(key, newValue);
+        keyChanged(key, newValue);
+    });
 }
 
 const Onyx = {
--- a/lib/OnyxCache.js
+++ b/lib/OnyxCache.js
@@ -103,6 +103,10 @@ class OnyxCache {
      */
     merge(data) {
         this.storageMap = lodashMerge({}, this.storageMap, data);
+
+        const storageKeys = this.getAllKeys();
+        const mergedKeys = _.keys(data);
+        this.storageKeys = new Set([...storageKeys, ...mergedKeys]);
     }
 
     /**
--- a/lib/OnyxCache.js
+++ b/lib/OnyxCache.js
@@ -98,7 +98,7 @@ class OnyxCache {
     }
 
     /**
-     * Deep merge data to cache, any non existing keys will be crated
+     * Deep merge data to cache, any non existing keys will be created
      * @param {Record<string, *>} data - a map of (cache) key - values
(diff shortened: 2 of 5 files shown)


kidroca:
I found an issue:

https://github.com/Expensify/react-native-onyx/blob/4f341cfca83bcc4f9a8d89b593d81ea78903f04c/lib/Onyx.js#L683-L687

`cache.merge` does not update available storage keys

https://github.com/Expensify/react-native-onyx/blob/4f341cfca83bcc4f9a8d89b593d81ea78903f04c/lib/OnyxCache.js#L100-L106

1. A request to fetch user reports and IOUs completes before LNH and chats are rendered 
2. It merges data to storage and cache, but cache does not update available keys from `getAllKeys` (bug)
3. Component tries to connect with Onyx
4. It first `getAllKeys` and see the key does not exist in storage (bug)
5. And gives the subscriber `null` as the initial value instead of reading from storage

This also reviews a small secondary issue - data can be merged to cache even when there are no subscribers for a given object
While we want to clear cache items when there are no subscribers for the data

kidroca:
Updated. Added a fix for the above issue

kidroca:
Regarding
> This also reviews a small secondary issue - data can be merged to cache even when there are no subscribers for a given object
While we want to clear cache items when there are no subscribers for the data

The above goal can be achieved if we skip merging keys that are not already in cache
If a key is not in cache - no one is subscribed for it, we should skip adding it to cache, because we want to keep only data that's used on screen, but at the same time we should still update `storageKeys` so that we keep it up to date and know the key is available 

But this is also the weak point of this strategy as we know for a fact that the collection is going to be used shortly, and if we don't store it in cache it will be read from file 
The collection is literally retrieved less than half a second after the `mergeCollection` call and because it's available in cache I don't have to read 10 collection items (chats) from file 

Should I update this with the logic to skip the collection keys that do not already exist in cache?

--- 

Here's what actually happens for clean startup in practice

1. A lot of information is fetched - LHN and chats are not yet rendered
2. Fetching reports is completed before LHN and chats are rendered
3. A call to `mergeCollection` is made - no `withOnyx` components use the data yet
4. LHN and Chat start to render
5. At this point a component `withOnyx` makes a connection for the above data
6. As this is the first connection data is read from storage, even though it was just fetched

marcaaron:
>This also reviews a small secondary issue - data can be merged to cache even when there are no subscribers for a given object While we want to clear cache items when there are no subscribers for the data

Not sure I entirely follow the explanation above or the consequences. But I would maybe just err on the side of caution and add the values to the cache whether they have active subscribers or not 🤷‍♂️

marcaaron (approved):

PR #81, opened 2021-06-09
Fix: `Onyx.cleanCache` 

kidroca:
@Julesssss @marcaaron 

### Details

Renamed `Cache.remove` to `Cache.drop` this method should not remove the key from `getAllKeys`'s result
Updated `Onyx.remove` and `Onyx.clear` they `set` the cached value to `null` this way 
the next connection does not have to wait to read `null` from storage

### Related Issues

https://github.com/Expensify/react-native-onyx/pull/76#issuecomment-857609265
https://github.com/Expensify/Expensify.cash/issues/3482

### Automated Tests

Updated the tests for `cache.drop`

### DIFF
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -471,8 +471,8 @@ function disconnect(connectionID, keyToRemoveFromEvictionBlocklist) {
  * @return {Promise}
  */
 function remove(key) {
-    // Remove from cache
-    cache.drop(key);
+    // Cache the fact that the value was removed
+    cache.set(key, null);
 
     // Optimistically inform subscribers
     keyChanged(key, null);
@@ -639,7 +639,7 @@ function clear() {
         .then((keys) => {
             _.each(keys, (key) => {
                 keyChanged(key, null);
-                cache.drop(key);
+                cache.set(key, null);
             });
         })
         .then(AsyncStorage.clear)
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -417,7 +417,7 @@ function cleanCache(key) {
 
     // When this is a collection - also recursively remove any unused individual items
     if (isCollectionKey(key)) {
-        cache.remove(key);
+        cache.drop(key);
 
         getAllKeys().then(cachedKeys => _.chain(cachedKeys)
             .filter(name => name.startsWith(key))
@@ -436,8 +436,8 @@ function cleanCache(key) {
         }
     }
 
-    // Otherwise remove the item from cache
-    cache.remove(key);
+    // Otherwise remove the value from cache
+    cache.drop(key);
 }
 
 /**
@@ -472,7 +472,7 @@ function disconnect(connectionID, keyToRemoveFromEvictionBlocklist) {
  */
 function remove(key) {
     // Remove from cache
-    cache.remove(key);
+    cache.drop(key);
 
     // Optimistically inform subscribers
     keyChanged(key, null);
@@ -639,7 +639,7 @@ function clear() {
         .then((keys) => {
             _.each(keys, (key) => {
                 keyChanged(key, null);
-                cache.remove(key);
+                cache.drop(key);
             });
         })
         .then(AsyncStorage.clear)
--- a/lib/OnyxCache.js
+++ b/lib/OnyxCache.js
@@ -34,7 +34,7 @@ class OnyxCache {
         // bind all methods to prevent problems with `this`
         _.bindAll(
             this,
-            'getAllKeys', 'getValue', 'hasCacheForKey', 'addKey', 'set', 'remove', 'merge',
+            'getAllKeys', 'getValue', 'hasCacheForKey', 'addKey', 'set', 'drop', 'merge',
             'hasPendingTask', 'getTaskPromise', 'captureTask',
         );
     }
@@ -89,11 +89,10 @@ class OnyxCache {
     }
 
     /**
-     * Remove data from cache
+     * Forget the cached value for the given key
      * @param {string} key
      */
-    remove(key) {
-        this.storageKeys.delete(key);
+    drop(key) {
         delete this.storageMap[key];
     }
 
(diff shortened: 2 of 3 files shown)


kidroca:
You might be concerned that cache will never remove values from `getAllKeys`  and so when cache fills up
this short-circuit to `null` would not work:

https://github.com/Expensify/react-native-onyx/blob/9b03d95a851fadf4c0a5ef0c942aba1a3e5bd413/lib/Onyx.js#L373-L382

I don't think it's a problem as `null` will be returned from `cache` almost instantly here:

https://github.com/Expensify/react-native-onyx/blob/9b03d95a851fadf4c0a5ef0c942aba1a3e5bd413/lib/Onyx.js#L389-L399

`get(key)` would return `null`  due to these changes: https://github.com/Expensify/react-native-onyx/pull/81/commits/b0cf26cb80aab05bddd5588e127647fec3e7ba7c

Julesssss (approved):
This tests well and the regression is solved, thanks for the quick fix!

lib/Onyx.js
```diff
@@ -471,8 +471,8 @@ function disconnect(connectionID, keyToRemoveFromEvictionBlocklist) {
  * @return {Promise}
  */
 function remove(key) {
-    // Remove from cache
-    cache.remove(key);
+    // Cache the fact that the value was removed
+    cache.set(key, null);
```
marcaaron, line 475: Unsure why we need to do this. 🤔

lib/Onyx.js
```diff
@@ -639,7 +639,7 @@ function clear() {
         .then((keys) => {
             _.each(keys, (key) => {
                 keyChanged(key, null);
-                cache.remove(key);
+                cache.set(key, null);
```
marcaaron, line 642: Same here.

marcaaron (changes requested):
Left a couple of comments. Overall I agree that this is a bug:

>on disconnect it tells cache to remove the iou from cache - because there are no more usages on screen
>but cache.remove would also remove it from the storageKeys (the bug)

But less sure why we are calling `cache.set('someKey', null)` in the two places we are.

marcaaron (approved):
Gonna merge this so we can unbreak the production release, but still curious about the `cache.set()` stuff. I don't think anything bad can happen from it just not sure I understand it.

kidroca:
@marcaaron 
> But less sure why we are calling `cache.set('someKey', null)` in the two places we are.

When we remove something using `Onyx.remove` we cache the fact that storage is now empty for that key - `null` 
If we remove the value from cache the next time the value is retrieved it will have to be read from disk
only to find out that it's `null` 
But if we `set` the value as `null` in cache we can tell the subscriber we have the value for this key and it is `null`

You can also verify this is the way AsyncStorage works by default for a missing key 

```js
await AsyncStorage.removeItem('myItem');
const value = await AsyncStorage.getItem('myItem'); // `null` 
```

We're skipping the call to `AsyncStorage.getItem` when we already know the value would be `null`

marcaaron:
>we cache the fact that storage is now empty for that key - null
If we remove the value from cache the next time the value is retrieved it will have to be read from disk
only to find out that it's null

Cool thanks, I get that part, I just don't see how it is related to the linked issue or regression and someone looking at the code might not understand the purpose. Reaching into storage is how we treat most other "gets" for a key that does not exist in Onyx so it's not clear why we are making an exception here or whether it makes any difference.

marcaaron:
>Cache the fact that the value was removed

comment could be improved  as it doesn't explain why we are doing this

kidroca:
> Reaching into storage is how we treat most other "gets" for a key that does not exist in Onyx so it's not clear why we are making an exception here or whether it makes any difference.

Well now instead of reaching for storage we know ahead of time that the key is not there - when something deliberately removed a key using `remove` or all values using `clear`

Previously (actually still is for some cases) this was achieved by using `getAllKeys` and seeing that the key is not there, now we no longer have to do it if it's solely for checking if there's a value (in `connect` it's not solely for that but to also get collection keys) 

As you can see we no longer remove keys for the result of `getAllKeys`. If we do nothing the code here might continue and read from storage only to find out the underlying value is `null`

https://github.com/Expensify/react-native-onyx/blob/9b03d95a851fadf4c0a5ef0c942aba1a3e5bd413/lib/Onyx.js#L373-L382

I've tried to explain this here: https://github.com/Expensify/react-native-onyx/pull/81#issuecomment-857788064

If you want to also remove the key from the result of `getAllKeys`, we'll have to have a separate method or add a 2nd parameter to `cache.drop`

Just like we can cache the fact that we have something, we can do the same to cache that we know there's no value in storage for a given key - and skip the call to storage

marcaaron:
I'll try being more direct here:

1. The PR introduced a change that is not related to the issue it is fixing
2. The code presented has inadequate documentation

Can you elaborate (in the code) so that others (not me) in the future can understand what your intentions were?

TomatoToaster:
I think a change here is causing the regression. More details here: https://github.com/Expensify/Expensify.cash/issues/3485#issuecomment-858300446

I think I have an idea for a fix, testing it out right now.

kidroca:
Are you sure? 

The issue you're talking about was discovered before this PR was even merged - "iOS - App crashes after turning off the internet connection"

And the update that bumps E.cash to use the Onyx changes made here happened even later that day: https://github.com/Expensify/Expensify.cash/pull/3483

Check the timing of this comment and the time of the merges - https://expensify.slack.com/archives/C01GTK53T8Q/p1623256929142800

TomatoToaster:
Ah yes sorry you are 100% right. I've got the wrong PR.

More grandly, it's a change from this diff between the versions of react-native-onyx as seen here: https://github.com/Expensify/react-native-onyx/compare/d0a5d12c08f2f029e3038fb3516a93a9403e9746..52c254fbc4b96224d2ac087e57ef6efc4210ccc1

(This is comparing the production react-native-onyx in production with what's on staging/master).

This just happened to be the most recent onyx change so I thought it was from here, but it's a little tricky to pinpoint the exact PR.

I'll put more details in the Cash issue I got this from.

PR #82, opened 2021-06-11
Notify subscribers on the next tick

alex-mechler:
cc @kidroca @marcaaron 

### Details
This updates subscribers of data changes on the next tick, rather than blocking on notifying subscribers. 

### Related Issues
https://github.com/Expensify/Expensify.cash/issues/3485#issuecomment-858705481
https://github.com/Expensify/Expensify.cash/issues/3488#issuecomment-859136914

### Automated Tests
Existing automated tests cover these changes.

### Linked PRs
https://github.com/Expensify/Expensify.cash/pull/3535

### DIFF
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -475,7 +475,7 @@ function remove(key) {
     cache.set(key, null);
 
     // Optimistically inform subscribers
-    keyChanged(key, null);
+    Promise.resolve().then(() => keyChanged(key, null));
 
     return AsyncStorage.removeItem(key);
 }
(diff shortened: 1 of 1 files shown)


Julesssss (dismissed):
Resolves the endless spinner issue on login. Even with the redirect to /r, this seems like a good improvement to me.

marcaaron (commented):
Looks good, but curious if we want to also make a similar change for `merge()`, `multiSet()` and `mergeCollection()` ? There are a few other methods that set data and also were calling `keyChanged()` or `keysChanged()` async before kid roca's changes.

alex-mechler:
Ah good call Marc, I can test adding those. That might solve some of the other issues I am seeing.

alex-mechler:
Updated to notify the subscribers on the next tick for `remove` and `multiset` to match behavior from prior to the cache.

I did not notice any difference, however this better matches the prior behavior, and passes all tests.

marcaaron (approved):

trjExpensify:
@roryabraham shall we CP this one?

PR #88, opened 2021-07-14
Onyx optimizations

kidroca:
cc @marcaaron @tgolen 

### Details
This PR contains a few optimizations and refactors
1. The first commit has and updated `Onyx.printMetrics` function. You can use it as a base to compare before/after results. It prints a general table on the console which is easy to copy and put in excel. 
2. Fix: some connections happen even before `Onyx.init` is called. This should not be allowed as the `initialKeyStates` are applied during init. Added a deferred tasks that would queue incoming connections and only make them after init, this also gives the rest of the app init some breathing space
3. Perf: Update `set` to check with cache and skip writing to storage. Cache reflects what's already on disc, if we have the exact value in cache it means it didn't change and we don't need to make a write
4. Perf: Updated cache clearing strategy. It's a mix of LRU and active connections. The cache works in the following way: each time a connection is terminated we run a `cleanCache` function. The function removes dated items from cache, if the item is still being used by a component it's kept in cache. By default it will try to keep up to ~150 keys in cache, can be a bit more due to active connections to old items
5. Refactor: `merge` was refactored so that it's promises are resolved when a queue for a key is done. This helps with measuring the function more accurately
6. Perf: Updated `defaultKeyStates` handling. Merging these keys one by one was taking a significant amount of time, even though they are only 4. Instead of writing them to storage, we read data from storage and merge it along with the defaults to the cache. Then we never allow for the `defaultKeys` to be removed from cache 

### Related Issues

https://github.com/Expensify/Expensify.cash/issues/2667

### Automated Tests
Added and updated tests related to changes in cache handling.
The rest of the changes are covered by the existing tests

### Linked PRs

TBD

### Benchmark results 
There are overall improvements that can be seen in the stats here
https://docs.google.com/spreadsheets/d/1kHRvFL1ITbr4p9TzT_nsmtvCY4KoPwKCXp84FTCsGzk/edit?usp=sharing

##### Android
|                       | Before Updates | After Updates |
|-----------------------|---------------:|--------------:|
| Total Onyx Time       |       55.06sec |      19.67sec |
| Heavy reading ends at |        8.27sec |       5.44sec |

##### iOS
|                       | Before Updates | After Updates |
|-----------------------|---------------:|--------------:|
| Total Onyx Time       |       16.37sec |       8.74sec |
| Heavy reading ends at |        2.96sec |       2.98sec |

- Total: the total time spent by Onyx functions data fetching/writing
- Heavy reading: the point since app launch at which hard reads from disk are over
- The link above contains device information and how the tests were performed

### DIFF
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -36,69 +36,37 @@ let MAX_CACHED_KEYS = 150;
 const deferredInit = createDeferredTask();
 
 /**
- * Get multiple keys in a single batch
- *
- * @param {string[]} keys
- * @returns {Promise<Record<string, *>>} - an object of key -> value mapping
- */
-function multiGet(keys) {
-    const [fromCache, hardReads] = _.partition(
-        keys,
-        key => cache.hasCacheForKey(key) || cache.hasPendingTask(`get:${key}`)
-    );
-
-    // Init a list of tasks containing any pending reads from storage
-    const tasks = fromCache
-        .filter(key => cache.hasPendingTask(`get:${key}`))
-        .map(key => Promise.resolve(cache.getTaskPromise(`get:${key}`)));
-
-    // Append a task for any remaining "hard reads" the current list
-    if (hardReads.length > 0) {
-        const readKeysFromStorage = AsyncStorage.multiGet(hardReads)
-            .then(pairs => pairs.forEach(([key, val]) => {
-                const parsed = val && JSON.parse(val);
-                cache.set(key, parsed || null);
-            }));
-
-        tasks.push(readKeysFromStorage);
-
-        /* Capture a resolver for potential multiGet calls that happen while we retrieve keys
-        * example one one call retrieves keys A,B,C another call retrieves B,C,D
-        * we don't need to start a new read for B,C we'll hook them from the first call
-        */
-        hardReads.forEach((key) => {
-            cache.captureTask(`get:${key}`, readKeysFromStorage.then(() => cache.getValue(key)));
-        });
-    }
-
-    // Return a grouped result
-    return Promise.all(tasks)
-        .then(() => {
-            /* At this point we have everything in cache
-            * Instead of searching for the value inside the result, we can retrieved it by key from cache */
-            const result = _.chain(keys)
-                .map((key) => {
-                    // eslint-disable-next-line no-use-before-define
-                    addLastAccessedKey(key);
-                    return [key, cache.getValue(key)];
-                })
-                .object()
-                .value();
-
-            return result;
-        });
-}
-
-/**
  * Get some data from the store
  *
  * @param {string} key
  * @returns {Promise<*>}
  */
 function get(key) {
-    /* We use multiGet as it will batch multiple calls and flush them on the next iteration of
-    * the event loop. AsyncStorage uses `setImmediate` to do this internally */
-    return multiGet([key]).then(result => result[key]);
+    // eslint-disable-next-line no-use-before-define
+    addLastAccessedKey(key);
+
+    // When we already have the value in cache - resolve right away
+    if (cache.hasCacheForKey(key)) {
+        return Promise.resolve(cache.getValue(key));
+    }
+
+    const taskName = `get:${key}`;
+
+    // When a value retrieving task for this key is still running hook to it
+    if (cache.hasPendingTask(taskName)) {
+        return cache.getTaskPromise(taskName);
+    }
+
+    // Otherwise retrieve the value from storage and capture a promise to aid concurrent usages
+    const promise = AsyncStorage.getItem(key)
+        .then((val) => {
+            const parsed = val && JSON.parse(val);
+            cache.set(key, parsed);
+            return parsed;
+        })
+        .catch(err => logInfo(`Unable to get item from persistent storage. Key: ${key} Error: ${err}`));
+
+    return cache.captureTask(taskName, promise);
 }
 
 /**
@@ -367,24 +335,6 @@ function sendDataToConnection(config, val, key) {
 }
 
 /**
- *  Retrieve all the connections that should be made for a given mapping
- * @param {object} mapping
- * @param {string} mapping.key
- *
- * @returns {Promise<string[]>}
- */
-function getKeysToConnectTo(mapping) {
-    if (isCollectionKey(mapping.key)) {
-        // Find all collection items keys matching this mapping
-        return getAllKeys()
-            .then(keys => _.filter(keys, key => isKeyMatch(mapping.key, key)));
-    }
-
-    // Otherwise it's just a single item, wrap it in a list
-    return Promise.resolve([mapping.key]);
-}
-
-/**
  * Subscribes a react component's state directly to a store key
  *
  * @param {object} mapping the mapping information to connect Onyx to the components state
@@ -417,27 +367,34 @@ function connect(mapping) {
             }
         }
     })
-        .then(() => getKeysToConnectTo(mapping))
-        .then((matchingKeys) => {
-            // When there are no matching keys, initialize the value with null
+        .then(getAllKeys)
+        .then((keys) => {
+            // Find all the keys matched by the config key
+            const matchingKeys = _.filter(keys, key => isKeyMatch(mapping.key, key));
+
+            // If the key being connected to does not exist, initialize the value with null
             if (matchingKeys.length === 0) {
                 sendDataToConnection(mapping, null);
-                return [];
+                return;
             }
 
-            return multiGet(matchingKeys);
-        })
-        .then((result) => {
-            /* For React components subscribed to collections we want to send the data
-            *  as a single object and trigger setState just once */
+            // When using a callback subscriber we will trigger the callback
+            // for each key we find. It's up to the subscriber to know whether
+            // to expect a single key or multiple keys in the case of a collection.
+            // React components are an exception since we'll want to send their
+            // initial data as a single object when using collection keys.
             if (mapping.withOnyxInstance && isCollectionKey(mapping.key)) {
-                return sendDataToConnection(mapping, result);
+                Promise.all(_.map(matchingKeys, key => get(key)))
+                    .then(values => _.reduce(values, (finalObject, value, i) => ({
+                        ...finalObject,
+                        [matchingKeys[i]]: value,
+                    }), {}))
+                    .then(val => sendDataToConnection(mapping, val));
+            } else {
+                _.each(matchingKeys, (key) => {
+                    get(key).then(val => sendDataToConnection(mapping, val, key));
+                });
             }
-
-            /* When using a callback subscriber we will trigger the callback
-            *  for each key we find. It's up to the subscriber to know whether
-            *  to expect a single key or multiple keys in the case of a collection. */
-            _.each(result, (val, key) => sendDataToConnection(mapping, val, key));
         });
 
     return connectionID;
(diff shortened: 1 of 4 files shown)


kidroca:
There's not much to be seen in a visual comparison. I expected to see the LHN much faster, but it seems Onyx is not the main bottleneck ATM 

#### SBS Before/After Android Physical Device
https://user-images.githubusercontent.com/12156624/125672045-fab3eb46-2234-4194-9bef-d092e560cd8d.mp4
- I've trimmed the videos to the point where the JS starts to run, otherwise there is like a ~10sec. splash screen 
- This should be faster with a `prod` build, but I don't have the signing certificate needed to make one

Overall I can feel the app is a bit smoother
There's a noticeable improvement on iOS as well, at one point I've removed the full screen loader, to test a bit and it wasn't even needed - chats rendered instantly

cc @quinthar 
Benchmark info attached in the main description

kidroca:
Here's the E.cash PR where you can build and try out the changes: https://github.com/Expensify/Expensify.cash/pull/4040 on E.cash

lib/Onyx.js
```diff
@@ -154,7 +151,7 @@ function removeLastAccessedKey(key) {
  */
 function addLastAccessedKey(key) {
     // Only specific keys belong in this list since we cannot remove an entire collection.
-    if (isCollectionKey(key) || !isSafeEvictionKey(key)) {
+    if (isCollectionKey(key)) {
```
kidroca, line 154: This was updated to allow to keep more items in the access list
To preserve the same functionality the check was moved here: [Line 481](https://github.com/Expensify/react-native-onyx/pull/88/files#diff-146c57cda221f8f5ee5c808eee0c6692b02c4ac2b8818d0104896a04be78e78aR481)

lib/Onyx.js
```diff
@@ -492,7 +476,10 @@ function remove(key) {
  */
 function evictStorageAndRetry(error, ionMethod, ...args) {
     // Find the first key that we can remove that has no subscribers in our blocklist
-    const keyForRemoval = _.find(recentlyAccessedKeys, key => !evictionBlocklist[key]);
+    const keyForRemoval = _.find(
+        recentlyAccessedKeys,
+        key => !evictionBlocklist[key] && isSafeEvictionKey(key),
```
kidroca, line 481: This check `isSafeEvictionKey` was moved here to allow to use `recentlyAccessedKeys` to actually put all recently accessed keys to that list

marcaaron:
Hey @kidroca, Thanks for the changes + metrics! Super exciting stuff! 

Before diving into any reviews I'd like to focus on one change at a time to make them easier to evaluate. It seems like there's a mix of refactoring, cleanup, and performance stuff going on - which is great to see, but makes it difficult to get a sense for the impact of each individual change + also harder still to test and track any regressions if they pop up.

Let's create PRs for each of these individual changes, review each one in isolation, and discuss anything that requires further explanation in that PR. It would also be good to prioritize the PRs based on which ones have the largest impacts on performance so we can get those out sooner than later.

----
**Specific Notes / Questions:**

>The first commit has and updated Onyx.printMetrics function. You can use it as a base to compare before/after results. It prints a general table on the console which is easy to copy and put in excel.

>Refactor: merge was refactored so that it's promises are resolved when a queue for a key is done. This helps with measuring the function more accurately

Sounds like these two could maybe be moved into a separate PR that intends to clean up the timing stuff. Also sounds like this would be a lower priority since there is no direct impact on performance.

>Fix: some connections happen even before Onyx.init is called. This should not be allowed as the initialKeyStates are applied during init. Added a deferred tasks that would queue incoming connections and only make them after init, this also gives the rest of the app init some breathing space

Can you elaborate on breathing space? Is that another way of saying there was a performance improvement associated with this?

>Perf: Update set to check with cache and skip writing to storage. Cache reflects what's already on disc, if we have the exact value in cache it means it didn't change and we don't need to make a write

Slightly confused about this. If we have a value in the cache why would we assume that we don't need to make a write? To clarify, are you proposing we remove the ability to call `Onyx.set()` and overwrite a value?

>Perf: Updated cache clearing strategy. It's a mix of LRU and active connections. The cache works in the following way: each time a connection is terminated we run a cleanCache function. The function removes dated items from cache, if the item is still being used by a component it's kept in cache. By default it will try to keep up to ~150 keys in cache, can be a bit more due to active connections to old items

How is this related to performance? Seems like maybe it will help with memory consumption? I think this would be a good change to evaluate separate from the other changes so we can look at impacts on memory. But let me know if there's some other benefit here.

>Perf: Updated defaultKeyStates handling. Merging these keys one by one was taking a significant amount of time, even though they are only 4. Instead of writing them to storage, we read data from storage and merge it along with the defaults to the cache. Then we never allow for the defaultKeys to be removed from cache

This would also be good to see as a separate PR. I'm not sure what the default behavior of this is now, but surprised that it is taking up a significant amount of time since, as you mentioned, there are only 4 keys.

kidroca:
@marcaaron 

> **Specific Notes / Questions:**
> 
> > The first commit has and updated Onyx.printMetrics function. You can use it as a base to compare before/after results. It prints a general table on the console which is easy to copy and put in excel.
> 
> > Refactor: merge was refactored so that it's promises are resolved when a queue for a key is done. This helps with measuring the function more accurately
> 
> Sounds like these two could maybe be moved into a separate PR that intends to clean up the timing stuff. Also sounds like this would be a lower priority since there is no direct impact on performance.

These are things that would help anyone making benchmarks and using the `Onyx.printMetrics()` function
If I'm the only one that's using it I'm fine to just have this change live locally and not be committed
But if more people use it, it would be best to merge this first so the same print format can be used on the before/after comparisons



> > Fix: some connections happen even before Onyx.init is called. This should not be allowed as the initialKeyStates are applied during init. Added a deferred tasks that would queue incoming connections and only make them after init, this also gives the rest of the app init some breathing space
> 
> Can you elaborate on breathing space? Is that another way of saying there was a performance improvement associated with this?

Yes, this carries performance improvements
1.  Postponing the connections until after `Onyx.init` finishes improved loading
2. Also it gives a better batching opportunity. Since all the connections that happen will wait for init, when that happens they would be batched together (by AsyncStorage) and resolve with one `multiGet` (if we implement this) 


> > Perf: Update set to check with cache and skip writing to storage. Cache reflects what's already on disc, if we have the exact value in cache it means it didn't change and we don't need to make a write
> 
> Slightly confused about this. If we have a value in the cache why would we assume that we don't need to make a write? To clarify, are you proposing we remove the ability to call `Onyx.set()` and overwrite a value?

Quite a few times we're trying to "overwrite" something that didn't change, the most notable one is `modal` where I would see 19 consecutive calls to write `false` and on top of that the value in storage is already `false` 

When we `set` something we `cache` the updated value
https://github.com/Expensify/react-native-onyx/blob/86be75945a47f9015b9a37ca738e8fdd36e42de3/lib/Onyx.js#L515-L517
Even if this happens via other means like `mergeCollection` - if cache has a value it's always the most recent up to date value 
So when we try to write a value, but we have a cached key and the exact same value in `cache` we can skip the write
https://github.com/Expensify/react-native-onyx/pull/88/commits/c5a4ba6e8fe851493ef674263b4f40ac031cba97
I've tested with a `log` in the condition here and very often we try to write the same value for `modal`, `network`, `iou`, `session` 


> > Perf: Updated cache clearing strategy. It's a mix of LRU and active connections. The cache works in the following way: each time a connection is terminated we run a cleanCache function. The function removes dated items from cache, if the item is still being used by a component it's kept in cache. By default it will try to keep up to ~150 keys in cache, can be a bit more due to active connections to old items
> 
> How is this related to performance? Seems like maybe it will help with memory consumption? I think this would be a good change to evaluate separate from the other changes so we can look at impacts on memory. But let me know if there's some other benefit here.

This helps on a few fronts
1. it helps with the optimization above - having more items in cache, we have more knowledge of what's already written on disk and prevent unnecessary writes
2. Switching between the recently used chats happen faster as the information is still in cache

I've disabled clearing cache entirely and it improved speeds, then I came up with this LRU update, which gives pretty much the same performance while still clears cache when it can


> > Perf: Updated defaultKeyStates handling. Merging these keys one by one was taking a significant amount of time, even though they are only 4. Instead of writing them to storage, we read data from storage and merge it along with the defaults to the cache. Then we never allow for the defaultKeys to be removed from cache
> 
> This would also be good to see as a separate PR. I'm not sure what the default behavior of this is now, but surprised that it is taking up a significant amount of time since, as you mentioned, there are only 4 keys.

Yes I was surprised too, but it was a repeating pattern from the numerous benches that I did (Android) - these 4 keys would start fetching at 1.5sec after launch and resolve at the 3rd or 4th second after init - ~2sec to retrieve them
I think it's down to the fact that many thing are running through the bridge at that time
With the update I'm proposing it takes about 0.6sec and only one trip through the bridge

---

I'm ready to start opening PRs, just need to know whether I should start with the `Onyx.printMetrics()` changes (1st item in my response) or skip it altogether

marcaaron:
> These are things that would help anyone making benchmarks and using the Onyx.printMetrics() function
If I'm the only one that's using it I'm fine to just have this change live locally and not be committed
But if more people use it, it would be best to merge this first so the same print format can be used on the before/after comparisons

Good point, let's start with that then!

>Yes, this carries performance improvements Postponing the connections until after Onyx.init finishes improved loading Also it gives a better batching opportunity. Since all the connections that happen will wait for init, when that happens they would be batched together (by AsyncStorage) and resolve with one multiGet (if we implement this)

Ok I'm not too sure I get what we want to do here yet, but sounds like it should be it's own PR and we can evaluate the impact on app start time?

>Quite a few times we're trying to "overwrite" something that didn't change, the most notable one is modal where I would see 19 consecutive calls to write false and on top of that the value in storage is already false

Nice. I agree it doesn't make sense to write data that hasn't changed. The part I was missing from this proposal is how should we tell the difference between the data that has changed and the data that has not changed? Looking at the commit you shared and it seems like we'd use `_.isEqual()` for that. 

This makes sense to me, but it seems like there might be some cases where we know this data has changed and are going to perform an additional check on it now? e.g. merging some large JSON blob would now require that we perform this equality check on it in addition to saving it to storage.

>it helps with the optimization above - having more items in cache, we have more knowledge of what's already written on disk and prevent unnecessary writes

Ok so sounds like we can maybe start with that first optimization and then add this one in after?

>Yes I was surprised too, but it was a repeating pattern from the numerous benches that I did (Android) - these 4 keys would start fetching at 1.5sec after launch and resolve at the 3rd or 4th second after init - ~2sec to retrieve them I think it's down to the fact that many thing are running through the bridge at that time. With the update I'm proposing it takes about 0.6sec and only one trip through the bridge

Nice. It sounds like maybe this improvement can be combined with the deferred connection logic since they are both focusing on the app start time?

**Let me know if you are good with this game plan...**

PR1 - Add the new metrics improvements so that we can use them in the benchmarking the rest of the changes
PR2 - Focus on `defaultKeyStates ` handling and postpone connections until after `Onyx.init` finishes (benchmark app start time)
PR3 - Prevent unnecessary writes when data has not changed by adding `_.isEqual()`
PR4 - Improve cache clearing logic so that data is retained longer and more unnecessary writes can be prevented

kidroca:
> Let me know if you are good with this game plan...

Yes that works

> e.g. merging some large JSON blob would now require that we perform this equality check on it in addition to saving it to storage.

The check happens really fast - we're dealing with objects available in memory, and we're not computing the full diff between the objects the check is over as soon the first difference is found.
Maybe we can find some information on how big a JSON object needs to be in order to negatively affects this, but IMO for data in memory it shouldn't be a problem.
Writing times improved (Android) like 3x for the worst case - 984ms down to 342ms, and 6x for the average case 384ms down to 64ms, but what's better is that we're removing this noise from the bridge channel and other stuff can run faster too
iOS improved as well - from 98ms on average down to 29ms

marcaaron:
Sounds great, thanks

PR #89, opened 2021-07-16
Update Onyx.printMetrics format

kidroca:
@marcaaron 

### Details
This PR does not change how Onyx work and only affects dev usages
The changes here update the print format so that a summary table is printed
This table can be easily selected and pasted in xls

#### Sample Output
Screenshot 2021-07-16 at 2 57 28

### Related Issues

https://github.com/Expensify/Expensify.cash/issues/2667

### Automated Tests

Covered by existing tests

### DIFF
--- a/lib/decorateWithMetrics.js
+++ b/lib/decorateWithMetrics.js
@@ -62,9 +62,12 @@ function sum(list, prop) {
 }
 
 /**
- * Returns total, average time and all captured stats mapped under
- * summaries.methodName -> method stats
+ * Aggregates and returns benchmark information
  * @returns {{summaries: Record<string, Object>, totalTime: number, lastCompleteCall: *}}
+ * An object with
+ * - `totalTime` - total time spent by decorated methods
+ * - `lastCompleteCall` - millisecond since launch the last call completed at
+ * - `summaries` - mapping of all captured stats: summaries.methodName -> method stats
  */
 function getMetrics() {
     const summaries = _.chain(stats)
@@ -156,7 +159,7 @@ function printMetrics(raw = false) {
             return {
                 methodName,
                 ...prettyTimes,
-                last: toDuration(lastCall.endTime, raw),
+                'time last call completed': toDuration(lastCall.endTime, raw),
                 calls: calls.length,
                 prettyCalls,
             };
@@ -166,7 +169,7 @@ function printMetrics(raw = false) {
     /* eslint-disable no-console */
     console.group('Onyx Benchmark');
     console.info('  Total: ', toDuration(totalTime, raw));
-    console.info('  Last call made at: ', toDuration(lastCompleteCall.endTime, raw));
+    console.info('  Last call finished at: ', toDuration(lastCompleteCall.endTime, raw));
 
     console.table(prettyData.map(({prettyCalls, ...summary}) => summary));
 
--- a/lib/decorateWithMetrics.js
+++ b/lib/decorateWithMetrics.js
@@ -64,7 +64,7 @@ function sum(list, prop) {
 /**
  * Returns total, average time and all captured stats mapped under
  * summaries.methodName -> method stats
- * @returns {{averageTime: number, summaries: Record<string, Object>, totalTime: number}}
+ * @returns {{summaries: Record<string, Object>, totalTime: number}}
  */
 function getMetrics() {
     const summaries = _.chain(stats)
@@ -88,11 +88,9 @@ function getMetrics() {
         .value();
 
     const totalTime = sum(_.values(summaries), 'total');
-    const averageTime = (totalTime / _.size(summaries)) || 0;
 
     return {
         totalTime,
-        averageTime,
         summaries,
     };
 }
@@ -117,7 +115,7 @@ function toHumanReadableDuration(millis) {
  * and a table of individual calls
  */
 function printMetrics() {
-    const {totalTime, averageTime, summaries} = getMetrics();
+    const {totalTime, summaries} = getMetrics();
 
     const prettyData = _.chain(summaries)
         .filter(method => method.avg > 0)
@@ -148,7 +146,6 @@ function printMetrics() {
     /* eslint-disable no-console */
     console.group('Onyx Benchmark');
     console.info('  Total: ', toHumanReadableDuration(totalTime));
-    console.info('  Average: ', toHumanReadableDuration(averageTime));
 
     console.table(prettyData.map(({prettyCalls, ...summary}) => summary));
 
--- a/lib/decorateWithMetrics.js
+++ b/lib/decorateWithMetrics.js
@@ -64,7 +64,7 @@ function sum(list, prop) {
 /**
  * Returns total, average time and all captured stats mapped under
  * summaries.methodName -> method stats
- * @returns {{summaries: Record<string, Object>, totalTime: number}}
+ * @returns {{summaries: Record<string, Object>, totalTime: number, lastCompleteCall: *}}
  */
 function getMetrics() {
     const summaries = _.chain(stats)
@@ -74,6 +74,7 @@ function getMetrics() {
             const avg = (total / calls.length) || 0;
             const max = _.max(calls, 'duration').duration || 0;
             const min = _.min(calls, 'duration').duration || 0;
+            const lastCall = _.max(calls, 'endTime');
 
             return [methodName, {
                 methodName,
@@ -81,6 +82,7 @@ function getMetrics() {
                 max,
                 min,
                 avg,
+                lastCall,
                 calls,
             }];
         })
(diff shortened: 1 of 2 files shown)


lib/decorateWithMetrics.js
```diff
@@ -74,30 +74,42 @@ function getMetrics() {
...
                 avg,
+                lastCall,
                 calls,
             }];
         })
         .object() // Create a map like methodName -> StatSummary
         .value();
 
     const totalTime = sum(_.values(summaries), 'total');
-    const averageTime = (totalTime / _.size(summaries)) || 0;
```
kidroca, line 91: Removed this as it doesn't provide anything useful 
it's just `total time spent / number of decorated methods` 

Added `lastCall` per each summary and `lastCompleteCall` that is the most
recent operation of all that ended 

I use this to track when we're done making operations with Onyx e.g. 
- when we're done reading from Onyx during init 
- when we're done reading during after switching to a new chat

lib/decorateWithMetrics.js
```diff
@@ -115,30 +127,55 @@ function toHumanReadableDuration(millis) {
  * Print extensive information on the dev console
  * max, min, average, total time for each method
  * and a table of individual calls
+ *
+ * @param {boolean} [raw=false] setting this to true will print raw instead of human friendly times
+ * Useful when you copy the printed table to excel and let excel do the number formatting
```
kidroca, line 132: Having the raw number in excel means you can also do sorting or other number related operations which aren't easily performed with a value like `"565.344ms"`

lib/decorateWithMetrics.js
```diff
@@ -115,30 +127,55 @@ function toHumanReadableDuration(millis) {
...
+                calls: calls.length,
+                prettyCalls,
+            };
+        })
+        .value();
+
+    /* eslint-disable no-console */
+    console.group('Onyx Benchmark');
+    console.info('  Total: ', toDuration(totalTime, raw));
+    console.info('  Last call made at: ', toDuration(lastCompleteCall.endTime, raw));
```
marcaaron, line 169: Not really when it was made but when it _finished_ right?

marcaaron (changes requested):
This looks good to me. But I am hoping we can maybe clear up what `last` means. Probably just needs a small language tweak here and there to make things more obvious for the next ones to use this tool.

lib/decorateWithMetrics.js
```diff
@@ -115,30 +127,55 @@ function toHumanReadableDuration(millis) {
...
  * max, min, average, total time for each method
  * and a table of individual calls
+ *
+ * @param {boolean} [raw=false] setting this to true will print raw instead of human friendly times
+ * Useful when you copy the printed table to excel and let excel do the number formatting
  */
-function printMetrics() {
-    const {totalTime, averageTime, summaries} = getMetrics();
+function printMetrics(raw = false) {
+    const {totalTime, summaries, lastCompleteCall = {endTime: -1}} = getMetrics();
```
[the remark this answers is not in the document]
kidroca: This is a precaution for when `printMetrics` is called before any calls are made at all
Since `lastCompleteCall` is derived from `calls` data, and there are none it will be `undefined` 
I've picked `-1` as it looks clear that no call has happened, I guess `0` can work as well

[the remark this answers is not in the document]
marcaaron: Ah, I see. In that case, I could expect to see something like `"Unavailable"`, but the solution you have here works.

lib/decorateWithMetrics.js
```diff
@@ -6,6 +6,11 @@ import _ from 'underscore';
  */
 let stats = {};
 
+/* For some reason `performance.now()` does not start from `0` but a very large offset
+* like `508,080,000` see: https://github.com/facebook/react-native/issues/30069
+* Capturing an offset allows us to record start/ends times relative to app launch time */
+const APP_LAUNCH_TIME = performance.now();
```
kidroca, line 12: I've found this issue after I switched to the "new" debugging flow today: https://expensify.slack.com/archives/C01GTK53T8Q/p1626422647390900?thread_ts=1626297548.276300&cid=C01GTK53T8Q

kidroca: It's related to the "old" chrome debugger not being used

lib/decorateWithMetrics.js
```diff
@@ -6,6 +6,11 @@ import _ from 'underscore';
  */
 let stats = {};
 
+/* For some reason `performance.now()` does not start from `0` but a very large offset
+* like `508,080,000` see: https://github.com/facebook/react-native/issues/30069
+* Capturing an offset allows us to record start/ends times relative to app launch time */
```
marcaaron, line 11: Interesting, since it is supposed to be [more reliable than `Date`](https://github.com/Expensify/Expensify.cash/issues/3889#issuecomment-877395187), but still has this issue? 🤔 

Not a blocker, but I'm not sure I'd expect it this to be `0`. `performance.now()` returns elapsed since the time origin (or is supposed to at least).

kidroca:
@marcaaron 
I'm about to open a continuation PR for `Onyx.printMetrics` 

Since the update for Reanimated 2 (TurboModules/JSI)
It's no longer possible to use the old (Chrome) debugging flow

`react-native` runtime does not implement the `console.table` we're using to print information
It was provided by Chrome, by launching the (old) debug flow

The other PR adds MD table formatting like this 
image

Which would allow you to paste data like this:

### Onyx Benchmark
  - Total: 13.4min
  - Last call finished at: 31.95sec

|        method        | total time spent |   max    |    min    |    avg    | time last call completed | calls made |
|----------------------|-----------------:|---------:|----------:|----------:|-------------------------:|-----------:|
| Onyx:set             |           1.1min | 19.48sec | 120.607ms |   2.25sec |                 31.95sec |         30 |
| Onyx:getAllKeys      |           8.8min |  6.27sec |   0.140ms |   1.90sec |                 30.64sec |        277 |
| Onyx:merge           |         48.43sec | 19.49sec |  31.937ms |   1.73sec |                 31.95sec |         28 |
| Onyx:mergeCollection |          7.83sec |  2.60sec | 663.751ms |   1.57sec |                 31.95sec |          5 |
| Onyx:get             |           2.6min | 19.12sec |   0.064ms | 751.025ms |                 30.63sec |        207 |

lib/decorateWithMetrics.js
```diff
@@ -20,7 +25,7 @@ function decorateWithMetrics(func, alias = func.name) {
     stats[alias] = [];
 
     function decorated(...args) {
-        const startTime = performance.now();
+        const startTime = performance.now() - APP_LAUNCH_TIME;
```
marcaaron, line 28: I'm trying to understand the reliability of this next usage of `performance.now()` here and which metric this will make more reliable. Is it just the `endTime`? Since all the other metrics are relative to `performance.now()` but that one will tell us the actual time from when Onyx started doing work and when it finished for each method?

marcaaron (approved):
Changes look good to me. Left a couple questions about the `performance.now()` workaround, but won't block on them.

marcaaron:
@kidroca ah that's annoying, I wasn't expecting an effect like that from updating that library 🤔  but the markdown table looks very slick 👏

lib/decorateWithMetrics.js
kidroca: Yeah it will never be `0` but the origin is definitely not app launch the benchmark would say something like "Last call finished at: 3940min" which is useless

lib/decorateWithMetrics.js
kidroca: This doesn't make anything more reliable, just keeps our start/end times relative to app launch 

Differences between `performance.now()` are reliable but the initial time given by `performance.now()` is ridiculously high - and I can't match the value to anything - it's not the current unix ts, it's not the time since app launch
It's a value like 508 million and increases as time passes. Might be relative to midnight 00:00 UTC ? But that doesn't help us track boot time 😄 

On the old flow, when this was provided by Chrome, the value will be very close to `0` initially and so you know a value of `1000` means something started/ended a second past app launch

By capturing our own reference point APP_LAUNCH_TIME we keep start/end times relative to that point
so we have more useful information

kidroca: If I remove this, I'll get this ATM: "Last call finished at: 8907.4min" so it's not midnight
It might be the time since last reboot, let me check
Yep seems to be it: "Last call finished at: 10.0min"
I guess it uses some clock managed by the OS and that's what it returns

lib/decorateWithMetrics.js
marcaaron: Got it that makes more sense to me now. Thanks for the explanation.

lib/decorateWithMetrics.js
marcaaron: Haha, ok awesome, thanks for the explanation. 
Very weird, but now that we are interested in something besides relative times also very necessary.

PR #90, opened 2021-07-16
`Onxy.printMetrics` MD, CSV and JSON formats

kidroca:
@marcaaron 

### Details

This is a continuation of #89
This PR adds more print options: 
- print data as Markdown tables 
- as csv strings (easy import to Excel)
- as json

It necessary to have something like this as it aids capturing and saving benchmark information 

This PR does not change how Onyx works and only affects dev, debug, benchmark usages

#### MD (Console)
image

#### CSV
image

When the data is too much it's not printed in full, but can be expanded and/or copied
image

#### JSON
image

#### MD (String)
image
This is useful when `__DEV__` is `false` as calling `console.info` would not output anything 
image

### Related Issues

N/A

### Automated Tests

Covered by existing tests, 
Manually tested the printing functionality

### DIFF
--- a/lib/MDTable.js
+++ b/lib/MDTable.js
@@ -1,0 +1,38 @@
+import AsciTable from 'ascii-table';
+
+class MDTable extends AsciTable {
+    static factory({
+        title, heading, leftAligned = [], rows = []
+    }) {
+        const table = new MDTable({title, heading, rows});
+        table.leftAligned = leftAligned;
+
+        /* By default we want everything aligned to the right as most values are numbers
+         * we just override the columns that are not right aligned */
+        heading.forEach((name, idx) => table.setAlign(idx, AsciTable.RIGHT));
+        leftAligned.forEach(idx => table.setAlign(idx, AsciTable.LEFT));
+
+        return table;
+    }
+
+    toString() {
+        let idx = -1;
+        const ascii = super.toString()
+            .replace(/-\|/g, () => {
+                /* we replace "----|" with "---:|" to align the data to the right in MD */
+                idx++;
+
+                if (this.leftAligned.includes(idx)) {
+                    return '-|';
+                }
+
+                return ':|';
+            });
+
+        // strip the top and the bottom row (----) to make an MD table
+        const md = ascii.split('\n').slice(1, -1).join('\n');
+        return md;
+    }
+}
+
+export default MDTable;
--- a/lib/MDTable.js
+++ b/lib/MDTable.js
@@ -1,28 +1,26 @@
 import AsciTable from 'ascii-table';
 
 class MDTable extends AsciTable {
-    static factory({
-        title, heading, leftAligned = [], rows = []
-    }) {
-        const table = new MDTable({title, heading, rows});
-        table.leftAligned = leftAligned;
-
-        /* By default we want everything aligned to the right as most values are numbers
-         * we just override the columns that are not right aligned */
-        heading.forEach((name, idx) => table.setAlign(idx, AsciTable.RIGHT));
-        leftAligned.forEach(idx => table.setAlign(idx, AsciTable.LEFT));
-
-        return table;
-    }
-
+    /**
+     * Create a CSV string from the table data
+     * @returns {string}
+     */
     toCSV() {
         return [this.getHeading(), ...this.getRows()].join('\n');
     }
 
+    /**
+     * Create a JSON string from the table data
+     * @returns {string}
+     */
     toJSON() {
         return JSON.stringify(super.toJSON());
     }
 
+    /**
+     * Create a MD string from the table data
+     * @returns {string}
+     */
     toString() {
         // Ignore modifying the first |---| for titled tables
         let idx = this.getTitle() ? -2 : -1;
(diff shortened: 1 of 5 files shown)


kidroca:
Oops some leftovers that should be cleaned up, will push another commit

lib/MDTable.js
```diff
@@ -0,0 +1,69 @@
...
+            .replace(/-\|/g, () => {
+                /* we replace "----|" with "---:|" to align the data to the right in MD */
+                idx++;
+
+                if (idx < 0 || this.leftAlignedCols.includes(idx)) {
+                    return '-|';
+                }
+
+                return ':|';
+            });
```
kidroca, line 37: I guess this is the most weird part of the PR

Markdown tables are left aligned by default 
To align a column you add a `:` column on the side you want it to align (or on both sides for centered)
```
|----------|---------:|----------:|
```

When we're printing a table with a title we want to skip the first row which looks like:
```
|            Onyx:set            |
|--------------------------------|
```

That's why we start with `-2` on that case

lib/MDTable.js
```diff
@@ -0,0 +1,69 @@
...
+
+        // strip the top and the bottom row (----) to make an MD table
+        const md = ascii.split('\n').slice(1, -1).join('\n');
+        return md;
```
kidroca, line 41: The tables are surrounded by `------` border on the top and the bottom `------`
I was always removing them when I was copying as it didn't render correctly in MD

lib/decorateWithMetrics.js
```diff
@@ -136,55 +137,83 @@ function toDuration(millis, raw = false) {
...
+ * @param {Object} [options]
+ * @param {boolean} [options.raw=false] - setting this to true will print raw instead of human friendly times
  * Useful when you copy the printed table to excel and let excel do the number formatting
+ * @param {'console'|'csv'|'json'|'string'} [options.format=console] The output format of this function
+ * `string` is useful when __DEV__ is set to `false` as writing to the console is disabled, but the result of this
+ * method would still get printed as output
+ * @returns {string|undefined}
  */
-function printMetrics(raw = false) {
-    const {totalTime, summaries, lastCompleteCall = {endTime: -1}} = getMetrics();
```
kidroca, line 143: Changed this to print `Last call finished at: N/A` when there are no calls made yet (or were cleared)

lib/decorateWithMetrics.js
```diff
@@ -136,55 +137,83 @@ function toDuration(millis, raw = false) {
...
+
+            const callsTable = MDTable.factory({
+                title: methodName,
+                heading: ['start time', 'end time', 'duration', 'args'],
+                leftAlignedCols: [3],
+                rows: calls.map(call => ([
+                    toDuration(call.startTime, raw),
+                    toDuration(call.endTime, raw),
+                    toDuration(call.duration, raw),
+                    call.args.map(String).join(', ').slice(0, 60), // Restrict cell width to 60 chars max
```
kidroca, line 179: We're limiting this as it breaks the table when it's too large 
I find it better to not JSON.stringify the data and instead keep it simple like: `account, [object Object]` 
otherwise it adds too much noise

lib/decorateWithMetrics.js
```diff
@@ -136,55 +137,83 @@ function toDuration(millis, raw = false) {
...
-        console.table(method.prettyCalls);
+    /* eslint-disable no-console */
+    console.info(mainOutput.join('\n'));
+    methodCallTables.forEach((table) => {
+        console.groupCollapsed(table.getTitle());
+        console.info(table.toString());
         console.groupEnd();
     });
-
-    console.groupEnd();
```
kidroca, line 187: I wonder whether we need console printing at all 
The only convenience here is `console.groupCollapsed`

package.json
```diff
@@ -12,6 +12,7 @@
     "test": "jest"
   },
   "dependencies": {
+    "ascii-table": "0.0.9",
```
kidroca, line 15: I was expecting to find something that will build directly as MD, but this was the closest thing I've found 
This library required the least modifications to achieve an MD table
I've tried these:
- https://www.npmjs.com/package/tty-table
- https://www.npmjs.com/package/console-table-printer
- https://www.npmjs.com/package/as-table

johnmlee101:
Is this ready for review?

kidroca:
@johnmlee101 Yes

lib/decorateWithMetrics.js
```diff
@@ -136,62 +138,91 @@ function toDuration(millis, raw = false) {
...
+        return allTables.map((table) => {
+            switch (format.toLowerCase()) {
+                case 'csv':
+                    return table.toCSV();
+                case 'json':
+                    return table.toJSON();
+                default:
+                    return table.toString();
+            }
+        }).join('\n\n');
```
johnmlee101, line 200: Any specific reason we join with 2 new lines?

johnmlee101 (approved):
Looks good to me AFAICT, I asked @marcaaron to do final review

lib/MDTable.js
marcaaron: maybe could use a different token than `Onyx:set` like `Onyx~set` but this seems fine.

marcaaron (approved):

lib/decorateWithMetrics.js
kidroca: To distinct separate tables easier

lib/MDTable.js
kidroca: Individual table names would be title with whatever alias is provided here: 
https://github.com/Expensify/react-native-onyx/blob/master/lib/Onyx.js#L771-L777

PR #91, opened 2021-07-22
Defer initialization and wait for defaultKeyStates

kidroca:
@marcaaron 

##### Make connections wait for Onyx.init
Some connections start really early before Onyx.init is even called
They should wait as Onyx.init adds some initial values that might be needed
Also waiting helps as every call uses `getAllKeys` and the keys are retrieved during init - if connections wait, every following connection would instantly retrieve them from cache, otherwise even though they hook to the same storage read they still create a new promise

##### Read defaults in one go and keep them in cache
Instead of reading default keys, then merging default values to them and then writing them back in storage
We read the default keys in one go, and then write the merged result to cache
We also make sure to never remove the default keys from cache - they don't take much space and are used often

### Related Issues

https://github.com/Expensify/App/issues/2667

### Automated Tests

Covered by existing tests

##### Android Before 
|                    |     Run 1 |     Run 2 |     Run 3 |     Run 4 |     Run 5 |
|--------------------|----------:|----------:|----------:|----------:|----------:|
| Total              | 124.06sec | 118.71sec | 118.43sec | 116.57sec | 114.24sec |
| Last call ended at |   8.26sec |   7.83sec |   7.25sec |   7.13sec |   7.89sec |
- Full data: https://docs.google.com/spreadsheets/d/1lEZ1dZD77D6zUQgB5A6-QTlX3JdVcyARThKh7-dru_4/edit?usp=sharing

##### Android After
|                    |    Run 1 |    Run 2 |    Run 3 |    Run 4 |    Run 5 |
|--------------------|---------:|---------:|---------:|---------:|---------:|
| Total              | 92.22sec | 92.44sec | 92.15sec | 89.81sec | 91.98sec |
| Last call ended at |  7.77sec |  7.41sec |  7.03sec |  7.71sec |  7.55sec |
- Full data: https://docs.google.com/spreadsheets/d/1USrACRs5grnnBZwL9y2eU-QNLr8y1tWzacMCd9hFSWo/edit?usp=sharing

##### iOS Before
|                    |    Run 1 |    Run 2 |    Run 3 |    Run 4 |    Run 5 |
|--------------------|---------:|---------:|---------:|---------:|---------:|
| Total              | 87.91sec | 87.41sec | 87.21sec | 86.29sec | 86.18sec |
| Last call ended at |  4.96sec |  4.47sec |  4.47sec |  4.77sec |  4.71sec |
- Full data: https://docs.google.com/spreadsheets/d/1quU0R5xhRWmH3xsatU-uUMOIPqSy7v7rlZobjwtJk38/edit?usp=sharing

##### iOS After
|                    |    Run 1 |    Run 2 |    Run 3 |    Run 4 |    Run 5 |
|--------------------|---------:|---------:|---------:|---------:|---------:|
| Total              | 66.36sec | 61.48sec | 58.39sec | 67.08sec | 65.94sec |
| Last call ended at |  4.63sec |  4.76sec |  4.72sec |  4.78sec |  4.90sec |
- Full data: https://docs.google.com/spreadsheets/d/1B4k4lidFnIZgngKmjI5EW7MTZ8ORqEUqpdG8taoZtJU/edit?usp=sharing

#### Benchmark info: 
Add `global.Onyx = Onyx` in Eexpensify.js 
Open the dev menu and make sure dev=false 
Disable all breakpoints in the debugger 
Wait for any syncing (top right corner) to end before starting any tests

##### Test   
1. Reload the app 
2. Wait processes to settle 
3. Run `Onyx.printMetrics({ format: 'csv', raw: true })`  
4. Copy the results to file 
5. Import file as new sheet in excel 
 
Run 7 times
Remove the fastest and the slowest runs 

All benchmarks were performed against this hash in Expensify/App: https://github.com/Expensify/App/commit/fc965f68d308167d19d00a29b91e85ca33707882

### DIFF
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -28,6 +28,13 @@ const evictionBlocklist = {};
 // Optional user-provided key value states set when Onyx initializes or clears
 let defaultKeyStates = {};
 
+// Connections can be made before `Onyx.init`. They would wait for this promise before resolving
+const deferredInit = {};
+deferredInit.promise = new Promise((res, rej) => {
+    deferredInit.resolve = res;
+    deferredInit.reject = rej;
+});
+
 /**
  * Get some data from the store
  *
@@ -359,18 +366,20 @@ function connect(mapping) {
         return connectionID;
     }
 
-    // Check to see if this key is flagged as a safe eviction key and add it to the recentlyAccessedKeys list
-    if (mapping.withOnyxInstance && !isCollectionKey(mapping.key) && isSafeEvictionKey(mapping.key)) {
+    // Commit connection only after init passes
+    deferredInit.promise.then(() => {
+        // Check to see if this key is flagged as a safe eviction key and add it to the recentlyAccessedKeys list
+        if (mapping.withOnyxInstance && !isCollectionKey(mapping.key) && isSafeEvictionKey(mapping.key)) {
         // All React components subscribing to a key flagged as a safe eviction
         // key must implement the canEvict property.
-        if (_.isUndefined(mapping.canEvict)) {
+            if (_.isUndefined(mapping.canEvict)) {
             // eslint-disable-next-line max-len
-            throw new Error(`Cannot subscribe to safe eviction key '${mapping.key}' without providing a canEvict value.`);
+                throw new Error(`Cannot subscribe to safe eviction key '${mapping.key}' without providing a canEvict value.`);
+            }
+            addLastAccessedKey(mapping.key);
         }
-        addLastAccessedKey(mapping.key);
-    }
-
-    getAllKeys()
+    })
+        .then(getAllKeys)
         .then((keys) => {
             // Find all the keys matched by the config key
             const matchingKeys = _.filter(keys, key => isKeyMatch(mapping.key, key));
@@ -742,6 +751,9 @@ function init({
         cache.set(key, newValue);
         keyChanged(key, newValue);
     });
+
+    // Give green light to any pending connections
+    deferredInit.resolve();
 }
 
 const Onyx = {
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -4,6 +4,7 @@ import Str from 'expensify-common/lib/str';
 import lodashMerge from 'lodash/merge';
 import {registerLogger, logInfo, logAlert} from './Logger';
 import cache from './OnyxCache';
+import createDeferredTask from './createDeferredTask';
 
 // Keeps track of the last connectionID that was used so we can keep incrementing it
 let lastConnectionID = 0;
@@ -29,11 +30,7 @@ const evictionBlocklist = {};
 let defaultKeyStates = {};
 
 // Connections can be made before `Onyx.init`. They would wait for this promise before resolving
-const deferredInit = {};
-deferredInit.promise = new Promise((res, rej) => {
-    deferredInit.resolve = res;
-    deferredInit.reject = rej;
-});
+const deferredInit = createDeferredTask();
 
 /**
  * Get some data from the store
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -203,9 +203,11 @@ function addToEvictionBlockList(key, connectionID) {
  * the recently accessed list when initializing the app. This
  * enables keys that have not recently been accessed to be
  * removed.
+ *
+ * @returns {Promise}
  */
 function addAllSafeEvictionKeysToRecentlyAccessedList() {
-    getAllKeys()
+    return getAllKeys()
         .then((keys) => {
             _.each(evictionAllowList, (safeEvictionKey) => {
                 _.each(keys, (key) => {
@@ -636,7 +638,7 @@ function merge(key, val) {
 /**
  * Merge user provided default key value pairs.
  *
- * @returns {Promise<void>}
+ * @returns {Promise}
  */
 function initializeWithDefaultKeyStates() {
     return AsyncStorage.multiGet(_.keys(defaultKeyStates))
@@ -757,11 +759,13 @@ function init({
 
     // Let Onyx know about which keys are safe to evict
     evictionAllowList = safeEvictionKeys;
-    addAllSafeEvictionKeysToRecentlyAccessedList();
 
     // Initialize all of our keys with data provided then give green light to any pending connections
-    initializeWithDefaultKeyStates()
-        .finally(deferredInit.resolve); // Proceed even if the above task fails
+    Promise.all([
+        addAllSafeEvictionKeysToRecentlyAccessedList(),
+        initializeWithDefaultKeyStates()
+    ])
+        .then(deferredInit.resolve);
 
     // Update any key whose value changes in storage
     registerStorageEventListener((key, newValue) => {
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -414,6 +414,11 @@ function connect(mapping) {
  * @param {string} key
  */
 function cleanCache(key) {
+    // Don't remove default keys from cache, they don't take much memory and are accessed frequently
+    if (_.has(defaultKeyStates, key)) {
+        return;
+    }
+
     const hasRemainingConnections = _.some(callbackToStateMapping, {key});
 
     // When the key is still used in other places don't remove it from cache
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -630,9 +630,21 @@ function merge(key, val) {
 
 /**
  * Merge user provided default key value pairs.
+ *
+ * @returns {Promise<void>}
  */
 function initializeWithDefaultKeyStates() {
-    _.each(defaultKeyStates, (state, key) => merge(key, state));
+    return AsyncStorage.multiGet(_.keys(defaultKeyStates))
+        .then((pairs) => {
+            const asObject = _.chain(pairs)
+                .map(([key, val]) => [key, val && JSON.parse(val)])
+                .object()
+                .value();
+
+            const merged = lodashMerge(asObject, defaultKeyStates);
+            cache.merge(merged);
+            _.each(merged, (val, key) => keyChanged(key, val));
+        });
 }
 
 /**
@@ -707,7 +719,9 @@ function mergeCollection(collectionKey, collection) {
 /**
  * Initialize the store with actions and listening for storage events
  *
- * @param {Object} [options]
+ * @param {Object} options
+ * @param {Object} [options.keys]
+ * @param {Object} [options.initialKeyStates]
  * @param {String[]} [options.safeEvictionKeys] This is an array of keys
  * (individual or collection patterns) that when provided to Onyx are flagged
  * as "safe" for removal. Any components subscribing to these keys must also
@@ -740,17 +754,15 @@ function init({
     evictionAllowList = safeEvictionKeys;
     addAllSafeEvictionKeysToRecentlyAccessedList();
 
-    // Initialize all of our keys with data provided
-    initializeWithDefaultKeyStates();
+    // Initialize all of our keys with data provided then give green light to any pending connections
+    initializeWithDefaultKeyStates()
+        .finally(deferredInit.resolve); // Proceed even if the above task fails
 
     // Update any key whose value changes in storage
     registerStorageEventListener((key, newValue) => {
         cache.set(key, newValue);
         keyChanged(key, newValue);
     });
-
-    // Give green light to any pending connections
-    deferredInit.resolve();
 }
 
 const Onyx = {
@@ -784,6 +796,7 @@ function applyDecorators() {
     merge = decorate.decorateWithMetrics(merge, 'Onyx:merge');
     mergeCollection = decorate.decorateWithMetrics(mergeCollection, 'Onyx:mergeCollection');
     getAllKeys = decorate.decorateWithMetrics(getAllKeys, 'Onyx:getAllKeys');
+    initializeWithDefaultKeyStates = decorate.decorateWithMetrics(initializeWithDefaultKeyStates, 'Onyx:defaults');
     /* eslint-enable */
 
     // Re-expose decorated methods
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -29,8 +29,8 @@ const evictionBlocklist = {};
 // Optional user-provided key value states set when Onyx initializes or clears
 let defaultKeyStates = {};
 
-// Connections can be made before `Onyx.init`. They would wait for this promise before resolving
-const deferredInit = createDeferredTask();
+// Connections can be made before `Onyx.init`. They would wait for this task before resolving
+const deferredInitTask = createDeferredTask();
 
 /**
  * Get some data from the store
@@ -366,18 +366,19 @@ function connect(mapping) {
     }
 
     // Commit connection only after init passes
-    deferredInit.promise.then(() => {
-        // Check to see if this key is flagged as a safe eviction key and add it to the recentlyAccessedKeys list
-        if (mapping.withOnyxInstance && !isCollectionKey(mapping.key) && isSafeEvictionKey(mapping.key)) {
-        // All React components subscribing to a key flagged as a safe eviction
-        // key must implement the canEvict property.
-            if (_.isUndefined(mapping.canEvict)) {
-            // eslint-disable-next-line max-len
-                throw new Error(`Cannot subscribe to safe eviction key '${mapping.key}' without providing a canEvict value.`);
+    deferredInitTask.promise
+        .then(() => {
+            // Check to see if this key is flagged as a safe eviction key and add it to the recentlyAccessedKeys list
+            if (mapping.withOnyxInstance && !isCollectionKey(mapping.key) && isSafeEvictionKey(mapping.key)) {
+                // All React components subscribing to a key flagged as a safe eviction
+                // key must implement the canEvict property.
+                if (_.isUndefined(mapping.canEvict)) {
+                    // eslint-disable-next-line max-len
+                    throw new Error(`Cannot subscribe to safe eviction key '${mapping.key}' without providing a canEvict value.`);
+                }
+                addLastAccessedKey(mapping.key);
             }
-            addLastAccessedKey(mapping.key);
-        }
-    })
+        })
         .then(getAllKeys)
         .then((keys) => {
             // Find all the keys matched by the config key
@@ -765,7 +766,7 @@ function init({
         addAllSafeEvictionKeysToRecentlyAccessedList(),
         initializeWithDefaultKeyStates()
     ])
-        .then(deferredInit.resolve);
+        .then(deferredInitTask.resolve);
 
     // Update any key whose value changes in storage
     registerStorageEventListener((key, newValue) => {
--- a/lib/createDeferredTask.js
+++ b/lib/createDeferredTask.js
@@ -1,16 +1,15 @@
 /**
  * Create a deferred task that can be resolved when we call `resolve()`
- * The returned promise will complete only when we call `resolve` or `reject`
+ * The returned promise will complete when we call `resolve`
  * Useful when we want to wait for a tasks that is resolved from an external action
  *
  * @template T
- * @returns {{ resolve: function(*), reject: function(Error), promise: Promise<T|void> }}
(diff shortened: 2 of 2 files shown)


lib/Onyx.js
```diff
@@ -199,9 +203,11 @@ function addToEvictionBlockList(key, connectionID) {
  * the recently accessed list when initializing the app. This
  * enables keys that have not recently been accessed to be
  * removed.
+ *
+ * @returns {Promise}
  */
 function addAllSafeEvictionKeysToRecentlyAccessedList() {
-    getAllKeys()
+    return getAllKeys()
```
kidroca, line 210: return the Promise here so we can wait for it as well
Since every `connect` uses `getAllKeys` it would be best to wait for the read here before any connections are made
Then every connection would instantly get all keys from cache

lib/Onyx.js
```diff
@@ -359,18 +365,20 @@ function connect(mapping) {
         return connectionID;
     }
 
-    // Check to see if this key is flagged as a safe eviction key and add it to the recentlyAccessedKeys list
-    if (mapping.withOnyxInstance && !isCollectionKey(mapping.key) && isSafeEvictionKey(mapping.key)) {
+    // Commit connection only after init passes
+    deferredInit.promise.then(() => {
```
kidroca, line 369: This promise is resolved after Onyx.init runs
Moved the original logic inside the `then` due to the `isSafeEvictionKey` check - it seems those keys are set during `Onyx.init` and this check would not work correctly before `Onyx.init` is complete

lib/Onyx.js
```diff
@@ -624,9 +637,21 @@ function merge(key, val) {
...
+    return AsyncStorage.multiGet(_.keys(defaultKeyStates))
+        .then((pairs) => {
+            const asObject = _.chain(pairs)
+                .map(([key, val]) => [key, val && JSON.parse(val)])
+                .object()
+                .value();
+
+            const merged = lodashMerge(asObject, defaultKeyStates);
+            cache.merge(merged);
+            _.each(merged, (val, key) => keyChanged(key, val));
```
kidroca, line 653: We don't need to write back to storage as `defaultKeyStates` is something that is always overwritten during init
Another reason to preserve in cache forever is that the constant `defaultKeyStates` is never cleared, which means these keys were already in memory for the entire life of the app

lib/Onyx.js
```diff
@@ -775,6 +805,7 @@ function applyDecorators() {
     merge = decorate.decorateWithMetrics(merge, 'Onyx:merge');
     mergeCollection = decorate.decorateWithMetrics(mergeCollection, 'Onyx:mergeCollection');
     getAllKeys = decorate.decorateWithMetrics(getAllKeys, 'Onyx:getAllKeys');
+    initializeWithDefaultKeyStates = decorate.decorateWithMetrics(initializeWithDefaultKeyStates, 'Onyx:defaults');
```
kidroca, line 808: Track the time added by `initializeWithDefaultKeyStates` 
We don't need to add `addAllSafeEvictionKeysToRecentlyAccessedList` as it's already covered by `getAllKeys`

kidroca (commented):
Added some explanations for the changes

lib/Onyx.js
```diff
@@ -28,6 +29,9 @@ const evictionBlocklist = {};
 // Optional user-provided key value states set when Onyx initializes or clears
 let defaultKeyStates = {};
 
+// Connections can be made before `Onyx.init`. They would wait for this promise before resolving
+const deferredInit = createDeferredTask();
```
marcaaron, line 33: Can we rename this to `deferredInitTask` if we are creating a task when we call the method? The comment mentions that it's a promise. But after reviewing the code I see it's more than just a promise.

lib/Onyx.js
```diff
@@ -732,10 +759,13 @@ function init({
...
-    addAllSafeEvictionKeysToRecentlyAccessedList();
 
-    // Initialize all of our keys with data provided
-    initializeWithDefaultKeyStates();
+    // Initialize all of our keys with data provided then give green light to any pending connections
+    Promise.all([
+        addAllSafeEvictionKeysToRecentlyAccessedList(),
+        initializeWithDefaultKeyStates()
+    ])
+        .then(deferredInit.resolve);
```
marcaaron, line 768: It took me some time to wrap my head around this but this is a pretty cool trick we are doing with the deferred task stuff. I'm curious, how much better this would be compared to making `Onyx.init()` itself a promise and calling it before allowing any connections to happen? Does it take a significant amount of time to run this code?

lib/Onyx.js
marcaaron: Let's put this `.then()` on a new line if we're going to chain  ?

lib/createDeferredTask.js
```diff
@@ -0,0 +1,17 @@
...
+ * Useful when we want to wait for a tasks that is resolved from an external action
+ *
+ * @template T
+ * @returns {{ resolve: function(*), reject: function(Error), promise: Promise<T|void> }}
+ */
+export default function createDeferredTask() {
+    const deferred = {};
+    deferred.promise = new Promise((res, rej) => {
+        deferred.resolve = res;
+        deferred.reject = rej;
```
marcaaron, line 13: Let's implement this or remove it?

marcaaron (changes requested):
These changes look fine and make sense. But I'm curious to get your interpretation of the data collected. Does this improve performance in some way?

lib/createDeferredTask.js
kidroca: What do you mean the `reject` functionality?
It should be up to the caller - if they need to reject they should have the option exposed here

All the deferred tasks provide the reject/resolve/promise interface - this is the standard pattern

lib/Onyx.js
kidroca: > Does it take a significant amount of time to run this code?

Before the change the initial `get` calls would start at 2ms past app launch and resolve at ~250ms 
After the change they would start at 212ms and resolve at 255ms 

So even though we delay things by 200ms they resolve at the same time, the speed up comes from the rest of the calls happening and in the end the last storage call is resolved 0.3-0.7 sec earlier

> I'm curious, how much better this would be compared to making `Onyx.init()` itself a promise and calling it before allowing any connections to happen?

Calling Onyx.init before any other Onyx calls would require many changes - every usage that is not a component based `withOnyx` can happen before Onyx.init is called, depending on script load order 
Even if we decide to do so it would not bring anything more to the table in terms of performance

kidroca:
Updated, ready for another review

lib/createDeferredTask.js
marcaaron: >It should be up to the caller - if they need to reject they should have the option exposed here

I get what you are saying but I only see one caller. As a general rule, we either implement code or we remove it.

lib/Onyx.js
marcaaron: >the speed up comes from the rest of the calls happening

Hmm which calls? Sorry not sure I understand. Could you maybe say it another way?

>in the end the last storage call is resolved 0.3-0.7 sec earlier

It looks like iOS experienced almost no changes (and in some cases got worse)? Does that sound right? If so, can you account for the inconsistency there?

>Calling Onyx.init before any other Onyx calls would require many changes - every usage that is not a component based withOnyx can happen before Onyx.init is called, depending on script load order

👍

marcaaron:
Left a couple more comments. Changes look good and looks like there will be some improvement for Android. 

Not a blocker, but I am concerned about the results on iOS and curious why there is inconsistency there.

@timszot feel free to give this a look 🙇

kidroca:
> Not a blocker, but I am concerned about the results on iOS and curious why there is inconsistency there.

If the reason is "Last call ended at" metrics, I find those to depend on network calls and render as well and that they're not a good indication of performance gains/regression - the time there can vary by up to a second for two consecutive runs

Maybe we can identify a point in loading that does not depend on factors as network and use that as a reference

I'll update with the requested changes on Monday

kidroca:
Ready: Addressed all requested changes

marcaaron:
>I find those to depend on network calls and render as well and that they're not a good indication of performance gains/regression - the time there can vary by up to a second for two consecutive runs

Ah so the data is not reliable since network calls can interfere with their accuracy? Is the same true for Android?

>Maybe we can identify a point in loading that does not depend on factors as network and use that as a reference

Perhaps we can test Onyx in isolation somehow, by creating a sample app or "test" page in our main app for running diagnostics without making API calls? Anyways, I don't doubt that this PR will be helpful and I'm not against merging it so we can move on to the other changes, but feels like the benchmarks could be improved.

kidroca:
> Ah so the data is not reliable since network calls can interfere with their accuracy? Is the same true for Android?

Yes, now when I've inspected the data again, the "Last call ended at" is from `Onyx.set` and sometimes from `Onyx.mergeCollection`

> Perhaps we can test Onyx in isolation somehow, by creating a sample app or "test" page in our main app for running diagnostics without making API calls?

I've made something similar working on the ticket here: Expensify/App#2667 but it only tests AsyncStorage and it turned out it works very fast in isolation 
It might help to have a dedicated test page or app, but IMO the problem in Expensify is due to the many things that happen at the same time, and if we improve the parts involved we should see an improvement in general. 
E.g. if we just take the ReportScreen and make a test out of it, it would probably render and work a lot faster, because the rest of the app is not there

My next PR is on Onyx.set and preventing duplicate writes. 
This reduces (depending on chat/device) the time spent in writes from 30sec to 10sec, which can be measured by the existing benchmark, besides the 30 to 10 improvement this should give some more freedom to the rest of the processes that happen at that time

As a summary I think it would be best to capture more metrics in Expensify, but only such that give consistent result like the total time spent in `Onyx.set` this is very repeatable and varies only if you use different chat
Maybe we can hook the benchmark for `withOnyx` time spent waiting and time spent rendering, this should also provide consistent results given the same chat

marcaaron:
>it only tests AsyncStorage and it turned out it works very fast in isolation

So, it sounds like at least some of Onyx's "waiting time" is time it spends waiting for other JS calls to finish?

>Maybe we can hook the benchmark for withOnyx time spent waiting and time spent rendering, this should also provide consistent results given the same chat

Could work, but seems like we also would still have that same problem.

marcaaron:
>the problem in Expensify is due to the many things that happen at the same time, and if we improve the parts involved we should see an improvement in general

I think I'm starting to see this as well. The app tries to do too many things simultaneously and everything ends up waiting for everything else.

timszot:
@marcaaron looked this over and agree with what's being said in the comments. Not sure what the next steps currently are here though.

marcaaron:
I think we should leave this PR open until there is a way to prove it has value in one form or another. 
We generally don't merge code that adds complexity without value. I think this change has value but it feels more like it's solving a problem we don't have yet. Which is something like:

>Some connections start really early before Onyx.init is even called
>They should wait as Onyx.init adds some initial values that might be needed

That makes sense but there's no open issue for that problem and I don't see the relationship to performance.

@kidroca Can we get started with the next PR please? And focus on something that helps performance?

kidroca:
> I think we should leave this PR open until there is a way to prove it has value in one form or another.

What do you mean? Isn't 30sec less time spent in Onyx enough of an evidence?

> That makes sense but there's no open issue for that problem and I don't see the relationship to performance.

It's a fix by coincidence
Eviction keys are set during init and are used in the `connect` logic, if `connect` happens before `init` eviction related keys are not set yet so checks like `isSafeEvictionKey` aren't working correctly 

> @kidroca Can we get started with the next PR please? And focus on something that helps performance?

How would we prove the value of the remaining PRs when they use the same metric (time spent) - if it's unreliable for this PR surely it would be unreliable for the rest?

marcaaron:
>What do you mean? Isn't 30sec less time spent in Onyx enough of an evidence?

I went off of how long it takes the last Onyx call to finish and seems like there is not much difference (and in some cases taking longer e.g. iOS). So, I presumed this PR would help reduce boot time (when Onyx init is called) or help in some other obvious way. But it looks like Onyx finishes working at or around the same time as before. Can you help explain what are we achieving?

>It's a fix by coincidence Eviction keys are set during init and are used in the connect logic, if connect happens before init eviction related keys are not set yet so checks like isSafeEvictionKey aren't working correctly

Right, I think the change makes sense for those reasons (meaning we can merge it with an appropriate linked issue). It might be a perfectly reasonable thing to do, but doesn't really match up with the problem we set out to solve and our top focus which is performance.

>How would we prove the value of the remaining PRs when they use the same metric (time spent) - if it's unreliable for this PR surely it would be unreliable for the rest?

To clarify, are you saying that you agree that the metrics are flawed and that we can't prove that any future changes we're making here are valuable? Or did you want to work together to create some better metrics?

quinthar:
I like the idea of being guided by @kidroca's overall metric of time spent waiting on Onyx, even though it does not directly correlate to user time.  Sorta like how when you use `top` there's a difference between wall clock and system clock time, "wall clock" is how long the user waits -- but even if the user waits the same amount of time, making "system clock" time go down is helpful.

Granted, "wall clock" (ie, user wait time) is by far the more important, but if we can measurably reduce system time without introducing complexity, then I think we should take those wins as we get them with the belief that efficiency floats all boats in the end.

marcaaron:
> Sorta like how when you use top there's a difference between wall clock and system clock time, "wall clock" is how long the user waits -- but even if the user waits the same amount of time, making "system clock" time go down is helpful.

I understand this theory, but it seems more practical to prioritize "user time" over "system time". 
Why would we make the improvements that are difficult to observe first?

quinthar:
> I understand this theory, but it seems more practical to prioritize "user time" over "system time".
Why would we make the improvements that are difficult to observe first?

I agree we should prioritize user time over system time, but I'm suggesting we want both.  However, that assumes there's no real downside to merging this.  Is there a downside you are concerned about?

I 100% agree we shouldn't ship anything that has no defined/measurable benefit.  But this *is* a measurable benefit, to system time, just not to user time.

marcaaron:
>Is there a downside you are concerned about?

If the bar for merging changes is "has no downsides" then we'd be making a lot of useless changes that have no downsides. But the answer is no, I think the changes are great and valuable for other reasons besides performance that may not be priorities. But I'm confused about the metrics and what they are showing.

I'm mainly trying to: 

- understand how this will make things go faster
- get us to prioritize things that will make things go faster

quinthar:
The bar is "has a measurable benefit AND has no downsides".  I think we are
just debating on whether or not "reducing system time" is a benefit if it
doesn't also reduce user time.  I think we obviously would prefer reducing
user time, but if we have a change that "only" reduces system time and
doesn't have any downsides, I don't see why we wouldn't take that small win.

kidroca:
> To clarify, are you saying that you agree that the metrics are flawed and that we can't prove that any future changes we're making here are valuable? 

Only the "Last call ended at" is not very reliable, it helps but only if you know what you're doing. 
For the `init` tests it can vary due to to everything that's going on
But if you use it to benchmark chat A to B switch time it works well

The "time spent" metric is reliable, both for total and individual method calls. So I see an improvement here even though it doesn't seem to improve boot time

> Or did you want to work together to create some better metrics?

I'm trying different things to capture better metrics. When I find something that works I'll bring it up.

marcaaron:
>I think we are just debating on whether or not "reducing system time" is a benefit if it doesn't also reduce user time

I agree that reducing "system time" is valuable on its own and I think you agree that it it's more valuable to reduce "user time".

>if we have a change that "only" reduces system time and doesn't have any downsides, I don't see why we wouldn't take that small win.

The problem I'm experiencing is that I'm not sure how to prioritize things that only reduce system time and don't add complexity.

If this is not the highest value performance improvement we could make right now or its value is not well understood my instinct is to suggest we find something better to work on first. We can take any small wins that have no downsides, but it feels like we are digging in our couch for spare change or something.

kidroca:
The task I'm working on is to optimize AsyncStorage usage. This won't necessarily fix boot time, but helps with everything in general.

The total time spent was decreased by 1/4.
3 more PRs like this and it will be 0 😁

I've made many experiments and managed to speed up reading - with batching, but even then boot time didn't drop significantly, but things like chat switching are becoming faster.
Onyx or at least data access might not be what's causing the slow down. I'm leaning more and more towards `withOnyx` like the other research discovered. But in the mean time why not take the improvements we have. 

The thing is measuring chat switch is hard to do manually. But since app init is loading a chat from storage, benchmarking it provides data that captures improvements in that direction as well

marcaaron:
>things like chat switching are becoming faster

Now you are speaking my language! Where is that PR? Let's do it!! 😂 

>The thing is measuring chat switch is hard to do manually

I agree it's hard to do (and also time consuming if you are doing release builds - which I think is the most reliable way). I've spent the entire week doing benchmarks with the production API and release builds on iOS and Android. So, let me know if I can help test anything.

marcaaron (approved):

kidroca:
Have you tried the hash 93c30bebba9a0595405e0e8b3bc103221af81e05 with all the changes from this PR: #88 

Unfortunately since then "total time" have increased, a lot due to `withLocalize`, but these changes should give you a glimpse is the end result and chat switch speed

I think all changes complement each other, but for chat switching the set and cache improvement should be the most valuable 

What could speed boot time is persisting the recent access list and loading it with a single get during init (@quinthar had the idea some time ago). I can try it (the changes needed are in one of the following PRs) and report what I've found. But then we'll definitely have to defer init so that we capture the full benefit of pre-filling cache, so we might as well merge the defer logic now

marcaaron:
Merged this since I don't have any reason to block and agree with the deferred Onyx connections stuff regardless of performance.

>I think all the changes complement each other, but for chat switching the set and cache improvement should be the most valuable

Sorry if I'm coming off as skeptical here. I just prefer to have a clear idea which single action we are taking has the greatest impact on performance.

>we'll definitely have to defer init so that we capture the full benefit of pre-filling cache

That's true. I'm also curious to do an experiment where we do a deferred init then a big multiGet with ALL of the available data and never clear the cache at all. I'd imagine that would reduce virtually all movement across the bridge. Not sure if that would be a good idea or not but I'd expect massive improvements in chat switching since we will effectively spend 0 time crossing the bridge. If Onyx's biggest bottleneck is `AsyncStorage` - should we try to make a version of Onyx that doesn't ever interact with `AsyncStorage` and see how the app runs?

kidroca:
Yes we can get all keys and then multi get the values. That's pretty easy to test and should review how much the app can improve with just changes in Onyx 

That's pretty much how Redux works with local storage 

During init the Redux store is created and hydrated with everything from storage 
Then when a change happens it's applied to data in memory and scheduled to be persisted in storage 

Now with Expensify there are concerns that data might become too much, so a hybrid approach can be made where we hydrate everything but chat messages...


> If Onyx's biggest bottleneck is `AsyncStorage` - should we try to make a version of Onyx that doesn't ever interact with `AsyncStorage` and see how the app runs?

I think if we read evening at start and then only write to storage the effect would be the same 
The bottleneck IMO is caused by `withOnyx` and this "get all during init" might prove it if even then boot does not improve

quinthar:
Yes, we are aiming to build a platform that has shit tons of data stored
locally -- let's not limit ourselves to what fits in RAM.

marcaaron:
>Yes, we are aiming to build a platform that has shit tons of data stored locally -- let's not limit ourselves to what fits in RAM.

Ah sorry, to clarify, the suggestion was to _experimentally_ do this. Then we can see how the app runs without a dependency on `AsyncStorage` and deduce how much room for improvement there is.

lib/Onyx.js
```diff
@@ -624,9 +638,21 @@ function merge(key, val) {
 
 /**
  * Merge user provided default key value pairs.
+ *
+ * @returns {Promise}
  */
 function initializeWithDefaultKeyStates() {
-    _.each(defaultKeyStates, (state, key) => merge(key, state));
+    return AsyncStorage.multiGet(_.keys(defaultKeyStates))
```
marcaaron, line 645: Just caught this. But do the "after" metrics in the description include this time at all?

kidroca: Yes, because we decorate and track this method here:
https://github.com/kidroca/react-native-onyx/blob/9549e4057abaf2a8f96bad1f255acd1c0b002cd2/lib/Onyx.js#L809

marcaaron: Ah ok cool. Thanks.

PR #93, opened 2021-08-03
Perf: Lru cache

kidroca:
cc @marcaaron 

### Details

PR4. Of the changes planned here: https://github.com/Expensify/react-native-onyx/pull/88#issuecomment-880916315

Keep a list of recently accessed keys in `OnyxCache` - `recentKeys`
This serves to clean "Least Recently Used" keys from cache

Keys are added to `recentKeys` for any cache access operation - get/set/merge

Added a configurable limit of 150 keys in cache
This looks like a sane value from observations during chat browsing

#### Note on print/benchmark updates:
This helps with extracting information to excel in a consistent matter

Add an optional `methods` list parameter
When such a list is provided only the methods from this list will be printed

Removed any sorting
Tables order varies and makes it harder to deal with in Excel

### Related Issues

Expensify/App#2667

### Automated Tests

Added test covering that recent keys are available in cache 
and that older keys past the LRU limit are cleared from cache when safe eviction connection happens

### DIFF
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -683,8 +683,8 @@ function mergeCollection(collectionKey, collection) {
  * @param {function} registerStorageEventListener a callback when a storage event happens.
  * This applies to web platforms where the local storage emits storage events
  * across all open tabs and allows Onyx to stay in sync across all open tabs.
- * @param {Number} [options.maxCachedKeysCount=150] Sets how many recent keys should we try to keep in cache
- * Setting this to 0 would only keep active connections in cache
+ * @param {Number} [options.maxCachedKeysCount=55] Sets how many recent keys should we try to keep in cache
+ * Setting this to 0 would keep cache forever
  * @param {Boolean} [options.captureMetrics]
  */
 function init({
@@ -692,7 +692,7 @@ function init({
     initialKeyStates,
     safeEvictionKeys,
     registerStorageEventListener,
-    maxCachedKeysCount = 150,
+    maxCachedKeysCount = 55,
     captureMetrics = false,
 }) {
     if (captureMetrics) {
@@ -701,7 +701,7 @@ function init({
         applyDecorators();
     }
 
-    if (_.isNumber(maxCachedKeysCount)) {
+    if (maxCachedKeysCount > 0) {
         setInterval(() => {
             // Try to free anything dated from cache
             cache.removeLeastRecentUsedKeys(maxCachedKeysCount);
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -356,14 +356,21 @@ function connect(mapping) {
     deferredInitTask.promise
         .then(() => {
             // Check to see if this key is flagged as a safe eviction key and add it to the recentlyAccessedKeys list
-            if (mapping.withOnyxInstance && !isCollectionKey(mapping.key) && isSafeEvictionKey(mapping.key)) {
-                // All React components subscribing to a key flagged as a safe eviction
-                // key must implement the canEvict property.
-                if (_.isUndefined(mapping.canEvict)) {
-                    // eslint-disable-next-line max-len
-                    throw new Error(`Cannot subscribe to safe eviction key '${mapping.key}' without providing a canEvict value.`);
+            if (isSafeEvictionKey(mapping.key)) {
+                // Try to free some cache whenever we connect to a safe eviction key
+                cache.removeLeastRecentlyUsedKeys();
+
+                if (mapping.withOnyxInstance && !isCollectionKey(mapping.key)) {
+                    // All React components subscribing to a key flagged as a safe eviction
+                    // key must implement the canEvict property.
+                    if (_.isUndefined(mapping.canEvict)) {
+                        throw new Error(
+                            `Cannot subscribe to safe eviction key '${mapping.key}' without providing a canEvict value.`
+                        );
+                    }
+
+                    addLastAccessedKey(mapping.key);
                 }
-                addLastAccessedKey(mapping.key);
             }
         })
         .then(getAllKeys)
@@ -684,8 +691,9 @@ function mergeCollection(collectionKey, collection) {
  * This applies to web platforms where the local storage emits storage events
  * across all open tabs and allows Onyx to stay in sync across all open tabs.
  * @param {Number} [options.maxCachedKeysCount=55] Sets how many recent keys should we try to keep in cache
- * Setting this to 0 would keep cache forever
- * @param {Boolean} [options.captureMetrics]
+ * Setting this to 0 would practically mean no cache
+ * We try to free cache when we connect to a safe eviction key
+ * @param {Boolean} [options.captureMetrics] Enables Onyx benchmarking and exposes the get/print/reset functions
  */
 function init({
     keys,
@@ -702,10 +710,7 @@ function init({
     }
 
     if (maxCachedKeysCount > 0) {
-        setInterval(() => {
-            // Try to free anything dated from cache
-            cache.removeLeastRecentUsedKeys(maxCachedKeysCount);
-        }, 10 * 1000);
+        cache.setRecentKeysLimit(maxCachedKeysCount);
     }
 
     // Let Onyx know about all of our keys
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -29,6 +29,9 @@ const evictionBlocklist = {};
 // Optional user-provided key value states set when Onyx initializes or clears
 let defaultKeyStates = {};
 
+// Cache cleaning uses this to remove least recently accessed keys
+let MAX_CACHED_KEYS = 150;
+
 // Connections can be made before `Onyx.init`. They would wait for this task before resolving
 const deferredInitTask = createDeferredTask();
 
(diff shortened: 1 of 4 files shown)


lib/OnyxCache.js
```diff
@@ -144,6 +154,34 @@ class OnyxCache {
...
+        // Get the last N keys by doing a negative slice
+        const recentlyAccessed = new Set([...this.recentKeys].slice(-recentKeysSize));
+
+        _.chain(this.storageMap)
+            .keys()
+            .each((key) => {
+                if (!recentlyAccessed.has(key)) {
+                    this.drop(key);
+                }
+            });
```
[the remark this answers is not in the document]
marcaaron: Rather than loop over every key it might be clearer to get the [difference](https://underscorejs.org/#difference) of these keys first then just drop all of them?

lib/Onyx.js
```diff
@@ -412,48 +402,6 @@ function connect(mapping) {
...
-function cleanCache(key) {
-    // Don't remove default keys from cache, they don't take much memory and are accessed frequently
-    if (_.has(defaultKeyStates, key)) {
-        return;
-    }
-
-    const hasRemainingConnections = _.some(callbackToStateMapping, {key});
-
-    // When the key is still used in other places don't remove it from cache
-    if (hasRemainingConnections) {
```
marcaaron, line 428: Just to confirm after this change - if we theoretically have something like 200 keys simultaneously subscribed and disconnected just one that we'd drop 50 of the least recently accessed whether a subscriber is accessing them or not?

The previous logic only drops keys that are not being subscribed to so I was surprised to see that removed. Just curious to hear thought about why we shouldn't need logic like that anymore since it it being removed.

marcaaron (commented):
Just leaving some comments as this is still in draft.

lib/Onyx.js
kidroca: > Just to confirm after this change - if we theoretically have something like 200 keys simultaneously subscribed and disconnected just one that we'd drop 50 of the least recently accessed whether a subscriber is accessing them or not?

Yes this is correct, but it would be an extremely rare case for this to happen
IRL we do have 200+ connections but they involve about 50-60 unique keys

> The previous logic only drops keys that are not being subscribed to so I was surprised to see that removed. Just curious to hear thought about why we shouldn't need logic like that anymore since it it being removed.

Good point. I forgot to bring that up

In short if something is "so old" yet still connected after we've cycled 100-150 keys and it didn't re-appear at least ones to revive itself - it's not something that will benefit from the "cache while connected" strategy - no other connection/component wants to use it anyway

---

The limit of 150 keys is actually too big. I've browsed storage on Android and see that for my account with 10 chats I have 58 unique keys in storage so I'll never reach the LRU limit of 150
Of these 58 keys only 10 can grow indefinitely - `reportActions_...` (I guess that's why they are the "eviction keys") the rest are of finite and small size - precisely 34.1kB (24.1kB if we go and remove the report objects)

With this information I think we should consider 3 items for the future:
1. Don't drop non eviction keys from cache - these are the frequently used keys and often all of them are in memory. 
2. Load all non-eviction keys in one multiGet during init
3. Apply the LRU strategy only for the eviction keys

#### Here's what I'm planning for the current PR:

Reduce the LRU limit to 50-60 keys - this will still let cache fill with all the reusable keys during init (48 keys) 
Cache will start to free up after you switch between some chats

lib/OnyxCache.js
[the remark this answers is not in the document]
kidroca: Good idea, thanks

lib/decorateWithMetrics.js
```diff
@@ -24,8 +24,6 @@ function decorateWithMetrics(func, alias = func.name) {
         throw new Error(`"${alias}" is already decorated`);
     }
 
-    stats[alias] = [];
-
```
kidroca, line 28: We don't need this here, it's created during the `finally` call

lib/Onyx.js
```diff
@@ -757,6 +701,13 @@ function init({
         applyDecorators();
     }
 
+    if (maxCachedKeysCount > 0) {
+        setInterval(() => {
+            // Try to free anything dated from cache
+            cache.removeLeastRecentUsedKeys(maxCachedKeysCount);
+        }, 10 * 1000);
```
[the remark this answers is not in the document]
kidroca: Instead of interval based solution perhaps we can hook to something that happens less frequently
I think it would be OK to clean cache every minute or even less frequently as long as we clean it

lib/Onyx.js
marcaaron: Good thoughts thanks for the explanation. If it helps at all, I have 310 keys in `localStorage` for my main account that is maybe more of a typical business usage.

It seems reasonable to go with something more aggressive that has an arbitrary limit for now (which seems better than what we had before).

It does feel like the most important case is switching to a chat we've already accessed in a session and will need to retrieve the `reportActions_*` from storage instead of cached (grabbing and parsing the big object will take time). So this should help with that.

I do agree it seems less value to drop anything finite and may be better to front load the cost of getting finite keys to a `multiGet` on init and then just never drop those keys.

This is maybe thinking ahead a bit. But if we do go that direction we can potentially grab values synchronously for `withOnyx` connected components that are only interested in cached values - which might be a huge improvement for `withOnyx`. But that's only possible if we can guarantee which keys will be cached and which are subject to eviction.

lib/Onyx.js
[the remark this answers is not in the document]
marcaaron: I feel maybe the interval based solution is too aggressive since most of the time we don't really need to clean the cache at all.

What if we hook the cache clean into the connection of a safe eviction key like the `reportActions_*`. If we are connecting to another large object that seems like a good opportunity to clean the cache in preparation.

lib/decorateWithMetrics.js
kidroca: Reverting this change as it's part of a test and functionality to prevent duplicate decorations
I don't think we need this functionality but it will be perhaps to much changes to remove it under this PR

lib/Onyx.js
kidroca: > I do agree it seems less value to drop anything finite and may be better to front load the cost of getting finite keys to a multiGet on init and then just never drop those keys.

I'll open follow up ticket to discuss the points above

> This is maybe thinking ahead a bit. But if we do go that direction we can potentially grab values synchronously for withOnyx connected components that are only interested in cached values - which might be a huge improvement for withOnyx. But that's only possible if we can guarantee which keys will be cached and which are subject to eviction.

Yes, that sounds about right 
We can use the deferred task, and release it after we've prefilled cache with the non-eviction keys. This will happens faster than letting components asks for these keys one by one during init

Now `withOnyx` can pick values synchronously from cache
Hmm, but actually only the connections are waiting `init` to resolve
Components would mount right away and initiate a connection during the `didMount` phase - component construction and initial state won't wait for Onyx.init
There are ways around this though - like rendering the WrappedComponent until Onyx finishes init and then passing the state in one update

BTW it might be better to connect to Onyx in the constructor instead of during `componentDidMount`. 
In `didMount` it means that at least one commit happened and only then we start the connection

lib/Onyx.js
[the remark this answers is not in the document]
kidroca: Thanks for the idea, that sounds much better, will update

kidroca:
Updated 
Replaced the set interval cache clean trigger with cache clean triggered on safe eviction keys connections

kidroca:
I've moved to test the changes here on the latest Expensify 
But after capturing some metrics I see there's no much difference in before/after 

The cache already contains everything from storage (For the Before run)
image

But my chat list is small - I've got only 10

Here's why the cache already contains everything from storage 
1. This connection loads all chat actions and is kept for the duration of the app - the old cache logic does not remove items that are connected:  https://github.com/Expensify/App/blob/e88b7f984ab4e89e33844260c3a334622bf23023/src/libs/actions/ReportActions.js#L20-L22
2. General report data used in LHN connects to all report items (and other keys) 
3. Recent optimizations keep some keys in cache forever 
   - The OnxyProvider related keys (network, personalDetails, report action drafts)
   - The initial keys (sesson, account, network, IOU, is sidebar loaded)

The changes in this PR set a default limit of 55 recent keys in cache so they should help reduce the cache footprint for users that have more than 10 Chats (With 10 chats I have 58 unique keys in storage)

I've explained why we don't need to keep items in cache just because they have an active connection here: https://github.com/Expensify/react-native-onyx/pull/93#discussion_r682747562
> In short if something is "so old" yet still connected after we've cycled 100-150 keys and it didn't re-appear at least once to revive itself - it's not something that will benefit from the "cache while connected" strategy - no other connection/component wants to use it anyway

As a summary the changes here don't seem to provider any performance benefit over the existing strategy, but they are simplifying the code and can help with freeing more cache

marcaaron (approved):
Thanks for the breakdown. I don't mind merging these changes anyway since they are an improvement over the previous logic.

Should we get these changes into Expensify/App?

kidroca:
> Should we get these changes into Expensify/App?

I can open a PR and test all platforms on Monday

kidroca:
@marcaaron Opened this PR: https://github.com/Expensify/App/pull/4509

PR #97, opened 2021-08-12
Call setState once when handling initial withOnyx data to reduce JS scripting time

marcaaron:
### Details

Implements the idea discussed in the link below cc @kidroca 

**Before change**

**After change:**

Small difference of about 81ms here to total scripting time when the app first becomes active. But this should have positive affects in all other places where a `withOnyx` connection is set up.

### Related Issues
https://github.com/Expensify/App/issues/4592#issuecomment-897672983

### DIFF
--- a/lib/withOnyx.js
+++ b/lib/withOnyx.js
@@ -25,9 +25,13 @@ export default function (mapOnyxToState) {
             constructor(props) {
                 super(props);
 
+                this.setInitialData = this.setInitialData.bind(this);
+
                 // This stores all the Onyx connection IDs to be used when the component unmounts so everything can be
                 // disconnected. It is a key value store with the format {[mapping.key]: connectionID}.
                 this.activeConnectionIDs = {};
+
+                this.tempState = {};
 
                 this.state = {
                     loading: true,
@@ -39,7 +43,7 @@ export default function (mapOnyxToState) {
                 _.each(mapOnyxToState, (mapping, propertyName) => {
                     this.connectMappingToOnyx(mapping, propertyName);
                 });
-                this.checkAndUpdateLoading();
+                this.checkEvictableKeys();
             }
 
             componentDidUpdate(prevProps) {
@@ -55,7 +59,7 @@ export default function (mapOnyxToState) {
                         this.connectMappingToOnyx(mapping, propertyName);
                     }
                 });
-                this.checkAndUpdateLoading();
+                this.checkEvictableKeys();
             }
 
             componentWillUnmount() {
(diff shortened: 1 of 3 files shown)


tests/unit/withOnyxTest.js
```diff
@@ -68,9 +68,13 @@ describe('withOnyx', () => {
         })(ViewWithCollections);
         const onRender = jest.fn();
         render(<TestComponentWithOnyx onRender={onRender} />);
-
-        Onyx.mergeCollection(ONYX_KEYS.COLLECTION.TEST_KEY, {test_1: {ID: 123}, test_2: {ID: 234}, test_3: {ID: 345}});
-        return waitForPromisesToResolve()
+        waitForPromisesToResolve()
```
marcaaron, line 71: I think these tests are fixed correctly. Did not observe any issues with `mergeCollection()` yet

lib/withOnyx.js
```diff
@@ -67,12 +78,33 @@ export default function (mapOnyxToState) {
...
+             * @param {String} statePropertyName
+             * @param {*} val
+             */
+            setInitialState(statePropertyName, val) {
+                this.tempState[statePropertyName] = val;
+
+                // All state keys should exist and at least have a value of null
+                if (!_.every(requiredKeysForInit, key => !_.isUndefined(this.tempState[key]))) {
+                    return;
+                }
```
kidroca, line 95: IMO something like this is more readable: 
```js
if (_.some(requiredKeys, key => _.isUndefined(this.tempState[key])) {
  return;
}
```

lib/withOnyx.js
```diff
@@ -67,12 +78,33 @@ export default function (mapOnyxToState) {
...
+
+                this.setState({...this.tempState, loading: false}, () => {
+                    this.tempState = {};
+                });
```
kidroca, line 99: We don't need to reset `tempState` in a callback
This is equivalent and more performant

```js
this.setState({...this.tempState, loading: false});
this.tempState = {};
```

Also we could `delete this.tempState`

lib/withOnyx.js
```diff
@@ -67,12 +78,33 @@ export default function (mapOnyxToState) {
...
+            /**
+             * This method is used externally by sendDataToConnection to prevent unnecessary renders while a component
+             * still in a loading state. The temporary initial state is saved to the component instance and setState()
+             * only called once all the necessary data has been collected.
+             *
+             * @param {String} statePropertyName
+             * @param {*} val
+             */
+            setInitialState(statePropertyName, val) {
+                this.tempState[statePropertyName] = val;
```
kidroca, line 90: I think this method should throw an Error or at least print a warning and skip execution if it's called after `loading` is set to `false`

kidroca (commented):
Good work!
It seems we're pretty much moving/reusing existing code and gaining some performance
Left some minor notes

nickmurray47 (approved):
lgtm

lib/withOnyx.js
```diff
@@ -67,12 +78,37 @@ export default function (mapOnyxToState) {
...
+             * only called once all the necessary data has been collected.
+             *
+             * @param {String} statePropertyName
+             * @param {*} val
+             */
+            setInitialState(statePropertyName, val) {
+                if (!this.state.loading) {
+                    console.error('withOnyx.setInitialState() called after loading: false');
+                    return;
+                }
```
kidroca, line 93: @marcaaron 
Sorry we've made a mistake here 
This method is not called only for initial state but for updates caused by the logic in `componentDidUpdate`: https://github.com/Expensify/react-native-onyx/pull/97/files#diff-364fd7349a1af7b52091427c5a0734da8ff56057fbc7fe2be10734a06c97264cL55

We should update the logic to address calls triggered by updates

```jsx
setInitialState(statePropertyName, val) {
  if (this.state.loading) {
      this.tempState[statePropertyName] = val;

    if (_.some(requiredKeysForInit, key => _.isUndefined(this.tempState[key]))) {
      return;
    }
    
    this.setState({...this.tempState, loading: false});
    delete this.tempState;
  }
  else {
     this.setState({[statePropertyName]: val});
  }
}
```

Also we should maybe change the method name and description to reflect that it's not called only for the initial state

marcaaron: Oh I see. The `Onyx.connect()` method may be called again. That solution makes sense to me.

marcaaron: Thanks for. catching this one. Created a follow up here -> https://github.com/Expensify/react-native-onyx/pull/98

PR #98, opened 2021-08-16
Fix prop dependent withOnyx updates + add test

marcaaron:
### Details
cc @kidroca See this comment here -> https://github.com/Expensify/react-native-onyx/pull/97#discussion_r689287381
We've caused a regression with prop dependent updates.

### Related Issues
There are probably some if we've already started regressions.

### Automated Tests
A new test to make sure withOnyx always works with prop dependent keys was added.

### Linked PRs
https://github.com/Expensify/App/pull/4676

### DIFF
--- a/lib/withOnyx.js
+++ b/lib/withOnyx.js
@@ -31,7 +31,7 @@ export default function (mapOnyxToState) {
             constructor(props) {
                 super(props);
 
-                this.setInitialState = this.setInitialState.bind(this);
+                this.setWithOnyxState = this.setWithOnyxState.bind(this);
 
                 // This stores all the Onyx connection IDs to be used when the component unmounts so everything can be
                 // disconnected. It is a key value store with the format {[mapping.key]: connectionID}.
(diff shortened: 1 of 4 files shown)


lib/withOnyx.js
```diff
@@ -87,20 +87,20 @@ export default function (mapOnyxToState) {
              * @param {*} val
              */
             setInitialState(statePropertyName, val) {
-                if (!this.state.loading) {
```
Gonals, line 90: Mostly out of curiosity: why are we reversing this `if`?

marcaaron: The logic has changed from what we had before which was incorrect.

marcaaron: We don't need to invert the logic though it's just how I did it. Updated so this is unchanged but the logic is the same.

luacmartins (approved):
LGTM!

luacmartins:
Merging since this will remove the hold on a deploy blocker fix. Sorry @Gonals

PR #101, opened 2021-08-18
Capture metrics using `react-native-performance`

kidroca:
### Details
We're already using `react-native-performance` in Expensify/App
Using it here allows us to visualize Onyx measures in Flipper

image

### Related Issues

https://github.com/Expensify/App/issues/4656#issuecomment-900212408

### Automated Tests

Covered by existing tests


timszot:
This looks good to me, but I'd like @marcaaron to take a look here since I know he's more involved in the performance metrics.

lib/metrics.js
```diff
@@ -1,17 +1,31 @@
...
-let PERFORMANCE_OFFSET = performance.now();
+/**
+ * Capture a measurement between the start mark and now
+ * @param {{name: string, startTime:number, detail: {args: []}}} startMark
+ * @param {*} detail
+ */
+function measureMarkToNow(startMark, detail) {
+    performance.measure(`${startMark.name} [${startMark.detail.args.toString()}]`, {
+        start: startMark.startTime,
+        end: performance.now(),
```
marcaaron, line 25: Is using `performance.now()` the same as doing 

```js
performance.mark('endMark');
performance.measure('someMeasure', 'startMark', 'endMark');
```

lib/metrics.js
```diff
@@ -100,7 +106,7 @@ function getMetrics() {
         .value();
 
     const totalTime = sum(_.values(summaries), 'total');
-    const lastCompleteCall = _.max(_.values(summaries), ['lastCall', 'endTime']).lastCall;
+    const lastCompleteCall = _.max(_.values(summaries), ['lastCall', 'startTime']).lastCall;
```
marcaaron, line 109: I'm having trouble figuring out if this is changing and how... what's the point of this metric? Can we maybe add some code comments to help explain it. I had trouble with this last time and I'm even more confused now so it would be good to add some additional context.

lib/metrics.js
```diff
@@ -156,6 +162,7 @@ function printMetrics({raw = false, format = 'console', methods} = {}) {
     });
 
     const methodNames = _.isArray(methods) ? methods : _.keys(summaries);
+    const timeOrigin = performance.timeOrigin;
```
marcaaron, line 165: Not super clear why we need to worry about `timeOrigin` here. I recall reading something about this in the docs but maybe we can leave a mention somewhere. e.g. why do we only subtract the `timeOrigin` from a "start time" but not a "duration" with some brief explanation might be valuable?

marcaaron (changes requested):
This looks good to me, but I spotted a few places where I think it couldn't hurt to add some additional context or clarifying code comments to make it easier to work on for the next person.

lib/metrics.js
kidroca: Yes, and originally I did it like that but then I noticed we're not really using the marks - only measures appear on the timeline, so we skip capturing a mark here
Do you want to capture an end mark first and then use marks to make the measure?

lib/metrics.js
kidroca: Can we remove the `lastCompleteCall` altogether? 
This metric would capture the last call made per each alias - get/set/merge etc... 
I don't find it very useful now that we see it on the timeline
The other place we might use it is in excel and it's pretty easy to derive from calls

lib/metrics.js
kidroca: `timeOrigin` replaces `PERFORMANCE_OFFSET`: https://github.com/Expensify/react-native-onyx/pull/101/files#diff-d5767fc7b9c776a9683e5059ef6f67713b836b8dbf6781461ae950228a478aa1L14

`react-native-performance` "suffers" from the same issue that made `PERFORMANCE_OFFSET` necessary
the `startTime` of the marks would not be relative to app launch, but relative to system uptime 
In order to see a `startTime` (or `performance.now()`) relative to app launch we need to subtract `timeOrigin` 
I'll add a comment about it

lib/metrics.js
marcaaron: Nah mostly just asking to see if these were equivalent - I didn't realize we could pass start and end to measure as options.

lib/metrics.js
marcaaron: I am not opposed to removing it. But if it's helpful maybe just explain what it is. I think the part I'm struggling with is the change from `endTime` to `startTime`. Just sort of hard to tell what we are measuring or something from looking at the code so a comment would help.

lib/metrics.js
marcaaron: Sounds good, thanks!

lib/metrics.js
kidroca: Updated to capture the endTime and added documentation

kidroca:
Addressed requested changes. Ready for review

marcaaron (approved):

PR #102, opened 2021-08-30
File handling POC

kidroca:
cc @thienlnam 

### Details

Add file handling capabilities to Onyx

#### Uses [localforage](https://localforage.github.io/localForage/) for web/desktop. 
Extracted a web/desktop storage provider that saves all our key/value pairs in IndexedDB through `localforage`
This allows saving files as part of Onyx items

#### FIle saving on native
When an attachment is picked the `AttahcmentPicker` will persist it to App/cache folder. It is then available until cache is cleared
This lets us save only the reference to the File in our key/val storage
This reference is enough to work with the file - use it in an `<Image />` or in a upload request

### Related Issues

$ https://github.com/Expensify/App/issues/3867

### Automated Tests

TBD


lib/Onyx.js
```diff
@@ -1,7 +1,8 @@
 import _ from 'underscore';
-import AsyncStorage from '@react-native-async-storage/async-storage';
 import Str from 'expensify-common/lib/str';
 import lodashMerge from 'lodash/merge';
+import Storage from './storage';
```
kidroca, line 4: Instead of using AsyncStorage directly, we're exposing a `Storage` provider that works differently depending on the Platform

lib/Onyx.js
```diff
@@ -52,11 +53,10 @@ function get(key) {
...
 
     // Otherwise retrieve the value from storage and capture a promise to aid concurrent usages
-    const promise = AsyncStorage.getItem(key)
+    const promise = Storage.getItem(key)
         .then((val) => {
-            const parsed = val && JSON.parse(val);
-            cache.set(key, parsed);
-            return parsed;
+            cache.set(key, val);
+            return val;
```
kidroca, line 59: parsing and stringifying is encapsulated on the provider level now

lib/Onyx.js
```diff
@@ -722,13 +718,25 @@ function init({
...
         .then(deferredInitTask.resolve);
 
-    // Update any key whose value changes in storage
-    registerStorageEventListener((key, newValue) => {
-        cache.set(key, newValue);
-        keyChanged(key, newValue);
-    });
+    if (synchronizeInstances) {
+        Storage.synchronizeInstances({keyChanged});
+    }
```
kidroca, line 723: Relaying updates to other tabs happens in a similar manner as before, but since it's tied to IndexedDB (and localforage) usage it's better to manage it in Onyx

lib/storage/index.native.js
```diff
@@ -0,0 +1,43 @@
...
+        // Create result optimistically so the method can called as sync method
+        const uri = FileHandler.getLocalDestination(file);
+        const optimisticResult = {...file, uri};
+
+        // Todo: do something about the error, e.g. run eviction logic if it's due to lack of space
+        FileHandler.storeLocally(file)
+            .catch(console.error);
+
+        return optimisticResult;
+    }
```
kidroca, line 40: There's an alternative approach, where instead of having to call `prepareFile` we make a check every time we call `setItem` and recursively search the `value` for files:

https://github.com/Expensify/react-native-onyx/blob/86d8ad9f607254d33ef53df2bdf9d22786ff2e99/lib/storage/index.native.js#L25-L76

kidroca (commented):
Add some notes on the code to help review

lib/Onyx.js
```diff
@@ -722,13 +718,25 @@ function init({
...
+ * Then save it as usual by Onyx.set or Onyx.merge
+ *
+ * @example
+ * const preparedFile = Onyx.prepareFile(someFile);
+ * Onyx.merge('someKey', {
+ *   attachment: preparedFile,
+ *   hasAttachment: true,
+ * });
+ */
+const prepareFile = Storage.prepareFile;
```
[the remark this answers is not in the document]
kidroca: My original idea was to to have a collection key like `FILE_` and store everything there.  I explored that direction as well. It ends up with something similar as `prepareFile` 

If we always put files into a special collection like `FILE_` it means that everything that uses a file, like a comment with an attachment or a persisted request for file upload would have to first save the file to `FILE_` (very similar to `prepareFile`) and then associate a key on the comment to a key in storage e.g. `comment.attachment = 'FILE_specific_file_key'` 

So in the end we have something like:

##### Example usage if we had "FILE_" collection
```js
const fileData = {
  isAttachment: _.isEmpty(text) && file !== undefined
}

if (fileData .isAttachment) {
  Onyx.merge(`${ONYXKEYS.COLLECTION.FILES}${file.name}`, file);
  
  // In order to store a file we need to move it to our app storage folder
  // This would change it's URI
  // We need to reference it by this updated URI
  const uri = getLocalFileURI(file);
  fileData.localAttachment = { ...file, uri };

  // Another approach is to save the onyx key pointing to the file
  fileData.localAttachmentKey = `${ONYXKEYS.COLLECTION.FILES}${file.name}`;

  // But this means that in order to access the file we need to first use the key to retrieve the file 
  // E.g. we'll either have a place that subscribe to the entire FILE_ collection 
  // Or we'll have withOnyx({ prop: {  localAttachment:  ({ localAttachmentKey }) =>  localAttachmentKey }})
  // Or both
}

// Save the reference to the file in a Report Action (or anywhere else)
Onyx.merge(`${ONYXKEYS.COLLECTION.REPORT_ACTIONS}${reportID}`, {
        [optimisticReportActionID]: {
  actionName: CONST.REPORT.ACTIONS.TYPE.ADDCOMMENT,
  actorEmail: currentUserEmail,
   ...fileData,
)}
```

##### Example usage with `prepareFile`
```js
const isAttachment = _.isEmpty(text) && file !== undefined;
const localAttachment = isAttachment && Onyx.prepareFile(file);

Onyx.merge(`${ONYXKEYS.COLLECTION.REPORT_ACTIONS}${reportID}`, {
        [optimisticReportActionID]: {
  actionName: CONST.REPORT.ACTIONS.TYPE.ADDCOMMENT,
  actorEmail: currentUserEmail,
  isAttachment,
  localAttachment,
)}
```

Some of the advantages of `prepareFile` over plain Onyx.merge 
- it hides details like how to formulated the Onyx path to save it - `${ONYXKEYS.COLLECTION.FILES}${file.name}`
- it returns an updated file reference with the updated URI
- on web/desktop when you delete something that has a nested file - the file is deleted with it (We still need to cover this on Native though)

---

I prefer to store the file reference to an object key (or nested key) written to AsyncStorage
- it's just a JSON with `uri`, `name`, etc... the actual file is on the FS
 - this JSON can be used everywhere out of the box
- show a comment local image -> withOnyx(comment) -> `<Image source={comment.localAttachment} />`
- persist a network request with a file: [working example](https://github.com/Expensify/App/blob/e9c5d5083519180043f4debfb5169b35d7482d8a/src/libs/actions/Report.js#L1083-L1091)
- for web/desktop it's a `File` instance and works the same way.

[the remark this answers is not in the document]
tgolen: OK, thanks! I think you've sold me on it now as well.

However, I think there is still a valid concern for:
- How do I know to use `prepareFile()`?
- What happens if I forget to use `prepareFile()`?

Granted, I don't know that those are actual problems yet, but they are worth considering if there is any kind of cheap/easy way to `throw` inside of `merge()` if images are not being stored correctly.

Oh, one other idea I just had. Instead of:
```
Onyx.prepareFile();
Onyx.merge();
```

What would you think about just having this? This might help with my first concern.
```
Onyx.mergeWithFile();
```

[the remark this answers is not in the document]
marcaaron: I don't mind `mergeWithFile()`, but has the same problem as `prepareFile()` and cannot be used with `set()`.

>How do I know to use `prepareFile()`?
>What happens if I forget to use `prepareFile()`?

It won't work and you'll be confused about why then you'll go and look at how other places handle it and see `prepareFile()` 💡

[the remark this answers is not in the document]
kidroca: Hmm, `prepareFile` is only needed for mobile - it's a noop in web/desktop, saving files to IndexedDB works out of the box and you can totally skip the `prepareFile` call with no negative consequences 

`mergeWithFile` and `setWithFile` seem a bit better than having to use `prepareFile` before a merge

One more thing regarding `prepareFile` - not using it still works, we just do it to have more control by moving the file from temp to our own storage dir. Opening a image with the image picker makes a copy to the (app) cache dir I think it's the same with the document picker - I just couldn't find reliable information regarding how long files will be kept in temp/cache

Perhaps we can investigate this more and maybe drop the `prepareFile` call altogether 
- Of course it's not very reliable, but...
- I've made tests and images I've picked were still accessible in temp/cache on the next day. I've persited them to Onyx without calling `prepareFile` and uploaded them on the next day
- Leaving files in the cache dir might help with cleanup
- We too make a copy - I think I chose that as a precaution instead of directly moving the file - in case we're dealing with the original, but if we're always dealing with a file coming from cache I think it would be safe to at least use `move` instead of `copy`

It might be best to make saving files to just work with `merge` and `set` 
- I don't know how much of a performance hit it adds to iterate every object being merged, it might be negligible. Maybe we can integrate `prepareFile` as an internal thing, and decide whether we need it at all
- Another reason I opted out of it was - it adds complication and this is only a POC that could have been scraped

lib/storage/index.js
```diff
@@ -0,0 +1,54 @@
...
+                /** We make a temp write to storage just to raise the storage event (below)
+                 * then remove the "dummy" value.
+                 * Listeners would receive everything they need to read the new value from the DB */
+                global.localStorage.setItem(SYNC_ONYX, key);
+                global.localStorage.removeItem(SYNC_ONYX);
+            });
+
+        // This event will be triggered on tabs other than the current tab
+        global.addEventListener('storage', (event) => {
+            if (!event.newValue) {
```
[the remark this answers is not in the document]
kidroca: Yeah I try to explain this with the added code comments 

```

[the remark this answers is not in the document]
Comment (kidroca) [reply lib/storage/index.js:30]: So you're question is whether this would still raise the two events and raise them in the correct order 

```js
global.localStorage.setItem(SYNC_ONYX, key);
global.localStorage.removeItem(SYNC_ONYX);
```

Yes it will
It will first raise an event with something like `{ key: 'SYNC_ONYX', value: 'session' }` 
And immediately after `{ key: 'SYNC_ONYX', value: null }`   
It won't directly skip to the `null` case or fire them in reverse order

lib/storage/index.js
```diff
@@ -0,0 +1,54 @@
...
+            // Notify any listeners in case an Onyx key did change
+            if (event.key === SYNC_ONYX) {
+                Storage.getItem(event.newValue)
+                    .then(value => config.keyChanged(event.newValue, value));
+            }
+        });
+    },
+
+    /**
+     * On web we just return the passed file
```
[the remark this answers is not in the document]
kidroca: Oh, I mean that we return the exact same file we're passing - we're not mutating it, we're not returning another File or a copy

Compared to mobile where we estimate the new uri and return a new object

I guess I'm explaining how/why this method functions differently compared to `index.native.js` and you're not interested in that in the method docs

lib/storage/index.native.js
```diff
@@ -0,0 +1,43 @@
...
+
+    /**
+     * On native it's a no-op because it's not possible to have multiple instances of Onyx (e.g. multiple tabs)
+     */
+    synchronizeInstances() {},
+
+    /**
+     * On native we copy the file locally and return a reference to the new file
+     * This allows us to stringify and save the reference in our regular storage
+     * e.g. as part of a persisted request or an optimistic report action
```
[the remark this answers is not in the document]
kidroca: Will remove unnecessary explanation 
There's already an example usage in `Onyx.js`

image

lib/Onyx.js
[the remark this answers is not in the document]
marcaaron: >One more thing regarding prepareFile - not using it still works, we just do it to have more control by moving the file from temp to our own storage dir.

Maybe we should just remove it for now and add it when we experience some problem on mobile related to needing to store files for a long time.

lib/storage/index.native.js
```diff
@@ -0,0 +1,43 @@
...
+import Storage from './providers/AsyncStorage';
+import FileHandler from './providers/NativeFileHandler';
+
+const nativeStorage = {
+    ...Storage,
+
+    /**
+     * On native it's a no-op because it's not possible to have multiple instances of Onyx (e.g. multiple tabs)
+     */
+    synchronizeInstances() {},
```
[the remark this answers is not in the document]
marcaaron: I am suggesting we put it there so that there is some default. We are overwriting it anyway and then we won't have to worry about "implementing" it here.

lib/storage/index.js
```diff
@@ -0,0 +1,54 @@
...
+        const SYNC_ONYX = 'SYNC_ONYX';
+
+        // Override `setItem` to raise storage events that we can intercept in other tabs
+        this.setItem = (key, value) => Storage.setItem(key, value)
+            .then(() => {
+                /** We make a temp write to storage just to raise the storage event (below)
+                 * then remove the "dummy" value.
+                 * Listeners would receive everything they need to read the new value from the DB */
+                global.localStorage.setItem(SYNC_ONYX, key);
+                global.localStorage.removeItem(SYNC_ONYX);
```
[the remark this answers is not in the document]
kidroca: I've revised the comments, do you still want to extract a helper method for this?

lib/storage/index.native.js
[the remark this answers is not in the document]
kidroca: I've added this change and removed the empty `synchronizeInstances` 

##### Onyx.init
```js
    if (shouldSyncMultipleInstances && _.isFunction(Storage.keepInstancesSync)) {
        Storage.keepInstancesSync(keyChanged);
    }
```

lib/storage/providers/LocalForage.js
```diff
@@ -0,0 +1,63 @@
...
+        })
+            .then(() => pairs);
+    },
+    multiMerge(pairs) {
+        const tasks = pairs.map(([key, delta]) => this.getItem(key)
+            .then((existing) => {
+                let merged = delta;
+
+                if (existing) {
+                    if (_.isObject(existing)) {
```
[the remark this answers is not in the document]
kidroca: I don't see it documented on AsyncStorage but I think this is the same behavior as AsyncStorage's `multiMerge` 

The strategy as I remember is replaces or add keys, with objects it's a no brainer new keys are added existing keys get replaced, but for arrays it's always a bummer
I think this is what happens (and `lodashMerge` works that way) - if we imagine an array with 3 keys as object it will be an object like `{ '0': 'val', '1': 'val', '2': 'val' }` so you can see where merging an array with 2 or 4 keys would get you

lib/storage/providers/NativeFileHandler.js
```diff
@@ -0,0 +1,85 @@
+/**
+ * The native FS handler
+ * We store files in our app's directory space, which does not require special
+ * permission
+ */
+
+import * as FileSystem from 'expo-file-system';
```
kidroca, line 7: I saw `react-native-unimodules` are removed from Expensify/App. Unfortunately Expo-File-System won't work without them (it actually needs the new package `expo-modules`)...

I see the following options 
1. Removing `prepareFile` means we'll not be moving files to "our space" so we also won't need to delete files from our space and it seems we won't need any file handling
2. Replace `expo-file-system` with `react-native-fs`, I've opted for `expo-file-system` because we had already installed `react-native-unimodules` at the time and it was included there. Also I think as an Expo package it would have better support
3. Keep using `expo-file-system` which means installing `expo-modules` in Expensify/App

kidroca (commented):
@marcaaron @tgolen 
I think I've addressed all (or almost) of the required changes
I've just saw the it seems we're dropping the `prepareFile` and we'll just try to keep things simple and use FS/app cache folder for the moment

I'll apply that change tomorrow and re-test
If I've missed anything please bring it up

kidroca:
Update 

Instead of having `Onyx.prepareFile` and make Onyx move/copy file to a location we leave that to be handled by App
When an attachment is picked the `AttahcmentPicker` will persist it to App/cache folder, it is then available until `cache` is cleared
This makes it available to save only the reference to the File in our key/val storage
This information is enough to work with the file - use it in an `<Image />` or in a upload request

Some points to think about
1. clear the `cache` folder when the user logs out.
2. remove items from the `cache` folder when the object that references them from Onyx is removed

2) is a bit tricky.
When we call `Onyx.remove` for some key, we can traverse the `value` for files and remove them from `cache` but there's catch - we don't know whether the file isn't reference in some other Onyx key and that we shouldn't remove it
Perhaps we don't need to do anything about it as cache should be auto cleared

package.json
```diff
@@ -38,6 +38,7 @@
     "jest": "^26.5.2",
     "jest-cli": "^26.5.2",
     "jsdoc-to-markdown": "^7.1.0",
+    "localforage": "^1.10.0",
```
marcaaron, line 41: Can this be removed?

tests/unit/cacheEvictionTest.js
```diff
@@ -1,4 +1,4 @@
-import AsyncStorage from '@react-native-async-storage/async-storage';
+import Storage from '../../lib/storage';
```
marcaaron, line 1: So, this will use the `AsyncStorage` not `localforage`? Does `localforage` have mocks and should we write tests that show the providers are equivalent?

lib/Onyx.js
```diff
@@ -547,7 +547,7 @@ function set(key, value) {
  * @return {Array} an array of stringified key - value pairs
```
marcaaron, line 547: This comment is out of date now that we are not using `JSON.stringify()`

lib/storage/WebStorage.js
```diff
@@ -0,0 +1,49 @@
...
+        this.setItem = (key, value) => Storage.setItem(key, value)
+            .then(() => raiseStorageSyncEvent(key));
+
+        this.removeItem = key => Storage.removeItem(key)
+            .then(() => raiseStorageSyncEvent(key));
+
+        // If we just call Storage.clear other tabs will have no idea which keys were available previously
+        // so that they can call keysChanged for them. That's why we iterate and remove keys one by one
+        this.clear = () => Storage.getAllKeys()
+            .then(keys => _.forEach(keys, key => this.removeItem(key)));
```
marcaaron, line 33: Just a thought, but curious to see how this performs in the case of a sign out. There is some temporary code added [here](https://github.com/Expensify/App/blob/main/src/libs/SignoutManager.js) to ensure the sign outs are synced and I think we can remove it if this works how I think it does.

lib/storage/providers/AsyncStorage.js
```diff
@@ -0,0 +1,97 @@
...
+    },
+
+    /**
+     * Get multiple key-value pairs for the give array of keys in a batch
+     * @param {String[]} keys
+     * @return {Promise<Array<[key, value]>>}
+     */
+    multiGet(keys) {
+        return AsyncStorage.multiGet(keys)
+            .then(pairs => pairs.map(([key, value]) => [key, value && JSON.parse(value)]));
```
marcaaron, line 44: use `_`

lib/storage/providers/AsyncStorage.js
```diff
@@ -0,0 +1,97 @@
...
+        return AsyncStorage.setItem(key, prepareValueForStorage(value));
+    },
+
+    /**
+     * Stores multiple key-value pairs in a batch
+     * @param {Array<[key, value]>} pairs
+     * @return {Promise<void>}
+     */
+    multiSet(pairs) {
+        const stringPairs = pairs.map(([key, value]) => [key, prepareValueForStorage(value)]);
```
marcaaron, line 63: `_.map()`

lib/storage/providers/AsyncStorage.js
```diff
@@ -0,0 +1,97 @@
...
+        return AsyncStorage.multiSet(stringPairs);
+    },
+
+    /**
+     * Multiple merging of existing and new values in a batch
+     * @param {Array<[key, value]>} pairs
+     * @return {Promise<void>}
+     */
+    multiMerge(pairs) {
+        const stringPairs = pairs.map(([key, value]) => [key, prepareValueForStorage(value)]);
```
marcaaron, line 73: `_.map()`

marcaaron (dismissed):
LGTM just left a few comments about style and a question about tests.

lib/storage/__mocks__/index.js
```diff
@@ -0,0 +1,3 @@
+import WebStorage from '../WebStorage';
```
tgolen, line 1: I see we have have `App/__mocks__`. Should we be consistent with where we place the mocks?

kidroca: I can move this to `{root}/__mocks__/lib/storage/index.js` as long as it represents the correct folder structure the mock will be used. 

Keep in mind Jest suggest to put the `__mocks__` directory adjacent to the module being mocked so that's how everybody's been taught. In this case we're mocking our own `index.js` file so I put it there
https://jestjs.io/docs/manual-mocks#mocking-user-modules

In App we're only mocking external dependencies coming from `node_modules` and so `__mocks__` is at the root

I supposed you would ask "why do we mock something that's not external?". It's because we're testing functionality that doesn't depend on the actual storage implementation, we're asserting that certain calls are made at the expected times with the expected data. It was the same with AsyncStorage - we used the mock of AsyncStorage to verify we're calling `setItem` with the expected parameters. Here we're passing `WebStorage` which would internally use the mock of AsyncStorage, but would also execute the custom logic we've added (e.g. stringify JSONs)

kidroca (commented):
Pushing an update regarding the requested changes

It's seems this PR/POC is going to be accepted so I can add some tests tomorrow?

lib/storage/WebStorage.js
kidroca: I've altered the logic slightly 
1. WebStorage: wait for all remove calls to finish and then resolve clear - use `Promise.all`
2. Bug fix: on the new tab we were only `keysChanged` - which kept the old value in `cache`. Added an update to set the new value in cache and call `keysChanged` as well 

> There is some temporary code added [here](https://github.com/Expensify/App/blob/main/src/libs/SignoutManager.js) to ensure the sign outs are synced and I think we can remove it if this works how I think it does.

That `true` I've tested this and sign out works correctly with or without the `SignoutManager`

https://user-images.githubusercontent.com/12156624/140195176-0a584ac7-648c-45e1-8dc5-3b0f7a4ef15c.mp4

tgolen (approved):
It's looking very good to me at this stage!

tests/unit/cacheEvictionTest.js
marcaaron: >We should write tests that show the providers do what they are intended to do

Cool. I think it would be good to write some. But not a blocker for me if this is otherwise ready to get merged. I guess I just have a vague concern that by having two storage implementations we need to remember to always keep the interface the same and make sure we are not testing changes against only one provider, but I don't know if that will be an issue in practice.

kidroca:
I've prepared the App PR with videos and retested everything: https://github.com/Expensify/App/pull/4908
Found one issue on [mobile **web** ios](https://github.com/Expensify/App/pull/4908#issuecomment-961115405)
Otherwise everything seems ok 

I'm not sure how to proceed, here are some questions
- should I remove the Draft / Hold state and mark the PR as ready?
- should I write tests now?
- or try to fix the mobile web problem first?

tgolen:
I'd say it would be best to fix the mobile web problem first, write some
super basic tests, then remove hold/draft and get it merged.

kidroca:
@tgolen 
I've found out what's causing the problem - the issues occurs because once we read the persisted requests we immediately delete them from storage, which also deletes file data
When we read a file from IndexedDB we receive only the metadata, we don't get the binary data - we receive a reference with the necessary hooks to read it if we want to. When we clear storage we lose the binary data. Here are more details: https://github.com/Expensify/App/pull/4908#issuecomment-967139238

I don't know why the issue occurs only on iOS Safari/Chrome, perhaps the other web platforms have some protection that doesn't delete data immediately

Any thought on how to proceed?

Since we've identified the problem I think we can continue and merge this the way it is. The offline file upload feature would work everywhere but mobile safari/chrome. Regular file upload still works there. 
Or we can decide and apply an update to the offline request queue and then continue here

tgolen:
I'm pretty hesitant to merge this with that big of a known bug. I'd like to see if we can fix it, or at least prevent the user from being affected (ie. disable the feature in those platforms until we fix the bug). I'm having trouble understanding the exact nature and reproduction steps of the bug though. Can you go into some more detail on why it happens and then maybe with me and @marcaaron we can try to figure something out.

> the issues occurs because once we read the persisted requests we immediately delete them from storage, which also deletes file data

Does this happen because of changes in this PR, or because of localForage, or why? Can we control this at all to have it work better for us somehow? Where does this happen in the code?

kidroca:
It's not something that this PR introduced, it's how the network queue works

The details are here: https://github.com/Expensify/App/pull/4908#issuecomment-967139238 steps, code causing the issue and some ideas on how to address it

tgolen:
Ah, perfect! Just what I needed. I'll comment over there.

kidroca:
As per this comment: https://github.com/Expensify/App/pull/4908#issuecomment-992594869 I've removed the draft status of this PR

thienlnam:
@kidroca This shouldn't be on hold anymore right? Is this good to be reviewed/merged?

kidroca:
Yes, 
Nothing had actually changed here since the last review - but I found a way to get this be merged - attachment persistence will remain disabled (in App) until the issue (in App) with the attachment getting deleted prematurely is resolved (working on that here: https://github.com/Expensify/App/pull/6556)

As a summary:
Tim approved the PR, Marc left a comment but it's not a blocker: https://github.com/Expensify/react-native-onyx/pull/102#pullrequestreview-795798405
We can open a follow up PR to add a few more tests

thienlnam (approved):

thienlnam:
Cool cool, I took a look over the PR and things look good to me.

I opened up this issue as a reminder to add tests and add some documentation regarding these changes https://github.com/Expensify/react-native-onyx/issues/112

Since the PR is approved by both Tim and Marc, I will merge to push this forward

PR #117, opened 2022-01-22
Fix: handle `null` values and stringify them

kidroca:
### Details

AsyncStorage would throw an error when it is called with a `null` value instead of storing `null` for the key
In the past we stringified every value we tried to set in storage, this PR brings back the original behavior 

### Related Issues

$ https://github.com/Expensify/App/issues/7361

### Automated Tests

Updated AsyncStorage set tests to cover `null` input and verify it's getting stringified


sketchydroide (approved):
I'm happy with this change, it seems that you are familiar with the code (having done it yourself). and the the solution seems solid

PR #120, opened 2022-02-25
Prevent setting state on unmounted withOnyx components

marcaaron:
### Details
Calls to `Onyx.connect()` can call `setState()` after a component unmounts. This stashes a reference to the `connectionID` in the connection map. When `withOnyx()` unmounts it should be calling `disconnect()` which will remove the connection from the map. However, we don't check to see if it's been deleted when we call `sendDataToConnection()`.

### Related Issues

https://github.com/Expensify/Expensify/issues/198762

### DIFF
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -358,6 +358,12 @@ function keyChanged(key, data) {
  * @param {String} key
  */
 function sendDataToConnection(config, val, key) {
+    // If the mapping no longer exists then we should not send any data.
+    // This means our subscriber disconnected or withOnyx wrapped component unmounted.
+    if (!callbackToStateMapping[config.connectionID]) {
+        return;
+    }
+
     if (config.withOnyxInstance) {
         config.withOnyxInstance.setWithOnyxState(config.statePropertyName, val);
     } else if (_.isFunction(config.callback)) {
@@ -388,6 +394,7 @@ function sendDataToConnection(config, val, key) {
 function connect(mapping) {
     const connectionID = lastConnectionID++;
     callbackToStateMapping[connectionID] = mapping;
+    callbackToStateMapping[connectionID].connectionID = connectionID;
 
     if (mapping.initWithStoredValues === false) {
         return connectionID;


marcaaron:
@neil-marcellini after this is merged we can update your PR with the merge commit

TomatoToaster (commented):
Great find! If this fixes that console error forever that's great

TomatoToaster (approved):

lib/Onyx.js
```diff
@@ -388,6 +394,7 @@ function sendDataToConnection(config, val, key) {
 function connect(mapping) {
     const connectionID = lastConnectionID++;
     callbackToStateMapping[connectionID] = mapping;
+    callbackToStateMapping[connectionID].connectionID = connectionID;
```
roryabraham, line 397: Do we need to give the `mapping` itself a `connectionID` property here? Because it looks like we call `sendDataToConnection(mapping)`, and then in that function try to access `config.connectionID`, which I _think_ would always be `undefined` unless I'm missing something?

lib/Onyx.js
```diff
@@ -358,6 +358,12 @@ function keyChanged(key, data) {
  * @param {String} key
  */
 function sendDataToConnection(config, val, key) {
+    // If the mapping no longer exists then we should not send any data.
+    // This means our subscriber disconnected or withOnyx wrapped component unmounted.
+    if (!callbackToStateMapping[config.connectionID]) {
```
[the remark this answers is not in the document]
marcaaron: Gets added in the `connect()`method.

lib/Onyx.js
marcaaron: `callbackToStateMapping[connectionID] === mapping` the assignment creates an object reference.

neil-marcellini (approved):
This solves the errors I was getting with Onyx, so as far as I can tell it's good to go.

lib/Onyx.js
roryabraham: Oh, so since we do `callbackToStateMapping[connectionID] = mapping;`, then doing `callbackToStateMapping[connectionID].connectionID = connectionID` _also_ sets the `connectionID` on the `mapping` object? TIL

marcaaron: Good overview here: https://eloquentjavascript.net/04_data.html#h_C3n45IkMhg

kidroca: Mutating input parameters is considered a bad practice, some more strict languages would not even allow it
It might not cause problems here, but someone else can see the code and apply the same in other places
Besides it's possible to achieve the same thing without mutating the input by introducing a local variable

```js
const privateMapping = {
  ...mapping,
  connectionID, 
};

callbackToStateMapping[connectionID] = privateMapping;
```

Also because of referential equality it should be possible to achieve the same without capturing `connectionID`

```js
    // If the mapping no longer exists then we should not send any data.
    // This means our subscriber disconnected or withOnyx wrapped component unmounted.
    if (_.every(callbackToStateMapping(mapping => mapping !== config)) {
      return;
    }
```

marcaaron: Why is it considered a bad practice and what kinds of problems do you think it can cause? We are not accessing the connection mapping besides internally once it is passed here so I did not worry about mutations in this case.

kidroca: > what kinds of problems do you think it can cause?

For this particular usage primarily confusion
You know the code a great deal and that a mutation won't cause a problem in this case. 
Rory knows the code as well yet it wasn't obvious `mapping` received `connectionID` through mutation 

> Why is it considered a bad practice

In particular for mutating input arguments:
1. Unexpected behavior - unless the function's job is to mutate the object. You would not expect a function like `print(object)` to mutate the passed object, why should a function like `connect` be any different?
2. You don't see the whole object shape (reading code) if some properties are added later with mutation 

small airbnb thread on the subject of param reasinging: https://github.com/airbnb/javascript/issues/641

In general:
It depends on the case. Contexts that allow mutations usually enforce it in an action, so it's clear it's intentional and you capture a stack trace

marcaaron: Agree it depends on the context/case, but not sure if I agree there is anything inherently confusing about this. I might not be understanding the full argument about why this is "bad" or what the unexpected behavior would be in this case.

Confusion is subjective. Someone could be equally confused about why we are using the spread syntax or checking for equality on an object reference (not very obvious IMO). I'm not convinced there's anything we need to do here, but do appreciate the thoughts.

kidroca: Less confusing way with mutation:

```js
    const connectionID = lastConnectionID++;
    mapping.connectionID = connectionID;
    callbackToStateMapping[connectionID] = mapping;
```

This also makes the mutation obvious

marcaaron: Noted and PRs Welcome 😆

Issue #119, opened 2022-02-23
[Performance] `localforage.setItem()` is causing blocking behavior on init for web clients


marcaaron:
## Problem

When the app inits and many keys need to be set things get really slow on web. It currently takes me about 10-12 seconds to switch to a new chat when the app inits.

We've narrowed down the source of the problem to:

1. Not stringifying IndexedDB values. For whatever reason IDB takes a while to store structured JSON.
2. Calling `setItem()` too many times in a row

Addressing either one should work. Both seem to be causing similar rates of pain.

## Solution

1. Start stringifying values - complicated because some of the JSON is not serializable i.e. stuff with attachments.
2. Defer the writing to IndexedDB until we are idle or just slow it down or something at least. We are using a memory cache already so I can't think of a reason why our writes must happen immediately on page load.

More ideas here: https://rxdb.info/slow-indexeddb.html, but the general suggestion is to use fewer transactions. That is not possible with localforage so we might want to look elsewhere - maybe the plugin mentioned here can help https://github.com/localForage/localForage/issues/315

I did a test here where I just made all the writes synchronous (probably not the solution we want), but it improves things immensely

https://github.com/Expensify/react-native-onyx/pull/118

roryabraham:
closing as stale

PR #122, opened 2022-03-10
Add sync writes to multiMerge() in WebStorage. Fix caching when multiMerge() runs.

marcaaron:
### Details

cc @Julesssss @kidroca Second attempt at alleviating the IDB issues

### Related Issues
https://github.com/Expensify/react-native-onyx/issues/119

### Automated Tests
Added a new automated test specifically for the weirdness reported [here](https://github.com/Expensify/App/pull/8053#pullrequestreview-904686478).

### Linked PRs
https://github.com/Expensify/App/pull/8053

### DIFF
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -808,7 +808,7 @@ function init({
     keys = {},
     initialKeyStates = {},
     safeEvictionKeys = [],
-    maxCachedKeysCount = 55,
+    maxCachedKeysCount = 1000,
     captureMetrics = false,
     shouldSyncMultipleInstances = Boolean(global.localStorage),
 } = {}) {
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -768,11 +768,12 @@ function mergeCollection(collectionKey, collection) {
                 promises.push(Storage.multiSet(keyValuePairsForNewCollection));
             }
 
-            // Merge original data to cache
-            cache.merge(collection);
-
-            // Optimistically inform collection subscribers
-            keysChanged(collectionKey, collection);
+            // Prefill cache if necessary by calling get() on any existing keys and then merge original data to cache
+            // and update all subscribers
+            Promise.all(_.map(existingKeys, get)).then(() => {
+                cache.merge(collection);
+                keysChanged(collectionKey, collection);
+            });
 
             return Promise.all(promises)
                 .catch(error => evictStorageAndRetry(error, mergeCollection, collection));
--- a/lib/SyncQueue.js
+++ b/lib/SyncQueue.js
@@ -1,4 +1,20 @@
+/**
+ * Synchronous queue that can be used to ensure promise based tasks are run in sequence.
+ * Pass to the constructor a function that returns a promise to run the task then add data.
+ *
+ * @example
+ *
+ *     const queue = new SyncQueue(({key, val}) => {
+ *         return someAsyncProcess(key, val);
+ *     });
+ *
+ *     queue.push({key: 1, val: '1'});
+ *     queue.push({key: 2, val: '2'});
+ */
 export default class SyncQueue {
+    /**
+     * @param {Function} run - must return a promise
+     */
     constructor(run) {
         this.queue = [];
         this.isProcessing = false;
--- a/lib/SyncQueue.js
+++ b/lib/SyncQueue.js
@@ -1,0 +1,35 @@
+export default class SyncQueue {
+    constructor(run) {
+        this.queue = [];
+        this.isProcessing = false;
+        this.run = run;
+    }
+
+    process() {
+        if (this.isProcessing || this.queue.length === 0) {
+            return;
+        }
+
+        this.isProcessing = true;
+
+        const {data, resolve, reject} = this.queue.shift();
+        this.run(data)
+            .then(resolve)
+            .catch(reject)
+            .finally(() => {
+                this.isProcessing = false;
+                this.process();
+            });
+    }
+
+    /**
+     * @param {*} data
+     * @returns {Promise}
+     */
+    push(data) {
+        return new Promise((resolve, reject) => {
+            this.queue.push({resolve, reject, data});
+            this.process();
+        });
+    }
+}
--- a/lib/storage/providers/LocalForage.js
+++ b/lib/storage/providers/LocalForage.js
@@ -19,18 +19,18 @@ const provider = {
      * So, we are slowing this process down by waiting until one write is complete before moving on
      * to the next.
      */
-    setItemQueue: new SyncQueue(({key, value}) => localforage.setItem(key, value)),
-    multiMergeQueue: new SyncQueue((pairs) => {
-        const tasks = _.map(pairs, ([key, partialValue]) => this.default.getItem(key)
-            .then((existingValue) => {
-                const newValue = _.isObject(existingValue)
-                    ? lodashMerge({}, existingValue, partialValue)
-                    : partialValue;
-                return this.default.setItem(key, newValue);
-            }));
+    setItemQueue: new SyncQueue(({key, value, shouldMerge}) => {
+        if (shouldMerge) {
+            return localforage.getItem(key)
+                .then((existingValue) => {
+                    const newValue = _.isObject(existingValue)
+                        ? lodashMerge({}, existingValue, value)
+                        : value;
+                    return localforage.setItem(key, newValue);
+                });
+        }
 
-        // We're returning Promise.resolve, otherwise the array of task results will be returned to the caller
-        return Promise.all(tasks).then(() => Promise.resolve());
+        return localforage.setItem(key, value);
     }),
 
     /**
@@ -54,7 +54,10 @@ const provider = {
      * @return {Promise<void>}
      */
     multiMerge(pairs) {
-        return this.multiMergeQueue.push(pairs);
+        const tasks = _.map(pairs, ([key, value]) => this.setItemQueue.push({key, value, shouldMerge: true}));
+
+        // We're returning Promise.resolve, otherwise the array of task results will be returned to the caller
+        return Promise.all(tasks).then(() => Promise.resolve());
     },
 
     /**
--- a/lib/storage/providers/LocalForage.js
+++ b/lib/storage/providers/LocalForage.js
@@ -7,39 +7,32 @@
 import localforage from 'localforage';
 import _ from 'underscore';
 import lodashMerge from 'lodash/merge';
+import SyncQueue from '../../SyncQueue';
 
 localforage.config({
     name: 'OnyxDB'
 });
 
-const writeQueue = [];
-let isQueueProcessing = false;
-
-/**
- * Writing very quickly to IndexedDB causes performance issues and can lock up the page and lead to jank.
- * So, we are slowing this process down by waiting until one write is complete before moving on
- * to the next.
- */
-function processNextWrite() {
-    if (isQueueProcessing || writeQueue.length === 0) {
-        return;
-    }
-
-    isQueueProcessing = true;
-
-    const {
-        key, value, resolve, reject,
-    } = writeQueue.shift();
-    localforage.setItem(key, value)
-        .then(resolve)
-        .catch(reject)
-        .finally(() => {
-            isQueueProcessing = false;
-            processNextWrite();
-        });
-}
-
 const provider = {
+    /**
+     * Writing very quickly to IndexedDB causes performance issues and can lock up the page and lead to jank.
+     * So, we are slowing this process down by waiting until one write is complete before moving on
+     * to the next.
+     */
+    setItemQueue: new SyncQueue(({key, value}) => localforage.setItem(key, value)),
+    multiMergeQueue: new SyncQueue((pairs) => {
+        const tasks = _.map(pairs, ([key, partialValue]) => this.default.getItem(key)
+            .then((existingValue) => {
+                const newValue = _.isObject(existingValue)
+                    ? lodashMerge({}, existingValue, partialValue)
+                    : partialValue;
+                return this.default.setItem(key, newValue);
+            }));
+
+        // We're returning Promise.resolve, otherwise the array of task results will be returned to the caller
+        return Promise.all(tasks).then(() => Promise.resolve());
+    }),
+
     /**
      * Get multiple key-value pairs for the give array of keys in a batch
      * @param {String[]} keys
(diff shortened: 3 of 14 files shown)


tests/unit/onyxMultiMergeWebStorageTest.js
```diff
@@ -0,0 +1,70 @@
...
+    let Storage;
+    beforeAll(() => {
+        jest.mock('../../lib/storage');
+        Onyx = require('../../index').default;
+        Storage = require('../../lib/storage').default;
+        jest.useRealTimers();
+    });
+
+    it('mergeCollection', () => {
+        expect(Storage.name).toBe('WebStorage');
```
kidroca, line 31: Does the success of the test depend on the underlying storage provider? - it shouldn't 
`mergeCollection` should be considered working correctly if it correctly delegates work to a Storage provider

Testing that `mergeCollection` is interacting correctly with cache is valuable

tests/unit/onyxMultiMergeWebStorageTest.js
```diff
@@ -0,0 +1,70 @@
...
+
+                // Expect that our new data has merged with the existing data in the cache
+                expect(OnyxCache.getValue('test_1')).toEqual(finalObject);
+                expect(OnyxCache.getValue('test_2')).toEqual(finalObject);
+                expect(OnyxCache.getValue('test_3')).toEqual(finalObject);
+
+                // Check the storage to make sure our data has been saved
+                expect(localforageMock.storageMap.test_1).toEqual(finalObject);
+                expect(localforageMock.storageMap.test_2).toEqual(finalObject);
+                expect(localforageMock.storageMap.test_3).toEqual(finalObject);
```
kidroca, line 67: What I'm saying is changes made around mocking are not necessary here, because which mock is used shouldn't matter
We can assert that `Storage.multiSet` and/or `Storage.multiMerge` are getting called correctly

The changes made to `__mocks__/localforage.js` can help us test `LocalForage.multiSet` and `LocalForage.multiMerge` are working correctly with the queues here: [LocalForageProviderTest](https://github.com/Expensify/react-native-onyx/blob/032ff54b4374c93ee12676792a6e029c61ea1e06/tests/unit/storage/providers/LocalForageProviderTest.js#L1-L6)

lib/storage/providers/LocalForage.js
```diff
@@ -7,39 +7,32 @@
...
+     * to the next.
+     */
+    setItemQueue: new SyncQueue(({key, value}) => localforage.setItem(key, value)),
+    multiMergeQueue: new SyncQueue((pairs) => {
+        const tasks = _.map(pairs, ([key, partialValue]) => this.default.getItem(key)
+            .then((existingValue) => {
+                const newValue = _.isObject(existingValue)
+                    ? lodashMerge({}, existingValue, partialValue)
+                    : partialValue;
+                return this.default.setItem(key, newValue);
```
kidroca, line 29: Surprising to see `this` usages are still working

kidroca (commented):
I think we still have a problem with multiMerge, though it might be more subtle now

lib/Onyx.js
```diff
@@ -768,11 +768,12 @@ function mergeCollection(collectionKey, collection) {
...
-            cache.merge(collection);
-
-            // Optimistically inform collection subscribers
-            keysChanged(collectionKey, collection);
+            // Prefill cache if necessary by calling get() on any existing keys and then merge original data to cache
+            // and update all subscribers
+            Promise.all(_.map(existingKeys, get)).then(() => {
+                cache.merge(collection);
+                keysChanged(collectionKey, collection);
+            });
```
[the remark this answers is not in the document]
marcaaron: >if get has to read from disk, data might not yet be written to disk (when write is still pending), and still result in partial data being added to cache

If you can, please explain how exactly it can result in partial data being added to the cache? I would love to write a test for it and we can come up with a solution if it's possible, but I want to make sure there we understand how it can happen.

>delaying cache update opens a door for problems - other get/connect calls might use stale cache information

One alternative I can think of is to prefill the cache.

>it might delay the UI update a bit, since now keysChanged would be delayed if get actually has to read something

Yes, but better for something to be delayed than send partial data without previous data initially in storage.

>The goal with cache is to update it sync as fast as possible to avoid concurrency issues

I think I understand this argument, but I'm not sure that's how `merge()` works today. Here's the implementation for `merge()` which also potentially blocks with a `get()` call

https://github.com/Expensify/react-native-onyx/blob/032ff54b4374c93ee12676792a6e029c61ea1e06/lib/Onyx.js#L672-L681

`set()` is the only thing which modifies the data in the cache first before reading existing values 

https://github.com/Expensify/react-native-onyx/blob/032ff54b4374c93ee12676792a6e029c61ea1e06/lib/Onyx.js#L548-L551

>Perhaps you'd like this alternative:
>Keep code here unchanged
>Update code in cache.merge to exclude merging any keys that are not currently present in cache

I do like this idea and it solves this problem in a different way, but we would defer the filling of the cache. I'm not sure if that is bad or good or doesn't make a difference and would be willing to try it if it sounds simpler. Thanks!

lib/storage/providers/LocalForage.js
```diff
@@ -7,39 +7,32 @@
...
+    /**
+     * Writing very quickly to IndexedDB causes performance issues and can lock up the page and lead to jank.
+     * So, we are slowing this process down by waiting until one write is complete before moving on
+     * to the next.
+     */
+    setItemQueue: new SyncQueue(({key, value}) => localforage.setItem(key, value)),
+    multiMergeQueue: new SyncQueue((pairs) => {
+        const tasks = _.map(pairs, ([key, partialValue]) => this.default.getItem(key)
+            .then((existingValue) => {
+                const newValue = _.isObject(existingValue)
```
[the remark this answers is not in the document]
marcaaron: Let me see if I can test this scenario, but I think you are correct.

tests/unit/onyxMultiMergeWebStorageTest.js
marcaaron: Since the implementation we are having issues with is the storage provider then I want to make sure things work correctly - it also took me some time to get this working. Testing against AsyncStorage would not touch any of the code we're proposing to add. Sorry, I think you are trying to say that we should maybe:

1. Have one set of tests
2. Run them against both storage providers

But not totally sure. Please clarify if there is a specific suggestion, thanks!

tests/unit/onyxMultiMergeWebStorageTest.js
marcaaron: I am probably not understanding the message. The test is helping me to verify the changes by using a mock that works similar to storage. Why is it not necessary? The previous tests did not test this scenario and things were broken. So I'm adding a case to cover the thing that was breaking. What should we do instead?

lib/Onyx.js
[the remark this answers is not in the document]
marcaaron: Ok so pulling on that last thread and I can't get it working. It leads to a stale cache that is missing the data. Here are the changes I made:

- Reverted the change here.
- Added this so we would not set data if no existing key was in the cache

```diff
@@ -107,11 +110,19 @@ class OnyxCache {
     }

     /**
-     * Deep merge data to cache, any non existing keys will be created
+     * Deep merge data to cache, any non existing keys will not be merged
      * @param {Record<string, *>} data - a map of (cache) key - values
      */
     merge(data) {
-        this.storageMap = lodashMerge({}, this.storageMap, data);
+        // Filter any keys we have not yet cached
+        const newData = _.reduce(data, (prev, curr, key) => {
+            if (this.hasCacheForKey(key)) {
+                return {...prev, [key]: curr};
+            }
+            return prev;
+        }, {});
+
+        this.storageMap = lodashMerge({}, this.storageMap, newData);

         const storageKeys = this.getAllKeys();
```
- Re-ran tests and added a subscriber (which should trigger the cache fill)
- Storage ends up correct, but cache is wrong 

I think it's because the subscriber can be added at anytime while we are adding new data. We already skipped the cache for those changes and yes they are making it to storage correctly. But if you add a subscriber at the same time it will only get the result of what has been written and save that to the cache.

Julesssss:
Sorry, I didn't get to this today -- I stopped due to a family issue. Adding it to my list for tomorrow.

lib/Onyx.js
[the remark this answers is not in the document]
marcaaron: Hey @kidroca, I couldn't figure this one out, but got your other idea working. Let me know if you have any other thoughts here.

lib/storage/providers/LocalForage.js
marcaaron: Just realized we can use `localforage.setItem` directly now

lib/storage/providers/LocalForage.js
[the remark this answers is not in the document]
marcaaron: Yep got this to break the tests by using some delays and `jest.useRealTimers()` wrapped on `setItem()`... should be working correctly now.

>yes: is any of the pending writes for a key we're merging

The specificity might be good here, but I just decided to create tasks for each key we are merging and have them all block. Could be optimized later, but seems to solve the issue for now.

tests/unit/onyxMultiMergeWebStorageTest.js
kidroca: Actually I don't think it would be worth it to run against all storage providers

We've introduced 2 changes: 
1. Queue LocalForage writes - this can be tested outside of `mergeCollection`
2. We changed the way `meregeCollection` updates cache - this shouldn't depend on specific storage provider

Specifically for the bug with partial cache I think the issue should be happening with any storage provider, because we evicted cache keys

tests/unit/onyxMultiMergeWebStorageTest.js
kidroca: By testing what's stored inside `storageMap` of the mock we're asserting that `WebStorage` works correctly when `Storage.multiMerge` is called within `mergeCollection` 

We can have a test where we import `LocalForage` and invoke `multiMerge` on it directly

tests/unit/onyxMultiMergeWebStorageTest.js
```diff
@@ -0,0 +1,70 @@
...
+            test_2: additionalDataOne,
+            test_3: additionalDataOne,
+        });
+
+        const additionalDataTwo = {d: 'd'};
+        Onyx.mergeCollection(ONYX_KEYS.COLLECTION.TEST_KEY, {
+            test_1: additionalDataTwo,
+            test_2: additionalDataTwo,
+            test_3: additionalDataTwo,
+        });
```
kidroca, line 51: I'm not sure this captures the bug with cache
If you revert `mergeCollection` changes would the test fail because `finalObject` is not in OnyxCache? 

I think to cover the cache bug the test should:
1. merge `additionalDataOne` 
2. wait for the merge to complete
3. evict test_1, test_2 from cache
4. merge `additionalDataTwo`
5. Assert `finalObject` is in cache

lib/Onyx.js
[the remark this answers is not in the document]
kidroca: I think prefilling cache might be our safest and optimal option here

It seems that for Native:
1. Calling AsyncStorage.get
2. Then calling AsyncStorage.set, before the get is completed
3. `get` would resolve with the value from step 2)

I'm not certain this would work out the same way on Web/Desktop
It used to work that way when we used `window.localStorage` because local storage calls are sync and happen sequentially

So we're back to: "Oh, Cache save us from concurrency issues" 😄 
And to do that it might be best to leverage `multiGet` and get everything but safe eviction keys in cache during init
Then we don't have to delay or change logic that updates cache and cache should never become stale

---

Another strategy could be to recreated the above 3 steps ourselves 
- capture pending writes (we already have a queue)
- when a read happens check whether we're not writing or are about to write the requested key
- if we are - resolve the get with the value from the write queue, otherwise read from disk

tests/unit/onyxMultiMergeWebStorageTest.js
marcaaron: That sounds like a good additional test. Just to clarify this is covering the case where we are not _first_ grabbing _existing data_ and merging it into the cache. We start with some initial data, then call mergeCollection, then call it again. Before the changes we would end up skipping the initial value and merging partial data into the cache.

tests/unit/onyxMultiMergeWebStorageTest.js
marcaaron: I like this test because it more accurately represents the code flow that went wrong here. It's a combination of Onyx changes and storage provider changes so I think it is reasonable to test them both. But I am possibly being lazy 😄

marcaaron:
Still need to make sure there are no major issues for native with this change. But the performance is pretty nice on web and the cache issues seem resolved. Gonna take it out of draft.

tests/unit/onyxMultiMergeWebStorageTest.js
kidroca: Yeah, that makes sense it's just that we're not doing it anywhere else - using specific storage provider and mock

So it's not clear when or why tests need to use a specific storage provider, e.g. why this tests does but others don't?

What we see changed in Onyx.mergeCollection is a change that should work regardless of the storage provider
We can write 2x tests with the other provider as well, but then it results in more maintenance outweighing the value of the tests

I think we can achieve the same test by:
1. Mock `Storage.getItem`
2. Verity `Storage.getItem` is being called when a given key is not available in cache
3. Verify returned mock data is used in cache - e.g. cache matches expected final value

To me this gives us more information regarding what we expect to happen 
_(First time here I didn't even realize we we were testing the case where we don't have data in cache but have it on disk)_

Julesssss:
Seems like we're still discussing solutions. @marcaaron is this still the blocking PR?

thienlnam (approved):
Changes look good! Just had one small question/comment

lib/SyncQueue.js
```diff
@@ -0,0 +1,51 @@
...
+        if (this.isProcessing || this.queue.length === 0) {
+            return;
+        }
+
+        this.isProcessing = true;
+
+        const {data, resolve, reject} = this.queue.shift();
+        this.run(data)
+            .then(resolve)
+            .catch(reject)
```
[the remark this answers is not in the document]
marcaaron: We do have some handling here (but it could be better). In this case, we're just letting the original caller handle the error though and don't need more specific handling (at least not for now since the error will reach this code here)..

https://github.com/Expensify/react-native-onyx/blob/032ff54b4374c93ee12676792a6e029c61ea1e06/lib/Onyx.js#L778

marcaaron:
I'm gonna move ahead with this with a caveat that there's probably more work to be done here. The performance is suffering a lot for web/desktop users internally so this should be a welcome temporary improvement

Things to follow up on:

- Look into improving the tests further
- Explore a more robust solution to fix IndexedDB slow transactions

Issue #125, opened 2022-04-19
initializeWithDefaultKeyStates overrides data that exists in the Onyx cache


neil-marcellini:
## Problem
When Onyx.clear() is called the storage is cleared and then Onyx is initialized with the defaultKeyStates provided by the application. In NewDot transition flows, when a user is signing into a different account, Onyx is cleared to sign out the old user and then the new user is immediately signed in, setting a new authToken on the session. By the time initializeWithDefaultKeyStates is called, the new user has been signed in. 

In the method, the defaultKeyStates are merged into the key states existing in storage, which are all null after clearing the storage. Then keyChanged is called for each (key, value) pair from the defaultKeyStates. The default key state for session is `{loading: false, shouldShowComposeInput: true}`, and therefore the subscribers are notified that the session has been updated to this value, and the user is signed out.

Here you can see how the currentValue from the cache contains the full, merged, session information, but the val from defaultKeyStates only contains limited information for initializing.
<img width="334" alt="image" src="https://user-images.githubusercontent.com/26260477/163896241-39572908-d7dc-490e-929a-3b50bb1b647e.png">

## Solution
Notify the subscribers with merged cache data when initializing with default key states.

marcaaron:
I think this bug needs to be reframed a bit.

I would actually expect a call to `Onyx.clear()` to be _blocking_ in that if we attempted to set any new data to Onyx while we were clearing it would throw these updates away to prevent cache pollution and cause the conflicts we are describing here.

If you need to set data after storage and cache is cleared then the obvious answer is to just wait for the storage and cache to clear then add the new data.

The problem with the solution and linked PR is that we are making assumptions about the data we are setting when `Onyx.clear()` is running. Is it something that should be in the store after `Onyx.clear()` finishes or not? Was it from the previous user session or meant for the next?

neil-marcellini:
After a lot of [discussion in Slack](https://expensify.slack.com/archives/C03TQ48KC/p1651522644295769) we agreed on a solution where we set the default key states while clearing the cache, and remove the call to initializeWithDefaultKeyStates.

marcaaron:
Fixed in https://github.com/Expensify/react-native-onyx/pull/129

PR #126, opened 2022-04-19
Trigger callbacks with merged cache data when initializing with default key states 

neil-marcellini:
cc @marcaaron
### Details

Merge user provided default key value pairs with not only the data in storage, but also the data in the cache. See the issue for more context.
### Related Issues

https://github.com/Expensify/react-native-onyx/issues/125

### Linked PRs

https://github.com/Expensify/App/issues/8676

### DIFF
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -699,9 +699,9 @@ function initializeWithDefaultKeyStates() {
 
             const merged = lodashMerge(asObject, defaultKeyStates);
             cache.merge(merged);
-            _.each(merged, (_, key) => {
+            _.each(merged, (val, key) => {
                 const currentValue = cache.getValue(key);
-                keyChanged(key, currentValue)
+                keyChanged(key, currentValue);
             });
         });
 }
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -699,7 +699,10 @@ function initializeWithDefaultKeyStates() {
 
             const merged = lodashMerge(asObject, defaultKeyStates);
             cache.merge(merged);
-            _.each(merged, (val, key) => keyChanged(key, val));
+            _.each(merged, (_, key) => {
+                const currentValue = cache.getValue(key);
+                keyChanged(key, currentValue)
+            });
         });
 }
 


neil-marcellini:

lib/Onyx.js
```diff
@@ -699,7 +699,10 @@ function initializeWithDefaultKeyStates() {
 
             const merged = lodashMerge(asObject, defaultKeyStates);
             cache.merge(merged);
-            _.each(merged, (val, key) => keyChanged(key, val));
+            _.each(merged, (val, key) => {
+                const currentValue = cache.getValue(key);
+                keyChanged(key, currentValue);
+            });
```
marcaaron, line 705: **question:** It looks like `merged` is going to end up as the combined result of any existing values for these keys + the `defaultKeyStates` object? I might just be missing it, but why would it be necessary to ask the cache for this data if `merged` is the combined result already?

marcaaron (changes requested):

lib/Onyx.js
neil-marcellini: In the case that I'm trying to fix, `merged` contains only the `defaultKeyStates` as the `Storage` is empty. However, the `cache` is not empty. So if the `defaultKeyStates` are not merged with the value in the `cache` then `keyChanged` is called without up to date values.

marcaaron: >However, the cache is not empty.

**thought:** If we are signing out then both the cache and storage should be empty. Am I missing something?

marcaaron: https://github.com/Expensify/react-native-onyx/blob/2d42566779884bb445aa59dd48b5bcdff6c0a4e6/lib/Onyx.js#L711-L721

neil-marcellini: It is also surprising to me that the cache is not empty but the storage is. My guess is that `Storage.clear` takes a while to complete, and during that time the new user is logged in and lots of values are set in Onyx. 

When are values written from the cache to storage? Can the storage be out of date temporarily? I will try to look through the code and answer my own questions, but if you can give me a quick explanation / point me in the right direction that would be great.

neil-marcellini:
I'm going to convert this to a draft while I add a unit test for updating a key with the initialKeyStates after Onyx.clear().

lib/Onyx.js
marcaaron: >My guess is that Storage.clear takes a while to complete, and during that time the new user is logged in and lots of values are set in Onyx.

That seems possible. Here's where we clear on log out:

https://github.com/Expensify/App/blob/f6dc28f6a6c3d110a65e5caa33f94d441be3b630/src/libs/actions/SignInRedirect.js#L22-L39

But in the case where we sign someone out and sign in a new person we might not be waiting for that `Onyx.clear()` before logic elsewhere signs in the user? Probably would check to see if this code:

https://github.com/Expensify/App/blob/f6dc28f6a6c3d110a65e5caa33f94d441be3b630/src/pages/LogInWithShortLivedTokenPage.js#L57-L62

Runs before this promise resolves:

https://github.com/Expensify/App/blob/f6dc28f6a6c3d110a65e5caa33f94d441be3b630/src/libs/actions/SignInRedirect.js#L28

marcaaron: >When are values written from the cache to storage? 

I don't think values from the cache are even written to storage.

>Can the storage be out of date temporarily?

I'm not sure if I have enough context to explain a situation where the storage could be out of date. But it certainly seems possible.

neil-marcellini: I have verified that signing in the new user occurs before Onyx.clear() finishes. Waiting to sign the user in until the storage clears could work, and I guess we could do that, but wouldn't it be better to not wait for that? I also think that this solution makes more sense. Please see the issue for more context and an example if you haven't seen that already.

neil-marcellini:
I couldn't figure out how to add a test where I read a value when the cache is out of date.

lib/Onyx.js
marcaaron: >Waiting to sign the user in until the storage clears could work, and I guess we could do that, but wouldn't it be better to not wait for that?

**clarification:** By better do we mean faster? Or something else. 

**thought:** Signing out a user and signing in a new one seems like a rare event. So maybe performance is not a concern (assuming that is the concern).

>I also think that this solution makes more sense.

I do see how it solves the immediate issue. But just want to make sure we're picking the best solution. I'm not sure if allowing data to be set while we are in the process of re-initializing storage makes sense. I think my expectation is that we'd defer those updates until the app has finished one "clean" cycle.

Mind if we go back to this question?

**Should we completely sign the previous user out before starting a new sign in process?**

marcaaron:
>I couldn't figure out how to add a test where I read a value when the cache is out of date.

Can you describe the conditions that would lead the storage to go out of date?

lib/Onyx.js
neil-marcellini: > clarification: By better do we mean faster? Or something else.

Yes. I think you're right that it probably wouldn't slow things down too much and that this flow is rare.

> I'm not sure if allowing data to be set while we are in the process of re-initializing storage makes sense.

It would be up to the user of Onyx to make sure that no data is set while re-initializing storage right? That seems slightly less robust. If any initialKey is set while re-initializing storage then it could send an out of date value to its subscribers. How do you make your whole app wait while Onyx clears and re-initializes?

> Should we completely sign the previous user out before starting a new sign in process?

Probably, yes. I will work on a PR to do that. If that works and we agree that it's a better solution, then I'm happy to do that instead.

marcaaron: >It would be up to the user of Onyx to make sure that no data is set while re-initializing storage right?

Yep!

>That seems slightly less robust. If any initialKey is set while re-initializing storage then it could send an out of date value to its subscribers. How do you make your whole app wait while Onyx clears and re-initializes?

`Onyx.clear()` returns a promise so that would be my first instinct. But maybe there's a different solution. Might be something we can get some more opinions on too? I'm not 100% what the best answer would be, but just wanted to think about it for a second. Thanks for hearing me out!

neil-marcellini:
> Can you describe the conditions that would lead the storage to go out of date?

1. [Storage is cleared](https://github.com/Expensify/react-native-onyx/blob/d4e0d10fd711142cdbe407d95d1dd4d84d7bfcad/lib/Onyx.js#L719)
2. While storage is being cleared Onyx.set is called
3. [The value is set in the cache](https://github.com/Expensify/react-native-onyx/blob/d4e0d10fd711142cdbe407d95d1dd4d84d7bfcad/lib/Onyx.js#L548)
4. initializeWithDefaultKeyStates runs
5. [The object](https://github.com/Expensify/react-native-onyx/blob/d4e0d10fd711142cdbe407d95d1dd4d84d7bfcad/lib/Onyx.js#L698) created from Storage.multiGet(_.keys(defaultKeyStates)) is empty. The **storage is out of date**.
6. [keyChanged is called](https://github.com/Expensify/react-native-onyx/blob/d4e0d10fd711142cdbe407d95d1dd4d84d7bfcad/lib/Onyx.js#L702) with the value being only the defaultKeyStates

Does that make sense? Am I missing something?

neil-marcellini:
Closing in favor of https://github.com/Expensify/react-native-onyx/pull/129.

PR #129, opened 2022-05-03
Set default key states when clearing the cache

neil-marcellini:
cc @marcaaron
### Details

When we clear Onyx and reset the cache set every value to `null` OR the default key state if one exists for that key. Remove the call to `initializeWithDefaultKeyStates` since it is no longer needed. These changes prevent updating Onyx subscribers with out of date information by sending only the defaultKeyStates. See the issue for more info and an example.

### Related Issues

https://github.com/Expensify/react-native-onyx/issues/125

### Automated Tests

[onyxClearNativeStorageTest.js](https://github.com/Expensify/react-native-onyx/blob/dcb7f0139f8b5371affdd3bb2f9ce84e4687f087/tests/unit/onyxClearNativeStorageTest.js) and [onyxClearWebStorageTest.js](https://github.com/Expensify/react-native-onyx/blob/068a98a9afd45a53114754cc96c6928ec9fccbb3/tests/unit/onyxClearWebStorageTest.js)

### Manual Tests
I had a lot of trouble adding an automated tests earlier so I also added manual tests.
1. Checkout the App branch for **Case 1**
2. `npm i` to install [this Onyx version](https://github.com/Expensify/react-native-onyx/tree/neil-test-clear) that has been modified for testing

For each case below:
1. Checkout the App branch
2. `npm run web`
3. Sign in to any account
4. Open the console
5. Sign out
6. Make sure there are no assertion errors in the console
7. For cases 1 and 2 only:
8. `command + f` in the console and search 'Onyx test'
9. Make sure that the cache clears, then data is set or merged, and then Onyx finishes clearing.

- [ ] **Case 1:** [neil-test-onyx-clear-merge](https://github.com/Expensify/App/tree/neil-test-onyx-clear-merge)

```
Onyx.clear();
Onyx.merge('someDefaultKey', 'someValue');

// Expect the value of the merge to replace the default key state
// Expect that storage matches what we have in the cache and that our call to 
// `Storage.clear()` did not clear the value of merge from storage after we called Onyx.clear()
```

- [ ] **Case 2:** [neil-test-onyx-clear-set](https://github.com/Expensify/App/tree/neil-test-onyx-clear-set)
```
Onyx.clear();
Onyx.set('someDefaultKey', 'someValue');

// Expect the value of the set to replace the default key state
// Expect that storage matches what we have in the cache and that our call 
// to `Storage.clear()` did not prevent any values from being set
```

- [ ] **Case 3:** [neil-test-onyx-set-clear](https://github.com/Expensify/App/tree/neil-test-onyx-set-clear)
```
Onyx.set('someDefaultKey', 'someValue');
Onyx.clear();

// Expect the value from the `set()` to be replaced with the default key state value
```

- [ ] **Case 4:** [neil-test-onyx-merge-clear](https://github.com/Expensify/App/tree/neil-test-onyx-merge-clear)
```
Onyx.merge('someDefaultKey', 'someValue');
Onyx.clear();

// Expect the value from the `merge()` to be replaced with the default key state value
```

### Linked PRs

https://github.com/Expensify/App/pull/8855

### DIFF
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -707,11 +707,11 @@ function initializeWithDefaultKeyStates() {
 /**
  * Clear out all the data in the store
  *
- * Note that calling Onyx.clear and then Onyx.set on a key with a default
+ * Note that calling Onyx.clear() and then Onyx.set() on a key with a default
  * key state may store an unexpected value in Storage.
  *
- * Onyx.set might call Storage.setItem before Onyx.clear calls
- * Storage.setItem. Use Onyx.merge instead if possible.
+ * Onyx.set() might call Storage.setItem() before Onyx.clear() calls
+ * Storage.setItem(). Use Onyx.merge() instead if possible.
  *
  * @returns {Promise<void>}
  */
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -707,6 +707,12 @@ function initializeWithDefaultKeyStates() {
 /**
  * Clear out all the data in the store
  *
+ * Note that calling Onyx.clear and then Onyx.set on a key with a default
+ * key state may store an unexpected value in Storage.
+ *
+ * Onyx.set might call Storage.setItem before Onyx.clear calls
+ * Storage.setItem. Use Onyx.merge instead if possible.
+ *
  * @returns {Promise<void>}
  */
 function clear() {
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -718,6 +718,7 @@ function clear() {
                 cache.set(key, resetValue);
             });
         })
+        .then(() => console.log('[Onyx test] Done clearing the Onyx cache'))
         .then(Storage.clear);
 }
 
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -718,7 +718,6 @@ function clear() {
                 cache.set(key, resetValue);
             });
         })
-        .then(() => console.log('[Onyx test] Done clearing the Onyx cache'))
         .then(Storage.clear);
 }
 
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -735,8 +735,8 @@ function clear() {
                 // Optimistically inform subscribers on the next tick
                 Promise.resolve().then(() => keyChanged(key, resetValue));
             });
-        })
-        .then(Storage.clear);
+            Storage.clear();
+        });
 }
 
 /**
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -710,8 +710,18 @@ function initializeWithDefaultKeyStates() {
  * Note that calling Onyx.clear() and then Onyx.set() on a key with a default
  * key state may store an unexpected value in Storage.
  *
+ * E.g.
+ * Onyx.clear();
+ * Onyx.set(ONYXKEYS.DEFAULT_KEY, 'default');
+ * Storage.getItem(ONYXKEYS.DEFAULT_KEY)
+ *     .then((storedValue) => console.log(storedValue));
+ * null is logged instead of the expected 'default'
+ *
  * Onyx.set() might call Storage.setItem() before Onyx.clear() calls
- * Storage.setItem(). Use Onyx.merge() instead if possible.
+ * Storage.setItem(). Use Onyx.merge() instead if possible. Onyx.merge() calls
+ * Onyx.get(key) before calling Storage.setItem() via Onyx.set().
+ * Storage.setItem() from Onyx.clear() will have already finished and the merged
+ * value will be saved to storage after the default value.
  *
  * @returns {Promise<void>}
  */
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -714,8 +714,10 @@ function clear() {
         .then((keys) => {
             _.each(keys, (key) => {
                 const resetValue = lodashGet(defaultKeyStates, key, null);
-                keyChanged(key, resetValue);
                 cache.set(key, resetValue);
+
+                // Optimistically inform subscribers on the next tick
+                Promise.resolve().then(() => keyChanged(key, resetValue));
             });
         })
         .then(Storage.clear);
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -1,6 +1,7 @@
 import _ from 'underscore';
 import Str from 'expensify-common/lib/str';
 import lodashMerge from 'lodash/merge';
+import lodashGet from 'lodash/get';
 import Storage from './storage';
 
 import {registerLogger, logInfo, logAlert} from './Logger';
@@ -712,12 +713,12 @@ function clear() {
     return getAllKeys()
         .then((keys) => {
             _.each(keys, (key) => {
-                keyChanged(key, null);
-                cache.set(key, null);
+                const resetValue = lodashGet(defaultKeyStates, key, null);
+                keyChanged(key, resetValue);
+                cache.set(key, resetValue);
             });
         })
-        .then(Storage.clear)
-        .then(initializeWithDefaultKeyStates);
+        .then(Storage.clear);
 }
 
 /**
--- a/tests/unit/onyxClearNativeStorageTest.js
+++ b/tests/unit/onyxClearNativeStorageTest.js
@@ -8,7 +8,7 @@ const ONYX_KEYS = {
 
 jest.useFakeTimers();
 
-const storageCallResolveList = [];
+let storageCallResolveList = [];
 function addStorageCallResolve(name) {
     storageCallResolveList.push(name);
 }
@@ -17,7 +17,7 @@ function storageCallResolveOrder(methodName) {
     return storageCallResolveList.indexOf(methodName) + 1;
 }
 
-const storageCallQueue = [];
+let storageCallQueue = [];
 
 // Mock clear to wait for promises and add a delay
 Storage.clear = jest.fn(() => Promise.all(storageCallQueue)
@@ -62,6 +62,8 @@ describe('Set data while storage is clearing', () => {
     });
 
     afterEach(() => {
+        storageCallResolveList = [];
+        storageCallQueue = [];
         Onyx.disconnect(connectionID);
         Onyx.clear();
         jest.runAllTimers();
--- a/tests/unit/onyxClearNativeStorageTest.js
+++ b/tests/unit/onyxClearNativeStorageTest.js
@@ -1,4 +1,4 @@
-import AsyncStorageMock from '@react-native-async-storage/async-storage';
+import AsyncStorageMock from '../../__mocks__/@react-native-async-storage/async-storage';
 import Storage from '../../lib/storage';
 import waitForPromisesToResolve from '../utils/waitForPromisesToResolve';
 
--- a/tests/unit/onyxClearNativeStorageTest.js
+++ b/tests/unit/onyxClearNativeStorageTest.js
@@ -127,4 +127,31 @@ describe('Set data while storage is clearing', () => {
                 expect(storedValue).resolves.not.toBe(setValue);
             });
     });
+
+    it('should replace the value of Onyx.set with the default key state in the cache', () => {
+        let defaultValue;
+        connectionID = Onyx.connect({
+            key: ONYX_KEYS.DEFAULT_KEY,
+            initWithStoredValues: false,
+            callback: val => defaultValue = val,
+        });
+        const setValue = 'set';
+        expect.assertions(6);
+        Onyx.set(ONYX_KEYS.DEFAULT_KEY, setValue);
+        Storage.clear = jest.fn(() => AsyncStorageMock.clear());
+        Onyx.clear();
+        return waitForPromisesToResolve()
+            .then(() => {
+                expect(storageCallResolveOrder('setItem')).toBe(1);
+                expect(storageCallResolveOrder('clear')).toBe(2);
+                expect(defaultValue).toBe('default');
+                const cachedValue = cache.getValue(ONYX_KEYS.DEFAULT_KEY);
+                expect(cachedValue).toBe('default');
+                const storedValue = Storage.getItem(ONYX_KEYS.DEFAULT_KEY);
+
+                // The default key state is never stored during Onyx.clear
+                expect(storedValue).resolves.not.toBe('default');
+                expect(storedValue).resolves.toBeNull();
+            });
+    });
 });
--- a/tests/unit/onyxClearNativeStorageTest.js
+++ b/tests/unit/onyxClearNativeStorageTest.js
@@ -10,41 +10,6 @@ const SET_VALUE = 'set';
 const MERGED_VALUE = 'merged';
 const DEFAULT_VALUE = 'default';
 
-let storageCallResolveList = [];
-
-/**
- * Get the order in which the storage method call resolved so we can assert that
- * updates to storage are made in the expected order.
- * @param {String} methodName The name of the storage method
- * @returns {Number}
- */
-function getStorageCallResolveOrder(methodName) {
-    return storageCallResolveList.indexOf(methodName) + 1;
-}
-
-// Add all storage calls to a queue and make them wait for each other to match the real implementation.
-let storageCallQueue = [];
-
-// Track when AsyncStorageMock.clear calls resolve.
-const asyncStorageMockClear = AsyncStorageMock.clear;
-AsyncStorageMock.clear = jest.fn(() => Promise.all(storageCallQueue)
-    .then(() => {
-        const clearPromise = asyncStorageMockClear()
-            .then(storageCallResolveList.push('clear'));
-        storageCallQueue.push(clearPromise);
-        return clearPromise;
-    }));
-
-// Track when AsyncStorageMock.setItem calls resolve.
-const asyncStorageMockSetItem = AsyncStorageMock.setItem;
-AsyncStorageMock.setItem = jest.fn((key, value) => Promise.all(storageCallQueue)
-    .then(() => {
-        const setItemPromise = asyncStorageMockSetItem(key, value)
-            .then(storageCallResolveList.push('setItem'));
-        storageCallQueue.push(setItemPromise);
-        return setItemPromise;
-    }));
-
 describe('Set data while storage is clearing', () => {
     let connectionID;
     let onyxValue;
@@ -74,15 +39,11 @@ describe('Set data while storage is clearing', () => {
     afterEach(() => {
         Onyx.disconnect(connectionID);
         Storage.clear.mockReset();
-        return Onyx.clear()
-            .then(() => {
-                storageCallResolveList = [];
-                storageCallQueue = [];
-            });
+        return Onyx.clear();
     });
 
     it('should persist the value of Onyx.merge when called between the cache and storage clearing', () => {
-        expect.assertions(5);
+        expect.assertions(3);
 
         // GIVEN that Onyx is completely clear
         // WHEN Onyx.clear() is called
@@ -95,10 +56,6 @@ describe('Set data while storage is clearing', () => {
         Onyx.clear();
         return waitForPromisesToResolve()
             .then(() => {
-                // THEN the storage finishes clearing before Storage.setItem finishes for the merged value
-                expect(getStorageCallResolveOrder('clear')).toBe(1);
-                expect(getStorageCallResolveOrder('setItem')).toBe(2);
-
                 // THEN the value in Onyx, the cache, and the storage is the merged value
                 expect(onyxValue).toBe(MERGED_VALUE);
                 const cachedValue = cache.getValue(ONYX_KEYS.DEFAULT_KEY);
@@ -109,7 +66,7 @@ describe('Set data while storage is clearing', () => {
     });
 
     it('should replace the value of Onyx.set with the default key state in the cache', () => {
-        expect.assertions(5);
+        expect.assertions(3);
 
         // GIVEN that Onyx is completely clear
         // WHEN set then clear is called on a key with a default key state
@@ -118,10 +75,6 @@ describe('Set data while storage is clearing', () => {
         Onyx.clear();
         return waitForPromisesToResolve()
             .then(() => {
-                // THEN the Storage.setItem finishes for the set value before the storage finishes clearing
-                expect(getStorageCallResolveOrder('setItem')).toBe(1);
-                expect(getStorageCallResolveOrder('clear')).toBe(2);
-
                 // THEN the value in Onyx and the cache is the default key state
                 expect(onyxValue).toBe(DEFAULT_VALUE);
                 const cachedValue = cache.getValue(ONYX_KEYS.DEFAULT_KEY);
@@ -135,7 +88,7 @@ describe('Set data while storage is clearing', () => {
     });
 
     it('should replace the value of Onyx.merge with the default key state in the cache', () => {
-        expect.assertions(5);
+        expect.assertions(3);
 
         // GIVEN that Onyx is completely clear
         // WHEN merge then clear is called on a key with a default key state
@@ -144,10 +97,6 @@ describe('Set data while storage is clearing', () => {
         Onyx.clear();
         return waitForPromisesToResolve()
             .then(() => {
-                // THEN the Storage.setItem finishes for the merged value before the storage finishes clearing
-                expect(getStorageCallResolveOrder('setItem')).toBe(1);
-                expect(getStorageCallResolveOrder('clear')).toBe(2);
-
                 // THEN the value in Onyx and the cache is the default key state
                 expect(onyxValue).toBe(DEFAULT_VALUE);
                 const cachedValue = cache.getValue(ONYX_KEYS.DEFAULT_KEY);
--- a/tests/unit/onyxClearNativeStorageTest.js
+++ b/tests/unit/onyxClearNativeStorageTest.js
@@ -5,6 +5,9 @@ import waitForPromisesToResolve from '../utils/waitForPromisesToResolve';
 const ONYX_KEYS = {
     DEFAULT_KEY: 'defaultKey',
 };
+const SET_VALUE = 'set';
+const MERGED_VALUE = 'merged';
+const DEFAULT_VALUE = 'default';
 
 let storageCallResolveList = [];
 function addStorageCallResolve(name) {
@@ -50,7 +53,7 @@ describe('Set data while storage is clearing', () => {
             keys: ONYX_KEYS,
             registerStorageEventListener: () => {},
             initialKeyStates: {
-                [ONYX_KEYS.DEFAULT_KEY]: 'default',
+                [ONYX_KEYS.DEFAULT_KEY]: DEFAULT_VALUE,
             },
         });
     });
@@ -77,11 +80,10 @@ describe('Set data while storage is clearing', () => {
             initWithStoredValues: false,
             callback: val => defaultValue = val,
         });
-        const mergedValue = 'merged';
         expect.assertions(5);
         Storage.clear = jest.fn(() => {
             // Call merge between the cache and storage clearing
-            Onyx.merge(ONYX_KEYS.DEFAULT_KEY, mergedValue);
+            Onyx.merge(ONYX_KEYS.DEFAULT_KEY, MERGED_VALUE);
             AsyncStorageMock.clear();
             return waitForPromisesToResolve();
         });
@@ -90,11 +92,11 @@ describe('Set data while storage is clearing', () => {
             .then(() => {
                 expect(storageCallResolveOrder('clear')).toBe(1);
                 expect(storageCallResolveOrder('setItem')).toBe(2);
-                expect(defaultValue).toBe(mergedValue);
+                expect(defaultValue).toBe(MERGED_VALUE);
                 const cachedValue = cache.getValue(ONYX_KEYS.DEFAULT_KEY);
-                expect(cachedValue).toBe(mergedValue);
+                expect(cachedValue).toBe(MERGED_VALUE);
                 const storedValue = Storage.getItem(ONYX_KEYS.DEFAULT_KEY);
-                expect(storedValue).resolves.toBe(mergedValue);
+                expect(storedValue).resolves.toBe(MERGED_VALUE);
             });
     });
 
@@ -105,11 +107,10 @@ describe('Set data while storage is clearing', () => {
             initWithStoredValues: false,
             callback: val => defaultValue = val,
         });
-        const setValue = 'set';
         expect.assertions(5);
         Storage.clear = jest.fn(() => {
             // Call set between the cache and storage clearing
-            Onyx.set(ONYX_KEYS.DEFAULT_KEY, setValue);
+            Onyx.set(ONYX_KEYS.DEFAULT_KEY, SET_VALUE);
             AsyncStorageMock.clear();
             return waitForPromisesToResolve();
         });
@@ -120,11 +121,11 @@ describe('Set data while storage is clearing', () => {
                 // AsyncStorage.setItem resolves before AsyncStorage.clear
                 expect(storageCallResolveOrder('setItem')).toBe(1);
                 expect(storageCallResolveOrder('clear')).toBe(2);
-                expect(defaultValue).toBe(setValue);
+                expect(defaultValue).toBe(SET_VALUE);
                 const cachedValue = cache.getValue(ONYX_KEYS.DEFAULT_KEY);
-                expect(cachedValue).toBe(setValue);
+                expect(cachedValue).toBe(SET_VALUE);
                 const storedValue = Storage.getItem(ONYX_KEYS.DEFAULT_KEY);
-                expect(storedValue).resolves.not.toBe(setValue);
+                expect(storedValue).resolves.not.toBe(SET_VALUE);
             });
     });
 
@@ -135,22 +136,21 @@ describe('Set data while storage is clearing', () => {
             initWithStoredValues: false,
             callback: val => defaultValue = val,
         });
-        const setValue = 'set';
         expect.assertions(6);
-        Onyx.set(ONYX_KEYS.DEFAULT_KEY, setValue);
+        Onyx.set(ONYX_KEYS.DEFAULT_KEY, SET_VALUE);
         Storage.clear = jest.fn(() => AsyncStorageMock.clear());
         Onyx.clear();
         return waitForPromisesToResolve()
             .then(() => {
                 expect(storageCallResolveOrder('setItem')).toBe(1);
                 expect(storageCallResolveOrder('clear')).toBe(2);
-                expect(defaultValue).toBe('default');
+                expect(defaultValue).toBe(DEFAULT_VALUE);
                 const cachedValue = cache.getValue(ONYX_KEYS.DEFAULT_KEY);
-                expect(cachedValue).toBe('default');
+                expect(cachedValue).toBe(DEFAULT_VALUE);
                 const storedValue = Storage.getItem(ONYX_KEYS.DEFAULT_KEY);
 
                 // The default key state is never stored during Onyx.clear
-                expect(storedValue).resolves.not.toBe('default');
+                expect(storedValue).resolves.not.toBe(DEFAULT_VALUE);
                 expect(storedValue).resolves.toBeNull();
             });
     });
--- a/tests/unit/onyxClearNativeStorageTest.js
+++ b/tests/unit/onyxClearNativeStorageTest.js
@@ -50,6 +50,12 @@ describe('Set data while storage is clearing', () => {
 
     beforeAll(() => {
         Onyx = require('../../index').default;
+    });
+
+    // Always use a "fresh" cache instance
+    beforeEach(() => {
+        onyxValue = null;
+        cache = require('../../lib/OnyxCache').default;
         Onyx.init({
             keys: ONYX_KEYS,
             registerStorageEventListener: () => {},
@@ -57,12 +63,6 @@ describe('Set data while storage is clearing', () => {
                 [ONYX_KEYS.DEFAULT_KEY]: DEFAULT_VALUE,
             },
         });
-    });
-
-    // Always use a "fresh" cache instance
-    beforeEach(() => {
-        onyxValue = null;
-        cache = require('../../lib/OnyxCache').default;
         connectionID = Onyx.connect({
             key: ONYX_KEYS.DEFAULT_KEY,
             initWithStoredValues: false,
--- a/tests/unit/onyxClearNativeStorageTest.js
+++ b/tests/unit/onyxClearNativeStorageTest.js
@@ -82,8 +82,10 @@ describe('Set data while storage is clearing', () => {
 
     it('should persist the value of Onyx.merge when called between the cache and storage clearing', () => {
         expect.assertions(5);
+
+        // WHEN Onyx.clear() is called
         Storage.clear = jest.fn(() => {
-            // Call merge between the cache and storage clearing
+            // WHEN merge is called between the cache and storage clearing, on a key with a default key state
             Onyx.merge(ONYX_KEYS.DEFAULT_KEY, MERGED_VALUE);
             AsyncStorageMock.clear();
             return waitForPromisesToResolve();
@@ -91,8 +93,11 @@ describe('Set data while storage is clearing', () => {
         Onyx.clear();
         return waitForPromisesToResolve()
             .then(() => {
+                // THEN the storage finishes clearing before Storage.setItem finishes for the merged value
                 expect(getStorageCallResolveOrder('clear')).toBe(1);
                 expect(getStorageCallResolveOrder('setItem')).toBe(2);
+
+                // THEN the value in Onyx, the cache, and the storage are the merged value
                 expect(onyxValue).toBe(MERGED_VALUE);
                 const cachedValue = cache.getValue(ONYX_KEYS.DEFAULT_KEY);
                 expect(cachedValue).toBe(MERGED_VALUE);
--- a/tests/unit/onyxClearNativeStorageTest.js
+++ b/tests/unit/onyxClearNativeStorageTest.js
@@ -146,32 +146,15 @@ describe('Set data while storage is clearing', () => {
     });
 
     it('should replace the value of Onyx.merge with the default key state in the cache', () => {
-        let onyxValue;
-        connectionID = Onyx.connect({
-            key: ONYX_KEYS.DEFAULT_KEY,
-            initWithStoredValues: false,
-            callback: val => onyxValue = val,
-        });
         expect.assertions(6);
         Onyx.merge(ONYX_KEYS.DEFAULT_KEY, MERGED_VALUE);
         Storage.clear = jest.fn(() => AsyncStorageMock.clear());
         Onyx.clear();
         return waitForPromisesToResolve()
             .then(() => {
-                // Onnyx.merge calls Onyx.set which sets "merged" in the cache.
-                // The within Onyx.clear as the cache clears keyChanged is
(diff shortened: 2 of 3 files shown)


lib/Onyx.js
```diff
@@ -712,12 +713,12 @@ function clear() {
...
-                keyChanged(key, null);
-                cache.set(key, null);
+                const resetValue = lodashGet(defaultKeyStates, key, null);
+                keyChanged(key, resetValue);
+                cache.set(key, resetValue);
             });
         })
-        .then(Storage.clear)
-        .then(initializeWithDefaultKeyStates);
+        .then(Storage.clear);
```
marcaaron, line 721: Looks good so far. I still have questions about what the effects of both `getAllKeys()` and `Storage.clear` would be and would like to see automated tests for the following cases:

**Case 1:**
```js
Onyx.clear();
Onyx.merge('someDefaultKey', 'someValue');

// Expect the value of the merge to replace the default key state
// Expect that storage matches what we have in the cache and that our call to 
// `Storage.clear()` did not clear the value of merge from storage after we called Onyx.clear()
```

**Case 2:**
```js
Onyx.clear();
Onyx.set('someDefaultKey', 'someValue');

// Expect the value of the set to replace the default key state
// Expect that storage matches what we have in the cache and that our call 
// to `Storage.clear()` did not prevent any values from being set
```

**Case 3:**
```js
Onyx.set('someDefaultKey', 'someValue');
Onyx.clear();

// Expect the value from the `set()` to be replaced with the default key state value
// Expect that storage matches what we have in the cache exactly
```

**Case 4:**
```js
Onyx.merge('someDefaultKey', 'someValue');
Onyx.clear();

// Expect the value from the `merge()` to be replaced with the default key state value
// Expect that storage matches what we have in the cache exactly
```

neil-marcellini: These are great thanks, I will create some tests.

neil-marcellini:
How can I get Onyx.merge or Onyx.set to run between clearing the cache and clearing the storage? I tried inserting a lot of data into storage before clearing, but that didn't work. I need some kind of callback in between, or I need to make Storage.clear take a really long time.

Right now I have this test for Case 1:
```javascript
it('should store merged values when calling merge on a default key after clear', () => {
        let defaultValue;
        connectionID = Onyx.connect({
            key: ONYX_KEYS.DEFAULT_KEY,
            initWithStoredValues: false,
            callback: val => defaultValue = val,
        });
        console.log('Calling clear in test');
        const afterClear = Onyx.clear();
        console.log('Calling merge');
        const afterMerge = Onyx.merge(ONYX_KEYS.DEFAULT_KEY, 'merged');
        return Promise.all([afterClear, afterMerge])
            .then(() => {
                console.log('defaultValue = ', defaultValue);
                expect(defaultValue).toEqual('merged');
                return Storage.getItem(ONYX_KEYS.DEFAULT_KEY);
            })
            .then((storedValue) => {
                const cachedValue = cache.getValue(ONYX_KEYS.DEFAULT_KEY);
                console.log('cachedValue = ', cachedValue);
                console.log('storedValue = ', storedValue);
                expect(cachedValue).toEqual(storedValue);
            });
    });
```

I'm also console logging when Onyx starts clearing Storage and when it stops:
```
console.log
      Calling clear in test

      at Object.<anonymous> (tests/unit/onyxTest.js:329:17)

    console.log
      [Onyx.clear] clearing storage

      at Object.clear (lib/Onyx.js:721:23)

    console.log
      [Onyx.clear] finished clearing storage

      at Object.clear (lib/Onyx.js:723:23)

    console.log
      Calling merge

      at Object.<anonymous> (tests/unit/onyxTest.js:331:17)

    console.log
      defaultValue =  merged

      at fn (tests/unit/onyxTest.js:335:25)

    console.log
      cachedValue =  undefined

      at fn (tests/unit/onyxTest.js:341:25)
```

marcaaron:
>How can I get Onyx.merge or Onyx.set to run between clearing the cache and clearing the storage? I tried inserting a lot of data into storage before clearing, but that didn't work. I need some kind of callback in between, or I need to make Storage.clear take a really long time.

I'm not sure if I understand the question exactly, but `merge()` or `set()` will start working as soon as it can. Which would be when any synchronous logic in `Onyx.clear()` finishes - which looks like `getAllKeys()`. I'd expect something like this:

1. `clear()` -> getAllKeys() - which is a Promise
2. `merge()` -> push value into merge queue or possibly call get() - which is a Promise
3. What happens next depends on when those promises resolve...

We can't call really call `Onyx.merge()` _while_ `Storage.clear()` is running if we are using `Onyx.clear()`. We can only wait for that entire `Onyx.clear()` promise to resolve.
 
One idea - though I'm not sure if it will work - is to add the ability to make the `Storage.clear` mock delayed. Then we can use [fake timers](https://jestjs.io/docs/timer-mocks) to have more control over how long different parts of a larger promise are taking.

neil-marcellini:
> One idea - though I'm not sure if it will work - is to add the ability to make the `Storage.clear` mock delayed. Then we can use [fake timers](https://jestjs.io/docs/timer-mocks) to have more control over how long different parts of a larger promise are taking.

I can understand how we could create a version of Storage.clear that is delayed, but I don't get how we would make Onyx.clear() call that version.

I thought about adding an optional callback param to Onyx.clear(). By doing this I can get merge or set to run in between. The problem with this approach of course is that the code is only used for the tests. I also tried injecting the code into Onyx.clear during the test and then calling eval, based on [this SO answer](https://stackoverflow.com/a/2109007/17686499), but I couldn't get it to work and it's kind of a messy solution.
```javascript
function clear(testingCallback = null) {
    return getAllKeys()
        .then((keys) => {
            _.each(keys, (key) => {
                const resetValue = lodashGet(defaultKeyStates, key, null);
                keyChanged(key, resetValue);
                cache.set(key, resetValue);
            });
        })
        .then(() => new Promise((resolve) => {
            if (testingCallback) {
                resolve(testingCallback());
            } else {
                resolve();
            }
        }))
        .then(Storage.clear)
}
```

marcaaron:
>I can understand how we could create a version of Storage.clear that is delayed, but I don't get how we would make Onyx.clear() call that version.

With jest you can mock whatever you want which will force `Onyx.clear()` to use that instead of what it would have used.

Actually, there are some tests using a mock already since node doesn't have "storage"...

https://github.com/Expensify/react-native-onyx/blob/f6c2b03f29193bbceca68a9997fa94cc998f52a6/__mocks__/localforage.js#L29-L34

Maybe just create something like `Storage.setDelayForClear()` and call it from your test?

https://jestjs.io/docs/manual-mocks

marcaaron:
Here's an example where we took the mock and added our own test functionality to it by setting some "initial data" 

https://github.com/Expensify/react-native-onyx/blob/2d42566779884bb445aa59dd48b5bcdff6c0a4e6/tests/unit/onyxMultiMergeWebStorageTest.js#L3-L18

neil-marcellini:
Awesome, thanks so much for steering me in the right direction here. Jest is new to me and I struggled with it quite a bit yesterday, but hopefully I can get it working today.

neil-marcellini:
I went down a bit of a rabbit hole again unfortunately, but I learned a few things in the process.

When calling merge or set after clear the value will get set in the cache right away. However, `Storage.setItem` from set won't get called until `Storage.clear` finishes. I refactored my test for Case 1 to address that. I'll finish up the other tests later.

If we merge and then clear storage the defaultKeyState will not be set in the storage. Default key states are not set in storage after clearing on this branch or on `main`. I don't think this is a problem because if we close the tab and kill the cache, the default key states will be re-set in the cache when Onyx initializes.

neil-marcellini:
Ready for review now. For cases 3 and 4, I did not include an assertion that the stored value matches the cached value after setting data and then clearing Onyx.

> If we merge and then clear storage the defaultKeyState will not be set in the storage. Default key states are not set in storage after clearing on this branch or on `main`. I don't think this is a problem because if we close the tab and kill the cache, the default key states will be re-set in the cache when Onyx initializes.

marcaaron:
Sorry, @neil-marcellini I'm going OOO for a bit. 

List of all contributors here might be a good bet for a replacement reviewer: 

https://github.com/Expensify/react-native-onyx/graphs/contributors

@kidroca and @tgolen may be able to help as well

tgolen:
Where are the unit tests being added?

tgolen (dismissed):
These changes look fine.

kidroca:
> I had a lot of trouble adding an automated test where I could mock the Storage.clear method because [keepInstancesSync overrides the clear method](https://github.com/Expensify/react-native-onyx/blob/ce5bd2c203fc054d53841f049e8276fbecaf7c0a/lib/storage/WebStorage.js#L34) when Onyx initializes. I added manual tests instead.

Depending on how you want to mock Storage, you can either init Onyx with `shouldSyncMultipleInstances: false` and then monitor `localForage` calls. Initializing Onyx with `shouldSyncMultipleInstances: false` would skip applying `keepInstancesSync`

https://github.com/Expensify/react-native-onyx/blob/ce5bd2c203fc054d53841f049e8276fbecaf7c0a/lib/Onyx.js#L844-L848

Or you can override the storage mock like so: 

##### unit-test.js
```js
import Storage from '../../lib/storage';

// makes `Storage` imported above to be the one from `storage/__mocks__`
jest.mock('../../lib/storage');

// we can now override methods like remove the `keepInstancesSync` and/or change the `clear` method
delete Storage.keepInstancesSync;
Storage.clear = jest.fn(() => Promise.resolve());

// The important bit is to import/init Onyx only after `Storage` had been overridden
const Onyx = require('../../index').default;

describe('My tests', () => {
  /* ... */
});
```

I can help writing unit tests if you'd like

neil-marcellini:
Based on [this slack thread](https://expensify.slack.com/archives/C03TQ48KC/p1651691723596049?thread_ts=1651686942.533839&cid=C03TQ48KC) Marc suggested that we should not set `shouldSyncMultipleInstances: false` since we wouldn't be testing the actual Storage.clear behavior.

However, you made me realize that I could mock `Storage.keepInstancesInSync` and add a delay to the clear method when I mock it. Also, how could I create a storage mock for AsyncStorage rather than WebStorage and tell Onyx to use that mock for Storage?

Since I have a lot of fixes blocked on this change and the manual tests work well, may I go ahead and merge it now? Or should we add the unit tests first? @tgolen @kidroca

tgolen:
I personally feel like any change to Onyx (especially fixing or changing behavior) should always be done from a unit-test first approach. So, unless there is any reason to move forward without having unit tests (like it's fixing a fire or something), then I think we should wait until the unit tests are done.

Admittedly, I don't know the context of all the fixes blocked on this change though.

kidroca:
Since in unit tests it's not possible to use a real storage provider, trying to mimic specific storage provider behavior (like `keepInstancesInSync`) shouldn't be necessary - that's internal to the storage provider and can be tested separately 

The functionality you're changing depends solely on `Storage.clear` - e.g. it assumes that `Storage.clear` works no matter the underlying storage provider, so we should just provide a mock that stubs `Storage.clear` in all desired scenarios we want to test

---

To assert that `keepInstancesInSync` we should write a test for the `WebStorage` provider where we verify that it `clear` as expected
- notifies other tabs
- returns a promise
- etc...

neil-marcellini:
I uploaded a test with console logs to make sure that it was working. Is that a good test? If so I'll follow the same pattern to add cases 2-4. I'm going to work on getting a web test to work in the meantime.

neil-marcellini:
@chiragsalian could you please give a review when you have a chance?

kidroca:
> I uploaded a test with console logs to make sure that it was working. Is that a good test? If so I'll follow the same pattern to add cases 2-4. I'm going to work on getting a web test to work in the meantime.

console logs might be handy to logs what's happening but as you can see they are considered lint errors 

I'm not sure whether we need to capture a Promise for each onyx method used, or `setTimeout(..., 0)` to run on the next tick - instead of `Promise.all([promise1, promise2])` we use `waitForPromisesToResolve()` and `then` we make the assertions 

Something like this:

```js
Onyx.clear()
Onyx.merge(...)

  .then(() => {
    expect(defaultValue).toEqual('merged');
    const cachedValue = cache.getValue(ONYX_KEYS.DEFAULT_KEY);
    expect(cachedValue).toEqual('merged');
  })
```

You can use `setTimeout` if you want to add some delay between calls but it would probably require jest's fake timers 

I think what we're trying to test is 
- `Onyx.clear` iterates a list of keys 
- `Onyx.merge` is called shortly after clear (e.g. in parallel) to change something
- maybe `merge` can change some key before `clear` has iterated over it, then `clear` would overwrite it

I think IndexedDB (web) and AsyncStorage (native) has mechanisms to queue transactions so `merge` would happen after `clear`, but our cache has no such logic

neil-marcellini:
Thanks for the help again @kidroca!

> console logs might be handy to logs what's happening but as you can see they are considered lint errors

I will take them out in my next commits. They've been handy for me and I thought they might help for review.
> 
> I'm not sure whether we need to capture a Promise for each onyx method used, or `setTimeout(..., 0)` to run on the next tick - instead of `Promise.all([promise1, promise2])` we use `waitForPromisesToResolve()` and `then` we make the assertions
> 
> Something like this:
> 
> ```js
> Onyx.clear()
> Onyx.merge(...)
> 
> return waitForPromisesToResolve()
>   .then(() => {
>     expect(defaultValue).toEqual('merged');
>     const cachedValue = cache.getValue(ONYX_KEYS.DEFAULT_KEY);
>     expect(cachedValue).toEqual('merged');
>   })
> ```
> 

I've seen `waitForPromisesToResolve()` used all over the place and I see that it wraps `setImmediate` in a promise. How does that work? Why does it only resolve when all previous promises resolve? I'll use that instead.

I'm using `setTimeout` is so that merge runs **after** the Onyx cache clears but **before** the storage clears. If I don't add that then it runs before the cache clears and that is not the case I want to test.

> _We also need to assert the value written in storage, it seems like Onyx connections are notified correctly, but it doesn't verify that the correct value is captured in storage Since `AsyncStorageMock.clear()` is set to run with a `.5sec` delay it would seem that connections might get one value, but storage would be cleared from it_
> 
> You can use `setTimeout` if you want to add some delay between calls but it would probably require jest's fake timers
> 
> I think what we're trying to test is
> 
> * `Onyx.clear` iterates a list of keys
> * `Onyx.merge` is called shortly after clear (e.g. in parallel) to change something
> * maybe `merge` can change some key before `clear` has iterated over it, then `clear` would overwrite it
> 
> I think IndexedDB (web) and AsyncStorage (native) has mechanisms to queue transactions so `merge` would happen after `clear`, but our cache has no such logic

Yeah that's pretty much it, with the added condition that merge is called after the cache clears but before the storage finishes clearing. I guess I would need to figure out how to mock the storage so that  merge happens after the storage finishes clearing? However, as I outlined in [this comment](https://github.com/Expensify/react-native-onyx/pull/129#issuecomment-1119933719), I think it doesn't matter if the storage matches the cache since this is only for default key states which are reset on Onyx init.

I will work on refining this more tomorrow.

neil-marcellini:
I have implemented the native test for case 1 such that:
1. I'm using waitForPromisesToResolve
2. I have mocked the AsyncStorage so that Storage.setItem and Storage.clear will wait for each other. I'm also asserting that first the storage clears and then set item is called.
3. I've changed the structure so that I know for a fact that merge is called in between the cache and storage clear

tests/unit/onyxClearNativeStorageTest.js
```diff
@@ -0,0 +1,184 @@
...
+                // Onnyx.merge calls Onyx.set which sets "merged" in the cache.
+                // The within Onyx.clear as the cache clears keyChanged is
+                // called with the resetValue of 'defaut' so the onyxValue is
+                // 'default' and same for the cache. Then keyChanged is called
+                // from Onyx.set and the onyxValue is 'merged'. At that point
+                // the cache is done clearing. Storage.setItem finishes setting
+                // the storage to 'merged'. Then clear finishes, setting it to
+                // null.
+                // The end result is that the cachedValue is 'default',
+                // the onyxValue is 'merged', and the storage is null.
```
neil-marcellini, line 170: I have made this test pass for now, but it's not the desired behavior. Tomorrow I will have to think about how to fix it.

neil-marcellini:
I learned that I had mocked AsyncStorage.setItem incorrectly and so I was never actually setting data in the storage mock. I was also not waiting for assertions on the stored values. I fixed both of those errors, and now all tests are passing expect for the test with set and then clear.

neil-marcellini:
@kidroca @tgolen @marcaaron Bump on my last comment. I'm not sure how to proceed.

neil-marcellini:
Marc recommended that I go ahead and make my App PR wait for Onyx while we figure out what to do here.

tgolen:
Can you point to where the code is in your PR that is failing? I'm a little lost on the different cases, so it will help me to see the code in context.

My initial thought is that we shouldn't be testing Onyx in a way that is inconsistent with how it is used in production. This leads me to believe that we would never call `set()` and `clear()` at the same time (or immediately after each other).

neil-marcellini:
> Can you point to where the code is in your PR that is failing?

@tgolen this is the line where the test is failing. https://github.com/Expensify/react-native-onyx/blob/ebeb20a104c7ae44ab10c0769d1fa6262f93f81c/tests/unit/onyxClearNativeStorageTest.js#L124

> My initial thought is that we shouldn't be testing Onyx in a way that is inconsistent with how it is used in production.

I agree.

> This leads me to believe that we would never call `set()` and `clear()` at the same time (or immediately after each other).
Ya, I don't think we do that in the app at all. We do call `clear()` and then quickly 'merge()' during the transition flow. We [signOutAndRedirectToSignIn](https://github.com/Expensify/App/blob/2996c88959a216fd7a33468e7dccbf3731947ec1/src/pages/LogInWithShortLivedTokenPage.js#L101) and then [signInWithShortLivedToken](https://github.com/Expensify/App/blob/2996c88959a216fd7a33468e7dccbf3731947ec1/src/pages/LogInWithShortLivedTokenPage.js#L60) which calls [merge](https://github.com/Expensify/App/blob/51fb61808d84f4aae4c6cbb3f6875d43d355fbb5/src/libs/actions/Session/index.js#L282) on the session before the storage finishes clearing, which is how I noticed this bug.

Anyways, I don't the failing test is a problem here, so I want to make sure that you all agree.

marcaaron (commented):
Left a quick question for now.

lib/Onyx.js
```diff
@@ -712,12 +713,14 @@ function clear() {
     return getAllKeys()
         .then((keys) => {
             _.each(keys, (key) => {
-                keyChanged(key, null);
-                cache.set(key, null);
+                const resetValue = lodashGet(defaultKeyStates, key, null);
+                cache.set(key, resetValue);
+
+                // Optimistically inform subscribers on the next tick
```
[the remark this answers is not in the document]
tgolen: +1

tests/unit/onyxClearNativeStorageTest.js
```diff
@@ -0,0 +1,165 @@
...
+describe('Set data while storage is clearing', () => {
+    let connectionID;
+    let Onyx;
+    let onyxValue;
+
+    /** @type OnyxCache */
+    let cache;
+
+    beforeAll(() => {
+        Onyx = require('../../index').default;
```
tgolen, line 52: Why not just do this at the top of the file?

tgolen (dismissed):
Some of the tests are not really clear to me what the flow is that they are testing. In order to improve this, please use the following comment format. You can find this in some of the other tests, and we should always write test comments in this way:

```
// GIVEN <explain the initial state of the system>
...code that sets the initial state

// WHEN <something is done>
...code that does something

// THEN <what is expected to be the final state of the system>
...assertions
```

tests/unit/onyxClearNativeStorageTest.js
```diff
@@ -0,0 +1,165 @@
+import AsyncStorageMock from '@react-native-async-storage/async-storage';
```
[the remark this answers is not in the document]
tgolen: Ah, that's interesting and quite magical. 🤔 

I would probably favor the second import since it makes it more clear. If we stick with the first import, I think just adding a comment to explain that it's magically loading the mock and not the real thing would help.

neil-marcellini:
Tim, Marc, and I had a call about this PR and the related App PR to fix the transition flows. We decided that it's ok if the automated test case 2 fails, where we call `set` and then `clear`. We don't currently do this for default key states in the App and we want to use `merge` over `set` whenever possible. I will add a comment about this into the JSDoc for `clear`.

On our call I forgot to mention that I added this line to fix test case 4 (`merge` then `clear`). This line sends the default value to subscribers after the merge value. I will update the comment to explain that.
https://github.com/Expensify/react-native-onyx/blob/27b0983e5bf31a3b069989b663a358e89fe70280/lib/Onyx.js#L719-L720

Is that change ok, or should I remove that automated test? @marcaaron

lib/Onyx.js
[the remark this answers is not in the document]
neil-marcellini: See https://github.com/Expensify/react-native-onyx/pull/129#issuecomment-1136467520

tests/unit/onyxClearNativeStorageTest.js
```diff
@@ -0,0 +1,165 @@
...
+    DEFAULT_KEY: 'defaultKey',
+};
+const SET_VALUE = 'set';
+const MERGED_VALUE = 'merged';
+const DEFAULT_VALUE = 'default';
+
+let storageCallResolveList = [];
+function addStorageCallResolve(name) {
+    storageCallResolveList.push(name);
+}
```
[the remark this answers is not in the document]
neil-marcellini: Yeah good point, it's not needed anymore. I had it doing more initially. I'll remove it.

tests/unit/onyxClearNativeStorageTest.js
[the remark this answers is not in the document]
neil-marcellini: Importing it directly is a good idea.

tests/unit/onyxClearNativeStorageTest.js
neil-marcellini: I put that there because I am mocking `AsyncStorage.clear` and `AsyncStorage.setItem`, and I thought that I had to do that before requiring Onyx. I put it at the top of the file and tested it, and I was wrong.

neil-marcellini:
I addressed Tim's comments and I followed up with the plan from our call. I still have a question from https://github.com/Expensify/react-native-onyx/pull/129#issuecomment-1136467520 and I would appreciate another draft review. I'm assuming that I should still add these test cases for web storage as well?

tests/unit/onyxClearNativeStorageTest.js
```diff
@@ -0,0 +1,162 @@
...
+
+    it('should persist the value of Onyx.merge when called between the cache and storage clearing', () => {
+        expect.assertions(5);
+
+        // GIVEN that Onyx is completely clear
+        // WHEN Onyx.clear() is called
+        Storage.clear = jest.fn(() => {
+            // WHEN merge is called between the cache and storage clearing, on a key with a default key state
+            Onyx.merge(ONYX_KEYS.DEFAULT_KEY, MERGED_VALUE);
+            AsyncStorageMock.clear();
```
marcaaron, line 92: I'm not following how this is possible or why we are creating a mock of `Storage.clear()` that does a weird thing that never actually happens. So I think we should test the actual code that will run and simplify this test.

The examples I had in mind for something like this were more like:

```js
const clearingPromise = Onyx.clear();
Onyx.merge('test', {test: 1});
return clearingPromise.then(() => {
    expect(onyxValue).toEqual({test: 1});
});
``` 
Tthat should replicate the condition where `Onyx.clear()` was called and then `Onxy.merge()` was called _before_ the `Storage` was actually cleared.

Otherwise, I'm not sure what situation we're trying to test exactly. I'm making some assumptions (i.e. that we don't have to dip into storage to get any values) but this is what I think the order of operations would typically be...

- `clear()` called and `getAllKeys()` is called (returns a promise so the merge logic is what runs next)
- `merge()` called and we call `get()` here (also returns a promise)
https://github.com/Expensify/react-native-onyx/blob/f11e7c1949916a81c36c17dd87919dd532d4620f/lib/Onyx.js#L672
- I would think that call to get will pull the cached value _before_  any of these keys are modified here
https://github.com/Expensify/react-native-onyx/blob/a37d7661a8702d75eb6244fb6ff6f47d004e1eb6/lib/Onyx.js#L721-L723
- The `getAllKeys()` promise will returns next and the above logic runs ([we shouldn't have to dip into storage to get these keys so it will just return](https://github.com/Expensify/react-native-onyx/blob/a37d7661a8702d75eb6244fb6ff6f47d004e1eb6/lib/Onyx.js#L75-L78))
- `Storage.clear()` promise starts
- Then the `get()` call will return to trigger the merge and call `setItem()` [here](https://github.com/Expensify/react-native-onyx/blob/a37d7661a8702d75eb6244fb6ff6f47d004e1eb6/lib/Onyx.js#L682) this will modify the values in the cache and start the `Storage.setItem()` promise.
- At this point, the cache should be clear except for the merged value?

tests/unit/onyxClearNativeStorageTest.js
```diff
@@ -0,0 +1,162 @@
...
+            });
+    });
+
+    it('should replace the value of Onyx.set with the default key state in the cache', () => {
+        expect.assertions(5);
+
+        // GIVEN that Onyx is completely clear
+        // WHEN set then clear is called on a key with a default key state
+        Onyx.set(ONYX_KEYS.DEFAULT_KEY, SET_VALUE);
+        Storage.clear = jest.fn(() => AsyncStorageMock.clear());
```
marcaaron, line 117: I would move this up to the top of the test or reset it in `beforeEach()` so it's easier to see that we are testing the final outcome of doing something like this...

```js
Onyx.set(ONYX_KEYS.DEFAULT_KEY, SET_VALUE);
Onyx.clear();
```

tests/unit/onyxClearNativeStorageTest.js
```diff
@@ -0,0 +1,162 @@
...
+        Onyx.clear();
+        return waitForPromisesToResolve()
+            .then(() => {
+                // THEN the Storage.setItem finishes for the set value before the storage finishes clearing
+                expect(getStorageCallResolveOrder('setItem')).toBe(1);
+                expect(getStorageCallResolveOrder('clear')).toBe(2);
```
marcaaron, line 123: Confused about why the order in which these resolve is relevant. Shouldn't it always be the same? What would change this? Are we trying to prevent a specific situation by adding this logic?

tests/unit/onyxClearWebStorageTest.js
```diff
@@ -0,0 +1,62 @@
...
+
+        Onyx.init({
+            keys: ONYX_KEYS,
+            registerStorageEventListener: () => {},
+            initialKeyStates: {
+                [ONYX_KEYS.DEFAULT_KEY]: 'default',
+            },
+        });
+        const webStorageClear = Storage.clear;
+        Storage.clear = jest.fn(() => new Promise(resolve => setTimeout(resolve, 20))
```
marcaaron, line 27: Why are we delaying this?

tests/unit/onyxClearWebStorageTest.js
```diff
@@ -0,0 +1,62 @@
...
+    beforeEach(() => {
+        cache = require('../../lib/OnyxCache').default;
+    });
+
+    afterEach(() => {
+        Onyx.disconnect(connectionID);
+        return Onyx.clear();
+    });
+
+    it('should store merged values when calling merge on a default key after clear', () => {
```
marcaaron, line 41: Why only test the "`Onyx.clear()` then `Onyx.merge()`" condition with the web storage but testing 2 additional case in the native storage test?

**NAB question:** Is there some way to have them run the same tests just twice and use a different provider each time?

tests/unit/onyxClearWebStorageTest.js
```diff
@@ -0,0 +1,62 @@
...
+            initWithStoredValues: false,
+            callback: val => defaultValue = val,
+        });
+        const afterClear = Onyx.clear();
+
+        // Merge just after clear, on the next tick
+        const afterMerge = new Promise(resolve => setTimeout(() => {
+            Onyx.merge(ONYX_KEYS.DEFAULT_KEY, 'merged')
+                .then(resolve);
+        }, 0));
```
marcaaron, line 54: So we are trying to simulate what exactly...? I think it's important to walk through the order of operations to see if it matters. This intentions need to be explained better or simplified.

1. `Onyx.clear()` is called and calls triggers a `getAllKeys()`
2. -- nothing happens with `merge()` because we are arbitrarily waiting
3.  `getAllKeys()` returns and we set the cached values then schedules the keys changed call and `Storage.clear()`
4. `merge` can start now and calls `get()`
5. ... does it matter? and how is this different from just calling `Onyx.merge()` right after `Onyx.clear()`?

marcaaron (changes requested):
Thanks @neil-marcellini, I gave this a more thorough review. Ultimately, I think the code is looking fine, but I'm having a really difficult time making sense out of the tests (but appreciate you taking the time to look into them).

I think at this point it would be best if we greatly simplified the testing portion here and just test whatever we reasonably can without introducing delays and creative mocks. If we run into an actually issue with the code we can try to resolve it and react by creating a test.

tests/unit/onyxClearNativeStorageTest.js
kidroca: I initially used this type of expressions for one of my tests

Re-requiring Onyx gives us a fresh instance (empty cache, no connections etc...) 
It makes sense to `require` like that in a `beforeEach` blocks, but not in `beforeAll`

neil-marcellini:
Thanks for the review Marc! I should have mentioned that I'm not done with the WebStorage tests yet. Overall I think I made these tests too complicated. I was tracking the resolve order of the Storage calls to be explicit about when each call completes, but it's really not necessary so I will remove all of that complexity. 

I was putting the Storage calls into a queue based off of [this comment](https://github.com/Expensify/react-native-onyx/pull/129#issuecomment-1122603886), but maybe I interpreted this wrong and it is also unnecessary.
> I think IndexedDB (web) and AsyncStorage (native) has mechanisms to queue transactions so merge would happen after clear, but our cache has no such logic

I will try to greatly simplify these tests while still making sure that `merge()` or `set()` happens between the cache and storage clearing.

tests/unit/onyxClearNativeStorageTest.js
neil-marcellini: > I'm not following how this is possible or why we are creating a mock of Storage.clear() that does a weird thing that never actually happens. So I think we should test the actual code that will run and simplify this test.

I was doing this to make sure that `merge()` is called between the cache and storage clearing. Your explanation of the flow makes sense, so I guess I don't need that mock.

tests/unit/onyxClearNativeStorageTest.js
```diff
@@ -0,0 +1,104 @@
...
+        // WHEN merge is called between the cache and storage clearing, on a key with a default key state
+        Onyx.merge(ONYX_KEYS.DEFAULT_KEY, MERGED_VALUE);
+        return waitForPromisesToResolve()
+            .then(() => {
+                // THEN the value in Onyx, the cache, and the storage is the merged value
+                expect(onyxValue).toBe(MERGED_VALUE);
+                const cachedValue = cache.getValue(ONYX_KEYS.DEFAULT_KEY);
+                expect(cachedValue).toBe(MERGED_VALUE);
+                const storedValue = Storage.getItem(ONYX_KEYS.DEFAULT_KEY);
+                return expect(storedValue).resolves.toBe(MERGED_VALUE);
```
neil-marcellini, line 59: The storedValue is resolving to `null` and I have no idea why. From some logging I added it looks like `Storage.setItem(ONYXKEYS.DEFAULT_KEY, MERGED_VALUE)` called from `Onyx.set` is the last call to resolve before the assertions, so why is the value `null`?

lib/Onyx.js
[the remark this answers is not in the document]
neil-marcellini: > Not sure I get it. If we are calling a `merge()` then a `clear()` in the same tick then wouldn't the correct behavior be for the `clear()` to cancel the `merge()`... ?

Without this, `clear()` runs before `merge()` since `merge()` is delayed by `get(key)` and `set()`. What are you thoughts on that?
https://github.com/Expensify/react-native-onyx/blob/5478ec947dbbbfaaaa29923c90d47314117ff83d/lib/Onyx.js#L673-L688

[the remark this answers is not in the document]
marcaaron: Sorry, not seeing the issue. `Onyx.clear()` runs before `Onyx.merge()` when we do this?

```js
Onyx.merge('something', {test: 'wat'});
Onyx.clear();
```

>clear() runs before merge() since merge() is delayed by get(key) and set()

Can we maybe break down this explanation a bit further ?

tests/unit/onyxClearNativeStorageTest.js
marcaaron: Let's look at what `Storage.getItem()` doing in the context of this test...

It will get the value inside the mock [here](https://github.com/react-native-async-storage/async-storage/blob/master/jest/async-storage-mock.js) so I'd imagine that this code is [running](https://github.com/react-native-async-storage/async-storage/blob/f0c1ef595fdb7df5168476ac82122c4e9b008e8e/jest/async-storage-mock.js#L83-L84) and clearing the value.

It seems like this validates my concern that a value merged after `Onyx.clear()` will not actually make it into device storage, but does make it to the cache. Though, I'm not sure whether this is a problem or not for our applications using Onyx.

If it causes no issues for us in `App` I think it would be best to create an issue to improve this later and document that it's a known issue.

neil-marcellini: > It seems like this validates my concern that a value merged after `Onyx.clear()` will not actually make it into device storage, but does make it to the cache. 

Yeah I think you are right. I did some logging with the web storage test and what I'm seeing is that `merge()` is called before the cache is cleared from `clear()`. I would like to go back to my previous test where I call `merge()` in between the cache and storage clearing. That test works and it is the case that this PR aims to fix. https://github.com/Expensify/react-native-onyx/blob/17812239616797b3e331031cc0817cb4d98321f5/tests/unit/onyxClearNativeStorageTest.js#L50-L56

> If it causes no issues for us in `App` I think it would be best to create an issue to improve this later and document that it's a known issue.

I don't think it creates any issues in the App and I agree that we should create an issue to improve it later.

lib/Onyx.js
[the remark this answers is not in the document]
neil-marcellini: Alright, ignore my last explanation. Here is what I have found from logging.
```javascript
Onyx.merge('something', {test: 'wat'});
Onyx.clear();
```
Without the change in question, `keyChanged()` from `Onyx.set()` called via `Onyx.merge()` is called **after** `keyChanged()` from `Onyx.clear()`. This happens because in `Onyx.merge()` we call `Onyx.get()` which finishes before `Onyx.getAllKeys()`. Back in `Onyx.merge()` once we get the key we call `Onyx.set()`. We set the cache and then we call `keyChanged()` on the next tick and we call `Storage.setItem()` for the merged key. Next we go back to `Onyx.clear()` where we loop through all the keys we got and we set the cache and call `keyChanged()` on each key. Immediately after that `keyChanged()` runs with the merged value. Then we go back to `Onyx.clear()` and we call `Storage.clear()`.

That is the saga of merge then clear, and is how the Onyx value is the merged value while the cache and storage are the default value. Calling `keyChanged()` on the next tick fixes this problem.

tests/unit/onyxClearNativeStorageTest.js
marcaaron: >I would like to go back to my previous test where I call merge() in between the cache and storage clearing. That test works and it is the case that this PR aims to fix.

Can we back up and re-state what it is exactly that we are trying to test with recreating the "in between the cache and storage clearing" condition?

If this test is failing I think that's fine, but we must change the assertion and add some documentation that basically explains how there is remove possibility that values might not be saved to device storage and that it's a known issue.

neil-marcellini: > Can we back up and re-state what it is exactly that we are trying to test with recreating the "in between the cache and storage clearing" condition?

I think I was trying to test to see if `Storage.setItem()` waits for `Storage.clear()` to finish before it runs, based on [this Slack comment](https://expensify.slack.com/archives/C03TQ48KC/p1651533185923059?thread_ts=1651522644.295769&cid=C03TQ48KC). However, embarrassingly, I'm still not quite sure why I tried to test  in between the cache and storage clearing.

> If this test is failing I think that's fine, but we must change the assertion and add some documentation that basically explains how there is remove possibility that values might not be saved to device storage and that it's a known issue.

Ok. It seems that any Onyx action to set data around `Onyx.clear()` might be unreliable because the data that gets set in Onyx, the cache, and Storage depends on how long it takes async calls to run. The async calls include `Onyx.get()`, `Onyx.getAllKeys()` and `Storage.clear()` and `Storage.setItem()`.

marcaaron: >It seems that any Onyx action to set data around Onyx.clear() might be unreliable because the data that gets set in Onyx, the cache, and Storage depends on how long it takes async calls to run.

I think it's less drastic and probably more common that `Storage.clear()` happens _after_ we set a value because of the general order of operations and cached values that we are working with. 

`Onyx.set()` immediately saves to storage

https://github.com/Expensify/react-native-onyx/blob/b61c904a55d5410679e08a9fa6bd1bad10f1f894/lib/Onyx.js#L540-L556

`Onyx.merge()` has to wait for `get()` before calling `set()` (the get value is usually cached so we will 99% of the time just continue on the next tick and not a random time)

https://github.com/Expensify/react-native-onyx/blob/b61c904a55d5410679e08a9fa6bd1bad10f1f894/lib/Onyx.js#L667-L683

`Onyx.clear()` has to wait for `getAllKeys()` before it calls `Storage.clear()` (this value is also cached so we will continue on the next tick not a random time)

https://github.com/Expensify/react-native-onyx/blob/b61c904a55d5410679e08a9fa6bd1bad10f1f894/lib/Onyx.js#L713-L721

Here are my thoughts...

1. **clear -> set** - 100% of the time a set after clear will not get saved to storage
2. **clear -> merge** - If we have a merge after clear it is not possible for `Storage.clear()` to get called after the `set()` inside `merge()` because it will always run before the `get()` in `merge()` resolves - no matter how long the `get()` takes.
3. **merge -> clear** - Our `get()` in `merge()` would have to take longer than `getAllKeys()` and `Storage.clear()` and in that case it seems possible for it to set a stale value into storage. This would be very rare since most values will immediately return from the cache.
4. **set -> clear** - There is no promise in `set()` so it's impossible for us to set a stale value as `Storage.setItem()` will always run before `Storage.clear()`

Only 1 and 3 seem like problems we might want to solve, but it's not clear if we need to yet.

One idea for how to fix this would be something like:

**set after clear solution:**
- Set an internal flag that `Onyx.clear()` is still running and hasn't fully cleared storage.
- While that happens batch any new writes to storage or flag which keys need to be saved after we are done, but allow cached values to be set, merged, whatever.
- Once `Storage.clear()` is done we can sync up any queued "writes" with the values in the cache.

**merge before clear solution:**
- Add a "cancelled" state to any merges going into the merge queue here:
https://github.com/Expensify/react-native-onyx/blob/b61c904a55d5410679e08a9fa6bd1bad10f1f894/lib/Onyx.js#L668-L671
- When we call clear() flag all the things in the merge queue as cancelled. 
- Don't process any cancelled items when applying the merge.

There could be other solutions but I'm not sure if we need these ones or others.

neil-marcellini: Currently case 2 is the only one that fails in my automated tests. Case 1 also fails but I have deleted it. All cases work in the manual tests.

Thank you for outlining the potential solutions. Could we please leave those for a follow up? Should I add an example and documentation for case 2 to the `Onyx.clear()` JSDoc? Can I remove case 2 like I removed case 1?

lib/Onyx.js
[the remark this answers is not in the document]
marcaaron: >This happens because in Onyx.merge() we call Onyx.get() which finishes before Onyx.getAllKeys() called from Onyx.clear(). 

Ok so the answer is that we have to do this because we are doing this in `set()`:

https://github.com/Expensify/react-native-onyx/blob/b61c904a55d5410679e08a9fa6bd1bad10f1f894/lib/Onyx.js#L552-L553

We need to do it here as well otherwise a call to `Onyx.clear()` will call `keysChanged()` first despite running later.

That change is pretty old but I think we are probably scared to change it at this point :D 

I would suggest we change that usage to something like `notifySubscribersOnNextTick()` or something and refactor the few usages we have where we are doing `Promise.resolve().then(() => keyChanged())` - it might help contextualize better why we need to do this in some but not all places.

tests/unit/onyxClearNativeStorageTest.js
marcaaron: >Could we please leave those for a follow up?

Yeah I am not sure we even need to follow up. Creating an issue could be good but it doesn't seem like a priority right now.

>Should I add an example and documentation for case 2 to the Onyx.clear() JSDoc?

That sounds good and yes remove whichever cases you think make sense to remove based on this.

neil-marcellini:
@marcaaron I was tinkering around and found that if I call `Storage.clear()` in the same callback when the cache clears then **clear -> merge** works. I think this change makes sense because it makes `Onyx.clear()` more "atomic". Do you think that change is good? If not I will revert it and add some documentation for why this case doesn't work.

Also, with this change and for the native storage test exclusively, it is escaping the quotes in the value. Do you know why that is or how to get around it? I did find [this SO](https://stackoverflow.com/questions/56912226/why-does-jest-snapshots-contain-escaping-characters-because-of-double-quotes) that sort of explains why, but not how to fix it.
image

marcaaron:
>I was tinkering around and found that if I call Storage.clear() in the same callback when the cache clears then clear -> merge works.

Ok so moving the `Storage.clear()` out of the `.then()` move it up in the order of operations so that what I am describing here works?

>clear -> merge - If we have a merge after clear it is not possible for Storage.clear() to get called after the set() inside merge() because it will always run before the get() in merge() resolves - no matter how long the get() takes.

Maybe I was wrong about that, but the change you are suggesting now makes it true?

marcaaron:
>for the native storage test exclusively, it is escaping the quotes in the value

Hmm not entirely sure but seems maybe related to stringifying JSON. Maybe try something that isn't a string as a value - object e.g.

neil-marcellini:
> Maybe I was wrong about that, but the change you are suggesting now makes it true?
Yes

> Hmm not entirely sure but seems maybe related to stringifying JSON. Maybe try something that isn't a string as a value - object e.g.

It's super weird. It seems like the mock isn't very good. I wrote a work-around using integers and `parseInt` since the storage returned them as a string. Hopefully that's ok. All of the tests are passing now.

lib/Onyx.js
```diff
@@ -706,18 +707,36 @@ function initializeWithDefaultKeyStates() {
...
+                Promise.resolve().then(() => keyChanged(key, resetValue));
             });
-        })
-        .then(Storage.clear)
-        .then(initializeWithDefaultKeyStates);
+            Storage.clear();
```
marcaaron, line 738: NAB
```suggestion
            return Storage.clear();
```

lib/Onyx.js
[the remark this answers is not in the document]
marcaaron: I still think we should improve this later but the context is here so I'm good with this.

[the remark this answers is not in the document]
neil-marcellini: > I would suggest we change that usage to something like `notifySubscribersOnNextTick()` or something and refactor

Do you want me to do that now, in a follow up, or let someone else do that?

marcaaron (approved):

PR #130, opened 2022-05-13
Allow a version of Onyx that doesn't include any react-native components

tgolen:
### Related Issues

GH_LINK


kidroca:
We might need to do something about this

https://github.com/Expensify/react-native-onyx/blob/cb655b13fd2ffe1e1b4d957e8063ade175d61d4b/lib/Onyx.js#L821-L824

https://github.com/Expensify/react-native-onyx/blob/cb655b13fd2ffe1e1b4d957e8063ade175d61d4b/lib/Onyx.js#L875-L877

https://github.com/Expensify/react-native-onyx/blob/cb655b13fd2ffe1e1b4d957e8063ade175d61d4b/lib/metrics.js#L1-L3

Even thought that `react-native-performance` is imported dynamically and is an optional dependency, since Onyx is not bundled but lets the parent project bundle its source, bundler like webpack will follow these dynamic imports - they don't know that they won't be used, because at build time we don't know whether Onyx.init would be called with `captureMetrics: true` or `false` 

---

Bundling would again help here because we can stub `./metrics.web.js` to be empty or to have a web-specific metrics implementation

tgolen:
Closing this in favor of https://github.com/Expensify/react-native-onyx/pull/132

PR #132, opened 2022-05-16
Bundle Onyx with Webpack

kidroca:
### Details

Bundle Onyx with Webpack to allow using Onyx outside of react-native

Using a test project with the follow package.json config
<details>
<summary>package.json</summary>

```json
{
  "name": "test-onyx",
  "version": "0.1.0",
  "private": true,
  "dependencies": {
    "@testing-library/jest-dom": "^5.16.4",
    "@testing-library/react": "^13.2.0",
    "@testing-library/user-event": "^13.5.0",
    "expensify-common": "git+https://github.com/Expensify/expensify-common.git#2e5cff552cf132da90a3fb9756e6b4fb6ae7b40c",
    "localforage": "^1.10.0",
    "react": "^18.1.0",
    "react-dom": "^18.1.0",
    "react-native-onyx": "git+https://github.com/kidroca/react-native-onyx.git#11dc5443e4d6e7cac3a967344996466bb26966c8",
    "react-scripts": "5.0.1",
    "web-vitals": "^2.1.4"
  },
  "scripts": {
    "start": "react-scripts start"
  },
  "eslintConfig": {
    "extends": [
      "react-app",
      "react-app/jest"
    ]
  }
}
```
</details>

We can import and use `Onyx` in a non react-native project, like we do in App:

```js
import Onyx from 'react-native-onyx';
```

Using the following package.json fields we tell bundlers to load different Onyx configurations per environment 
- `browser` -> Onyx is loaded for non react-native projects and uses localforage to manage IndexedDb
- `react-native` -> Same as what we used to have, Onyx is loaded and uses the native storage
- `main` -> Right now this field only aids IDEs with code completion from the unmodified source

### Related Issues

Slack thread: https://expensify.slack.com/archives/C01GTK53T8Q/p1652389473489429

### Automated Tests

No new functionality added
Existing tests are still pass


package.json
```diff
@@ -6,22 +6,32 @@
...
     "test": "jest",
-    "build-docs": "node buildDocs.js"
+    "build": "npm run build:native && npm run build:web",
+    "build:docs": "node buildDocs.js",
+    "build:native": "webpack --config webpack.native.js",
+    "build:web": "webpack --config webpack.web.js"
   },
   "dependencies": {
     "ascii-table": "0.0.9",
-    "expensify-common": "git+https://github.com/Expensify/expensify-common.git#2e5cff552cf132da90a3fb9756e6b4fb6ae7b40c",
```
kidroca, line 17: Specifying `expensify-common` under peer dependencies ensures both `Expnesify/App` and `Expnsify/react-native-onyx` use the same `expensify-common` version (the one from App)

package.json
```diff
@@ -38,22 +49,33 @@
...
+    "webpack": "^5.72.1",
+    "webpack-cli": "^4.9.2",
+    "webpack-merge": "^5.8.0"
   },
   "peerDependencies": {
     "@react-native-async-storage/async-storage": "^1.15.5",
+    "expensify-common": ">=1",
     "localforage": "^1.10.0",
-    "react": "^17.0.2",
+    "react": ">=17.0.2",
```
kidroca, line 67: `>=` so React 18 is accepted as well

package.json
```diff
@@ -38,22 +49,33 @@
...
   },
   "peerDependenciesMeta": {
     "react-native-performance": {
       "optional": true
+    },
+    "@react-native-async-storage/async-storage": {
+      "optional": true
+    },
+    "localforage": {
+      "optional": true
```
kidroca, line 78: `async-storage` and `localforage` are optional to let consumers use one or the other

package.json
```diff
@@ -81,5 +103,6 @@
     "setupFilesAfterEnv": [
       "@testing-library/jest-native/extend-expect"
     ]
-  }
+  },
+  "sideEffects": false
```
kidroca, line 107: This lets Webpack bundlers know that our code is not doing anything funky like change the Prototype or adding stuff to `window` and aids tree shaking

webpack.common.js
```diff
@@ -0,0 +1,36 @@
...
+                use: {
+                    loader: 'babel-loader',
+                },
+            },
+        ],
+    },
+    externals: [
+        'react-native',
+        /^expensify-common\/.+$/,
+        /^lodash\/.+$/,
```
kidroca, line 25: Regex syntax is intentional since we import subpaths of these modules 
- `lodash/merge` 
- `expensify-common/str/...` 

If we don't use a pattern the whole package is bundled even if it's a peer dependency: https://webpack.js.org/guides/author-libraries/#external-limitations

webpack.common.js
```diff
@@ -0,0 +1,36 @@
...
+        ..._.keys(pkg.peerDependencies),
+        ..._.keys(pkg.dependencies),
```
kidroca, line 27: Our package `dependencies` are installed automatically into the consumer project `node_modules` 

`peerDependencies` would be installed as well, or flagged as not met depending on the `npm` version 
These are relatively easy to resolve if not installed, because we'll get an error that a module cannot be found

kidroca:
cc @tgolen

package.json
```diff
@@ -6,22 +6,32 @@
   "description": "State management for React Native",
   "license": "MIT",
   "private": true,
+  "files": [
+    "dist"
+  ],
```
tgolen, line 11:
```suggestion
  ],
  "main": "dist/index.native.js",
```

package.json
```diff
@@ -6,22 +6,32 @@
...
+  "exports": {
+    "./web": "dist/index.web.js",
+    ".": "dist/index.native.js"
```
tgolen, line 14:
```suggestion
    "./web": "dist/index.web.js"
```

tgolen (changes requested):
Found out we need the `main` field

package.json
kidroca: Added

kidroca:
✅ Fixed unit tests
This was a bit tricky to tackle. It turned out that since we're using `react-native` preset for `jest`, `jest` was configured to resolve `.native.js` extensions like `index.native.js` and it would prefer them over plain `index.js`, so the mock for Storage wasn't getting picked up

✅ Fixed eslint errors
Similar problems with letting eslint know that there's no `storage/index.js` but `storage/index.native.js` and `storage/index.web.js` 
Also ignored the `dist` directory since it's minified code

✅ Configured `exports` correctly
Looks like it's preferable to specify paths as `./dist/index...` instead of just `dist/index...` so that IDEs don't complain

image
_An IDE complaining the path `react-native-onyx/web` is not installed_

✅ Added production grade source maps
These help with stack traces and debugging Onyx sources through parent projects
Now that the code is minified we need these to trace calls like Onyx.merge from App

✅ Added script that generates typescript typings
Again since code is now bundled a the information about method signature cannot be resolved like it used to
In order to let IDEs continue to provide code completion we need to provide static typing of the library

image
_Code completion is suggested for both `react-native-onyx` and `react-native-onyx/web`_

✅ Verified all envs and entry points work in App and for new projects using Onyx
The package.json structure and exports/main are correctly resolved in 
- webpack (web/desktop)
- metro (ios/android)

babel.config.js
```diff
@@ -1,4 +1,8 @@
 module.exports = {
-    presets: [require('metro-react-native-babel-preset')],
-    plugins: [],
+    presets: [
+        require('metro-react-native-babel-preset'),
+    ],
+    plugins: [
+        '@babel/plugin-proposal-class-properties',
+    ],
```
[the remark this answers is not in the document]
tgolen: Ah, OK. Good to know. The plugin is required, yeah. Thanks!

tgolen (dismissed):
Awesome, this looks great! Thanks for taking this across the finishline @kidroca.

tgolen:
@kidroca When testing this with the web-only import... I'm getting the following webpack error:
image

This is what my import looks like:
```
import {withOnyx} from 'react-native-onyx/web';
```

It appears that the code is working, so I'm not sure what this error message means or is referring to. Google seemed equally confused.

tgolen:
Actually, I was wrong about the code not working (must have been the old version of code). This is the JS console error I get:

image

kidroca:
> @kidroca When testing this with the web-only import... I'm getting the following webpack error:

Hmm could it be because I added back `.` to exports? or because of the order of the imports 
I'll give it another test in a web only project

kidroca:
Regarding the _"Error: Default condition should be last one"_
looks like the default condition for exports should be the last one, like so

```json
"exports": {
    ".": {
      "require": "./dist/revue-draggable.cjs.js",
      "default": "./dist/revue-draggable.es.js"
    },
```

kidroca:
I've recreated the error, looks like it wasn't a problem in App, but it was for the `/web` import 
Moving the `default` declarations to the bottom resolved the Error and the module is found

tgolen (approved):
Cool. Yep, that did the trick!

dist/types/lib/OnyxCache.d.ts.map
```diff
@@ -0,0 +1 @@
+{"version":3,"file":"OnyxCache.d.ts","sourceRoot":"","sources":["../../../lib/OnyxCache.js"],"names":[],"mappings":";AAgMA,kCAAiC;AA1LjC;;;GAGG;AACH;IAEQ;;;;OAIG;IACH,oBAA4B;IAE5B;;;;OAIG;IACH,mBAA2B;IAE3B;;;;OAIG;IACH,mBAAoB;IAEpB;;;;OAIG;IACH,wBAAyB;IAW7B;;;OAGG;IACH,cAFa,MAAM,EAAE,CAIpB;IAED;;;;OAIG;IACH,cAHW,MAAM,OAMhB;IAED;;;;OAIG;IACH,oBAHW,MAAM,GACJ,OAAO,CAInB;IAED;;;;OAIG;IACH,YAFW,MAAM,QAIhB;IAED;;;;;;OAMG;IACH,SAJW,MAAM,mBAUhB;IAED;;;OAGG;IACH,UAFW,MAAM,QAIhB;IAED;;;OAGG;IACH,YAFW,OAAO,MAAM,MAAI,QAS3B;IAED;;;;OAIG;IACH,yBAHW,MAAM,OAKhB;IAED;;;;;;;OAOG;IACH,4BAHW,MAAM,cAKhB;IAED;;;;;;;OAOG;IACH,2BAJW,MAAM,uCAUhB;IAED;;;;OAIG;IACH,0BAIC;IAED;;OAEG;IACH,oCASC;IAED;;;OAGG;IACH,0BAFW,MAAM,QAIhB;IADG,0BAA8B;CAErC"}
```
parasharrajat, line 1: Do we need these map files?

tgolen: I kind of feel like we should always just be creating unmodified source code with these bundles. Anything that imports them would be doing it's own bundling/minifying, so having the source code unmodified would help development, right?

kidroca: It's a bit more tricky in our case, because the bundler will have to be aware that Onyx source has `.web` and `.native` files 

See how React has `react.production.min.js` - they already provide a minified file
Even the development file comes from a pre-bundled location - `./cjs/...` 

Every library should be bundled, this relieves the consumer project form any quirks the library might otherwise require

---

The only thing helpful so far for me using unmodified Onyx source was that I can directly tweak it inside `node_modules` to experiment locally
(But I guess that's huge 😅)

Maybe we can experiment a bit and create a `.development.web.js` and `development.native.js` files, it should be as easy as setting `mode: development` 
And it should also be possible to do this:

```json
{
  "exports": {
    ".": {
       "development": "./dist/index.development.native.js"
       "default": "./dist/index.native.js"
     },
    "./web":  {
       "development": "./dist/index.development.web.js"
       "default": "./dist/index.web.js"
     },
  }
}
```

kidroca:
Removed `.d.ts.map` files 

Will experiment with the `.development.js` idea and post back

techievivek:
Removing myself as a reviewer as I don't have much experience with webpack thing and we have got a good amount of helping hands. Thanks.

webpack.config.js
```diff
@@ -0,0 +1,95 @@
...
+        filename: 'native.development.js',
+    },
+});
+
+module.exports = [
+    webConfig,
+    webDevConfig,
+    nativeConfig,
+    nativeDevConfig,
+];
```
kidroca, line 95: Exporting an array of configs let's webpack build them in parallel

tgolen (changes requested):
The changes look good to me. I didn't know that about webpack being able to build the configs in parallel. Neat!

I think the last thing is just adding a few comments and instructions so that it's clear what everything is for, and then this should be good to merge.

kidroca:
Hey @tgolen, while writing the documentation I was testing stuff and found out how to get rid of the `.d.ts` files, and also use the unmodified source for Onyx for `react-native` projects
Now the `native` source doesn't have to be bundled, which allows us to tweak it (experiment) from parent projects

Another change is we no longer have to import `react-native-onyx/web` for web projects, `package.json` has the following fields 
- `react-native` - react native cli would load package source from whatever path is added here 
- `main` - IDEs and other tools would load source from here (in our case it provides code completion for IDEs)
- `browser` - bundlers like Webpack would prefer to load code from here 

Using the `browser` field, bundlers would load the source from `dist/web`

Sorry for the new changes, but I think this makes the setup simpler and closer to what we previously had for `react-native` 
(I've triple tested these changes, but would be happy if someone else confirms the changes are good)

README.md
```diff
@@ -43,6 +43,14 @@ const config = {
 Onyx.init(config);
 ```
 
+### Usage in non react-native projects
+Onyx can be used in non react-native projects, by leveraging the `borwser` field in `package.json`
```
tgolen, line 47:
```suggestion
Onyx can be used in non react-native projects, by leveraging the `browser` field in `package.json`  
```

README.md
```diff
@@ -43,6 +43,14 @@ const config = {
 Onyx.init(config);
 ```
 
+### Usage in non react-native projects
+Onyx can be used in non react-native projects, by leveraging the `borwser` field in `package.json`  
+Bundlers like Webpack respect that field and import code from the path resolved there
+
+```javascript
+import Onyx, { withOnyx } from 'react-native-onyx/web';
```
tgolen:
```suggestion
import Onyx, { withOnyx } from 'react-native-onyx';
```

tgolen (changes requested):
Awesome, I like that better. Cleans it up some!

README.md
```diff
@@ -234,3 +239,20 @@ Sample output of `Onyx.printMetrics()`
 |    1.08sec |   2.20sec |   1.12sec | iou, [object Object]     |
 |    1.17sec |   2.20sec |   1.03sec | currentURL, /            |
 ```
+
+
+# Development
+
+`react-native` bundles source using the `metro` bundler. `metro` does not follow symlinks, so we can't use `npm link` to
+use a local version of Onyx during development
```
kidroca, line 247:
```suggestion
link a local version of Onyx during development
```

tgolen (approved):

tgolen:
I agree with you that the only real reason it's helpful is for messing with
Onyx code directly (strictly a time-saver). I don't think we need to spend
too much time trying to provide that functionality, but I'm curious about
the development.js idea and see how that works.

kidroca:
> I'm curious about the development.js idea and see how that works.

Right now for web we directly return the minified code from `dist/`
What we can do is add a new index script - `web.js` and load that instead
Inside `web.js` would be a `NODE_ENV` check and depending on the value we load the minified code or the non minified code 
Dev servers are running with NODE_ENV=development, while builds are configured for production

One thing to note is that we still load bundled code, so live changes to source code would still not be reflected - to make changes to Onyx live you'd have to modify the non-minified code from `dist/` (which is still a lot easier compared to dealing with minified code)

---

If we want to save time on testing new thing in Onyx, perhaps we can make a script in App that
1. `cd` to `node_modules/react-native-onyx`
2. runs webpack in `watch` mode so any changes to `lib` are rebuild and sent to `dist`
3. now make changes to `node_modules/react-native-onyx/lib` source and see them live in web/desktop

tgolen:
So strange. I wrote that reply months ago and it only now posted to GH.

I do still think it's a problem worth fixing to make development on Onyx source code easier. We've been making a lot of performance changes, so it's been in active development lately.

I like your second idea since that would allow us to edit the source code directly. Could we try that?

tgolen:
I think maybe we should open a new GH issue for that

kidroca:
@tgolen 
> I like your second idea since that would allow us to edit the source code directly. Could we try that?
> I think maybe we should open a new GH issue for that

Yep, we have almost everything in place, we just need to:
1. Add a `watch` script in Onyx, that runs webpack in watch mode
2. Add a `test-onyx` script in App. It goes to Onyx folder and install `devDependencies` so we can build and re-build the app with the build command (no-op if they are already installed), then trigger the `watch` script

puneetlath:
@kidroca @tgolen @parasharrajat just FYI, it looks like this PR likely caused this bug: https://github.com/Expensify/App/issues/10622

tgolen:
Do you mean [this PR](https://github.com/Expensify/react-native-onyx/pull/198) specifically, or is there another? There are multiple attached to that issue.

kidroca:
@puneetlath 
This PR caused the web part of the bug, because we forgot to test capturing metrics after the change
- capturing metrics on web was not supposed to work after the change, so we didn't test setting the flag `ONYX_METRICS: true` 
- capturing metrics on native continued to work until the introduction of `Onyx.update` method

The Android issue was because of the introduction of the `Onyx.update` method which (at the time) didn't return Promise but was added to the list of decorated methods (activated by env flag)
https://github.com/Expensify/react-native-onyx/pull/133/files#diff-146c57cda221f8f5ee5c808eee0c6692b02c4ac2b8818d0104896a04be78e78aR917

Initially `decorateWithMetrics` was implemented to be able to decorate any func - promise returning or not, but the code was slightly more complicated and I was asked to make it work only for promise returning (async) functions in order to simplify 

I guess we should have added documentation about only working with async function to
- the place where we decorate all the Onyx (async) functions
- and on `decorateWithMetrics` function description

puneetlath:
Ok good learnings! Anything we should do now?

kidroca:
We can update method documentation and comments to mention that only Promise returning methods can be decorated

We can update Onyx metrics to work on Web - we only disabled them so that Onyx can be used in other non react-native projects, because it caused issues with bundling - instead of applying a working around - disabling metrics for web (via a no-op web file), we can fix bundling

PR #133, opened 2022-05-20
Create onyx.update

luacmartins:
### Related Issues

GH_LINK

### DIFF
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -556,6 +556,42 @@ function set(key, value) {
 }
 
 /**
+ * Insert API responses and lifecycle data into Onyx
+ *
+ * @param {Object} data An object 
+ */
+ function update(data) {
+    // First, validate the Onyx object is in the format we expect
+    const isValidDataObject = _.every(data, ({onyxMethod, key, value}) => {
+        if (onyxMethod !== 'set' || onyxMethod !== 'merge') {
+            return false;
+        }
+        if (!_.isString(key)) {
+            return false;
+        }
+        if (!value) {
+            return false;
+        }
+        return true;
+    });
+
+    if (!isValidDataObject) {
+        // Throw error
+    }
+
+    _.each(data, ({onyxMethod, key, value}) => {
+        if (onyxMethod === 'set') {
+            Onyx.set(key, value);
+            return;
+        }
+        if (onyxMethod === 'merge') {
+            Onyx.merge(key, value);
+            return;
+        }
+    });
+}
+
+/**
  * Storage expects array like: [["@MyApp_user", value_1], ["@MyApp_key", value_2]]
  * This method transforms an object like {'@MyApp_user': myUserValue, '@MyApp_key': myKeyValue}
  * to an array of key-value pairs in the above format


luacmartins:

lib/Onyx.js
```diff
@@ -780,6 +780,39 @@ function mergeCollection(collectionKey, collection) {
...
+function update(data) {
+    // First, validate the Onyx object is in the format we expect
+    _.each(data, ({onyxMethod, key, value}) => {
+        if (!_.contains(['set', 'merge'], onyxMethod)) {
+            throw new Error(`Invalid onyxMethod ${onyxMethod} in Onyx update.`);
+        }
+        if (!_.isString(key)) {
+            throw new Error(`Invalid ${typeof key} key provided in Onyx update. Onyx key must be of type string.`);
+        }
+        if (typeof value === 'undefined') {
```
tgolen, line 800: Use `_.isUndefined()` to be consistent with other underscore methods.

lib/Onyx.js
```diff
@@ -780,6 +780,39 @@ function mergeCollection(collectionKey, collection) {
...
+    });
+
+    _.each(data, ({onyxMethod, key, value}) => {
+        if (onyxMethod === 'set') {
+            set(key, value);
+            return;
+        }
+        if (onyxMethod === 'merge') {
+            merge(key, value);
+        }
```
tgolen, line 812: NAB: I think doing a `switch()` is slightly better here, and it makes it a bit more extensible for the future.

lib/Onyx.js
```diff
@@ -780,6 +780,39 @@ function mergeCollection(collectionKey, collection) {
...
+            throw new Error(`Invalid ${typeof key} key provided in Onyx update. Onyx key must be of type string.`);
+        }
+        if (typeof value === 'undefined') {
+            throw new Error(`Undefined value passed to key ${key} in Onyx update.`);
+        }
+    });
+
+    _.each(data, ({onyxMethod, key, value}) => {
+        if (onyxMethod === 'set') {
+            set(key, value);
```
tgolen, line 807: NAB: It might be nice to collect all the things using set and pass them to `multiSet()`, but since we are trying to move away from using `set()` at all, the could be a preoptimization. I kinda wish we had a `multiMerge()` though.

tests/unit/onyxTest.js
```diff
@@ -306,4 +306,80 @@ describe('Onyx', () => {
...
+            Onyx.update(data);
+        } catch (error) {
+            expect(error.message).toEqual('Invalid onyxMethod multiSet in Onyx update.');
+        }
+
+        try {
+            data[1] = {onyxMethod: 'merge', key: true, value: {test2: 'test2'}};
+            Onyx.update(data);
+        } catch (error) {
+            // eslint-disable-next-line max-len
```
tgolen, line 374: Is this necessary? It looks like another liner higher up is longer than this, and I thought the length was something like 190, which this line looks to be an OK length.

lib/Onyx.js
tgolen: Also, I think it's OK for `undefined` to be passed with `set()` is that right? It would be expected to remove the Onyx key entirely.

tgolen (changes requested):
I always comment about this in PRs adding tests, and unfortunately, we don't do this consistently, but I think it GREATLY improves the readability of tests (especially when someone that didn't write them has the task of maintaining them). Could you please add comments in the test that follow this format:

```
// GIVEN <initial state of the system>
...code that sets up the initial state

// WHEN <something happens>
...code that does something

// THEN <explain the final expected state of the system>
...assertions
```

lib/Onyx.js
luacmartins: That seems right based on [this comment](https://github.com/Expensify/react-native-onyx/blob/8772195796b9a0c9c6fec5a3feeeb214f2b6a5ca/lib/Onyx.js#L631). However, [this line](https://github.com/Expensify/react-native-onyx/blob/8772195796b9a0c9c6fec5a3feeeb214f2b6a5ca/lib/Onyx.js#L634) seems to only check for `null` values. Maybe I'm missing something though.

lib/Onyx.js
luacmartins: Sounds good. I switch (pun intended)

lib/Onyx.js
luacmartins: I'm not sure how much of a pre-optimization that is, so I'll skip this for now.

tests/unit/onyxTest.js
tgolen: Yea, I agree with not having warnings. I guess I wonder why it's at 120 and not 190. Maybe try updating the eslint config to make sure it's using our latest?

lib/Onyx.js
luacmartins: I'll remove this check since it seems like we want to support `undefined` values as well.

tgolen: Hm, I don't think that's quite right. That line is for merging, and it just means that if a `null` value is passed to `merge()` then it will get ignored (hence, it is "deleted" from the data being merged into Onyx).

The [`setItem()` method in LocalForage](https://localforage.github.io/localForage/#data-api-setitem) doesn't seem to mention what happens if you pass `undefined` as the value, but it might be worth experimenting with to verify the expected behavior!

luacmartins: Hmm that makes sense. I experimented a bit and it seems that passing undefined does remove the key. I'm removing that check then. Thanks!

lib/Onyx.js
```diff
@@ -780,6 +780,39 @@ function mergeCollection(collectionKey, collection) {
         });
 }
 
+/**
+ * Insert API responses and lifecycle data into Onyx
+ *
+ * @param {Object} data An object containing the Onyx data to set
```
marcochavezf, line 786: This should be an `Array`, no?

```suggestion
 * @param {Array} data An array containing the Onyx data to set
```

tests/unit/onyxTest.js
luacmartins: Yea, we are really outdated on eslint-config. Updating it surfaces A LOT of errors, so I'll address that in a separate PR. I'll keep the comment for now then.

lib/Onyx.js
luacmartins: Good call! Thanks!

tests/unit/onyxTest.js
```diff
@@ -306,4 +306,84 @@ describe('Onyx', () => {
...
+                        onyxMethod: 'merge',
+                        key: ONYX_KEYS.ANOTHER_TEST,
+                        value: {test2: 'test2'},
+                    },
+                ]);
+                return waitForPromisesToResolve();
+            })
+            .then(() => {
+                // THEN the final Onyx state should be {test: 'one', anotherTest: {test1: 'test1', test2: 'test2'}}
+                expect(testKeyValue).toBe('one');
```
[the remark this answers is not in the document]
luacmartins: Haha I was testing something and shouldn't have committed that. I already reverted those changes.

luacmartins:
Updated!

marcochavezf (approved):
LGTM 
NAB: I think it will be great to add unit tests that set `undefined` and `null` values

tgolen (changes requested):

lib/Onyx.js
```diff
@@ -780,6 +780,36 @@ function mergeCollection(collectionKey, collection) {
...
+/**
+ * Insert API responses and lifecycle data into Onyx
+ *
+ * @param {Array} data An array of objects with shape {onyxMethod: oneOf('set', 'merge'), key: string, value: *}
+ */
+function update(data) {
+    // First, validate the Onyx object is in the format we expect
+    _.each(data, ({onyxMethod, key}) => {
+        if (!_.contains(['set', 'merge'], onyxMethod)) {
+            throw new Error(`Invalid onyxMethod ${onyxMethod} in Onyx update.`);
```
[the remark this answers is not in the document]
luacmartins: I'm not sure they are quite the same. If any of these checks fail, we don't run any set/merge at all vs if this happens in the switch, we could run a few set/merges before getting to the bad object. 

I'd imagine that we only want to process the update if the full API response is valid?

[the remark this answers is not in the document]
luacmartins: Removing the default case from the switch or just adding a `return` as default triggers lint errors. Options are:

1. Add this error handling, knowing that we should never hit that code.
2. Manually disable the lint error
3. Switch back to if statements
4. Something else?

lib/Onyx.js
```diff
@@ -883,6 +914,7 @@ function applyDecorators() {
     mergeCollection = decorate.decorateWithMetrics(mergeCollection, 'Onyx:mergeCollection');
     getAllKeys = decorate.decorateWithMetrics(getAllKeys, 'Onyx:getAllKeys');
     initializeWithDefaultKeyStates = decorate.decorateWithMetrics(initializeWithDefaultKeyStates, 'Onyx:defaults');
+    update = decorate.decorateWithMetrics(update, 'Onyx:update');
```
luacmartins, line 917: I had missed adding the update method to the decorators for capturing metrics

lib/Onyx.js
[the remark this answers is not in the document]
luacmartins: Chatted 1:1 with Tim and we decided to just add a `break` to the default method!

luacmartins:
Updated!

tests/unit/onyxTest.js
luacmartins: Got a [PR up](https://github.com/Expensify/react-native-onyx/pull/135) to update eslint-config

lib/Onyx.js
```diff
@@ -780,6 +780,36 @@ function mergeCollection(collectionKey, collection) {
         });
 }
 
+/**
+ * Insert API responses and lifecycle data into Onyx
+ *
+ * @param {Array} data An array of objects with shape {onyxMethod: oneOf('set', 'merge'), key: string, value: *}
```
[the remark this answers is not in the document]
luacmartins: That's a good idea. I'm not sure we want to block this PR on adding this reference though. Created an [issue](https://github.com/Expensify/react-native-onyx/issues/138) for it. Thanks!

luacmartins:
Ready for another review

lib/Onyx.js
```diff
@@ -780,6 +780,36 @@ function mergeCollection(collectionKey, collection) {
...
+    _.each(data, ({onyxMethod, key, value}) => {
+        switch (onyxMethod) {
+            case 'set':
+                set(key, value);
+                break;
+            case 'merge':
+                merge(key, value);
+                break;
+            default:
+                break;
```
Julesssss, line 808: Usually, I prefer switch/case, but this seems more verbose than two simple conditional checks. NAB though

Julesssss (approved):
Looks good to me, left a small comment but it's not a blocker. Feel free to resolve nobody else agrees 🙂

lib/Onyx.js
luacmartins: Haha I initially had it as a series of conditionals, but changed it because of @tgolen's comments [here](https://github.com/Expensify/react-native-onyx/pull/133#discussion_r879765933). I don't have strong feeling either way.

luacmartins: Sounds good! Let's not switch this then 🤣

lib/Onyx.js
```diff
@@ -780,6 +780,36 @@ function mergeCollection(collectionKey, collection) {
...
+ * @param {Array} data An array of objects with shape {onyxMethod: oneOf('set', 'merge'), key: string, value: *}
+ */
+function update(data) {
+    // First, validate the Onyx object is in the format we expect
+    _.each(data, ({onyxMethod, key}) => {
+        if (!_.contains(['set', 'merge'], onyxMethod)) {
+            throw new Error(`Invalid onyxMethod ${onyxMethod} in Onyx update.`);
+        }
+        if (!_.isString(key)) {
+            throw new Error(`Invalid ${typeof key} key provided in Onyx update. Onyx key must be of type string.`);
```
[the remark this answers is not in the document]
luacmartins: That's a good idea. We could certainly add that as a follow up if we come across this issue!

lib/Onyx.js
```diff
@@ -780,6 +780,36 @@ function mergeCollection(collectionKey, collection) {
         });
 }
 
+/**
+ * Insert API responses and lifecycle data into Onyx
+ *
+ * @param {Array} data An array of objects with shape {onyxMethod: oneOf('set', 'merge'), key: string, value: *}
+ */
+function update(data) {
```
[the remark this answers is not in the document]
luacmartins: @parasharrajat thanks for investigating this. So can we just return the respective methods [here](https://github.com/Expensify/react-native-onyx/blob/2eb3d74e7b581007d7c4dcc332b5bbc027145ce2/lib/Onyx.js#L837-L852)? e.g.

```
case 'merge':
    return merge(key, value);
```

PR #151, opened 2022-06-04
Add a debug line to help figure out what's happening

AndrewGable:
### Details

Adds a debug line to figure out what is not committed in the `gh pr create` step.

### Linked PRs

https://github.com/Expensify/Expensify/issues/212012


.github/workflows/publish.yml
```diff
@@ -61,6 +61,12 @@ jobs:
             - name: Push branch and publish tags
               run: git push --set-upstream origin ${{ env.BRANCH_NAME }} && git push --tags
 
+            - name: Git Status Debug
+              run: git status
+
+            - name: Git Diff Debug
+              run: git diff
```
roryabraham, line 68: Double-check that this will work without having to configure git to not use a pager. If I run git diff locally it will open vim, but we want this to just print the diff to stdout. Also maybe consider adding the flag to make the diff colored / easier to read.

roryabraham (approved):
other than those NAB suggestions, LGTM

.github/workflows/publish.yml
AndrewGable: Good point. I actually think we should remove this line just incase the file is sensitive.

ctkochan22 (approved):

roryabraham (approved):

PR #153, opened 2022-06-07
Add decrypted key to `.gitignore` to fix GitHub actions warning

AndrewGable:
### Details

We were seeing an warning when running the GitHub actions that said `Warning: 1 uncommitted change` - From debugging we figured out that it was the decrypted key. We should not commit this key, so we can add it to the gitignore.

### Related Issues

https://github.com/Expensify/Expensify/issues/212012

### Automated Tests

1. Merge this PR
2. I will check the logs for the GitHub action to verify that it fixes this warning

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,6 @@ dist/
 
 # IDEs
 .idea
+
+# Decrypted private key we do not want to commit
+.github/OSBotify-private-key.asc


parasharrajat:
I suggest removing is from the history of the repo as well. It sounds like a private key.

AndrewGable:
It wasn't ever committed decrypted, just encrypted.

roryabraham (approved):

PR #163, opened 2022-08-02
Add waitForCollectionCallback param to Onyx.connect

luacmartins:
### Details
Adds `waitForCollectionCallback` as a param to `Onyx.connect` to trigger a single callback when connecting to a collection

cc @tgolen I found that a `connectToCollection` method would share most of its logic with `connect` so I added a param instead of creating a new method.

### Related Issues
https://github.com/Expensify/App/pull/10051#discussion_r932712218

### Automated Tests
Added to `tests/unit/onyxTest.js`

### Linked PRs
https://github.com/Expensify/App/pull/10051

PR #167, opened 2022-08-08
Add waitForCollectionCallback to keysChanged

luacmartins:
cc @tgolen 

### Details
In https://github.com/Expensify/react-native-onyx/pull/163, we added a new param `waitForCollectionCallback` to `Onyx.connect` that waits for the collection to be read and triggers the callback only once. However, we forgot to add that param to `keysChanged` which means that the callback is being triggered for each key when we update the collection.

This PR fixes that and ensures that the callback is called only once when we update the collection.

### Related Issues
https://github.com/Expensify/react-native-onyx/pull/163

### Automated Tests
Added in `onyxTests`

### Linked PRs
https://github.com/Expensify/App/pull/10051

### DIFF
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -260,6 +260,12 @@ function keysChanged(collectionKey, collection) {
         if (isSubscribedToCollectionKey) {
             if (_.isFunction(subscriber.callback)) {
                 const cachedCollection = getCachedCollection(collectionKey);
+
+                if (subscriber.waitForCollectionCallback) {
+                    subscriber.callback(cachedCollection);
+                    return;
+                }
+
                 _.each(collection, (data, dataKey) => {
                     subscriber.callback(cachedCollection[dataKey], dataKey);
                 });
--- a/tests/unit/onyxTest.js
+++ b/tests/unit/onyxTest.js
@@ -489,12 +489,6 @@ describe('Onyx', () => {
             },
         };
 
-        // AND an API response for that collection
-        const apiResponse = {
-            ...initialCollectionData,
-            test_connect_collection_4: {ID: 123, value: 'four'},
-        };
-
         Onyx.mergeCollection(ONYX_KEYS.COLLECTION.TEST_CONNECT_COLLECTION, initialCollectionData);
         return waitForPromisesToResolve()
             .then(() => {
@@ -510,17 +504,34 @@ describe('Onyx', () => {
                 // THEN we expect the callback to be called only once and the initial stored value to be initialCollectionData
                 expect(mockCallback.mock.calls.length).toBe(1);
                 expect(mockCallback.mock.calls[0][0]).toEqual(initialCollectionData);
+            });
+    });
 
+    it('should return all collection keys as a single object when updating a collection key with waitForCollectionCallback = true', () => {
+        const valuesReceived = {};
+        const mockCallback = jest.fn(data => valuesReceived[data.ID] = data.value);
+        const collectionUpdate = {
+            test_connect_collection_3: {ID: 234, value: 'three'},
+            test_connect_collection_4: {ID: 123, value: 'four'},
+        };
+
+        // GIVEN an Onyx.connect call with waitForCollectionCallback=true
+        Onyx.connect({
+            key: ONYX_KEYS.COLLECTION.TEST_CONNECT_COLLECTION,
+            waitForCollectionCallback: true,
+            callback: mockCallback,
+        });
+        return waitForPromisesToResolve()
+            .then(() => {
                 // WHEN we update the collection, e.g. the API returns a response
-                Onyx.mergeCollection(ONYX_KEYS.COLLECTION.TEST_CONNECT_COLLECTION, apiResponse);
+                Onyx.mergeCollection(ONYX_KEYS.COLLECTION.TEST_CONNECT_COLLECTION, collectionUpdate);
                 return waitForPromisesToResolve();
             })
             .then(() => {
-                // THEN we expect the callback to have called twice, once for the initial connect call and once for the API response
+                // THEN we expect the callback to have called twice (once for the initial connect call and once for the collection update)
+                // AND the value should be equal to collectionUpdate
                 expect(mockCallback.mock.calls.length).toBe(2);
-
-                // AND all the updated collection data should be returned as a single object
-                expect(mockCallback.mock.calls[1][0]).toEqual(apiResponse);
+                expect(mockCallback.mock.calls[1][0]).toEqual(collectionUpdate);
             });
     });
 });
--- a/tests/unit/onyxTest.js
+++ b/tests/unit/onyxTest.js
@@ -7,6 +7,7 @@ const ONYX_KEYS = {
     COLLECTION: {
         TEST_KEY: 'test_',
         TEST_CONNECT_COLLECTION: 'test_connect_collection_',
+        TEST_POLICY: 'test_policy_',
     },
 };
 
@@ -470,8 +471,7 @@ describe('Onyx', () => {
     });
 
     it('should return all collection keys as a single object when waitForCollectionCallback = true', () => {
-        const valuesReceived = {};
-        const mockCallback = jest.fn(data => valuesReceived[data.ID] = data.value);
+        const mockCallback = jest.fn();
 
         // GIVEN some initial collection data
         const initialCollectionData = {
@@ -493,7 +493,7 @@ describe('Onyx', () => {
         return waitForPromisesToResolve()
             .then(() => {
                 // WHEN we connect to that collection with waitForCollectionCallback = true
-                Onyx.connect({
+                connectionID = Onyx.connect({
                     key: ONYX_KEYS.COLLECTION.TEST_CONNECT_COLLECTION,
                     waitForCollectionCallback: true,
                     callback: mockCallback,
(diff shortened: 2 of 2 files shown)


tgolen (changes requested):
Hm, maybe this test can be split up better for testing, because it's becoming a bit unclear to me what it's doing. Each test should have a very clear GIVEN/WHEN/THEN, but right now this has GIVEN/AND/WHEN/THEN/WHEN/THEN/AND 😅 

Could you try making the test more clear to test each case independently?

luacmartins:
Sure thing! I split the test into two, one to test `Onyx.connect` with initial data and one to test updating a collection after we are connected to it!Ready for another review!

lib/Onyx.js
```diff
@@ -260,6 +260,12 @@ function keysChanged(collectionKey, collection) {
...
                 const cachedCollection = getCachedCollection(collectionKey);
+
+                if (subscriber.waitForCollectionCallback) {
+                    subscriber.callback(cachedCollection);
+                    return;
+                }
+
                 _.each(collection, (data, dataKey) => {
                     subscriber.callback(cachedCollection[dataKey], dataKey);
                 });
```
aldo-expensify, line 271: `else` this stuff? otherwise your will call the callback with the collection and with each separate key... unless that is what you wanted :P

aldo-expensify (changes requested):

aldo-expensify:
lol, ignore my first deleted comment, I missed the `return`

tests/unit/onyxTest.js
```diff
@@ -488,23 +488,50 @@ describe('Onyx', () => {
...
                     waitForCollectionCallback: true,
                     callback: mockCallback,
                 });
                 return waitForPromisesToResolve();
             })
             .then(() => {
-                // THEN the callback should be triggered only once
+                // THEN we expect the callback to be called only once and the initial stored value to be initialCollectionData
                 expect(mockCallback.mock.calls.length).toBe(1);
+                expect(mockCallback.mock.calls[0][0]).toEqual(initialCollectionData);
```
aldo-expensify, line 506: Should we have a version of this test where you connect to a collection that has not been initialized? what would you have in `mockCallback.mock.calls[0][0]` in such case? In the test below this happens... but we don't check the value in `mockCallback.mock.calls[0][0]`

tests/unit/onyxTest.js
```diff
@@ -488,23 +488,50 @@ describe('Onyx', () => {
...
+            });
+    });
+
+    it('should return all collection keys as a single object when updating a collection key with waitForCollectionCallback = true', () => {
+        const valuesReceived = {};
+        const mockCallback = jest.fn(data => valuesReceived[data.ID] = data.value);
```
aldo-expensify, line 512: Why do we keep `valuesReceived`? this made sense in some tests before because we actually checked the values there... but since you are using `mockCallback.mock.calls[1][0]` to access this data... I guess we could just simplify to `const mockCallback = jest.fn();`

aldo-expensify:
Left 2 small comments, overall looks good!

tests/unit/onyxTest.js
luacmartins: That makes sense! I'll remove `valuesReceived`. Thanks!

tests/unit/onyxTest.js
luacmartins: The 2nd test does this and I think the expected value would be the object below. That's because we set those keys in the 1st test and we clear them (set the values to null, but the keys still exist) [after each test](https://github.com/Expensify/react-native-onyx/blob/b348a7d7b448c02d6994f31feccb208db4049070/tests/unit/onyxTest.js#L26). 

```
{"test_connect_collection_1": null, "test_connect_collection_2": null, "test_connect_collection_3": null}
```

We can have it be null if we use a new collection key though

luacmartins: I updated the tests with a new collection key, so we can check for a null value easily. Let me know if you had something different in mind!

aldo-expensify: yes, that sounds good... I think it is bad to reuse keys between different tests, it may cause some bad side effect / dependency between them

luacmartins: Cool! Updated!

luacmartins:
Updated!

aldo-expensify (approved):
I'll leave the merge to @tgolen so he can review

tests/unit/onyxTest.js
```diff
@@ -500,11 +501,40 @@ describe('Onyx', () => {
...
+
+        // GIVEN an Onyx.connect call with waitForCollectionCallback=true
+        connectionID = Onyx.connect({
+            key: ONYX_KEYS.COLLECTION.TEST_POLICY,
+            waitForCollectionCallback: true,
+            callback: mockCallback,
+        });
+        return waitForPromisesToResolve()
+            .then(() => {
+                // WHEN we update the collection, e.g. the API returns a response
```
tgolen, line 525:
```suggestion
                // WHEN mergeCollection is called with an updated collection
```

tests/unit/onyxTest.js
```diff
@@ -500,11 +501,40 @@ describe('Onyx', () => {
...
+                Onyx.mergeCollection(ONYX_KEYS.COLLECTION.TEST_POLICY, collectionUpdate);
+                return waitForPromisesToResolve();
+            })
+            .then(() => {
+                // THEN we expect the callback to have called twice, once for the initial connect call + once for the collection update
+                expect(mockCallback.mock.calls.length).toBe(2);
+
+                // AND the value for the first call should be null since the collection was not initialized at that point
```
tgolen, line 533: Does it seem weird to anyone else that:

1. There would be two calls
2. The first call passes `null`
3. The second call is really what we want with `waitForCollectionCallback=true`

?

Is it possible for us to remove that first call? What is the value in having the call triggered with `null` before the collection has been "initialized"?

tgolen (changes requested):
Those tests looks great now! Thank you

I wanted to question some assumptions, and my disclaimer is, it's been a long time since I've worked on that connect() code, so I don't fully remember what's feasible or not.

tests/unit/onyxTest.js
luacmartins: I think this is expected since we want to initialize that key's value to `null`, since we are connecting to that key, we just don't have a value for it. We leave it up to the subscriber to decide what to do with a null value. 

This happens [here](https://github.com/Expensify/react-native-onyx/blob/2eb3d74e7b581007d7c4dcc332b5bbc027145ce2/lib/Onyx.js#L440-L444) before we run any callback subscriber logic. I think it would make sense to keep the behavior consistent.

aldo-expensify: +1 to what @luacmartins said :), I would also prefer to avoid exceptions and let the subscriber handle that. What if `null` is a valid value and we want to be notified about it?

tgolen: Yeah, those are good points. I think it would be a change in behavior that could make other bugs in the process, so this was more of an exploratory question. Let's leave it as-is!

tgolen (approved):

marcaaron:
>In https://github.com/Expensify/react-native-onyx/pull/163, we added a new param waitForCollectionCallback to Onyx.connect that waits for the collection to be read and triggers the callback only once

I'm kind of confused about this decision though I understand the need that it fills and appreciate the improvement.

It almost feels like this should be the default behavior based on how subscribing to a collection key works when there is a `withOnyx` instance.

Just curious if anyone agrees that it would be more consistent to have something like:

```js
Onyx.connect({key: ONYXKEYS.COLLECTION.REPORT}, callback: (allReports) => {...})
```

because that's what this code would do:

```js
withOnyx({
    allReports: {key: ONYXKEYS.COLLECTION.REPORT},
})(MyComponent);
```

And if we want individual member values returned then we should do something like

```js
Onyx.connectToCollectionMembers({key: ONYXKEYS.COLLECTION.REPORT, callback: (value, key) => {...}});
```

Last, I'm just not really sure about the name `waitForCollectionCallback`. What are we waiting for exactly? 🤔

marcaaron:
Also unsure if this is known (and apologies in advance if I'm wrong about this) but the data returned changes based on whether `mergeCollection` or `merge` is used...

`keyChanged()` does this and won't look at the optional flag:

https://github.com/Expensify/react-native-onyx/blob/1f58aa5e2e649b82fe2d3604951f4a343f336f86/lib/Onyx.js#L330-L332
But `keysChanged()` (triggered when `mergeCollection` is used) will always return the collection

https://github.com/Expensify/react-native-onyx/blob/1f58aa5e2e649b82fe2d3604951f4a343f336f86/lib/Onyx.js#L260-L267

Kind of unpredictable, no? How do you know which method is used and what to expect in the callback?

tgolen:
> What are we waiting for exactly?

For the entire collection to be read out of Onyx into memory.

@marcaaron Can you clarify what you are proposing? It sounds like you're implying some kind of proposal (which looks good to me so far), but I'm just not clear what you're suggesting.

That's a good point about merge vs mergeCollection.

marcaaron:
I'm proposing that we make the behavior of `Onyx.connect()` and `withOnyx` for "collections" the same i.e. when you connect directly to a connection key it should return the entire updated collection when:

- The initial connection happens
- Any of the collection member keys update

That would be more consistent and not require us to use a flag like `waitForCollectionCallback` which doesn't describe the behavior that it enables well (when and why should we "wait"? what exactly is the "collection callback"?). This is the default behavior for `withOnyx`, but we have to opt-in when using `Onyx.connect()`.

tgolen:
Ah, got it. That makes sense to me 👍

luacmartins:
I agree with your proposed solution @marcaaron! Let's do it!

PR #171, opened 2022-08-11
Add additional documentation for handling collections. Fix behavior of `waitForCollectionCallback`.

marcaaron:
### Details

Fixes issue found here: https://github.com/Expensify/App/pull/10051#issuecomment-1211421037

Semi-related to: https://github.com/Expensify/Expensify/issues/222064

### Automated Tests

I can add some but the ones that exist are passing.

### Linked PRs

TODO

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,6 +137,70 @@ export default withOnyx({
 
 It is preferable to use the HOC over `Onyx.connect()` in React code as `withOnyx()` will delay the rendering of the wrapped component until all keys have been accessed and made available.
 
+## Collections
+
+Collections allow keys with similar value types to be subscribed together by subscribing to the collection key. To define one, it must be included in the `ONYXKEYS.COLLECTION` object and it must be suffixed with an underscore. Member keys should use a unique identifier or index after the collection key prefix (e.g. `chat_42`).
+
+```javascript
+const ONYXKEYS = {
+    COLLECTION: {
+        CHAT: 'chat_',
+    },
+};
+```
+
+### Setting Collection Values
+
+To save a new collection key we can either do:
+
+```js
+Onyx.merge(`${ONYXKEYS.COLLECTION.CHAT}${chat1.id}`, chat1);
+```
+
+or we can set many at once with `mergeCollection()`:
+
+```js
+Onyx.mergeCollection(ONYXKEYS.COLLECTION.CHAT, {
+    [`${ONYXKEYS.COLLECTION.CHAT}${chat1.id}`]: chat1,
+    [`${ONYXKEYS.COLLECTION.CHAT}${chat2.id}`]: chat2,
+    [`${ONYXKEYS.COLLECTION.CHAT}${chat3.id}`]: chat3,
+});
+```
+
+### Subscribing to Collections
+
+There are several ways to subscribe to these keys:
+
+```javascript
+withOnyx({
+    allChats: {key: ONYXKEYS.COLLECTION.CHAT},
+})(MyComponent);
+```
+
+This will return the initial collection as an object of collection member key/values. Changes to the individual member keys will modify the entire object and new props will be passed with each individual key update.
+
+```js
+Onyx.connect({key: ONYXKEYS.COLLECTION.CHAT}, callback: (memberValue, memberKey) => {...}});
+```
+
+This will fire the callback `n` times depending on how many collection member keys have been stored. Changes to those keys after the initial callbacks fire will occur when each individual key is updated.
+
+```js
+Onyx.connect({
+    key: ONYXKEYS.COLLECTION.CHAT,
+    waitForCollectionCallback: true,
+    callback: (allChats) => {...}},
+});
+```
+
+This final option forces `Onyx.connect()` to behave more like `withOnyx()` and only update the callback once with the entire collection initially and later with an updated version of the collection when individual keys update.
+
+### Performance Considerations When Using Collections
+
+Be cautious when using collections as things can get out of hand if you have a subscriber hooked up to a collection key that has large numbers of individual keys. If this is the case, it is critical to use `mergeCollection()` over `merge()`.
+
+Remember, `mergeCollection()` will notify a subscriber only *once* with the total collected values whereas each call to `Onyx.merge()` would re-render a connected component `n` times per key that you call it on.
+
 ## Clean up
 
 To clear all data from `Onyx` we can use `Onyx.clear()`.
--- a/README.md
+++ b/README.md
@@ -139,12 +139,12 @@ It is preferable to use the HOC over `Onyx.connect()` in React code as `withOnyx
 
 ## Collections
 
-Collections allow keys with similar value types to be subscribed together by subscribing to the collection key. To define one, it must be included in the `ONYXKEYS.COLLECTION` object and it must be suffixed with an underscore. Member keys should use a unique identifier or index after the collection key prefix (e.g. `chat_42`).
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -137,6 +137,70 @@ export default withOnyx({
 
 It is preferable to use the HOC over `Onyx.connect()` in React code as `withOnyx()` will delay the rendering of the wrapped component until all keys have been accessed and made available.
 
+## Collections
+
+Collections allow keys with similar value types to be subscribed together by subscribing to the collection key. To define one, it must be included in the `ONYXKEYS.COLLECTION` object and it must be suffixed with an underscore. Member keys should use a unique identifier or index after the collection key prefix (e.g. `chat_42`).
```
luacmartins, line 142: NAB but should we use an existing key as an example? That would make it easier for a contributor to search for examples in the code base and see how collections are used.

```suggestion
Collections allow keys with similar value types to be subscribed together by subscribing to the collection key. To define one, it must be included in the `ONYXKEYS.COLLECTION` object and it must be suffixed with an underscore. Member keys should use a unique identifier or index after the collection key prefix (e.g. `report_42`).
```

README.md
```diff
@@ -137,6 +137,70 @@ export default withOnyx({
...
+})(MyComponent);
+```
+
+This will return the initial collection as an object of collection member key/values. Changes to the individual member keys will modify the entire object and new props will be passed with each individual key update.
+
+```js
+Onyx.connect({key: ONYXKEYS.COLLECTION.CHAT}, callback: (memberValue, memberKey) => {...}});
+```
+
+This will fire the callback `n` times depending on how many collection member keys have been stored. Changes to those keys after the initial callbacks fire will occur when each individual key is updated.
```
dangrous, line 186: Where is `n` defined? I don't see it in the code above. Is it once per member key? Either way, might be good to clarify here. (Coming as someone with no context on this whatsoever.)

luacmartins (dismissed):
LGTM left a small NAB. Thanks for fixing the bug and adding more docs on collections!

README.md
```diff
@@ -137,6 +137,70 @@ export default withOnyx({
...
+});
+```
+
+This final option forces `Onyx.connect()` to behave more like `withOnyx()` and only update the callback once with the entire collection initially and later with an updated version of the collection when individual keys update.
+
+### Performance Considerations When Using Collections
+
+Be cautious when using collections as things can get out of hand if you have a subscriber hooked up to a collection key that has large numbers of individual keys. If this is the case, it is critical to use `mergeCollection()` over `merge()`.
+
+Remember, `mergeCollection()` will notify a subscriber only *once* with the total collected values whereas each call to `Onyx.merge()` would re-render a connected component `n` times per key that you call it on.
```
dangrous, line 202: Same thought as above - where is `n` coming from? If this is a silly comment, feel free to ignore me.

dangrous (commented):
Knowing very little about the context of this yet, this seems well written and understandable. Added a couple of places that could use clarification.

As for the actual content of it, I'll defer to someone who knows what they're talking about.

README.md
marcaaron: Ah so, my intention here was to try to keep this generic so that people could imagine using it for whatever their use case and not get thrown off by "expensify" language. But I think the point gets across well enough here either way though so I'll update it.

README.md
marcaaron: Hey, if you didn't get it the first time then the docs could be better! So, not silly at all 😄

README.md
luacmartins: Makes sense! Thanks for the changes

README.md
marcaaron: Added an example to help clarify.

dangrous:
Updates look great to me! Resolves my questions perfectly.

luacmartins (approved):

tgolen:
I was sort of holding off on reviewing this since @marcaaron and I were still having an ongoing discussion about whether a parameter is better or a separate method is better. Also, we should consider changing this param/method to be the default behavior for collections.

@marcaaron Did you want to make some of those changes in this PR?

marcaaron:
We can keep discussing and I can make them in a follow up (or just not at all). This PR largely just improves documentation and fixes a bug on `main` (and eventually `staging`).

README.md
```diff
@@ -137,6 +137,80 @@ export default withOnyx({
...
+
+### Setting Collection Values
+
+To save a new collection key we can either do:
+
+```js
+Onyx.merge(`${ONYXKEYS.COLLECTION.REPORT}${report1.reportID}`, report1);
+```
+
+or we can set many at once with `mergeCollection()`:
```
tgolen, line 160: Can you please add a little more guidance here so people know which one to choose? I think adding information to say that when there are multiple keys, you should always use mergeCollect() because it optimizes all the UI updates. If it's a single key, it's fine to use either.

README.md
```diff
@@ -137,6 +137,80 @@ export default withOnyx({
...
+
+There are several ways to subscribe to these keys:
+
+```javascript
+withOnyx({
+    allReports: {key: ONYXKEYS.COLLECTION.REPORT},
+})(MyComponent);
+```
+
+This will return the initial collection as an object of collection member key/values. Changes to the individual member keys will modify the entire object and new props will be passed with each individual key update.
```
tgolen, line 180:
```suggestion
This will add a prop to the component called `allReports` which is an object of collection member key/values. Changes to the individual member keys will modify the entire object and new props will be passed with each individual key update. The prop doesn't update on the initial rendering of the component until the entire collection has been read out of Onyx.
```

README.md
tgolen: Ah, I see you added all that information below. Maybe just add something here that says "See below for guidance on best practices"

README.md
```diff
@@ -137,6 +137,80 @@ export default withOnyx({
...
+
+### Performance Considerations When Using Collections
+
+Be cautious when using collections as things can get out of hand if you have a subscriber hooked up to a collection key that has large numbers of individual keys. If this is the case, it is critical to use `mergeCollection()` over `merge()`.
+
+Remember, `mergeCollection()` will notify a subscriber only *once* with the total collected values whereas each call to `merge()` would re-render a connected component *each time it is called*. Consider this example where `reports` is an array of reports that we want to index and save.
+
+```js
+// Bad
+_.each(reports, report => Onyx.merge(`${ONYXKEYS.COLLECTION.REPORT}${report.reportID}`, report)); // -> Connected component will update on each iteration
```
tgolen, line 206:
```suggestion
_.each(reports, report => Onyx.merge(`${ONYXKEYS.COLLECTION.REPORT}${report.reportID}`, report)); // -> A component using withOnyx() will have it's state updated with each iteration
```

README.md
```diff
@@ -137,6 +137,80 @@ export default withOnyx({
...
+
+// Good
+const values = {};
+_.each(reports, report => values[`${ONYXKEYS.COLLECTION.REPORT}${report.reportID}`] = report);
+Onyx.mergeCollection(ONYXKEYS.COLLECTION.REPORT, values); // -> Connected component will update just once
```
tgolen, line 211:
```suggestion
Onyx.mergeCollection(ONYXKEYS.COLLECTION.REPORT, values); // -> A component using withOnyx() will only have it's state updated once
```

tgolen (changes requested):

README.md
marcaaron: Nice!

marcaaron:
Updated. Thanks for the notes @tgolen 🙇

tgolen (approved):

PR #173, opened 2022-09-12
Change behavior of merging arrays 

iwiznia:
Please advise on what's the best way to test this in the app... I assume just playing around with it and making sure things work properly?

Also, there's 2 other `lodashMerge` calls in the Onyx file, but I am not sure if those need to change or not.

### Details
Context https://expensify.slack.com/archives/C03TQ48KC/p1662572468181579

### Related Issues

https://github.com/Expensify/Expensify/issues/211560

### Automated Tests
I've added 2 new tests for this behavior

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ Onyx.merge(ONYXKEYS.POLICY, {id: 1}); // -> {id: 1}
 Onyx.merge(ONYXKEYS.POLICY, {name: 'My Workspace'}); // -> {id: 1, name: 'My Workspace'}
 ```
 
-Arrays inside objects will be replaced fully:
+Arrays inside objects will NOT be concatenated and instead will be replaced fully:
 
 ```javascript
 Onyx.merge(ONYXKEYS.POLICY, {employeeList: ['Joe', 'Jack']}); // -> {employeeList: ['Joe', 'Jack']}
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ The data will then be cached and stored via [`AsyncStorage`](https://github.com/
 
 We can also use `Onyx.merge()` to merge new `Object` or `Array` data in with existing data.
 
-For `Array` the default behavior is to concatenate new items.
+For `Array` the default behavior is to concatenate replace it fully, effectively making it equivalent to set:
 
 ```javascript
 Onyx.merge(ONYXKEYS.EMPLOYEE_LIST, ['Joe']); // -> ['Joe']
@@ -77,7 +77,7 @@ Onyx.merge(ONYXKEYS.POLICY, {id: 1}); // -> {id: 1}
 Onyx.merge(ONYXKEYS.POLICY, {name: 'My Workspace'}); // -> {id: 1, name: 'My Workspace'}
 ```
 
-Arrays inside objects will NOT be concatenated and instead will be replaced fully:
+Arrays inside objects will NOT be concatenated and instead will be replaced fully, same as arrays not inside objects:
 
 ```javascript
 Onyx.merge(ONYXKEYS.POLICY, {employeeList: ['Joe', 'Jack']}); // -> {employeeList: ['Joe', 'Jack']}
--- a/README.md
+++ b/README.md
@@ -77,10 +77,10 @@ Onyx.merge(ONYXKEYS.POLICY, {id: 1}); // -> {id: 1}
 Onyx.merge(ONYXKEYS.POLICY, {name: 'My Workspace'}); // -> {id: 1, name: 'My Workspace'}
 ```
 
-One caveat to be aware of is that `lodash/merge` [follows the behavior of jQuery's deep extend](https://github.com/lodash/lodash/issues/2872) and will not concatenate nested arrays in objects. It might seem like this code would concat these arrays, but it does not.
+Arrays inside objects will be replaced fully:
 
 ```javascript
-Onyx.merge(ONYXKEYS.POLICY, {employeeList: ['Joe']}); // -> {employeeList: ['Joe']}
+Onyx.merge(ONYXKEYS.POLICY, {employeeList: ['Joe', 'Jack']}); // -> {employeeList: ['Joe', 'Jack']}
 Onyx.merge(ONYXKEYS.POLICY, {employeeList: ['Jack']}); // -> {employeeList: ['Jack']}
 ```
 
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -7,7 +7,7 @@ import Storage from './storage';
 import * as Logger from './Logger';
 import cache from './OnyxCache';
 import createDeferredTask from './createDeferredTask';
-import customizerForMergeWith from "./customizerForMergeWith";
+import customizerForMergeWith from './customizerForMergeWith';
 
 // Keeps track of the last connectionID that was used so we can keep incrementing it
 let lastConnectionID = 0;
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -1,6 +1,7 @@
 import _ from 'underscore';
 import Str from 'expensify-common/lib/str';
 import lodashMerge from 'lodash/merge';
+import lodashMergeWith from 'lodash/mergeWith';
 import lodashGet from 'lodash/get';
 import Storage from './storage';
 import * as Logger from './Logger';
@@ -636,6 +637,23 @@ function hasPendingMergeForKey(key) {
 }
 
 /**
+ * When merging 2 objects into onyx that contain an array, we want to completely replace the array instead of the default
+ * behavior which is to merge each item by its index.
+ * ie:
+ * merge({a: [1]}, {a: [2,3]}):
+ *  - In the default implementation would produce {a:[1,3]}
+ *  - With this function would produce: {a: [2,3]}
+ * @param {*} objValue
+ * @param {*} srcValue
+ * @return {*}
+ */
+function customizerForMergeWith(objValue, srcValue) {
+    if (_.isArray(objValue)) {
+        return srcValue;
+    }
+}
+
+/**
  * Given an Onyx key and value this method will combine all queued
  * value updates and return a single value. Merge attempts are
  * batched. They must occur after a single call to get() so we
@@ -661,7 +679,7 @@ function applyMerge(key, data) {
     if (_.isObject(data) || _.every(mergeValues, _.isObject)) {
         // Object values are merged one after the other
         return _.reduce(mergeValues, (modifiedData, mergeValue) => {
-            const newData = lodashMerge({}, modifiedData, mergeValue);
+            const newData = lodashMergeWith({}, modifiedData, mergeValue, customizerForMergeWith);
 
             // We will also delete any object keys that are undefined or null.
             // Deleting keys is not supported by AsyncStorage so we do it this way.
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -647,6 +647,7 @@ function hasPendingMergeForKey(key) {
  * @param {*} srcValue
  * @return {*}
  */
+// eslint-disable-next-line rulesdir/prefer-early-return
 function customizerForMergeWith(objValue, srcValue) {
     if (_.isArray(objValue)) {
         return srcValue;
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -734,7 +734,7 @@ function initializeWithDefaultKeyStates() {
         .then((pairs) => {
             const asObject = _.object(pairs);
 
-            const merged = lodashMerge(asObject, defaultKeyStates);
+            const merged = lodashMergeWith(asObject, defaultKeyStates, customizerForMergeWith);
             cache.merge(merged);
             _.each(merged, (val, key) => keyChanged(key, val));
         });
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -7,6 +7,7 @@ import Storage from './storage';
 import * as Logger from './Logger';
 import cache from './OnyxCache';
 import createDeferredTask from './createDeferredTask';
+import customizerForMergeWith from "./customizerForMergeWith";
 
 // Keeps track of the last connectionID that was used so we can keep incrementing it
 let lastConnectionID = 0;
@@ -634,24 +635,6 @@ const mergeQueue = {};
  */
 function hasPendingMergeForKey(key) {
     return Boolean(mergeQueue[key]);
-}
-
-/**
- * When merging 2 objects into onyx that contain an array, we want to completely replace the array instead of the default
- * behavior which is to merge each item by its index.
- * ie:
- * merge({a: [1]}, {a: [2,3]}):
- *  - In the default implementation would produce {a:[1,3]}
- *  - With this function would produce: {a: [2,3]}
- * @param {*} objValue
- * @param {*} srcValue
- * @return {*}
- */
-// eslint-disable-next-line rulesdir/prefer-early-return
-function customizerForMergeWith(objValue, srcValue) {
-    if (_.isArray(objValue)) {
-        return srcValue;
-    }
 }
 
 /**
--- a/lib/OnyxCache.js
+++ b/lib/OnyxCache.js
@@ -1,5 +1,6 @@
 import _ from 'underscore';
 import lodashMergeWith from 'lodash/mergeWith';
+import customizerForMergeWith from "./customizerForMergeWith";
 
 const isDefined = _.negate(_.isUndefined);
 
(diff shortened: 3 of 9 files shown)


lib/Onyx.js
```diff
@@ -635,6 +636,24 @@ function hasPendingMergeForKey(key) {
...
+ * @param {*} objValue
+ * @param {*} srcValue
+ * @return {*}
+ */
+// eslint-disable-next-line rulesdir/prefer-early-return
+function customizerForMergeWith(objValue, srcValue) {
+    if (_.isArray(objValue)) {
+        return srcValue;
+    }
+}
```
marcaaron, line 655: NAB, I'm new to `mergeWith()`. Any idea if this customizer work only on top level array values or is it deep? What would be the expected behavior for something like this:

```js
const existing = {a: {arr: [1, 2, 3]}};
const new = {a: {arr: [4, 5, 6]}};
```

README.md
```diff
@@ -77,10 +77,10 @@ Onyx.merge(ONYXKEYS.POLICY, {id: 1}); // -> {id: 1}
 Onyx.merge(ONYXKEYS.POLICY, {name: 'My Workspace'}); // -> {id: 1, name: 'My Workspace'}
 ```
 
-One caveat to be aware of is that `lodash/merge` [follows the behavior of jQuery's deep extend](https://github.com/lodash/lodash/issues/2872) and will not concatenate nested arrays in objects. It might seem like this code would concat these arrays, but it does not.
```
marcaaron, line 80: Maybe it is still useful to clarify that it will not concatenate arrays? The default behavior for merging an array (not object with array value) is to concat those values, but this is not the same for object's with array values.

iwiznia:

lib/Onyx.js
iwiznia: It is used on all levels, so the example above works as you expect (final result would be `{a: {arr: [4,5,6]}}`)

marcaaron: Nice. Much more predictable.

marcaaron (approved):
Changes look good. Unsure if we are concerned with this or not, but it crossed my mind that there could be some _possible_ inconsistency between places where we use `lodash.merge()` elsewhere in `react-native-onyx` (though haven’t confirmed it or tested)

**Default keys:** 

https://github.com/Expensify/react-native-onyx/blob/11119454c4c5340e0413ba0b9294d77aec2593e3/lib/Onyx.js#L754-L755

**Unsure about what the behavior is in `mergeCollection()`:**

`Storage.multiMerge()` - does it overwrite array values or not? And is it the same on all platforms? 🤔 
https://github.com/Expensify/react-native-onyx/blob/11119454c4c5340e0413ba0b9294d77aec2593e3/lib/Onyx.js#L835

And then the cache is updated with the entire collection:
https://github.com/Expensify/react-native-onyx/blob/11119454c4c5340e0413ba0b9294d77aec2593e3/lib/Onyx.js#L845

But will `cache.merge()` have the previous behavior...?
https://github.com/Expensify/react-native-onyx/blob/11119454c4c5340e0413ba0b9294d77aec2593e3/lib/OnyxCache.js#L113

**Merging the `withOnyx` state here seems like it would have the previous behavior when applying a new collection to one that is already in state:**

https://github.com/Expensify/react-native-onyx/blob/11119454c4c5340e0413ba0b9294d77aec2593e3/lib/Onyx.js#L285

I am not sure how deeply we want to look into these so if the current changes solve the issues then I am 👍 (but wanted to note the possibility that the weird array merging behavior could still happen in other places - but might not apply).

iwiznia:
> Default keys:

Yep, this is one of the ones I mentioned in the description.

> Unsure about what the behavior is in mergeCollection():

It works as expected, I added a test for it.

> Storage.multiMerge() - does it overwrite array values or not? And is it the same on all platforms? 🤔

hmmmm, not sure, but if I am re-reading the values and they are correct, I think the answer is yes? Or it could be different since the data might be coming from the cache?

> And then the cache is updated with the entire collection:
> But will cache.merge() have the previous behavior...?

hmmmm good question, but same as above, the test should be covering this? Or do the tests I modified don't use the cache?

> Merging the withOnyx state here seems like it would have the previous behavior when applying a new collection to one that is already in state:

hmmm I think you are right, will add a test for withOnyx

> I am not sure how deeply we want to look into these so if the current changes solve the issues then I am 👍 (but wanted to note the possibility that the weird array merging behavior could still happen in other places - but might not apply).

These are good comments and exactly what I wanted from the review, so will take a look at them, since we don't want weird behavior in this low level, as it can be hard to track the bug (if there is one).

iwiznia:
> hmmmm, not sure, but if I am re-reading the values and they are correct, I think the answer is yes? Or it could be different since the data might be coming from the cache?

Not sure what's going on in the other tests, but added a test for storage and it is failing and I don't see an easy way to update its behavior 😢

iwiznia:
> Default keys:

This one I don't know if we need to update or not.

> Merging the withOnyx state here seems like it would have the previous behavior when applying a new collection to one that is already in state:

I added tests and they are passing, not sure how/why that is not needed there.. maybe my tests are no good?

Additionally, I am struggling to make the cache work in the same way for top level arrays and I am thinking of just removing that feature from Onyx altogether. 
I would have to refactor places we rely on that like ACTIVE_CLIENTS and not sure if some other place (I looked at betas and I think that's fine as it is, since we always send all betas).

iwiznia:
ok, changed the behavior for merging arrays. Still missing these 2 doubts:
- Do I need to update the merge in Default keys?
- I am not sure if [this](https://github.com/Expensify/react-native-onyx/blob/11119454c4c5340e0413ba0b9294d77aec2593e3/lib/Onyx.js#L285) needs to be changed, all seems to work without changing it and I can't seem to write a test that does not work as expected

marcaaron:
>Do I need to update the merge in Default keys?

I think in theory it would be nice if it worked this way. But in practice it doesn't make any difference for us as we have no array values anywhere in any of this data...

https://github.com/Expensify/App/blob/68c50096675e02882e639107215e0920cdcbb77e/src/setup/index.js#L27-L37

>I am not sure if [this](https://github.com/Expensify/react-native-onyx/blob/11119454c4c5340e0413ba0b9294d77aec2593e3/lib/Onyx.js#L285) needs to be changed, all seems to work without changing it and I can't seem to write a test that does not work as expected

Try this:

```js
    it('mergeCollection() will replace array values on objects', () => {
        const mockCallback = jest.fn();
        const originalCollection = {
            test_policy_1: {arrayValues: [{test_a: 'a'}, {test_b: 'b'}]},
        };
        const modifiedCollection = {
            test_policy_1: {arrayValues: [{test_c: 'c'}, {test_d: 'd'}]},
        };

        connectionID = Onyx.connect({
            key: ONYX_KEYS.COLLECTION.TEST_POLICY,
            callback: mockCallback,
        });

            .then(() => {
                // WHEN mergeCollection is called with an updated collection
                Onyx.mergeCollection(ONYX_KEYS.COLLECTION.TEST_POLICY, originalCollection);
            })
            .then(() => {
                Onyx.mergeCollection(ONYX_KEYS.COLLECTION.TEST_POLICY, modifiedCollection);
            })
            .then(() => {
                // THEN we expect the callback to have called three times, once for the initial connect call + once for the collection update
                expect(mockCallback.mock.calls.length).toBe(3);

                // AND the value for the first call should not match anything since the collection was not initialized at that point
                expect(mockCallback.mock.calls[0][0]).not.toBe(originalCollection);

                // AND the value for the second call should be originalCollection since the collection was updated
                expect(mockCallback.mock.calls[1][0]).toEqual(originalCollection);

                // AND the value for the third call should be modifiedCollection since the collection was updated and it should not be
                // some weird frankenstein merge that nobody wants
                expect(mockCallback.mock.calls[2][0]).toEqual(modifiedCollection);
            });
    });
```

tests/unit/withOnyxTest.js
```diff
@@ -101,6 +101,34 @@ describe('withOnyx', () => {
...
+                return waitForPromisesToResolve();
+            })
+            .then(() => {
+                Onyx.mergeCollection(ONYX_KEYS.COLLECTION.TEST_KEY, {
+                    test_1: {list: [7]},
+                });
+                return waitForPromisesToResolve();
+            })
+            .then(() => {
+                expect(onRender.mock.calls.length).toBe(3);
```
iwiznia, line 127: OH! I also had doubts about this. Not sure why it is 3 times, I was kind of expecting only 2. Here's what `onRender.mock.calls` shows:
```
      [
        {
          onRender: [Function: mockConstructor] {
            _isMockFunction: true,
          },
          collections: {},
          testObject: { isDefaultProp: true }
        },
        {
          onRender: [Function: mockConstructor] {
          },
          text: { list: [ 1, 2 ] },
          collections: {},
          testObject: { isDefaultProp: true }
        },
        {
          onRender: [Function: mockConstructor] {
            _isMockFunction: true,
          },
          text: { list: [7] },
          collections: {},
          testObject: { isDefaultProp: true }
        }
      ]
```
Not sure what the first call is, I assume it's expected?

tests/unit/onyxTest.js
```diff
@@ -178,6 +178,42 @@ describe('Onyx', () => {
...
+
+        Onyx.merge(ONYX_KEYS.TEST_KEY, {something: [1, 2, 3]});
+        Onyx.merge(ONYX_KEYS.TEST_KEY, {something: [4]});
+        return waitForPromisesToResolve()
+            .then(() => {
+                expect(testKeyValue).toEqual({something: [4]});
+            });
+    });
+
+    it('should overwrite an array key nested inside an object when using mergeCollection', () => {
```
marcaaron, line 199:
```suggestion
    it('should overwrite an array key nested inside an object when using merge', () => {
```

lib/OnyxCache.js
```diff
@@ -1,5 +1,6 @@
 import _ from 'underscore';
-import lodashMerge from 'lodash/merge';
+import lodashMergeWith from 'lodash/mergeWith';
+import customizerForMergeWith from "./customizerForMergeWith";
```
marcaaron, line 3:
```suggestion
import customizerForMergeWith from './customizerForMergeWith';
```

tests/unit/withOnyxTest.js
marcaaron: Ah the first subscription always will trigger a render since it's like:

- withOnyx waits for data (there is none for this key)
- withOnyx renders child (1)
- data changes once and component renders again (2)
- data changes once more and component renders again (3)

marcaaron (changes requested):

tests/unit/withOnyxTest.js
marcaaron: If this test works (looks like it) then I am 👍 on not updating the `setState` call in `keysChanged()`

iwiznia: Ah, got it, thanks for the explanation!

iwiznia:
> I think in theory it would be nice if it worked this way. But in practice it doesn't make any difference for us as we have no array values anywhere in any of this data...

Changed this, because I agree it's nice it works in the same way.

> Try this:

I've added the test you suggest [here](https://github.com/Expensify/react-native-onyx/blob/da68b61f2c5b1bc49614212960efc2e93d8332b7/tests/unit/onyxTest.js#L216) (not committed) and it is passing, though for some reason I still don't understand it is making another test fail:
```
 FAIL  tests/unit/onyxTest.js (7.414 s)
  ● Onyx › should return all collection keys as a single object when updating a collection key with waitForCollectionCallback = true

    expect(received).toBe(expected) // Object.is equality

    Expected: null
    Received: {"test_policy_1": null}

      610 |
      611 |                 // AND the value for the first call should be null since the collection was not initialized at that point
    > 612 |                 expect(mockCallback.mock.calls[0][0]).toBe(null);
          |                                                       ^
      613 |
      614 |                 // AND the value for the second call should be collectionUpdate since the collection was updated
      615 |                 expect(mockCallback.mock.calls[1][0]).toEqual(collectionUpdate);

      at fn (tests/unit/onyxTest.js:612:55)
      at tryCallOne (node_modules/promise/lib/core.js:37:12)
      at call (node_modules/promise/lib/core.js:123:15)
      at flush (node_modules/asap/raw.js:50:29)
      at node_modules/@jest/fake-timers/build/legacyFakeTimers.js:505:18
```

iwiznia:
> If this test works (looks like it) then I am 👍 on not updating the setState call in keysChanged()

Ah, missed this. So this should be ready to merge? Do I need to publish a version or increment a number?

lib/Onyx.js
```diff
@@ -732,7 +734,7 @@ function initializeWithDefaultKeyStates() {
         .then((pairs) => {
             const asObject = _.object(pairs);
 
-            const merged = lodashMerge(asObject, defaultKeyStates);
+            const merged = lodashMergeWith(asObject, defaultKeyStates, customizerForMergeWith);
```
marcaaron, line 737: voodoo! how does this know which argument is the customizer? 😄

marcaaron (approved):

marcaaron:
>Do I need to publish a version or increment a number?

Nope. We have magic for that now.

lib/Onyx.js
iwiznia: Magic! heh I assume it looks at the type of the last param and if it is a function it assumes it's the customizer.

README.md
```diff
@@ -63,7 +63,7 @@ The data will then be cached and stored via [`AsyncStorage`](https://github.com/
 
 We can also use `Onyx.merge()` to merge new `Object` or `Array` data in with existing data.
 
-For `Array` the default behavior is to concatenate new items.
+For `Array` the default behavior is to concatenate replace it fully, effectively making it equivalent to set:
```
tgolen, line 66: I think you meant to remove the word "concatenate" here

README.md
```diff
@@ -77,10 +77,10 @@ Onyx.merge(ONYXKEYS.POLICY, {id: 1}); // -> {id: 1}
 Onyx.merge(ONYXKEYS.POLICY, {name: 'My Workspace'}); // -> {id: 1, name: 'My Workspace'}
 ```
 
-One caveat to be aware of is that `lodash/merge` [follows the behavior of jQuery's deep extend](https://github.com/lodash/lodash/issues/2872) and will not concatenate nested arrays in objects. It might seem like this code would concat these arrays, but it does not.
+Arrays inside objects will NOT be concatenated and instead will be replaced fully, same as arrays not inside objects:
```
tgolen, line 80: I would suggest not having the readme say how it used to work (ie. concatenation). I think it's fine to just say the arrays will be fully replaced.

lib/OnyxCache.js
```diff
@@ -110,7 +111,7 @@ class OnyxCache {
      * @param {Record<string, *>} data - a map of (cache) key - values
      */
     merge(data) {
-        this.storageMap = lodashMerge({}, this.storageMap, data);
+        this.storageMap = lodashMergeWith({}, this.storageMap, data, customizerForMergeWith);
```
tgolen, line 114: Since `lodashMergeWith` is always used in conjunction with `customizerForMergeWith`, it seems to me that it would be better to make a merge utility method that hides that a little bit. It would make it easier to work with.

lib/customizerForMergeWith.js
```diff
@@ -0,0 +1,21 @@
+import _ from 'underscore';
```
tgolen, line 1: Uh, we're using underscore to customize lodash? That's weird. I'd suggest just doing a vanilla JS check for `isArray()`. It looks like it's as simple as:

```
Array.isArray(arr);
```

tests/unit/onyxCacheTest.js
```diff
@@ -267,12 +267,25 @@ describe('Onyx', () => {
...
+            it('Should merge arrays inside objects correctly', () => {
+                // GIVEN cache with existing array data
+                cache.set('mockKey', {ID: [1]});
+
+                // WHEN merge is called with an array
+                cache.merge({
+                    mockKey: {ID: [2]},
+                });
+
+                // THEN the arrays should be merged as expected
```
tgolen, line 285: I think this should say more like:

```
// THEN the first array is completely replaced by the second array
```

tests/unit/onyxTest.js
```diff
@@ -178,6 +178,42 @@ describe('Onyx', () => {
...
+
+        Onyx.merge(ONYX_KEYS.TEST_KEY, {something: [1, 2, 3]});
+        Onyx.merge(ONYX_KEYS.TEST_KEY, {something: [4]});
+        return waitForPromisesToResolve()
+            .then(() => {
+                expect(testKeyValue).toEqual({something: [4]});
+            });
+    });
+
+    it('should overwrite an array key nested inside an object when using merge', () => {
```
tgolen, line 199: Could you please change this to say that the difference here is that it has to do with collections? I'm pretty confused on how this test is different than the one above it.

tgolen (commented):
I know we have an active debate about comments in unit tests (regarding the given/when/then), but in this PR it's being done inconsistently. Even though the debate hasn't been settled, I think it should at least maintain a consistent pattern of what is there before.

So, in short, please add given/when/then comments to all the tests.

README.md
iwiznia: hahaha yes

cead22:
That debate is settled afaict. So I think you shouldn't add comments with words in all caps, and since you're editing this file, you should update all comments with words in all caps to use normal capitalization

PR #179, opened 2022-09-19
Improve Onyx comments and clean up some underscore methods

marcaaron:
### Details

This is largely a clean up PR that achieves a few things:

- Improves comments in `react-native-onyx` in particular around the connect + keyChanged + keysChanged methods.
- Backfills some missing functionality for the `keysChanged` method where it would not update an `Onyx.connect()` subscriber to a collection member key (can't think of why this shouldn't work but maybe we are not using it in production)
- Replace several `_.each()` with for loops to try and speed things up at the lower levels.

### Related Issues

There is no issue here. Mostly trying to improve the readability of the code and remove some `_.each()` we I some suspicion will be less performant than for loops.

### Automated Tests

Some tests were added for the new behavior of notifying a collection member key subscriber. Automated tests should cover all of the other modifications.

### DIFF
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -257,11 +257,9 @@ function getCachedCollection(collectionKey) {
  * @param {Object} partialCollection - a partial collection of grouped member keys
  */
 function keysChanged(collectionKey, partialCollection) {
-    // We are iterating over all subscribers similar to keyChanged(). However, we are looking for subscribers who
-    // are subscribing to either a collection key or individual collection key member for the collection that is
-    // being updated. It is important to note that the collection parameter cane be a PARTIAL collection and does not
-    // represent all of the combined keys and values for a collection key. It is just the "new" data that was merged
-    // in via mergeCollection().
+    // We are iterating over all subscribers similar to keyChanged(). However, we are looking for subscribers who are subscribing to either a collection key or
+    // individual collection key member for the collection that is being updated. It is important to note that the collection parameter cane be a PARTIAL collection
+    // and does not represent all of the combined keys and values for a collection key. It is just the "new" data that was merged in via mergeCollection().
     const stateMappingKeys = _.keys(callbackToStateMapping);
     for (let i = stateMappingKeys.length; i--;) {
         const subscriber = callbackToStateMapping[stateMappingKeys[i]];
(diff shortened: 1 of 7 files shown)


joelbettner (changes requested):

lib/withOnyx.js
```diff
@@ -48,16 +49,22 @@ export default function (mapOnyxToState) {
 
             componentDidMount() {
                 // Subscribe each of the state properties to the proper Onyx key
-                _.each(mapOnyxToState, (mapping, propertyName) => {
+                const mapKeys = Object.keys(mapOnyxToState);
+                for (let i = 0; i < mapKeys.length; i++) {
+                    const propertyName = mapKeys[i];
+                    const mapping = mapOnyxToState[propertyName];
```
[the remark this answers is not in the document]
marcaaron: From what I have read, `for...in` performance is supposedly not as fast as a regular `for` loop. But I wonder if it makes a huge difference in practice. Since this code is really "low level" I'm mostly going with the idea that since we are calling these functions very often in the front end code that any small difference could yield a large one at scale.

joelbettner (approved):
LGTM

mountiny:
This has introduced a deploy blocker unfortunately something is off with the merge collection logic. https://github.com/Expensify/App/issues/11754

marcaaron:
I _think_ this PR just highlighted another problem which is that we are incorrectly using `mergeCollection` here:

https://github.com/Expensify/Web-Expensify/blob/23178afe4b30d66145173b1698c942d5cb4de4a5/lib/ReportAPI.php#L2043-L2062

cc @Julesssss

We're calling `mergeCollection()` with a key of `reportActions_273`. When it should only be used with `reportActions_` and a value of `{reportActions_273: {...}}`.

lib/Onyx.js
```diff
@@ -788,7 +887,7 @@ function clear() {
  */
 function mergeCollection(collectionKey, collection) {
     // Confirm all the collection keys belong to the same parent
-    _.each(collection, (data, dataKey) => {
+    _.each(collection, (_data, dataKey) => {
```
mountiny, line 890: @marcaaron This looks suspicious to me, but it was end of a day for me so I opted was quick solution. Sorry about that.

mountiny: I see now, had this open and was investigating why is this happening and havent seen the new comments

PR #183, opened 2022-09-23
Prevent unnecessary `setState()` calls on simple values when nothing has changed

marcaaron:
### Details

`setState()` calls trigger re-renders that are unnecessary. This is a small optimization. In reality, I found that we don't have tons of these in our App.

### Automated Tests

Added an automated test


marcaaron:
Gonna close this and open a new proposed PR.

PR #185, opened 2022-09-27
Prevent setting state for values that have not changed. Add additional Onyx debugging metrics.

marcaaron:
### Details

There are many calls to `setState()` in Onyx that can be avoided. The cost to calculate whether they have changed is lower than the cost to re-render the components that are subscribing to them. Therefore, it makes sense to use an optimized equality check to see if something has changed before we update a component.

I also added metrics that give us insights into why the `setState()` call is happening which can be enabled by using the `ONYX_METRICS=true` flag in the `.env` file.

**Note:** These metrics are currently broken on web due to some unrelated issue. And removing the `applyDecorators()` method will fix the build and allow our new logs to appear.

#### FCP/LCP/etc on Web

**without checks**

Screen Shot 2022-09-27 at 10 42 26 AM

**with checks**

Screen Shot 2022-09-27 at 10 20 14 AM

#### Chat Switch Time on Web

**without `dequal` checks:**

Screen Shot 2022-09-27 at 10 09 36 AM

**with `dequal` checks:**

Screen Shot 2022-09-27 at 10 06 16 AM

There's ~ 90 ms improvement in the chat switching from reducing set state.

So far... not super impressive... but check out Android profiles on chat switching...

#### Chat Switch Time on Android

**without checks**

Screen Shot 2022-09-27 at 10 59 30 AM

**with checks**

Screen Shot 2022-09-27 at 11 00 54 AM

That's 664 ms difference which is pretty great.

### DIFF
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -2,11 +2,13 @@
 import _ from 'underscore';
 import Str from 'expensify-common/lib/str';
 import lodashGet from 'lodash/get';
+import {dequal} from 'dequal/lite';
 import Storage from './storage';
 import * as Logger from './Logger';
 import cache from './OnyxCache';
 import createDeferredTask from './createDeferredTask';
 import mergeWithCustomized from './mergeWithCustomized';
+import * as PerformanceUtils from './metrics/PerformanceUtils';
 
 // Keeps track of the last connectionID that was used so we can keep incrementing it
 let lastConnectionID = 0;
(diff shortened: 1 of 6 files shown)


Szymon20000:
I'm not familiar with onyx but knowing other state management libraries I'm wondering if
would it be possible to make all data immutable so that it's enough to compare reference? 
So whenever you want to make a change you take the object from the state management and even if you want to just change one property you create a new object.

marcaaron:
Yeah, I think we can walk through that as an exercise. Probably first we need to understand whether we are passing around object references or new copies in some places to begin with. I am not actually 100% sure about this... 

**All the data a component initially receives will come from here:**

https://github.com/Expensify/react-native-onyx/blob/3f936d9a81350ee3763fcf70971ec70dafe6b72c/lib/Onyx.js#L563-L574

This next part is confusing because we have two kinds of "subscribers" the collection or just normal key. 

If we have a normal key the value in the cache it would be returned to the component otherwise we get it from storage, put it in the cache, and then return it:

https://github.com/Expensify/react-native-onyx/blob/3f936d9a81350ee3763fcf70971ec70dafe6b72c/lib/Onyx.js#L45-L62

When it goes in the cache we don't do anything special to it so it should be a good reference

https://github.com/Expensify/react-native-onyx/blob/3f936d9a81350ee3763fcf70971ec70dafe6b72c/lib/OnyxCache.js#L92-L95

If the key we are subscribing to is normal (non-"collection") we just set the value directly to state:

https://github.com/Expensify/react-native-onyx/blob/3f936d9a81350ee3763fcf70971ec70dafe6b72c/lib/Onyx.js#L571-L572
https://github.com/Expensify/react-native-onyx/blob/3f936d9a81350ee3763fcf70971ec70dafe6b72c/lib/Onyx.js#L442

And the reference to these simple object _I think_ should survive and be sent to the component:

https://github.com/Expensify/react-native-onyx/blob/3f936d9a81350ee3763fcf70971ec70dafe6b72c/lib/withOnyx.js#L90-L103

"Collection keys" work a bit different because we collect all the keys and then reduce them into a new object here:

https://github.com/Expensify/react-native-onyx/blob/3f936d9a81350ee3763fcf70971ec70dafe6b72c/lib/Onyx.js#L484-L488

Then the whole big object is set to a single `reports` key in the `withOnyx` state.

That's just to give an idea of the references the components start with:
    - simple key same object in cache
    - collection key new object with values referencing objects in the cache

**Next we should look at how they are updated...**

If a single key with object value is changed via a `merge()` then we batch all the merge calls and combine them with a reduce call (creating a new object each time):

https://github.com/Expensify/react-native-onyx/blob/3f936d9a81350ee3763fcf70971ec70dafe6b72c/lib/Onyx.js#L761-L771

Finally this new object is saved: 

https://github.com/Expensify/react-native-onyx/blob/3f936d9a81350ee3763fcf70971ec70dafe6b72c/lib/Onyx.js#L816

And then stored in the cache + subscribers are notified with the new value:

https://github.com/Expensify/react-native-onyx/blob/3f936d9a81350ee3763fcf70971ec70dafe6b72c/lib/Onyx.js#L685-L686

I think at this point it's safe to say that for the simple values this will always create a new object in the cache.

So even though we take the value from the cache and pass it to a subscriber here without modifying it when it is pulled from the cache:

https://github.com/Expensify/react-native-onyx/blob/3f936d9a81350ee3763fcf70971ec70dafe6b72c/lib/Onyx.js#L411-L413

So if we were to look at `prevState` I think we'd find that what is in the cache is not the same object and the simple compare wouldn't work (unless we first fix that or change how it works - I think that is the suggestion you are making 😄).

And if the key we changed belongs to a collection and the subscriber wants all the data then we do this:

https://github.com/Expensify/react-native-onyx/blob/3f936d9a81350ee3763fcf70971ec70dafe6b72c/lib/Onyx.js#L398-L406

Which in the case of `reports` would definitely be a new object since we are using a spread on the old collection.

Hopefully that's not TMI and helps give some ideas about how things work.

marcaaron:
Actually going to split these debugging metrics and anything related to equality checks into separate PRs.

PR #186, opened 2022-09-28
Use faster implementation for merging

Szymon20000:
The whole problem is described here https://github.com/Expensify/App/pull/11185

On Android merging takes a lot of time
I Initially thought that maybe it's caused by slow loads implementation but after looking more deeply into the problem it looks like require takes in lodash a lot of time. Have no idea how because it should be required only once and later not take time at all but it's not the case. Maybe some circular dependencies? 

I used the latest react-native-onyx package .18 with recent fixes like getting rid of merging items of array one by one. But it seems to no solve the problem.

Screenshot 2022-09-28 at 10 54 26
Screenshot 2022-09-28 at 10 54 50
Screenshot 2022-09-28 at 10 57 39
Screenshot 2022-09-28 at 10 57 33

### Related Issues

GH_LINK


Szymon20000:
@iwiznia Could you take a look as you recently optimised the same part?

lib/fastMerge.js
```diff
@@ -0,0 +1,99 @@
+// Mostly copied from  https://medium.com/@lubaka.a/how-to-remove-lodash-performance-improvement-b306669ad0e1
+
+/** // Mostly copied from  https://medium.com/@lubaka.a/how-to-remove-lodash-performance-improvement-b306669ad0e1
```
mountiny, line 3: Lets remove this comment here as it is already on the first line of the file

```suggestion
/**
```

lib/fastMerge.js
```diff
@@ -0,0 +1,99 @@
...
+* @returns {*|boolean}
+*/
+function isMergeableObject(val) {
+  var nonNullObject = val && typeof val === 'object';
+  return nonNullObject &&
+    Object.prototype.toString.call(val) !== '[object RegExp]' &&
+    Object.prototype.toString.call(val) !== '[object Date]';
+}
+/**
++* Empty the Target
```
mountiny, line 16: Typo
```suggestion
* Empty the Target
```

lib/fastMerge.js
```diff
@@ -0,0 +1,99 @@
...
+* @returns {*}
+*/
+function emptyTarget(val) {
+  return Array.isArray(val) ? [] : {}
+}
+/**
+* Clone if Necessary
+*
+* @param value
++* @param optionsArgument
```
mountiny, line 28: Typo
```suggestion
* @param optionsArgument
```

mountiny (commented):
@Szymon20000 You will also need to sign the CLA in this repo as well!

Regarding the JSDoc, lets update the indentation as you can see here, for example https://github.com/Expensify/App/blob/da6e3aac5b64154e1d87c261f34829e097c7b433/src/libs/ReportUtils.js#L388-L393

ie the next line has one whitespace at the start of the line. Thanks!

lib/fastMerge.js
```diff
@@ -0,0 +1,99 @@
...
+* @returns {*}
+*/
+function cloneIfNecessary(value, optionsArgument) {
+  var clone = optionsArgument && optionsArgument.clone === true;
```
tgolen, line 32: Let's remove the `clone` option code entirely since we will never use it

lib/fastMerge.js
```diff
@@ -0,0 +1,99 @@
...
+  * Merge Object and Arrays
+  *
+  * @param target
+  * @param source
+  * @param optionsArgument
+  * @returns {*}
+  */
+export function merge(target, source, optionsArgument) {
+  var array = Array.isArray(source);
+  var options = optionsArgument || { arrayMerge: defaultArrayMerge };
```
tgolen, line 92: Let's also remove this option as well since we also never use it.

tgolen (changes requested):

Szymon20000:

lib/fastMerge.js
```diff
@@ -0,0 +1,99 @@
+// Mostly copied from  https://medium.com/@lubaka.a/how-to-remove-lodash-performance-improvement-b306669ad0e1
+
+/**
+* Is the object Mergeable
+*
+* @param val
+* @returns {*|boolean}
+*/
```
mountiny, line 8: NAB: Can you please update the JSDocs so the next line has always a one whitespace at the start? 
```suggestion
 * Is the object Mergeable
 *
 * @param val
 * @returns {*|boolean}
 */
```

mountiny (commented):
@Szymon20000 I am also thinking if we should update the Docs for the functions, they always just copy what the function name is so it is not very useful. So we should either expand on it (why it is there/what it is used for - which would be handy imho) or remove them as they dont add any value as its current form

lib/OnyxCache.js
```diff
@@ -110,7 +111,7 @@ class OnyxCache {
      * @param {Record<string, *>} data - a map of (cache) key - values
      */
     merge(data) {
-        this.storageMap = mergeWithCustomized({}, this.storageMap, data);
+        this.storageMap = Object.assign({}, merge(this.storageMap, data));
```
tgolen, line 114: I think you can suppress the lint errors here. If we want this to be fast, we shouldn't use the spread operator.

lib/OnyxCache.js
```diff
@@ -1,5 +1,6 @@
 import _ from 'underscore';
 import mergeWithCustomized from './mergeWithCustomized';
+import { merge } from './fastMerge';
```
tgolen, line 3: Please fix all these lint errors

tgolen (changes requested):

lib/fastMerge.js
```diff
@@ -0,0 +1,40 @@
+// Mostly copied from  https://medium.com/@lubaka.a/how-to-remove-lodash-performance-improvement-b306669ad0e1
```
mountiny, line 1:
```suggestion
// Mostly copied from https://medium.com/@lubaka.a/how-to-remove-lodash-performance-improvement-b306669ad0e1
```

lib/fastMerge.js
```diff
@@ -0,0 +1,45 @@
...
+            const key = targetKeys[i];
+            destination[key] = target[key];
+        }
+    }
+    // eslint-disable-next-line rulesdir/prefer-underscore-method
+    const sourceKeys = Object.keys(source);
+    for (let i = 0; i < sourceKeys.length; ++i) {
+        const key = sourceKeys[i];
+        if (source[key] === undefined) {
+            // noop
```
mountiny, line 25: Maybe we could use `continue` instead of noop but that is just preference

mountiny (approved):
Thanks! Lgtm!

lib/fastMerge.js
Szymon20000: That was how I initially implemented it but you have rules that forbid `continue` :D

mountiny: :oh-nothing:

lib/Onyx.js
```diff
@@ -6,8 +6,8 @@ import Storage from './storage';
 import * as Logger from './Logger';
 import cache from './OnyxCache';
 import createDeferredTask from './createDeferredTask';
-import mergeWithCustomized from './mergeWithCustomized';
-import { merge } from './fastMerge';
+// eslint-disable-next-line rulesdir/prefer-import-module-contents
+import {fastMerge} from './fastMerge';
```
tgolen, line 10: By using a default export in `fastMerge.js` that will clean this up and make the ESLint warning go away.

```suggestion
import fastMerge from './fastMerge';
```

lib/fastMerge.js
```diff
@@ -1,99 +1,45 @@
...
-  var options = optionsArgument || { arrayMerge: defaultArrayMerge };
-  var arrayMerge = options.arrayMerge || defaultArrayMerge;
-  if (array) { // We may not even need that as mergeWithCustomized suggests it
-    return Array.isArray(target) ? arrayMerge(target, source, optionsArgument) : cloneIfNecessary(source, optionsArgument);
-  } else {
-    return mergeObject(target, source, optionsArgument);
-  }
+
+// eslint-disable-next-line import/prefer-default-export, rulesdir/no-inline-named-export
+export function fastMerge(target, source) {
```
tgolen, line 38: I think using a default export here is fine (rather than disabling the lint rule) since there is only one thing being exported.
```suggestion
export default (target, source) => {
```

lib/fastMerge.js
tgolen: I think since this lib is focused on being ultra performant, I would disable to `no-continue` rule and use it to make this as fast as possible.

mountiny: I agree with Tim, I would like to see `continue` here

lib/Onyx.js
Szymon20000: That's true but it's a bad practise in my opinion to use default exports. How can you find all occurrences of a function if it can have different name everywhere? I can change it to default exports but have you ever had issue with finding all occurrences of a function?

tgolen: We've never run into that problem, no. It's part of our style guide that all imports should have the same name as the module being imported.

Szymon20000: All good, will change it to default export and address all other comments tomorrow morning. 🙂

lib/fastMerge.js
luacmartins: +1 to continue

luacmartins (changes requested):

mountiny (dismissed):
Thanks for adding the comments, looks good to me

lib/OnyxCache.js
```diff
@@ -110,7 +110,8 @@ class OnyxCache {
      * @param {Record<string, *>} data - a map of (cache) key - values
      */
     merge(data) {
-        this.storageMap = mergeWithCustomized({}, this.storageMap, data);
+        // eslint-disable-next-line prefer-object-spread, rulesdir/prefer-underscore-method
+        this.storageMap = Object.assign({}, fastMerge(this.storageMap, data)); // lodash adds a small overhead so we don't use it here
```
luacmartins, line 114: NAB
```suggestion
        // lodash adds a small overhead so we don't use it here
        this.storageMap = Object.assign({}, fastMerge(this.storageMap, data));
```

luacmartins (dismissed):
Left a few NAB comments

lib/Onyx.js
```diff
@@ -761,7 +761,8 @@ function applyMerge(key, data) {
     if (_.isObject(data) || _.every(mergeValues, _.isObject)) {
         // Object values are merged one after the other
         return _.reduce(mergeValues, (modifiedData, mergeValue) => {
-            const newData = mergeWithCustomized({}, modifiedData, mergeValue);
+            // eslint-disable-next-line prefer-object-spread, rulesdir/prefer-underscore-method
+            const newData = Object.assign({}, fastMerge(modifiedData, mergeValue)); // lodash adds a small overhead so we don't use it here
```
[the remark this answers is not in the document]
tgolen: That comment will actually need to go above the eslint disable comment

lib/fastMerge.js
```diff
@@ -1,45 +1,69 @@
 // Mostly copied from https://medium.com/@lubaka.a/how-to-remove-lodash-performance-improvement-b306669ad0e1
 
+let mergeObject;
+
+/**
+* Is the object Mergeable
+*
+* @param {*} val
+* @returns {*|boolean}
+*/
```
tgolen, line 10:
```suggestion
/**
* @param {mixed} val
* @returns {mixed}
*/
```
Don't include a method description unless it's valuable (and what you had there just repeats the method name). For the return value, maybe we should cast it to always return a boolean?

lib/fastMerge.js
```diff
@@ -1,45 +1,69 @@
...
     return nonNullObject
     && Object.prototype.toString.call(val) !== '[object RegExp]'
     && Object.prototype.toString.call(val) !== '[object Date]';
 }
 
-function mergeObject(target, source) {
+/**
+  * Merge Object and Arrays
+  *
+  * @param {{}|[]} target
```
tgolen, line 21:
```suggestion
/**
  * @param {{}|[]} target
```

lib/fastMerge.js
```diff
@@ -0,0 +1,69 @@
...
+    return mergeObject(target, source);
+}
+
+/**
+* Merge Object
+*
+* @param {{}} target
+* @param {{}} source
+* @returns {{}}
+*/
```
tgolen, line 40:
```suggestion
/**
 * @param {Object} target
 * @param {Object} source
 * @returns {Object}
*/
```

lib/fastMerge.js
```diff
@@ -0,0 +1,69 @@
...
+mergeObject = function (target, source) {
+    const destination = {};
+    if (isMergeableObject(target)) {
+        // eslint-disable-next-line rulesdir/prefer-underscore-method
+        const targetKeys = Object.keys(target); // lodash adds a small overhead so we don't use it here
```
tgolen, line 45:
```suggestion
        // lodash adds a small overhead so we don't use it here
        // eslint-disable-next-line rulesdir/prefer-underscore-method
        const targetKeys = Object.keys(target);
```

lib/storage/providers/LocalForage.js
```diff
@@ -24,7 +24,8 @@ const provider = {
             return localforage.getItem(key)
                 .then((existingValue) => {
                     const newValue = _.isObject(existingValue)
-                        ? mergeWithCustomized({}, existingValue, value)
+                        // eslint-disable-next-line prefer-object-spread, rulesdir/prefer-underscore-method
+                        ? Object.assign({}, fastMerge(existingValue, value)) // lodash adds a small overhead so we don't use it here
```
tgolen, line 28:
```suggestion
                        // lodash adds a small overhead so we don't use it here
                        // eslint-disable-next-line prefer-object-spread, rulesdir/prefer-underscore-method
                        ? Object.assign({}, fastMerge(existingValue, value))
```

tgolen (changes requested):
Sorry to keep harping on the style, but we are sticklers for consistency. Please try to follow previous patterns and our style guides https://github.com/Expensify/Style-Guides/blob/main/javascript.md

lib/fastMerge.js
```diff
@@ -1,99 +1,45 @@
...
-    Object.keys(target).forEach(function (key) {
-      destination[key] = cloneIfNecessary(target[key], optionsArgument)
-    })
-  }
-  Object.keys(source).forEach(function (key) {
-    if (!isMergeableObject(source[key]) || !target[key]) {
-      destination[key] = cloneIfNecessary(source[key], optionsArgument)
-    } else {
-      destination[key] = merge(target[key], source[key], optionsArgument)
+    // eslint-disable-next-line rulesdir/prefer-underscore-method
```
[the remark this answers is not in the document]
kidroca: Why does the function has to be defined as `mergeObject = function (target, source)` instead of simply `function mergeObject(target, source)`?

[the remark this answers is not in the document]
luacmartins: We have to do that otherwise we would use it before it's defined, since we have the following circular dependency `mergeObject -> fastMerge -> mergeObject`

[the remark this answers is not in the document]
tgolen: +1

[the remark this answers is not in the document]
kidroca: when 2 functions are on the same level, they are registered at the same time, it's just that eslint would flag it as a code style issue, because our code style is to define stuff before it's used - impossible to address in case of a circular dependency, but we can just disable the warning

The actual problem is with variables - you can't use a variable before it is defined, but you can use a function before it's defined, because functions are always registered before any code runs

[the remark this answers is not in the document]
luacmartins: Thanks for clarifying @kidroca. In that case, I agree that we should declare it as `function mergeObject()` and disable the lint warning with a comment explaining why

mountiny (approved):
Thanks for addressing the style changes 👍

luacmartins (approved):

PR #202, opened 2022-10-25
Throw if we are trying to merge a non object value to the cache. noop if `mergeCollection()` is called with a bad value.

marcaaron:
### Details

~This should hold until we sent this to production -> https://github.com/Expensify/Web-Expensify/pull/35299~

cc @chiragsalian 

### Related Issues

https://github.com/Expensify/App/issues/12109

### Automated Tests

Yes

### Linked PRs

TODO

### DIFF
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -1046,7 +1046,7 @@ function clear() {
  * @returns {Promise}
  */
 function mergeCollection(collectionKey, collection) {
-    if (!_.isObject(collection) || _.isArray(collection)) {
+    if (!_.isObject(collection) || _.isArray(collection) || _.isEmpty(collection)) {
         Logger.logInfo('mergeCollection() called with invalid or empty value. Skipping this update.');
         return Promise.resolve();
     }
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -1046,7 +1046,7 @@ function clear() {
  * @returns {Promise}
  */
 function mergeCollection(collectionKey, collection) {
-    if (!_.isObject(collection) || _.isEmpty(collection)) {
+    if (!_.isObject(collection) || _.isArray(collection)) {
         Logger.logInfo('mergeCollection() called with invalid or empty value. Skipping this update.');
         return Promise.resolve();
     }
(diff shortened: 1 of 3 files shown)


amyevans (dismissed):

chiragsalian (approved):
Why is this on HOLD. It should work great even if the web-E code is merged. This is a safety check just like we have on web-E and both should work independently of each other, no?

marcaaron:
Hmm but the API is still sending array values for mergeCollection. So if we start throwing then errors will be thrown on the App side. If we push out a new version of onyx before the API is fixed then stuff will break.

chiragsalian:
> Hmm but the API is still sending array values for mergeCollection.

Isn't it sending array only when the array is empty but object otherwise? And in the case of empty array it sounds better if errors are thrown on the App side versus it merging incorrect values no? Also would throwing errors here really break the App? That sounds weird. I feel like merging empty array is breaking the app but throwing errors and preventing that sounds better 🤷‍♂️  If throwing Error can break the app then i wonder if we should just move this behind a console.debug or something. Anyway, I'm okay with waiting, mostly just trying to understand.

marcaaron:
>Isn't it sending array only when the array is empty but object otherwise?

It sends an array when an empty array is sent. If the array has a length > 0 it will show up as an object

>And in the case of empty array it sounds better if errors are thrown on the App side versus it merging incorrect values no?

Not really because the current version of App is not using `fastMerge()` yet so things aren't breaking anymore because the version of `react-native-onyx` was downgraded.

>Also would throwing errors here really break the App? That sounds weird. I feel like merging empty array is breaking the app but throwing errors and preventing that sounds better

Yes because we would be throwing an exception and that should trigger the error boundary in the App and show the oops screen we have. 

I think it's slightly better to throw an error than it is to let random stuff in the app work unexpectedly. But lmk whatever you think is best. My logic here is that the error is _for developers_ so if we want to only throw an error when we are in DEV then we can do that too. 

Mostly just want people to catch this and not release an app version where we do something that is not allowed and drops the entire cache making weird stuff happen.

marcaaron:
Updated so that we are just skipping updates when `mergeCollection()` is called with a bad value (h/t @amyevans). It was getting crazy to validate on the API side so we are just gonna let the values happen and then have clients filter out any that are bad.

amyevans (approved):

PR #208, opened 2022-10-28
Fix a bug where full state was being passed to a withOnyx selector

tgolen:
### Details
There was this code that was passing the entire state object to the selector. That's wrong. Instead, it should only be passing the piece of state that the withOnyx mapping is working with.

cc @marcaaron 

### Related Issues
Fixes an issue I found while testing https://github.com/Expensify/App/pull/12083

### Automated Tests
Covered by automated tests

### DIFF
--- a/tests/unit/subscribeToPropertiesTest.js
+++ b/tests/unit/subscribeToPropertiesTest.js
@@ -98,13 +98,18 @@ describe('Only the specific property changes when using withOnyx() and ', () => 
     });
 
     it('connecting to a single non-collection key with a selector function', () => {
+        const mockedSelector = jest.fn(obj => obj && obj.a);
         const TestComponentWithOnyx = withOnyx({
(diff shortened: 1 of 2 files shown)


tests/unit/subscribeToPropertiesTest.js
```diff
@@ -16,6 +16,23 @@ Onyx.init({
     registerStorageEventListener: () => {},
 });
 
+class ErrorBoundary extends React.Component {
```
tgolen, line 19: @marcaaron I ran into more cases where if there is a syntax error inside the selector function, then Jest isn't able to catch and report it.

marcaaron (approved):
:oh-nothing: good thing we're not using any selectors yet 😄 

Change LGTM!

tgolen:
@joelbettner bump for review please

PR #211, opened 2022-11-10
add quick-sqlite based Onyx storage provider

Szymon20000:
### Details
Why:
- about 3 times faster than async-storage in total cpu time (Js thread time is the same)
- it supports sync execution that maybe useful in the future (for instance fetch the whole DB state in one go)
- when JS thread is not busy it resolves faster
- As it just uses SQLite JSON API it's a good base for adding more complex functionalities in the future

### Related Issues

https://github.com/Expensify/App/pull/11066
https://github.com/Expensify/App/issues/10793


lib/storage/providers/SQLiteStorage.js
```diff
@@ -0,0 +1,106 @@
...
+import {QuickSQLite} from 'react-native-quick-sqlite';
+import _ from 'underscore';
+
+const DB_NAME = 'Expensify-new-db';
+QuickSQLite.open(DB_NAME);
+
+QuickSQLite.execute(DB_NAME, 'CREATE TABLE IF NOT EXISTS magic_map (record_key TEXT NOT NULL PRIMARY KEY , valueJSON JSON NOT NULL) WITHOUT ROWID;');
+QuickSQLite.execute(DB_NAME, 'PRAGMA CACHE_SIZE=-20000;');
+QuickSQLite.execute(DB_NAME, 'PRAGMA SYNCHRONOUS=NORMAL;');
+QuickSQLite.execute(DB_NAME, 'PRAGMA journal_mode=WAL;');
```
mountiny, line 14: Should we leave more information about these Pragmas here and why they were chosen these values? I know this was mostly from the call with SQLite and testing, but still some context would be handy.

lib/storage/providers/SQLiteStorage.js
```diff
@@ -0,0 +1,106 @@
...
+    getItem(key) {
+        return QuickSQLite.executeAsync(DB_NAME, 'SELECT record_key, valueJSON from magic_map where record_key=?;', [key]).then(({rows}) => {
+            // eslint-disable-next-line no-underscore-dangle
+            const res = rows._array[0];
+            return JSON.parse(res.valueJSON);
+        });
+    },
+
+    /**
+      * Get multiple key-value pairs for the give array of keys in a batch
```
mountiny, line 31:
```suggestion
      * Get multiple key-value pairs for the given array of keys in a batch
```

lib/storage/providers/SQLiteStorage.js
```diff
@@ -0,0 +1,106 @@
...
+      * @returns {Promise<String[]>}
+      */
+    getAllKeys: () => QuickSQLite.executeAsync(DB_NAME, 'SELECT record_key from magic_map;').then(({rows}) => {
+        // eslint-disable-next-line no-underscore-dangle
+        const res = _.map(rows._array, row => row.record_key);
+        return res;
+    }),
+
+    /**
+      * Remove given key and it's value from storage
```
mountiny, line 93:
```suggestion
      * Removes given key and it's value from storage
```

lib/storage/providers/SQLiteStorage.js
```diff
@@ -0,0 +1,106 @@
...
+      * @param {String} key
+      * @returns {Promise<void>}
+      */
+    removeItem: key => QuickSQLite.executeAsync(DB_NAME, 'DELETE FROM magic_map where record_key=?;', [key]),
+
+    /**
+      * Clear absolutely everything from storage
```
mountiny, line 100:
```suggestion
      * Clears absolutely everything from storage
```

mountiny (commented):
Great! Left couple of comment suggestions!

jasperhuangg:
Are there any tests or screen recordings you can provide of the old and new times?

lib/storage/providers/SQLiteStorage.js
```diff
@@ -0,0 +1,106 @@
...
+    /**
+      * Get the value of a given key or return `null` if it's not available in storage
+      * @param {String} key
+      * @return {Promise<*>}
+      */
+    getItem(key) {
+        return QuickSQLite.executeAsync(DB_NAME, 'SELECT record_key, valueJSON from magic_map where record_key=?;', [key]).then(({rows}) => {
+            // eslint-disable-next-line no-underscore-dangle
+            const res = rows._array[0];
+            return JSON.parse(res.valueJSON);
```
marcaaron, line 26: Will `rows._array` always return a non-empty array with JSON values?

lib/storage/providers/SQLiteStorage.js
```diff
@@ -0,0 +1,106 @@
...
+const provider = {
+    /**
+      * Get the value of a given key or return `null` if it's not available in storage
+      * @param {String} key
+      * @return {Promise<*>}
+      */
+    getItem(key) {
+        return QuickSQLite.executeAsync(DB_NAME, 'SELECT record_key, valueJSON from magic_map where record_key=?;', [key]).then(({rows}) => {
+            // eslint-disable-next-line no-underscore-dangle
+            const res = rows._array[0];
```
marcaaron, line 25:
```suggestion
            const result = rows._array[0];
```

lib/storage/providers/SQLiteStorage.js
marcaaron, line 26:
```suggestion
            return JSON.parse(result.valueJSON);
```

lib/storage/providers/SQLiteStorage.js
```diff
@@ -0,0 +1,106 @@
...
+      * Get multiple key-value pairs for the give array of keys in a batch
+      * @param {String[]} keys
+      * @return {Promise<Array<[key, value]>>}
+      */
+    multiGet(keys) {
+        return QuickSQLite.executeAsync(DB_NAME, `SELECT record_key, valueJSON from magic_map where record_key IN (${new Array(keys.length).fill('?').join(',')});`, keys)
+            .then(({rows}) => {
+                // eslint-disable-next-line no-underscore-dangle
+                const res = _.map(rows._array, row => [row.record_key, JSON.parse(row.valueJSON)]);
+                return res;
```
marcaaron, line 40:
```suggestion
                const result = _.map(rows._array, row => [row.record_key, JSON.parse(row.valueJSON)]);
                return result;
```

lib/storage/providers/SQLiteStorage.js
```diff
@@ -0,0 +1,106 @@
...
+            });
+    },
+
+    /**
+      * Sets the value for a given key. The only requirement is that the value should be serializable to JSON string
```
marcaaron, line 45: I recall some work that we did to allow files (like images) to be stored in Onyx. I can't remember if we are actually using this feature or not, but I assume we are. And also guessing that it may not work after this change.

cc @thienlnam @kidroca to refresh our memory

marcaaron (changes requested):
LGTM mostly had some suggestions for style improvements. We tend to avoid abbreviated variables and I spotted a few.

lib/storage/providers/SQLiteStorage.js
roryabraham: Also, did you ever try using `WAL2`? We can always do a follow-up to explore and compare benchmarks, but as I understand it can result in better performance when multiple threads are involved, particularly in the area of memory usage (preventing certain undesirable scenarios like an infinitely-growing WAL)

lib/storage/providers/SQLiteStorage.js
```diff
@@ -0,0 +1,106 @@
...
+    },
+
+    /**
+      * Returns all keys available in storage
+      * @returns {Promise<String[]>}
+      */
+    getAllKeys: () => QuickSQLite.executeAsync(DB_NAME, 'SELECT record_key from magic_map;').then(({rows}) => {
+        // eslint-disable-next-line no-underscore-dangle
+        const res = _.map(rows._array, row => row.record_key);
+        return res;
```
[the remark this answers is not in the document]
roryabraham: Oh, why not just?

```js
return _.map(rows._array, row => row.record_key);
```

package.json
```diff
@@ -70,13 +70,15 @@
     "react": "^17.0.2",
     "react-native": "0.64.1",
     "react-native-performance": "^2.0.0",
+    "react-native-quick-sqlite": "^5.0.3",
```
parasharrajat, line 73: shouldn't the package-lock.json file be updated as well?

lib/storage/providers/SQLiteStorage.js
thienlnam: Yeah good catch, we do seem to be using this feature currently to store image in onyx while we're offline and have it upload later

https://expensify.slack.com/archives/C01GTK53T8Q/p1665588920653759

lib/storage/providers/SQLiteStorage.js
```diff
@@ -0,0 +1,106 @@
...
+      * @return {Promise<void>}
+      */
+    multiMerge(pairs) {
+        return QuickSQLite.executeBatchAsync(DB_NAME,
+            [['INSERT into magic_map (record_key, valueJSON) VALUES (?, json(?)) ON CONFLICT DO UPDATE SET valueJSON = json_patch(valueJSON, json(?));',
+                _.map(pairs, (ele) => {
+                    const str = JSON.stringify(ele[1]);
+                    return [ele[0], str, str];
+                }),
+            ]]);
```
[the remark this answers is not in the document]
roryabraham: > Curious about the second str in this line. Is it needed? If so, maybe a quick comment about why would be nice to have.

It's because there's three `?` in the query, and you need to [bind params](https://www.sqlite.org/c3ref/bind_blob.html) for each. I think we could/should used named bindings like `:param` instead.

Few questions/comments:

- Can you include a note about why we're using `ON CONFLICT DO UPDATE`? It took me a bit to realize that the reason we're not using `INSERT OR REPLACE INTO` here is because if there's already a json value for that key, then we want to use `JSON_PATCH` to deep-merge the new json value into the existing one.
- Can we make sure we are consistent about using `ALL CAPS` for SQLite keywords?
- Can we dig into the QuickSQLite internals here and make sure that `executeBatchAsync` wraps all the separate queries it runs into a single transaction with `BEGIN TRANSACTION` and `COMMIT`?

So overall my suggestion might end up looking like this:

```js
multiMerge(pairs) {
    // Note: We use `ON CONFLICT DO UPDATE` here instead of `INSERT OR REPLACE INTO`
    // so the new JSON value is merged into the old one if there's an existing value
    const query = `
        INSERT OR REPLACE INTO magic_map (record_key, valueJSON)
        VALUES (:key, JSON(:value))
        ON CONFLICT DO UPDATE
    `;
    const queryArguments = _.map(timestamp, ([key, value]) => ({key, value}));
    return QuickSQLite.executeBatchAsync(DB_NAME, [query, queryArguments]);
}
```

That also makes me wonder if the `pairs` argument should be `Array<{key, value}>` instead of `Array<[key, value]>`, skipping the need for this `_.map` here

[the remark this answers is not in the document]
roryabraham: Looks like we are good with the `BEGIN TRANSACTION` and `COMMIT` stuff: https://github.com/ospfranco/react-native-quick-sqlite/blob/cea63588d702529c0655408754621cc110275618/cpp/sqlbatchexecutor.cpp#L59-L75

lib/storage/providers/SQLiteStorage.js
kidroca: It won't be a problem 

On native platforms images are stored on the FS while Onyx (db) only stores the path (URI) on the FS 

This happens automatically e.g. 
- pick image using the image picker
- whatever we select is resolved as `ImageSource` with `uri` key containing path to the file
- we store the `uri` in Onyx and not the actual file

kidroca (commented):
Left some comments 

One thing we might want to change in the future is to add an `init` method to the storage provider interface 

The idea is to let the app init and then call Onyx (storage) init which would in turn initialize the provider, instead of running the code eagerly like we do right here with the `open` and `execute` top level calls

lib/storage/providers/SQLiteStorage.js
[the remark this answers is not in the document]
kidroca: > That also makes me wonder if the `pairs` argument should be `Array<{key, value}>` instead of `Array<[key, value]>`, skipping the need for this `_.map` here

This is dictated by the Provider interface - Onyx expects to talk to a storage provider that accepts the current structure of `pairs`, if we change that, we'll have to update a bunch of other places

lib/storage/providers/SQLiteStorage.js
Szymon20000: @roryabraham I haven't tested it. That most likely would require to replace SQLite code that quick-sqlite uses with the version that supports WAL2. I happy to test it if you want :)

lib/storage/providers/SQLiteStorage.js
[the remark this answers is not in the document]
Szymon20000: I tried to remove it but then aslant expects everything to be in one line. It looks a bit weird then as 
`// eslint-disable-next-line no-underscore-dangle` needs to be between js docs and method definition. I can still do it this way if you want.

lib/storage/providers/SQLiteStorage.js
```diff
@@ -0,0 +1,106 @@
+/**
+ * The SQLiteStorage provider stores everything in a key/value store by
+ * converting the value to a JSON string
+ */
+import {QuickSQLite} from 'react-native-quick-sqlite';
+import _ from 'underscore';
+
+const DB_NAME = 'Expensify-new-db';
+QuickSQLite.open(DB_NAME);
```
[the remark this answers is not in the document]
Szymon20000: I tried it before and it didn't work. However, thanks to your comment I checked it more deeply and it turned out that `open` method is different from `QuickSQLite.open`. The first one expects options object and returns db however second one expects a db name and returns void :) . So after adjusting few thins it works all good.

Szymon20000:
Screenshot 2022-11-14 at 17 51 47
I see it's caused by old package-lock file. I have issues with running rpm install.

Szymon20000:
I removed all react17 deps. So I guess npm takes that deep from last published react-native-onyx version. Let me know if that's sth obvious otherwise I will fix it tomorrow morning.

lib/storage/providers/SQLiteStorage.js
roryabraham: > That most likely would require to replace SQLite code that quick-sqlite uses with the version that supports WAL2. I happy to test it if you want

Let's test out WAL2 in a follow-up

Szymon20000:
Tests are passing for me locally but it looks like sth is wrong with CI. 
I saw the you are testing AsyncStorage provider using AsyncStorage mock. I don't see a easy way for making sth similar for quick-sqlite. Right now I'm just using AsyncStorage provider for tests instead of SQLite one. I can try to find a way for mocking sqlite but I guess it's not easy.

hannojg:

hannojg:
recheck

mountiny (commented):
@Szymon20000 @hannojg Could you please update the comments to be aligned correctly

Otherwise I dont have any additional comments to this!

jestSetup.js
```diff
@@ -0,0 +1,3 @@
+jest.mock('./lib/storage/providers/SQLiteStorage', () =>
+  require('./lib/storage/providers/__mocks__/SQLiteStorage')
+);
```
thienlnam, line 3: newline missing

kidroca:
> I see it's caused by old package-lock file. I have issues with running npm install.

This is because until recently Expensify App and react-native-onyx were using Node 14 and package-lock version: 1 
App was updated to Node 16 (npm 8 and package-lock version 2), but I guess Onyx wasn't 

My advice would be to
1. revert changes to `package-lock.json`
2. switch to Node 14 using `nvm` 
3. Run `npm install` again 

This would result in a few changes in package-lock and we can migrate the Onyx project to Node 16 in a separate PR

jestSetup.js
kidroca, line 3: I'm not sure why this was added - to active a manual mock from the `__mocks__` folder you just call

```
jest.mock('./lib/storage/providers/SQLiteStorage')
```

In the test file that is supposed to use a mocked version

lib/storage/providers/SQLiteStorage.js
```diff
@@ -0,0 +1,120 @@
...
+const provider = {
+    /**
+        * Get the value of a given key or return `null` if it's not available in storage
+        * @param {String} key
+        * @return {Promise<*>}
+        */
+    getItem(key) {
+        return db.executeAsync('SELECT record_key, valueJSON FROM magic_map WHERE record_key=?;', [key]).then(({rows}) => {
+            // eslint-disable-next-line no-underscore-dangle
+            if (rows._array.length === 0) {
```
kidroca, line 28: There should be `rows.length` (as well as `rows._array`)
https://github.com/ospfranco/react-native-quick-sqlite/blob/654c8b44d166e198b5cfc41970e23c42fbab909d/src/index.ts#L60

```suggestion
            if (rows.length === 0) {
```

lib/storage/providers/SQLiteStorage.js
```diff
@@ -0,0 +1,120 @@
...
+                return null;
+            }
+            // eslint-disable-next-line no-underscore-dangle
+            const result = rows._array[0];
```
kidroca, line 32: We use `lodash/get` when accessing a chain of properties

To avoid the underscore dangle, you can use the interface shared in the previous comment
```suggestion
            const result = rows.item(0);
```

Or `lodashGet`

https://github.com/Expensify/react-native-onyx/blob/3f6cea066a74292bc47ba925113ad5ad529d72a3/lib/Onyx.js#L5

```suggestion
            const result = lodashGet(rows, ['_array', '0']);
```

lib/storage/providers/__mocks__/SQLiteStorage.js
```diff
@@ -0,0 +1,3 @@
+import provider from '../AsyncStorage';
+
+export default provider;
```
kidroca, line 3: We decided to put the mocks on the `lib/storage` level - that's why other storage providers don't have mocks 

Whether we have a mocked storage or a storage that returns a mocked provider it's the same thing, but it's more convenient to have the mock on the upper level and write just one mock, rather than having to write a mock for each provider

kidroca (commented):
There are a few usages of `// eslint-disable-next-line` that we can get rid of if we use `lodashGet`

We have already established a way of manually mock modules and the changes here introduce another way 
The way we mock things should already work and be applicable here as well 
I'll check it out locally in a bit and provide some tips

kidroca:
@Szymon20000 
I couldn't make tests work, without `jestSetup` either, I've submitted some suggestions as a PR on your fork here: https://github.com/Szymon20000/react-native-onyx/pull/1

---

IMO we have some problems with Onyx tests we should focus on in a separate PR

It looks like we relied on using Async storage and it's mock in our tests for most of the cases, while for a few we switch to web and the mock for localforage 
With the introduction of another storage provider it becomes more apparent that our mock/test structure is not ideal and we're kind of patching it up with duct tape (right now)

Problem - 2 or more ways to mock storage:
- Some of the mocking happens through root level mocks for external libraries 
- Other on the storage level which then mock from the point above
- And now on the provider level

In Onyx tests there should be only one mock - the `Storage` object that is imported in `lib/Onyx` should be mocked
It's not useful to have Storage unmocked or partially mocked, and maintain 2 or more mocks for providers 

Onyx tests should verify that Onyx is using the Storage (interface) correctly, it's not possible to verify that the actual storage is working correctly (in jest env) since we don't control the native storage code - we assume the actual provider already handles storage correctly for us, thus we don't need to write provider mocks per each provider class we implement

If something in the core Onyx logic should work differently based on platform, then it should be covered in a separate test with specific manual mocking (AFAIK there's no such logic), but in general one storage mock is enough

kidroca:
> > I see it's caused by old package-lock file. I have issues with running npm install.
> 
> This is because until recently Expensify App and react-native-onyx were using Node 14 and package-lock version: 1 App was updated to Node 16 (npm 8 and package-lock version 2), but I guess Onyx wasn't

It seems we're using this PR to upgrade React and a bunch of other packages, so I guess we might as well migrate to Node 16 and npm 8 and the new package-lock structure, in that case we should also ad the `engines` field to `package.json` with the same content from App and also an `.nvmrc` file with the same content

lib/storage/providers/__mocks__/SQLiteStorage.js
Szymon20000: Are you sure that mock is in use? I think it's just not used at all and you use AsyncStorage mock. Correct me if I'm wrong :).

Szymon20000:
Thx for next review iteration will address every comment tomorrow morning!

lib/storage/providers/__mocks__/SQLiteStorage.js
kidroca: Yeah, it's slightly more tangled than that - we do sometimes use the localforage mock 
Anyway I think it shouldn't be a concern for this PR.
Kudos for making the tests work with as little changes!

Szymon20000:
> 

I did what you suggested. Could you take a look if it's done properly?

Szymon20000:
CI is failing but the error seems to not be related to my changes.

kidroca:
Yeah, it's because of the package changes

The error in the lint job is about npm not having access to install `expensify-common`, but I also noticed the CI uses Node 14

https://github.com/Expensify/react-native-onyx/blob/91aabb7219ddc8722cc2955de0398b1ae70c546d/.github/workflows/lint.yml#L23

I'm not sure whether we should invest any more time trying to migrate to Node 16 in this PR, the other workflows probably need to be changed to Node 16 as well 

Some options:
- you can, drop the Node 16 changes and try to use `nvm` and node 14 as I initially suggested (before changing my mind 😃): https://github.com/Expensify/react-native-onyx/pull/211#issuecomment-1317167084
- you can try to update the workflow files to use node 16 and hope the problem is resolved
- someone can open a separate PR to update dependencies and migrate Onyx to Node 16

chrispader:

chrispader:
recheck

roryabraham:
> In Onyx tests there should be only one mock - the Storage object that is imported in lib/Onyx should be mocked
It's not useful to have Storage unmocked or partially mocked, and maintain 2 or more mocks for providers

Completely agree with your comments about tests, @kidroca

lib/storage/providers/SQLiteStorage.js
marcaaron: Thanks!

marcaaron:
@kidroca Been a while since I looked at the tests from a high level so probably better to create some issue or separate proposal for those improvements.

marcaaron (commented):
These changes look nice to me. I'm not entirely sure how to test them. @Szymon20000 can you include some test steps for how to test this PR?

And the only thought I'm having is that we could probably document the inclusion of quick sqlite, but not really a blocker as we can always go back and do that later.

Szymon20000:
> 

I tested it by running the app and also using snippets like this one https://github.com/Szymon20000/App/blob/%40szymon/raw_test/borrowed_modules/react-native-onyx/lib/storage/providers/SQLiteStorage.js#L113

I wanted to add tests similar to the one you have for AsyncStorageProvider but there is no mock for quick-sqlite. I can try to find node sqlite module to create one if you want me to.

lib/storage/providers/SQLiteStorage.js
```diff
@@ -0,0 +1,118 @@
...
+ */
+import {open} from 'react-native-quick-sqlite';
+import _ from 'underscore';
+
+const DB_NAME = 'Expensify-new-db';
+const db = open({name: DB_NAME});
+
+db.execute('CREATE TABLE IF NOT EXISTS magic_map (record_key TEXT NOT NULL PRIMARY KEY , valueJSON JSON NOT NULL) WITHOUT ROWID;');
+
+// All of the 3 pragmas below were suggested by SQLite team.
```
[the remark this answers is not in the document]
Szymon20000: They are described in the link that I added one line below. The only other reason I know for them is `speed` :D

lib/storage/providers/SQLiteStorage.js
```diff
@@ -0,0 +1,118 @@
...
+ * The SQLiteStorage provider stores everything in a key/value store by
+ * converting the value to a JSON string
+ */
+import {open} from 'react-native-quick-sqlite';
+import _ from 'underscore';
+
+const DB_NAME = 'Expensify-new-db';
+const db = open({name: DB_NAME});
+
+db.execute('CREATE TABLE IF NOT EXISTS magic_map (record_key TEXT NOT NULL PRIMARY KEY , valueJSON JSON NOT NULL) WITHOUT ROWID;');
```
iwiznia, line 11: Have we settled on this name for the DB?

lib/storage/providers/SQLiteStorage.js
```diff
@@ -0,0 +1,118 @@
...
+
+// All of the 3 pragmas below were suggested by SQLite team.
+// You can find more info about them here: https://www.sqlite.org/pragma.html
+db.execute('PRAGMA CACHE_SIZE=-20000;');
+db.execute('PRAGMA synchronous=NORMAL;');
+db.execute('PRAGMA journal_mode=WAL;');
```
iwiznia, line 17: Why WAL and not WAL2? I thought WAL2 was better overall, but maybe it only matters when you have multiple threads reading/writing in the DB?

lib/storage/providers/SQLiteStorage.js
```diff
@@ -0,0 +1,118 @@
...
+
+const provider = {
+    /**
+    * Get the value of a given key or return `null` if it's not available in storage
+    * @param {String} key
+    * @return {Promise<*>}
+    */
+    getItem(key) {
+        return db.executeAsync('SELECT record_key, valueJSON FROM magic_map WHERE record_key=?;', [key]).then(({rows}) => {
```
iwiznia, line 26: For consistency with other code we have, let's format all queries to have spaces before/after the comparisons. ie:
```suggestion
        return db.executeAsync('SELECT record_key, valueJSON FROM magic_map WHERE record_key = ?;', [key]).then(({rows}) => {
```

lib/storage/providers/SQLiteStorage.js
```diff
@@ -0,0 +1,118 @@
...
+    * @return {Promise<Array<[key, value]>>}
+    */
+    multiGet(keys) {
+        const placeholders = _.map(keys, () => '?').join(',');
+        const command = `SELECT record_key, valueJSON FROM magic_map WHERE record_key IN (${placeholders});`;
+        return db.executeAsync(command, keys)
+            .then(({rows}) => {
+                // eslint-disable-next-line no-underscore-dangle
+                const result = _.map(rows._array, row => [row.record_key, JSON.parse(row.valueJSON)]);
+                return result;
```
iwiznia, line 47:
```suggestion
                return _.map(rows._array, row => [row.record_key, JSON.parse(row.valueJSON)]);
```

lib/storage/providers/SQLiteStorage.js
```diff
@@ -0,0 +1,118 @@
...
+    },
+
+    /**
+    * Returns all keys available in storage
+    * @returns {Promise<String[]>}
+    */
+    getAllKeys: () => db.executeAsync('SELECT record_key FROM magic_map;').then(({rows}) => {
+        // eslint-disable-next-line no-underscore-dangle
+        const result = _.map(rows._array, row => row.record_key);
+        return result;
```
iwiznia, line 101:
```suggestion
        return result = _.map(rows._array, row => row.record_key);
```

lib/storage/providers/SQLiteStorage.js
```diff
@@ -0,0 +1,118 @@
...
+    }),
+
+    /**
+    * Removes given key and it's value from storage
+    * @param {String} key
+    * @returns {Promise<void>}
+    */
+    removeItem: key => db.executeAsync('DELETE FROM magic_map WHERE record_key=?;', [key]),
```
iwiznia, line 109:
```suggestion
    removeItem: key => db.executeAsync('DELETE FROM magic_map WHERE record_key = ?;', [key]),
```

iwiznia (changes requested):

lib/storage/providers/SQLiteStorage.js
Szymon20000: quick-sqlite doesn't support WAL2 I would need to modify that lib and will do but let's not do it in this pr.

lib/storage/providers/SQLiteStorage.js
Szymon20000: I wanted to do that but then eslint forces me to remove return keyword to what would require me to move (// eslint-disable-next-line no-underscore-dangle) between definition of the function and JSdoc.

lib/storage/providers/SQLiteStorage.js
Szymon20000: No, I had to pass a name and it was first what crossed my mind :). What would be a better name?

lib/storage/providers/SQLiteStorage.js
iwiznia: Thanks for the context. No need to add it for now, I assume it won't make much of a difference in this case anyway, so if it was available was worth the try, but not otherwise.

lib/storage/providers/SQLiteStorage.js
iwiznia: hmmmm how about `keyvaluepairs`?

lib/storage/providers/SQLiteStorage.js
iwiznia: Ah ok

marcaaron:
@kidroca @parasharrajat @thienlnam @mountiny @roryabraham @iwiznia 

Everyone mentioned is showing up in the Reviewers list. Can I get some thumbs up to show who wants to final review and merge this PR?

Have any of you been able to manually test this PR locally yet? I'm still not sure how to do it. I tried to pull down [this branch](https://github.com/Expensify/App/pull/11066), but didn't have much success building the app on iOS or Android.

@Szymon20000 please do let us know when this is ready for another review by leaving a comment that says "Updated". And if you can help with manual testing that would be great. It feels like we are very close to merging this and just need the ability to do some manual testing. Are you able to give us some step by step instructions?

mountiny:
I havent been able to test this :/

Szymon20000:
@Marc I prepared a branch where you can test it a few days ago. https://expensify.slack.com/archives/C035J5C9FAP/p1669991527095399?thread_ts=1669851838.427289&cid=C035J5C9FAP

Szymon20000:
1. checkout to https://github.com/margelo/expensify-app-fork/tree/%40szymon/test-new-provider
2. npm install
3. npm run start --reset-cached
4. run android or iOS

lib/storage/providers/SQLiteStorage.js
kidroca: I agree with @iwiznia 
Since `react-native-onyx` is an utility that can be used in any project and not specifically "New Expensify" the name should be more generic

BTW for web (IndexedDB) we use the name `OnyxDB` 

https://github.com/Expensify/react-native-onyx/blob/91aabb7219ddc8722cc2955de0398b1ae70c546d/lib/storage/providers/LocalForage.js#L12-L14

lib/storage/providers/SQLiteStorage.js
```diff
@@ -0,0 +1,118 @@
...
+    * @return {Promise<void>}
+    */
+    multiMerge(pairs) {
+        // Note: We use `ON CONFLICT DO UPDATE` here instead of `INSERT OR REPLACE INTO`
+        // so the new JSON value is merged into the old one if there's an existing value
+        const query = `INSERT INTO magic_map (record_key, valueJSON) 
+             VALUES (:key, JSON(:value)) 
+             ON CONFLICT DO UPDATE 
+             SET valueJSON = JSON_PATCH(valueJSON, JSON(:value)); 
+        `;
```
kidroca, line 86: Can you link to documentation regarding how exactly this works 

For example how are arrays merged - are they concatenated or merged like objects

kidroca (commented):
I'll try to test this locally as well tonight

Looks good, left a small note regarding `multiMerge` as I don't understand the merge strategy

lib/storage/providers/SQLiteStorage.js
marcaaron: This looks like a good explanation here -> https://www.sqlite.org/json1.html#jpatch

marcaaron:
@Szymon20000 sorry if I'm missing something obvious but the build is not working



Looked at the directory "borrowed_modules" and this checks out as it is empty. I noticed it's a submodule here:



Did I miss some step?

lib/storage/providers/SQLiteStorage.js
kidroca: Ok then this a potentially breaking change due to array merging 
`JSON_PATCH` would not merge arrays but use the array from the input as the final array, while the old logic (and the one used on web is):

https://github.com/Expensify/react-native-onyx/blob/91aabb7219ddc8722cc2955de0398b1ae70c546d/lib/fastMerge.js#L24-L28

So old logic would do this

```js
const src = { list: [1, 2, 3, 4] };
const dest = { list: [4, 3] };
merge(src, dest); // { list: [4, 3, 3, 4] }
```

While the new

```js
const src = { list: [1, 2, 3, 4] };
const dest = { list: [4, 3] };
merge(src, dest); // { list: [4, 3] }
```

---

This `JSON_PATCH` simple logic might be for the best as it's easier to understand, but we should test whether it doesn't break anything (merging collections)

marcaaron: I agree some quick testing (either manual or automated) would be good here. 

I recall that @iwiznia changed the behavior for array merging. And then we figured out later that it would be the same as the sqlite MergePatch behavior. That was before we implemented an even more custom merge function (`fastMerge()`). I would imagine that we would have tried to preserve that behavior, but don't know for sure.

Is it solved by this code here ?

https://github.com/Expensify/react-native-onyx/blob/91aabb7219ddc8722cc2955de0398b1ae70c546d/lib/fastMerge.js#L56-L64

marcaaron (dismissed):
Alright managed to get it working locally now with the latest changes pushed, thanks @Szymon20000!

Tested iOS  and Android and did not see any significant issues or regressions. Though, I am curious whether we feel the need to do any benchmarking to verify that things are improving (or not) - I admittedly found it hard to come to any sort of subjective conclusion about performance when comparing the current version of the app to the app with the sqlite provider.

I don't think this is a blocker personally (it gets my approval either way as it looks like a very solid change) but just wanted to toss it out there in case others have ideas or desires to validate more concretely with benchmarks.

Oh and one thing that will happen when we switch over is that all user data on the device will be lost, but not sure if this is something we need to create a migration for. People can just log in again to restore the app.

kidroca:
I've captured some metrics for initial load and then opening a chat

Very similar results, sometimes async storage is slightly faster - others sql edges a bit faster

The most frequent operation is `get` and as you can tell there's no big improvement there, though I should mention that `get` is reading from hard storage only for the initial key reads, otherwise it resolves instantly from cache.

###  SQLite
  - Sum: 1.7min

|        method        | total time spent |    max    |    min    |    avg    | time last call completed | calls made |
|----------------------|-----------------:|----------:|----------:|----------:|-------------------------:|-----------:|
| Onyx:getAllKeys      |          1.06sec | 142.166ms |   0.062ms |  11.056ms |                 11.34sec |         96 |
| Onyx:defaults        |        148.154ms | 148.154ms | 148.154ms | 148.154ms |                156.035ms |          1 |
| Onyx:get             |           1.4min | 969.491ms |   0.047ms | 115.817ms |                 13.51sec |        709 |
| Onyx:set             |          4.77sec | 972.717ms |   0.631ms | 183.269ms |                 13.72sec |         26 |
| Onyx:merge           |          7.78sec | 972.941ms |   0.572ms | 370.530ms |                 13.72sec |         21 |
| Onyx:update          |          1.23sec | 653.079ms | 132.798ms | 308.105ms |                 13.72sec |          4 |
| Onyx:mergeCollection |          1.94sec | 971.716ms | 971.376ms | 971.546ms |                  5.88sec |          2 |

### AsyncStorage
  - Sum: 1.6min

|        method        | total time spent |    max    |    min    |    avg    | time last call completed | calls made |
|----------------------|-----------------:|----------:|----------:|----------:|-------------------------:|-----------:|
| Onyx:getAllKeys      |          1.02sec | 127.659ms |   0.059ms |  10.595ms |                 11.84sec |         96 |
| Onyx:defaults        |        135.017ms | 135.017ms | 135.017ms | 135.017ms |                144.063ms |          1 |
| Onyx:get             |           1.3min | 994.087ms |   0.047ms | 111.082ms |                 12.47sec |        709 |
| Onyx:set             |          5.08sec | 994.964ms |   0.578ms | 195.468ms |                 12.68sec |         26 |
| Onyx:merge           |          8.09sec | 994.793ms |   0.813ms | 385.267ms |                 12.68sec |         21 |
| Onyx:update          |          1.31sec | 649.711ms | 185.640ms | 328.742ms |                 12.68sec |          4 |
| Onyx:mergeCollection |          2.24sec |   1.24sec | 993.759ms |   1.12sec |                  6.04sec |          2 |

_"Sum" is the sum of times for all calls and not how long it took to run the test_

Szymon20000:
@kidroca  Most likely you wrapped calls in before after performance.now() (based on the times above). The problem with that is that is also contains rerenders and so many other stuff that it's not reliable. However, both Async storage and new provider serialise data in the same why what is 99% of JS time used by them and because JS thread in the app is constantly busy there is not difference in JS Thread cpu usage. However, total CPU usage is few times better. In order to measure total CPU time you would need to run both solutions on a RN context without any business logic. I did it here:
https://expensify.slack.com/archives/C035J5C9FAP/p1667503389141599

lib/storage/providers/SQLiteStorage.js
kidroca: > Is it solved by this code here ?

Yes it is (I missed that part)
LocalForage provider seems to work the same way, though it uses `Object.assign`, so we should be good 

https://github.com/Expensify/react-native-onyx/blob/91aabb7219ddc8722cc2955de0398b1ae70c546d/lib/storage/providers/LocalForage.js#L28-L31

kidroca:
@Szymon20000 
The timings are for the total call duration including native code execution:
1. Timestamp at the start of the async call
2. JS time - passing to native read/write
3. Native time - doing the actual read/write
4. JS time - receiving from native read/write
5. Timestamp at the end of the async call (`Promise.finally`) 

Yes, in the meantime other operations draw CPU resources, but on average the end result is the same (avg for 10 runs per provider). 
Both AsyncStorage and SQLite were subject to the same test and same side effects 
React renders are idempotent - given the same input it would take the same amount of `renders` to produce the end result  

My point with this benchmark is that the storage driver is not the bottleneck and doesn't have a significant impact on Onyx performance in New Expensify (on average retrieving or writing a value takes the same amount of time)

The change still has a value if it leaves a door open for future optimizations, but to speed up App and Onyx I think we should explore other options:
- update batching
- improve React integration - less invalidating commits caused by `setState`
- retrieve initial App state in one go

Szymon20000:
That's true. I also see it more like a door opening approach which don't directly improve app startup time. 
I tried to retrieve initial App state in one go and it took around 400ms. If we could also modify withOnyx so it jest fetches initial state from onyx without setting a new state that could be a huge performance gain.

lib/storage/providers/SQLiteStorage.js
iwiznia: Let's change it

iwiznia (changes requested):

Szymon20000:
@iwiznia You mean change table name to `keyvaluepairs `?

iwiznia:
Yeah, if we use OnyxDB as the DB name currently, let's use the same here.

Szymon20000:
> Yeah, if we use OnyxDB as the DB name currently, let's use the same here.

Done :)

iwiznia (approved):

lib/storage/providers/SQLiteStorage.js
```diff
@@ -0,0 +1,118 @@
...
+    * @param {String} key
+    * @returns {Promise<void>}
+    */
+    removeItem: key => db.executeAsync('DELETE FROM keyvaluepairs WHERE record_key = ?;', [key]),
+
+    /**
+    * Clears absolutely everything from storage
+    * @returns {Promise<void>}
+    */
+    clear: () => db.executeAsync('DELETE FROM keyvaluepairs;', []),
```
yuwenmemon, line 115: We ran into an issue where clearing with `multiSet` [here](https://github.com/Expensify/react-native-onyx/blob/c26f6e5defdeb8183ff205053a20bde943b6f5d1/lib/Onyx.js#L1061) in Onyx caused iOS to crash because we were setting keys to `null`. 

I was looking through the AsyncStorage docs and saw there is a [`multiRemove`](https://reactnative.dev/docs/asyncstorage#multiremove) which I think would work better in this situation.

For SQLite I believe it would be as simple as:
```JS
`DELETE FROM keyvaluepairs WHERE record_key IN (${keysToRemove.join(',')});`;
```

Thoughts?

yuwenmemon: #expensify-open-source [thread](https://expensify.slack.com/archives/C01GTK53T8Q/p1671583867011199) for context.

Szymon20000: Sounds good! Do you want me to add that method?

yuwenmemon: If no one else is opposed 🙂

marcaaron: I don't have any strong feelings about why this shouldn't exist - but looks like it would need to be implemented in the [web storage](https://github.com/Expensify/react-native-onyx/blob/main/lib/storage/providers/LocalForage.js) as well.

mountiny: Maybe this could be a follow up given we need to add it to both providers

mountiny (approved):
Thanks for the persistent work on this one and glad we managed to get this tested locally. I think migration would be handy in case that would be super hard (otherwise it is probably not worth as users can sign out/ sign back)

lib/storage/providers/SQLiteStorage.js
kidroca: I think, since it's going to be a smaller change to rework how `clear` works it can be a PR that is merged and tested before the current PR 

But I'm also OK moving forward with this PR the way it is, as long as we're addressing the `clear` problem in a separate PR

kidroca (approved):

kidroca:
> Thanks for the persistent work on this one and glad we managed to get this tested locally. I think migration would be handy in case that would be super hard (otherwise it is probably not worth as users can sign out/ sign back)

IMO a migration would be far from straightforward to write and I commented about it here: https://github.com/Expensify/App/issues/10793#issuecomment-1363853454

Also just to be clear we're talking about a migration script written in App, like these: https://github.com/Expensify/App/tree/main/src/libs/migrations (so it's not something that can be done in the current PR, but in the one for App)

marcaaron:
>Also just to be clear we're talking about a migration script written in App, like these: https://github.com/Expensify/App/tree/main/src/libs/migrations (so it's not something that can be done in the current PR, but in the one for App)

Yes and agree the migration would not be super straightforward since both storage systems would need to be accessible. I personally think it would be OK to just have people get logged out and have to log back in. The main downside is that any queued messages they have would not be sent, but this is a one time event and not worth optimizing for.

lib/storage/providers/SQLiteStorage.js
marcaaron: yep, agree with @kidroca and @mountiny I think this PR has been open for long enough and we can do a follow up to look at improving `clear()`.

PR #213, opened 2022-12-06
Update Onyx.clear to accept passed in default values

yuwenmemon:
@neil-marcellini please review

### Details
Add an `additionalDefaults` param to `clear` to allow it to accept passed in default values

### Related Issues
https://github.com/Expensify/App/issues/13062

### Automated Tests
Added tests in [onyxClearWebStorageTest.js](https://github.com/Expensify/react-native-onyx/blob/317a6172b9c12ad435636af7c1e96e9387858d01/tests/unit/onyxClearWebStorageTest.js#L113-L151) and [onyxClearNativeStorageTest.js](https://github.com/Expensify/react-native-onyx/blob/317a6172b9c12ad435636af7c1e96e9387858d01/tests/unit/onyxClearNativeStorageTest.js#L111-L148)

### Linked PRs
https://github.com/Expensify/App/pull/13384

### DIFF
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -1025,13 +1025,18 @@ function initializeWithDefaultKeyStates() {
  * Storage.setItem() from Onyx.clear() will have already finished and the merged
  * value will be saved to storage after the default value.
  *
+ * You may pass in additionalDefaults to set/preserve onyx data that you want to keep
+ * after clearing.
+ *
+ * @param {Object} additionalDefaults additional default values for ONYXKEYS keys
  * @returns {Promise<void>}
  */
-function clear() {
+function clear(additionalDefaults = {}) {
+    const keyStatesToPreserve = fastMerge(defaultKeyStates, additionalDefaults);
     return getAllKeys()
         .then((keys) => {
             _.each(keys, (key) => {
-                const resetValue = lodashGet(defaultKeyStates, key, null);
+                const resetValue = lodashGet(keyStatesToPreserve, key, null);
                 cache.set(key, resetValue);
                 notifySubscribersOnNextTick(key, resetValue);
             });
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -661,7 +661,10 @@ function getCollectionDataAndSendAsObject(matchingKeys, mapping) {
  * @param {Boolean} [mapping.initWithStoredValues] If set to false, then no data will be prefilled into the
  *  component
  * @param {Boolean} [mapping.waitForCollectionCallback] If set to true, it will return the entire collection to the callback as a single object
- * @param {String|Function} [mapping.selector] THIS PARAM IS ONLY USED WITH withOnyx(). If included, this will be used to subscribe to a subset of an Onyx key's data. If the selector is a string, the selector is passed to lodashGet on the sourceData. If the selector is a function, the sourceData is passed to the selector and should return the simplified data. Using this setting on `withOnyx` can have very positive performance benefits because the component will only re-render when the subset of data changes. Otherwise, any change of data on any property would normally cause the component to re-render (and that can be expensive from a performance standpoint).
+ * @param {String|Function} [mapping.selector] THIS PARAM IS ONLY USED WITH withOnyx(). If included, this will be used to subscribe to a subset of an Onyx key's data.
+ *       If the selector is a string, the selector is passed to lodashGet on the sourceData. If the selector is a function, the sourceData is passed to the selector and should return the
+ *       simplified data. Using this setting on `withOnyx` can have very positive performance benefits because the component will only re-render when the subset of data changes.
+ *       Otherwise, any change of data on any property would normally cause the component to re-render (and that can be expensive from a performance standpoint).
  * @returns {Number} an ID to use when calling disconnect
  */
 function connect(mapping) {
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -1025,18 +1025,20 @@ function initializeWithDefaultKeyStates() {
  * Storage.setItem() from Onyx.clear() will have already finished and the merged
  * value will be saved to storage after the default value.
  *
- * You may pass in additionalDefaults to set/preserve onyx data that you want to keep
- * after clearing.
+ * You may pass in keyStatesToPreserve to preserve Onyx keys you don't want to clear.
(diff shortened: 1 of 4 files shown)


lib/Onyx.js
```diff
@@ -1025,13 +1025,18 @@ function initializeWithDefaultKeyStates() {
...
  *
+ * You may pass in additionalDefaults to set/preserve onyx data that you want to keep
+ * after clearing.
+ *
+ * @param {Object} additionalDefaults additional default values for ONYXKEYS keys
  * @returns {Promise<void>}
  */
-function clear() {
+function clear(additionalDefaults = {}) {
+    const keyStatesToPreserve = fastMerge(defaultKeyStates, additionalDefaults);
```
marcaaron, line 1035:
```suggestion
function clear(keyStatesToPreserve = {}) {
    const defaultValues = fastMerge(defaultKeyStates, keyStatesToPreserve);
```

I think this is clearer naming since the default keys are not necessarily things we are preserving.

marcaaron: I don't know if we want to address this here, but I find the existing code below this strange. We are clearing the storage entirely... won't this remove the default and preserved states from the device storage? If we want to "preserve" something would a more correct approach be to just skip it's deletion?

yuwenmemon: Yes, I think that would be a better approach. Kind of been thinking about switching it over to this too as I've been looking at the /App side of things.

neil-marcellini: Maybe the idea with keeping the defaultKeyStates in the cache only was that the defaultKeyStates are not saved to storage when Onyx is initialized?

I'm fine with switching to an approach where default keys are not cleared from the cache or storage.

yuwenmemon: Sounds good - gonna push that change I tested it with the linked PR and it works just the same :+1:

lib/Onyx.js
```diff
@@ -1025,12 +1028,19 @@ function initializeWithDefaultKeyStates() {
...
+ *
+ * @param {Array} keyStatesToPreserve ONYXKEYS values you want to keep
  * @returns {Promise<void>}
  */
-function clear() {
+function clear(keyStatesToPreserve = []) {
     return getAllKeys()
         .then((keys) => {
             _.each(keys, (key) => {
+                if (_.contains(keyStatesToPreserve, key)) {
```
neil-marcellini, line 1040: This can work, but I thought the idea going forward was to prevent default key states and keys to preserve from being cleared from storage.

tests/unit/onyxClearNativeStorageTest.js
```diff
@@ -105,4 +106,41 @@ describe('Set data while storage is clearing', () => {
                 return expect(storedValue).resolves.toBeNull();
             });
     });
+
+    it('should preserve the value of any keyStatesToPreserve passed in', () => {
+        expect.assertions(6);
+
+        // Given that Onyx has a value, and we have a variable listening to that value
+        Onyx.set(ONYX_KEYS.REGULAR_KEY, SET_VALUE);
```
neil-marcellini, line 114: NAB: It's weird to me to set the key before Onyx.connect.

lib/Onyx.js
yuwenmemon: There's just one issue though. For some reason the Storage assertion in the Native Storage test is _not_ passing while the Web version is 🤔 

Have to log off for a bit but will try to come back to this and see what's what.

lib/Onyx.js
yuwenmemon: > Sounds good - gonna push that change I tested it with the linked PR and it works just the same 👍

Uhhh... nvm I don't think I was quite understanding what I was doing 🤦‍♂️ 

I _think_ I understand this a bit clearer now...

> I'm fine with switching to an approach where default keys are not cleared from the cache or storage.

So I'm guessing this would necessitate removing each value one by one?

lib/Onyx.js
neil-marcellini: I think clearing the storage and then calling multiSet is a good approach.

neil-marcellini (commented):
Looking good, please let me know when it's ready for another review.

lib/Onyx.js
```diff
@@ -1025,17 +1028,27 @@ function initializeWithDefaultKeyStates() {
...
-                cache.set(key, resetValue);
-                notifySubscribersOnNextTick(key, resetValue);
+                if (_.contains(keyStatesToPreserve, key)) {
+                    keyValuesToPreserve.push([key, get(key)]);
+                } else {
+                    const resetValue = lodashGet(defaultKeyStates, key, null);
+                    cache.set(key, resetValue);
+                    notifySubscribersOnNextTick(key, resetValue);
+                }
+            });
```
marcaaron, line 1048:
```suggestion
                    keyValuesToPreserve.push([key, get(key)]);
                    return;
                }
                
                const resetValue = lodashGet(defaultKeyStates, key, null);
                cache.set(key, resetValue);
                notifySubscribersOnNextTick(key, resetValue);
            });
```

marcaaron: A bit more readable to me

lib/Onyx.js
```diff
@@ -1025,17 +1028,27 @@ function initializeWithDefaultKeyStates() {
...
         .then((keys) => {
+            const keyValuesToPreserve = [];
             _.each(keys, (key) => {
-                const resetValue = lodashGet(defaultKeyStates, key, null);
-                cache.set(key, resetValue);
-                notifySubscribersOnNextTick(key, resetValue);
+                if (_.contains(keyStatesToPreserve, key)) {
+                    keyValuesToPreserve.push([key, get(key)]);
+                } else {
+                    const resetValue = lodashGet(defaultKeyStates, key, null);
```
marcaaron, line 1044: Should we also add the defaultKeyStates into the `keyValuesToPreserve` array?

marcaaron: I can't think of a good reason to remove them from storage without adding them back in like we are doing with the key/values we want to preserve.

yuwenmemon:
I ended up branching off this PR to try out the multiSet idea we talked about [here in Slack](https://expensify.slack.com/archives/C01GTK53T8Q/p1670440304879179). Gonna close this out in favor of that one.

PR #220, opened 2023-01-03
Fix race condition with clearing Onyx and the SyncQueue

tgolen:
The problem is explained in detail in [this comment.](https://github.com/Expensify/App/issues/13884#issuecomment-1370202728)

### Related Issues
Related to https://github.com/Expensify/App/issues/13884

### Automated Tests
Mostly handled in existing unit tests. Some new tests being added.

### Linked PRs
Will be merged into App with https://github.com/Expensify/react-native-onyx/pull/220

### DIFF
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -1102,7 +1102,9 @@ function clear(keysToPreserve = []) {
                 notifyCollectionSubscribersOnNextTick(key, value);
             });
 
-            return Storage.multiRemove(keyValuesToReset);
+            // Extract all the keys and clear them out
+            const keysToReset = _.map(keyValuesToReset, keyValueToReset => keyValueToReset[0]);
+            return Storage.multiRemove(keysToReset);
         });
 }
 
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -1050,7 +1050,7 @@ function clear(keysToPreserve = []) {
             const keyValuesToReset = [];
             const defaultKeys = _.keys(defaultKeyStates);
             const keyValuesToPreserve = _.reduce(keysToPreserve, (finalArray, keyToPreserve) => {
-                finalArray.push([keyToPreserve, cache.getValue(keyToPreserve)])
+                finalArray.push([keyToPreserve, cache.getValue(keyToPreserve)]);
                 return finalArray;
             }, []);
 
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -1049,6 +1049,9 @@ function clear(keysToPreserve = []) {
         .then((keys) => {
             const keyValuesToReset = [];
             const defaultKeys = _.keys(defaultKeyStates);
+
+            // Get all the values for the keys that need to be preserved. These key/value pairs will be set
+            // in Onyx after the database is cleared().
             const keyValuesToPreserve = _.reduce(keysToPreserve, (finalArray, keyToPreserve) => {
                 finalArray.push([keyToPreserve, cache.getValue(keyToPreserve)]);
                 return finalArray;
@@ -1106,7 +1109,11 @@ function clear(keysToPreserve = []) {
                 notifyCollectionSubscribersOnNextTick(key, value);
             });
 
-            return Storage.clear(defaultKeyValuePairs.concat(keyValuesToPreserve));
+            // Call clear() and make sure that the default key/values and the key/values from the parameter
+            // are preserved in storage. This makes sure to always leave storage in a state that contains
+            // all the default values and any additional values that we want to remain after the database is cleared.
+            return Storage.clear()
+                .then(() => Storage.multiSet(defaultKeyValuePairs.concat(keyValuesToPreserve)));
         });
 }
 
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -1110,7 +1110,7 @@ function clear(keysToPreserve = []) {
             // are preserved in storage. This makes sure to always leave storage in a state that contains
             // all the default values and any additional values that we want to remain after the database is cleared.
             return Storage.clear()
-                .then(() => Storage.multiSet(defaultKeyValuePairs.concat(keyValuesToPreserve)));
+                .then(() => Storage.multiSet([...defaultKeyValuePairs, ...keyValuesToPreserve]));
         });
 }
 
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -1102,9 +1102,8 @@ function clear(keysToPreserve = []) {
                 notifyCollectionSubscribersOnNextTick(key, value);
             });
 
-            return Storage.multiSet(defaultKeyValuePairs).then(() => {
-                return Storage.multiRemove(keysToClear);
-            });
+            return Storage.multiSet(defaultKeyValuePairs)
+                .then(() => Storage.multiRemove(keysToClear));
         });
 }
 
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -1102,9 +1102,9 @@ function clear(keysToPreserve = []) {
                 notifyCollectionSubscribersOnNextTick(key, value);
             });
 
-            // Extract all the keys and clear them out
-            const keysToReset = _.map(keyValuesToReset, keyValueToReset => keyValueToReset[0]);
-            return Storage.multiRemove(keysToReset);
+            return Storage.multiSet(defaultKeyValuePairs).then(() => {
+                return Storage.multiRemove(keysToClear);
+            });
         });
 }
 
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -1052,10 +1052,7 @@ function clear(keysToPreserve = []) {
 
             // Get all the values for the keys that need to be preserved. These key/value pairs will be set
             // in Onyx after the database is cleared().
-            const keyValuesToPreserve = _.reduce(keysToPreserve, (finalArray, keyToPreserve) => {
-                finalArray.push([keyToPreserve, cache.getValue(keyToPreserve)]);
-                return finalArray;
-            }, []);
+            const keyValuesToPreserve = _.map(keysToPreserve, key => [key, cache.getValue(key)]);
 
             // The only keys that should not be cleared are:
             // 1. Anything specifically passed in keysToPreserve (because some keys like language preferences, offline
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -1102,7 +1102,7 @@ function clear(keysToPreserve = []) {
                 notifyCollectionSubscribersOnNextTick(key, value);
             });
 
-            return Storage.multiSet(keyValuesToReset);
+            return Storage.multiRemove(keyValuesToReset);
         });
 }
 
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -1049,6 +1049,10 @@ function clear(keysToPreserve = []) {
         .then((keys) => {
             const keyValuesToReset = [];
             const defaultKeys = _.keys(defaultKeyStates);
+            const keyValuesToPreserve = _.reduce(keysToPreserve, (finalArray, keyToPreserve) => {
+                finalArray.push([keyToPreserve, cache.getValue(keyToPreserve)])
+                return finalArray;
+            }, []);
 
             // The only keys that should not be cleared are:
             // 1. Anything specifically passed in keysToPreserve (because some keys like language preferences, offline
@@ -1102,8 +1106,7 @@ function clear(keysToPreserve = []) {
                 notifyCollectionSubscribersOnNextTick(key, value);
             });
 
-            return Storage.multiSet(defaultKeyValuePairs)
-                .then(() => Storage.multiRemove(keysToClear));
+            return Storage.clear(defaultKeyValuePairs.concat(keyValuesToPreserve));
         });
 }
 
--- a/lib/SyncQueue.js
+++ b/lib/SyncQueue.js
@@ -25,9 +25,6 @@ export default class SyncQueue {
      * Stop the queue from being processed and clear out any existing tasks
      */
     abort() {
-        if (!this.isProcessing || this.queue.length === 0) {
-            return;
-        }
         this.queue = [];
         this.isProcessing = false;
     }
--- a/lib/SyncQueue.js
+++ b/lib/SyncQueue.js
@@ -21,6 +21,17 @@ export default class SyncQueue {
         this.run = run;
     }
 
+    /**
+     * Stop the queue from being processed and clear out any existing tasks
+     */
+    abort() {
+        if (!this.isProcessing || this.queue.length === 0) {
+            return;
+        }
+        this.queue = [];
+        this.isProcessing = false;
+    }
+
     process() {
         if (this.isProcessing || this.queue.length === 0) {
             return;
--- a/lib/storage/WebStorage.js
+++ b/lib/storage/WebStorage.js
@@ -27,8 +27,18 @@ const webStorage = {
             .then(() => raiseStorageSyncEvent(key));
 
         // If we just call Storage.clear other tabs will have no idea which keys were available previously
-        // so that they can call keysChanged for them. That's why we iterate and remove keys one by one
-        this.clear = keysToPreserve => Storage.clear().then(() => Storage.multiSet(keysToPreserve));
+        // so that they can call keysChanged for them. That's why we iterate over every key that doesn't need to be
+        // preserved and raise a storage sync event for them
+        this.clear = keyValuesToPreserve => Storage.getAllKeys()
+            .then((allKeys) => {
+                const keysToPreserve = _.map(keyValuesToPreserve, keyValueToPreserve => keyValueToPreserve[0]);
+                return _.filter(allKeys, key => !_.contains(keysToPreserve, key));
+            })
+            .then(keysToRemove => _.map(keysToRemove, key => raiseStorageSyncEvent(key)))
+
+            // Clear out everything from storage and then set
+            .then(() => Storage.clear())
+            .then(() => Storage.multiSet(keyValuesToPreserve));
 
         // This listener will only be triggered by events coming from other tabs
         global.addEventListener('storage', (event) => {
--- a/lib/storage/WebStorage.js
+++ b/lib/storage/WebStorage.js
@@ -1,3 +1,8 @@
+/**
+ * This file is here to wrap LocalForage with a layer that provides data-changed events like the ones that exist
+ * when using LocalStorage APIs in the browser. These events are great because multiple tabs can listen for when
+ * data changes and then stay up-to-date with everything happening in Onyx.
+ */
 import _ from 'underscore';
 import Storage from './providers/LocalForage';
 
@@ -35,10 +40,7 @@ const webStorage = {
                 return _.filter(allKeys, key => !_.contains(keysToPreserve, key));
             })
             .then(keysToRemove => _.map(keysToRemove, key => raiseStorageSyncEvent(key)))
-
-            // Clear out everything from storage and then set
-            .then(() => Storage.clear())
-            .then(() => Storage.multiSet(keyValuesToPreserve));
+            .then(() => Storage.clear(keyValuesToPreserve));
 
         // This listener will only be triggered by events coming from other tabs
         global.addEventListener('storage', (event) => {
--- a/lib/storage/WebStorage.js
+++ b/lib/storage/WebStorage.js
@@ -28,9 +28,7 @@ const webStorage = {
 
         // If we just call Storage.clear other tabs will have no idea which keys were available previously
         // so that they can call keysChanged for them. That's why we iterate and remove keys one by one
-        this.clear = () => Storage.getAllKeys()
-            .then(keys => _.map(keys, key => this.removeItem(key)))
-            .then(tasks => Promise.all(tasks));
+        this.clear = keysToPreserve => Storage.clear().then(() => Storage.multiSet(keysToPreserve));
 
         // This listener will only be triggered by events coming from other tabs
         global.addEventListener('storage', (event) => {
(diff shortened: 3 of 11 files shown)


yuwenmemon:
@tgolen have you pushed the latest updates to this branch? Was looking into creating the test laid out here: https://github.com/Expensify/App/issues/13884#issuecomment-1370828497 but unsure if we're supposed to have any changes in [LocalForage](https://github.com/Expensify/react-native-onyx/blob/c58061ca81470ecac7e87ab96520dc30056889ff/lib/storage/providers/LocalForage.js#L77-L81) or not?

tgolen:
I've updated the remote branch now so that it's up to date with all my
(untested) changes. There is one small change in LocalForage.

tests/unit/storage/providers/LocalForageProviderTest.js
```diff
@@ -104,4 +104,26 @@ describe('storage/providers/LocalForage', () => {
                     });
             });
     });
+
+    it('clear', () => {
```
yuwenmemon, line 108: @kidroca - @tgolen asked me to help with implementing a test like what you described [here](https://github.com/Expensify/App/issues/13884#issuecomment-1370828497) - just want to double-check check I'm capturing this right since I'm a bit new to jest and it's "fake time"...

kidroca: Almost there.
What you've got right now won't resolve the promise after 1000ms
Also `setItemTimedOut` should be a function - right now it's just the result of a `setTimeout()` call - a numeric ID

I think you just need to change the mock like so:

```js
// Given an implementation of setItem that resolves after 1000ms
localForage.setItem = jest.fn(() => new Promise(
  (resolve) =>  setTimeout(resolve, 1000))
);
```

1. `setItem` should return a promise 
2. we make the returned promise resolve after `1000ms` - it would only resolve after we advance timers

lib/storage/WebStorage.js
```diff
@@ -27,10 +32,15 @@ const webStorage = {
...
-            .then(keys => _.map(keys, key => this.removeItem(key)))
-            .then(tasks => Promise.all(tasks));
+        // so that they can call keysChanged for them. That's why we iterate over every key that doesn't need to be
+        // preserved and raise a storage sync event for them
+        this.clear = keyValuesToPreserve => Storage.getAllKeys()
+            .then((allKeys) => {
+                const keysToPreserve = _.map(keyValuesToPreserve, keyValueToPreserve => keyValueToPreserve[0]);
+                return _.filter(allKeys, key => !_.contains(keysToPreserve, key));
+            })
+            .then(keysToRemove => _.map(keysToRemove, key => raiseStorageSyncEvent(key)))
```
kidroca, line 42: If these are all the keys to be removed, why do we `raiseStorageSyncEvent` for each, before they are even cleared?

Why do we use `_.map` to iterate a list? 
`raiseStorageSyncEvent` does not return a promise, but even if it did, you'd still need to have something like

```js
.then(keysToRemove => {
  const promises = _.map(keysToRemove, key => raiseStorageSyncEvent(key));
  return Promise.all(promises);
})
```

I think it would be better to create a `CLEAR_ONYX` event, similarly to how `SYNC_ONYX` works

Instead of having `this.clear` work with `keyValuesToPreserve` it can work with just the `keysToPreserve`
This would allow us to raise a `CLEAR_ONYX` event with `keysToPreserve` as value 
Then before clearing storage
1. storage would multi get `keysToPreserve` 
2. then call storage clear
3. and finally multi set the result of `multiget` at point 1

This should be simpler and more efficient that iterating all the keys

lib/storage/providers/LocalForage.js
```diff
@@ -75,10 +75,17 @@ const provider = {
...
+     *      is here to provide a performance boost when deleting a majority of keys, but wanting to keep only a few.
+     *      Example: It's faster than doing a big multiRemove() which is just a loop that calls removeItem().
      * @returns {Promise<void>}
      */
-    clear: localforage.clear,
+    clear(keyValuesToPreserve) {
+        this.setItemQueue.abort();
+        return localforage.clear()
+            .then(() => Storage.multiSet(keyValuesToPreserve));
+    },
```
kidroca, line 88: Following my prior comment this can become 

```js
    clear(keysToPreserve) {
        this.setItemQueue.abort();
        return this.multiGet(keysToPreserve)
            .then(keyValuesToPreserve =>
                localforage.clear()
                    .then(() => this.multiSet(keyValuesToPreserve))
             )
    },
```

kidroca (commented):
Added a few notes after examining the changes

lib/SyncQueue.js
```diff
@@ -21,6 +21,17 @@ export default class SyncQueue {
         this.run = run;
     }
 
+    /**
+     * Stop the queue from being processed and clear out any existing tasks
+     */
+    abort() {
+        if (!this.isProcessing || this.queue.length === 0) {
+            return;
+        }
```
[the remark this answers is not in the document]
tgolen: OK, sounds good. I'll go ahead and remove it entirely. I agree with you that it probably doesn't make sense to have it and is pretty useless.

lib/storage/WebStorage.js
tgolen: > Why do we use _.map to iterate a list?

You're right, it's not necessary, it's just what we were using before and it only needs to be a loop at a bare minimum.

> raiseStorageSyncEvent does not return a promise, but even if it did, you'd still need to have something like

Yes, that's almost exactly what was there before, as you can see in the diff, and it doesn't make any sense considering that `raiseStorageSyncEvent` doesn't return anything.

I'm not totally following the idea of `CLEAR_ONYX`, but the rest of what you said gave me some thoughts of a different way I can code this to keep this logic from having to deal with `keysToPreserve` at all. Let me think on it and I'll send in an update.

tgolen: > why do we raiseStorageSyncEvent for each, before they are even cleared?

Oh, to answer this... I am not sure. It's what was there before so I was trying to preserve the previous functionality (which I think was all added by yourself 😁 ).

I think it's so that other browser tabs will know that all the keys have been removed. Technically, it could happen either before or after calling `clear()`. If it's removed entirely, then other open tabs wouldn't sign out and they would retain all their existing data.

tests/unit/storage/providers/LocalForageProviderTest.js
yuwenmemon: Ah hm... in actuality, we want to measure how many times the promise is _resolved_ then, right? But there's no other way to do that other than having a callback get called when the promise resolves?

So my thought was to do something like this instead:
```JS
    it('clear', () => {
        // Use fake timers, so we can manipulate time at our will for this test.
        jest.useFakeTimers();

        // Given a mocked implementation of setItem that calls a callback function after 1000ms
        const callback = jest.fn();
        localforage.setItem = jest.fn(() => new Promise(resolve => setTimeout(resolve, 1000)).then(callback));

        // When we call setItem 5 times, but then call clear after only 1000ms
        for (let i = 0; i < 5; i++) {
            StorageProvider.setItem(`key${i}`, `value${i}`);
        }

        // We should not have called the mocked callback for setItem yet because we have not advanced the timer.
        expect(callback).not.toHaveBeenCalled();

        jest.advanceTimersByTime(1000);
        StorageProvider.clear();
        jest.advanceTimersByTime(4000);

        // The mocked callback for setItem should only have been called once since all other calls were aborted when we called clear()
        expect(callback).toHaveBeenCalledTimes(1);
    });
```

However I'm seeing that's not quite working as I would expect 🤔 
```
Error: expect(jest.fn()).toHaveBeenCalledTimes(expected)

Expected number of calls: 1
Received number of calls: 0
```

tests/unit/onyxClearWebStorageTest.js
```diff
@@ -50,26 +50,6 @@ describe('Set data while storage is clearing', () => {
         return Onyx.clear();
     });
 
-    it('should persist the value of Onyx.merge when called between the cache and storage clearing', () => {
```
tgolen, line 53: @neil-marcellini I saw you added these tests, and as far as I can tell, they are no longer necessary because the behavior of `clear()` has been changed. Can you please take a look at this and make sure that I am not introducing regressions that I'm not aware of? 

Basically, the way I read this is that the problem only happened because we were done something like this:
```
Onyx.clear();
Onyx.set(whatever);
```

My PR here changes this behavior so that no one should ever be doing that anymore. It should now become:

```
Onyx.clear(keysToPreserve);
```
So that's why I think it's OK to remove these tests since they no longer test a valid use case.

tests/unit/storage/providers/LocalForageProviderTest.js
yuwenmemon: Also, I noticed you had the comment [here](https://github.com/Expensify/react-native-onyx/blob/13eada48dc6d421f367df35fab34c001314d4e40/tests/unit/storage/providers/LocalForageProviderTest.js#L16) noting that "fake timers cause promises to hang" - wonder if that might be the same reason for the issue I'm seeing?

yuwenmemon: (Btw, just tested this again using real timers and I don't see this problem. So I _do_ have a suspicion this is somehow related to FakeTimers)

tests/unit/onyxClearWebStorageTest.js
neil-marcellini: These tests make sure that the following flow works for keys with a default key state, like the session.
1. Sign out the user with Onyx.clear()
2. Sign in a new user (before Onyx.clear finishes), and merge new data on the session key
3. Ensure that Onyx.clear doesn't set a default key state on the session key and wipe out the value saved in step 2.

I think it might still be valuable to test calling `Onyx.clear()` and then `Onyx.merge` on a key with a default state.

tgolen: OK, thanks! My point is that this...

> Sign in a new user (before Onyx.clear finishes),

...isn't possible to do. `Onyx.clear()` takes less that 200ms to complete now on an account with 4000 reports, as well as we are no longer optimistically redirecting to the sign-on page. We are waiting until `Onyx.clear()` is finished before the redirect happens.

If there is a valid use case of calling `clear -> merge`, I could see having a test, but I don't know if any time that would ever be a valid case. Do you have a specific one in mind? If that does exist in the code, it should be rewritten like I showed above.

neil-marcellini: > `Onyx.clear()` takes less that 200ms to complete now on an account with 4000 reports,

Ah ok nice. If that's the case for all platforms then I agree that we don't need this test any more 👍.



> we are no longer optimistically redirecting to the sign-on page. We are waiting until `Onyx.clear()` is finished before the redirect happens.

I don't think we are based on the code [here](https://github.com/Expensify/App/blob/14eee462c7da856fe5ef38a5e5343d655e9b62b4/src/libs/actions/SignInRedirect.js#L45-L46). As soon as the session key is cleared we are signed out, and that can happen before Onyx finishes clearing.

tgolen: > As soon as the session key is cleared we are signed out, and that can happen before Onyx finishes clearing.

Ah yes, you're technically right. However, the session key is removed at the same time as all other keys now (because it is once again using `LocalForage.clear()` under the hood rather than using a `multiSet()` or `multiRemove()`).

tests/unit/storage/providers/LocalForageProviderTest.js
kidroca: > Ah hm... in actuality, we want to measure how many times the promise is _resolved_ then, right?

I guess it might look like that from a certain point of view, but it's not necessary

If we make 5 `StorageProvider.setItem` calls they should be queued (this is what we test)
This should immediately call `localForage.setItem` (our mock) 1 time
then wait for the returned promise to resolve in order to move to the next item in the queue
In our test the queue would be emptied so then next 4 calls to `localForage.setItem` shouldn't happen

> Also, I noticed you had the comment [here](https://github.com/Expensify/react-native-onyx/blob/13eada48dc6d421f367df35fab34c001314d4e40/tests/unit/storage/providers/LocalForageProviderTest.js#L16) noting that "fake timers cause promises to hang" - wonder if that might be the same reason for the issue I'm seeing?

I'm can't remember what lead me to that conclusion...
Maybe we can try and update `jest` 

This might also need calls to `waitForPromisesToResolve`

> (Btw, just tested this again using real timers and I don't see this problem. So I _do_ have a suspicion this is somehow related to FakeTimers)

I'm not sure how you test this with real timers, but if you don't wait for the timers to expire the test should appear as passing successfully 

---

There's an alternative way to test this without using timers

```jsx
let i = 
const task = createDeferredTask();

localforage.setItem = jest.fn()
   . mockReturnValue(Promise.resolve()) // default value
   . mockReturnValueOnce(task.promise) // first call

for (let i = 0; i < 5; i++) {
  StorageProvider.setItem(`key${i}`, `value${i}`);
}

// at this point `localForage.setItem` should be called once, but we control when it resolves 
// and we keep it unresolved 

// this should queue any calls that follow, so we don't expect more than 1 `localForage.setItem` call
StorageProvider.clear();

return waitForPromisesToResolve().then(() => {
  expect(localForage.setItem).toHaveBeenCalledTimes(1);
})
```

lib/storage/WebStorage.js
kidroca: > I think it's so that other browser tabs will know that all the keys have been removed. Technically, it could happen either before or after calling `clear()`

Yeah I was just confused about that part
I thought that because we call it before we clear storage, other tabs might have a chance to re-sync keys from storage and read values before we delete them 

So I think it would be safer to first call `Storage.clear()` and only then raise the storage sync events 

> I'm not totally following the idea of `CLEAR_ONYX`

What you've got now seems to be less changes overall, so if it there are no issues it doesn't matter

tgolen:
Ah, OK. Thanks! I'll switch the order of these then and call it a day.

tests/unit/storage/providers/LocalForageProviderTest.js
yuwenmemon: Hm, yes I do see some issues with Jest's fake timers and Promises being reported by others like [here](https://stackoverflow.com/questions/52177631/jest-timer-and-promise-dont-work-well-settimeout-and-async-function) and [here](https://stackoverflow.com/questions/51126786/jest-fake-timers-with-promises).

Although I can't quite wrap my head around your alternative way... is there a way to demonstrate the "before" case (i.e. clear() doesn't call abort and thus setItem is called 5 times?)

tgolen:
OK, I've updated it to switch the order in which the storage event is raised

tests/unit/storage/providers/LocalForageProviderTest.js
kidroca: > is there a way to demonstrate the "before" case (i.e. clear() doesn't call abort and thus setItem is called 5 times?)

Running this test against the pre-PR code should produce exactly that error 
expected `setItem` to be called 1 time but it was called 5 times

> I can't quite wrap my head around your alternative way

1. We create a promise which we programmatically control when to resolve
2. We configure `localforage.setItem` to return this promise the first time it's called and to otherwise return resolved promises
3. We make 5 `StorageProvider.setItem` calls
4. This adds 5 items to the queue and starts executing the first `localForage.setItem`
5. This simulates the 1st `localForage.setItem` taking a random time. 
6. The important bit is we call `StorageProvider.clear()` while the first `localForage.setItem` isn't completed yet (and the queue is not empty, because the queue would only go to the next item after the 1st completes)
7. We call `StorageProvider.clear()` and only then we resolve the first `localForage.setItem` call
8. `waitForPromisesToResolve()` makes jest wait for any promises (even promises returned as the result of a promise) to resolve
9.If `StorageProvider.clear()` does not abort the queue, more `localForage.setItem` would be executed because `StorageProvider.setItem` works like this
      - `StorageProvider.setItem` (5th call)
          - Queue -> push -> process -> 
              - return promise for 5th call
                  - return promise for 4th call
                      - return promise for 3th call
                          - return promise for nth call

So the promise of the 5th call waits the promise of the 4th call and only then calls `localForage.setItem` and the promise of the 4th call waits the promise of the 3rd and so on

lib/storage/WebStorage.js
```diff
@@ -27,10 +32,23 @@ const webStorage = {
...
+                .then((keys) => {
+                    allKeys = keys;
+                })
+                .then(() => Storage.clear())
+                .then(() => {
+                    // Now that storage is cleared, the storage sync event can happen so that it is more of an atomic
+                    // action
+                    _.each(allKeys, raiseStorageSyncEvent);
+                });
+        };
```
kidroca, line 51: You can avoid the closure variable by doing something like

```js
        this.clear = () => {
            return Storage.getAllKeys()
                .then((allKeys) => {
                    return Storage.clear()
                        .then(() => _.each(allKeys, raiseStorageSyncEvent))
                });
        };
```

Although code has more nesting, it conveys the information you have to explain using comments here 

I guess this is opinion based and there's no hard rule on how to style such concepts

lib/storage/WebStorage.js
```diff
@@ -27,10 +32,23 @@ const webStorage = {
...
+
+            // They keys must be retreived before storage is cleared or else the list of keys would be empty
+            return Storage.getAllKeys()
+                .then((keys) => {
+                    allKeys = keys;
+                })
+                .then(() => Storage.clear())
+                .then(() => {
+                    // Now that storage is cleared, the storage sync event can happen so that it is more of an atomic
+                    // action
```
kidroca, line 48: Maybe it's only me, but it bothers me that this has to be a 2 line comments just because of 1 word

lib/Onyx.js
```diff
@@ -1050,6 +1050,13 @@ function clear(keysToPreserve = []) {
             const keyValuesToReset = [];
             const defaultKeys = _.keys(defaultKeyStates);
 
+            // Get all the values for the keys that need to be preserved. These key/value pairs will be set
+            // in Onyx after the database is cleared().
+            const keyValuesToPreserve = _.reduce(keysToPreserve, (finalArray, keyToPreserve) => {
+                finalArray.push([keyToPreserve, cache.getValue(keyToPreserve)]);
+                return finalArray;
+            }, []);
+
```
kidroca, line 1059: Instead of using `_.reduce` it's simpler to just use `_.map`

```js
const keyValuesToPreserve = _.map(keysToPreserve, key => [key, cache.getValue(key)]);
```

We remap every key we want to preserve to a key-value pair, in the end we still end up with the same list of key-value pairs

lib/Onyx.js
```diff
@@ -1102,7 +1109,11 @@ function clear(keysToPreserve = []) {
                 notifyCollectionSubscribersOnNextTick(key, value);
             });
 
-            return Storage.multiSet(keyValuesToReset);
+            // Call clear() and make sure that the default key/values and the key/values from the parameter
+            // are preserved in storage. This makes sure to always leave storage in a state that contains
+            // all the default values and any additional values that we want to remain after the database is cleared.
+            return Storage.clear()
+                .then(() => Storage.multiSet(defaultKeyValuePairs.concat(keyValuesToPreserve)));
```
kidroca, line 1116: Why do we use `.concat` instead of spreading (like we do with props)?

```js
Storage.multiSet([...defaultKeyValuePairs, ...keyValuesToPreserve])
```

both `.concat` and spreading result in a new array

kidroca (commented):
Added a few NAB comments

tests/unit/storage/providers/LocalForageProviderTest.js
yuwenmemon: Ah, amazing explanation - thanks for taking the time to explain that to me I appreciate it! Okay, let's go with this then since it doesn't require us to update jest in this PR.

lib/Onyx.js
tgolen: I try to stay away from using the spread operator on code that needs to be super performant. Babel will transpile to something that's pretty slow. It's probably not really necessary here, but it's the habit I've been in.

lib/Onyx.js
yuwenmemon: Agreed, makes things much easier to read.

yuwenmemon (changes requested):
One small comment about using _.map

lib/Onyx.js
kidroca: This might be `true` for objects, if some of the keys are destructed and the rest are `...rest`, but for Arrays, babel works like this 

Code
```js
const arr3 = [...arr1, ...arr2];
```

Transpiled
```js
var arr3 = [].concat(arr1, arr2);
```

You can check that out in the live https://babeljs.io/repl though I think babel stopped transpiling the spread operator as it is supported natively by all alive browsers: https://caniuse.com/?search=spread%20operator

I think if we tell webpack/babel to ignore IE11 it would just stop transpiling the spread operator

tgolen: Ah, that would be great to stop transpiling it altogether. We are not interested in supporting IE11. I didn't know that about arrays. I'll go ahead and switch this to a spread operator.

lib/storage/WebStorage.js
tgolen: Makes sense, yeah. I prefer the clarity of my current code than trying to reason about those nested promises. Thanks!

yuwenmemon (approved):
Tested locally, works like a dream

lib/Onyx.js
kidroca: > Ah, that would be great to stop transpiling it altogether. We are not interested in supporting IE11. I didn't know that about arrays. I'll go ahead and switch this to a spread operator.

Here's a small PR regarding that: https://github.com/Expensify/react-native-onyx/pull/226

tgolen:
I'm going to go ahead and merge this then. Thanks!

yuwenmemon:
Is there an App PR to update this there?

PR #223, opened 2023-01-06
Remove null values

abdulrahuman5196:
### Details
When Onyx.set() is called with null value we are storing it in OnyxCache and AsyncStorage without removing the data from the both. This is leading to iOS crashes for AsyncStorage.multiMerge() operation. More information can be found at the root cause analysis here - https://github.com/Expensify/App/issues/13907#issuecomment-1368779185
Reviewer - @jasperhuangg 

### Related Issues
https://github.com/Expensify/App/issues/13907

### Automated Tests
I have added added test naming 'should remove key value from OnyxCache/Storage when set is called with null value' which verifies if the data is removed from OnyxCache and Storage.

### Linked PRs
None. Currently Expensify is using '1.0.29' which is lower than current onyx version '1.0.32'. We have to confirm if we can update Expensify package.json with change or should we wait. Anyways it shouldn't block this PR.

### DIFF
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -799,7 +799,7 @@ function notifyCollectionSubscribersOnNextTick(key, value) {
  * @return {Promise}
  */
 function remove(key) {
-    cache.drop(key)
+    cache.drop(key);
     notifySubscribersOnNextTick(key, null);
     return Storage.removeItem(key);
 }
@@ -846,9 +846,11 @@ function evictStorageAndRetry(error, onyxMethod, ...args) {
  * @returns {Promise}
  */
 function set(key, value) {
-    if (!value) {
-        return remove(key)
+    if (_.isNull(value)) {
+        Logger.logInfo(`Removing the key: ${key} as it called set() with null value`);
+        return remove(key);
     }
+
     // Logging properties only since values could be sensitive things we don't want to log
     Logger.logInfo(`set() called for key: ${key}${_.isObject(value) ? ` properties: ${_.keys(value).join(',')}` : ''}`);
 
@@ -1296,6 +1298,7 @@ const Onyx = {
     mergeCollection,
     update,
     clear,
+    getAllKeys,
     init,
     registerLogger: Logger.registerLogger,
     addToEvictionBlockList,
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -847,12 +847,8 @@ function evictStorageAndRetry(error, onyxMethod, ...args) {
  */
 function set(key, value) {
     if (_.isNull(value)) {
-        Logger.logInfo(`Removing the key: ${key} as it called set() with null value`);
         return remove(key);
     }
-
-    // Logging properties only since values could be sensitive things we don't want to log
-    Logger.logInfo(`set() called for key: ${key}${_.isObject(value) ? ` properties: ${_.keys(value).join(',')}` : ''}`);
 
     // eslint-disable-next-line no-use-before-define
     if (hasPendingMergeForKey(key)) {
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -799,8 +799,7 @@ function notifyCollectionSubscribersOnNextTick(key, value) {
  * @return {Promise}
  */
 function remove(key) {
-    // Cache the fact that the value was removed
-    cache.set(key, null);
+    cache.drop(key)
     notifySubscribersOnNextTick(key, null);
     return Storage.removeItem(key);
 }
@@ -847,6 +846,9 @@ function evictStorageAndRetry(error, onyxMethod, ...args) {
  * @returns {Promise}
  */
 function set(key, value) {
+    if (!value) {
+        return remove(key)
+    }
     // Logging properties only since values could be sensitive things we don't want to log
     Logger.logInfo(`set() called for key: ${key}${_.isObject(value) ? ` properties: ${_.keys(value).join(',')}` : ''}`);
 
--- a/tests/unit/onyxTest.js
+++ b/tests/unit/onyxTest.js
@@ -51,7 +51,6 @@ describe('Onyx', () => {
         })
         .then((keys) => {
             expect(keys.includes(ONYX_KEYS.OTHER_TEST)).toBe(false);
-            return Onyx.set(ONYX_KEYS.OTHER_TEST, 42);
         }));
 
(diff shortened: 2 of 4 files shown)


abdulrahuman5196:

abdulrahuman5196:
recheck

abdulrahuman5196:
recheck

lib/Onyx.js
```diff
@@ -847,6 +846,11 @@ function evictStorageAndRetry(error, onyxMethod, ...args) {
  * @returns {Promise}
  */
 function set(key, value) {
+    if (_.isNull(value)) {
+        Logger.logInfo(`Removing the key: ${key} as it called set() with null value`);
```
tgolen, line 850: I think we should remove this log line, and also the log a couple of lines below. I don't think they are useful at all.

tests/unit/onyxTest.js
```diff
@@ -24,11 +24,37 @@ Onyx.init({
 describe('Onyx', () => {
     let connectionID;
 
+    /** @type OnyxCache */
+    let cache;
+
+    beforeEach(() => {
+        cache = require('../../lib/OnyxCache').default;
```
tgolen, line 31: Why is this required for each test and not imported at the top of the file?

tgolen (changes requested):

lib/Onyx.js
abdulrahuman5196: >also the log a couple of lines below
To confirm. Do you mean this line

'Logger.logInfo(`set() called for key: ${key}${_.isObject(value) ? ` properties: ${_.keys(value).join(',')}` : ''}`);'  850/854? 

I added the above log line w.r.t the current log line present 850/854. I can remove both if they are not required.

tests/unit/onyxTest.js
abdulrahuman5196: The OnyxCache resolves to undefined if we import it directly from the top of the file. I took this reference of using require() from existing unit tests found here and used same test logic.

https://github.com/Expensify/react-native-onyx/blob/7d6cdc339acc17aa90f50f61e3e825364163146b/tests/unit/onyxCacheTest.js#L17
https://github.com/Expensify/react-native-onyx/blob/af02f32b74ae7c870b75e029d3a16fe4079442cd/tests/unit/onyxClearWebStorageTest.js#L32
https://github.com/Expensify/react-native-onyx/blob/5686d24c198204d023a78c0780bb37850335cce6/tests/unit/onyxClearNativeStorageTest.js#L25

lib/Onyx.js
abdulrahuman5196: Removed the log lines and added a new commit

tgolen: Yes, let's please remove both. I've been meaning to remove that existing log line for a while.

tests/unit/onyxTest.js
tgolen: Ah, that's interesting. If other tests are doing it, I guess that's OK for now. Thanks!

lib/Onyx.js
abdulrahuman5196: Done. Removed the log lines with the latest commit.

tests/unit/onyxTest.js
```diff
@@ -24,11 +24,37 @@ Onyx.init({
...
+    });
+
     afterEach(() => {
         Onyx.disconnect(connectionID);
         return Onyx.clear();
     });
 
+    it('should remove key value from OnyxCache/Storage when set is called with null value',
+        () => Onyx.set(ONYX_KEYS.OTHER_TEST, 42)
```
tgolen, line 40:
```suggestion
    it('should remove key value from OnyxCache/Storage when set is called with null value', () => Onyx.set(ONYX_KEYS.OTHER_TEST, 42)
```
Just to match the style elsewhere.

tgolen (changes requested):
Just this one minor change, please!

amyevans (approved):
@tgolen Do we have a process for back-porting patches to older versions (e.g. 1.0.29)? Perhaps in this case it doesn't really matter and we can just hold until we're ready to use the newest release in App. But also just curious in general 😄

tgolen:
We don't, no. I think we should just HOLD this until the SQLite change is live in App, then it shouldn't really be an issue. It probably won't be an issue anyway (is my guess), but I am just being a little paranoid since there are a few changes in flux right now. I think really all that means is that maybe this won't get merged until Monday. I'm expecting that the SQLite change will pass testing and be merged today.

tests/unit/onyxTest.js
abdulrahuman5196: Doen with latest commit

tests/unit/onyxTest.js
```diff
@@ -24,11 +24,36 @@ Onyx.init({
...
+        .then(() => {
+            // Checks if cache value is removed.
+            expect(cache.getAllKeys().length).toBe(0);
+
+            // When cache keys length is 0, we fetch the keys from storage.
+            return Onyx.getAllKeys();
+        })
+        .then((keys) => {
+            expect(keys.includes(ONYX_KEYS.OTHER_TEST)).toBe(false);
+            return Onyx.set(ONYX_KEYS.OTHER_TEST, 42);
```
tgolen, line 54: Is this necessary? I don't think this value is ever verified anywhere, and the test is over at this point and then Onyx is cleared in `afterEach()`

tgolen (approved):

tests/unit/onyxTest.js
abdulrahuman5196: I don't think its absolutely necessary. I just added it to reset the OTHER_TEST state which i removed. 
Anyways added a commit to remove it since removing it doesn't have any effect.

abdulrahuman5196:
Thank you. @tgolen  and @amyevans  for reviewing and approving the PR quickly. I think this PR is good to be merged from your end(I wont have permission) after SQLite changes comes in. Kindly let me know if I have to do anything from my end.

jasperhuangg (approved):
looks good! nice job adding tests too :)

amyevans (approved):

tgolen:
This is off HOLD and I've merged it. @abdulrahuman5196 you will need to do a PR for App to bump the version of Onyx and test it out there.

abdulrahuman5196:
@tgolen  Thank you. Created PR to bump version - https://github.com/Expensify/App/pull/14178

PR #226, opened 2023-01-11
 Update babel config

kidroca:
### Details

This PR reduces the number of (de)optimizations Babel does with the web bundle, by explicitly telling babel to use `default` targets
Current browsers cover a lot modern syntax features natively and it does not need to be transpiled 
If we don't explicitly tell Babel to target defaults, it targets an unnecessary wide range of dead/obsolete browser versions
(This is due to backwards compatibility with previous babel versions. Starting with Babel 8 the default behavior of will be `targets: defaults`)

- Remove unneeded dev dependencies
- Add web specific babel presets
- Use separate babel config for web

### Related Issues

N/A

### Automated Tests

Tests aren't needed because there are no source changes
Only the build configuration changed and it can be verified by running build/test scripts
- `npm build` should build a useable we bundle
- `npm test` should still run the tests
- importing Onyx in projects (like App) should not raise errors/warnings


package.json
```diff
@@ -46,18 +46,15 @@
...
     "@react-native-async-storage/async-storage": "^1.17.11",
     "@react-native-community/eslint-config": "^2.0.0",
     "@testing-library/jest-native": "^3.4.2",
     "@testing-library/react-native": "^7.0.2",
     "babel-eslint": "^10.1.0",
     "babel-jest": "^26.2.2",
     "babel-loader": "^8.2.5",
-    "babel-plugin-module-resolver": "^4.0.0",
-    "babel-plugin-react-native-web": "^0.13.5",
-    "babel-plugin-transform-class-properties": "^6.24.1",
```
kidroca, line 60: These aren't needed/used
- babel-plugin-module-resolver
- babel-plugin-react-native-web

These are already covered by `metro-react-native-babel-preset`
- babel-plugin-transform-class-properties
- @babel/plugin-proposal-class-properties

babel.config.js
```diff
@@ -1,8 +1,15 @@
...
+module.exports = (api) => {
+    const isWeb = api.caller(caller => caller && caller.target === 'web');
+    if (isWeb) {
+        return {
+            // Default browser list is a reasonable preset covering a wide list of non-dead browsers
+            // https://github.com/browserslist/browserslist#best-practices
+            targets: 'defaults',
+            presets: ['@babel/preset-env', '@babel/preset-react'],
+        };
+    }
```
kidroca, line 10: We do something similar in App, where the webpack build is using preset env and preset react

tgolen (commented):
Looks pretty good to me 👍 I tested it quickly and it seems like it does the trick. I see that in `web.development.js` it now has the spread operator.

kidroca:
I've tried this in App locally, but I'm not sure how to prove that it works in App - I think in order to test this in App someone need to publish new Onyx version and then I can create a PR demonstrating the new version works

tgolen:
I'm still confused on what the plan is for this PR.

> someone need to publish new Onyx version

Can you do this? Either let's move forward with merging this, or just close it out.

kidroca:
> > someone need to publish new Onyx version
> 
> Can you do this? Either let's move forward with merging this, or just close it out.

I can't publish, but I can fix the PR conflicts so we can merge this

tgolen:
I am pretty sure react-native-onyx is automatically published every time a PR is merged.

tgolen (approved):

cristipaval (approved):

PR #228, opened 2023-01-12
Protect against undefined values

tgolen:
cc @marcaaron @chrispader 

### Details
Explained in this thread: https://expensify.slack.com/archives/C01GTK53T8Q/p1673550288930829

This caused a crash in the app when clearing Onyx due to one of the values passed to `multiSet()` being undefined (which is normal behavior for a value that we are trying to clear out).

image

### Related Issues
None

### Automated Tests
None

### Linked PRs
This will be tested and deployed with https://github.com/Expensify/App/pull/13886

### DIFF
--- a/lib/storage/providers/SQLiteStorage.js
+++ b/lib/storage/providers/SQLiteStorage.js
@@ -66,7 +66,7 @@ const provider = {
     multiSet(pairs) {
         const stringifiedPairs = _.map(pairs, pair => [
             pair[0],
-            JSON.stringify(pair[1]),
+            JSON.stringify(pair[1] || null),
         ]);
         return db.executeBatchAsync([['REPLACE INTO keyvaluepairs (record_key, valueJSON) VALUES (?, json(?));', stringifiedPairs]]);
     },
--- a/lib/storage/providers/SQLiteStorage.js
+++ b/lib/storage/providers/SQLiteStorage.js
@@ -66,7 +66,7 @@ const provider = {
     multiSet(pairs) {
         const stringifiedPairs = _.map(pairs, pair => [
             pair[0],
-            JSON.stringify(pair[1] || null),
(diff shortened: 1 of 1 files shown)


lib/storage/providers/SQLiteStorage.js
```diff
@@ -66,7 +66,7 @@ const provider = {
     multiSet(pairs) {
         const stringifiedPairs = _.map(pairs, pair => [
             pair[0],
-            JSON.stringify(pair[1]),
+            JSON.stringify(pair[1] || null),
```
marcaaron, line 69: Let's only short circuit to `null` in the case of an explicitly `undefined` value? Anything falsy will default to `null` with what you have here. It might be fun for a while, but will probably lose it's charm quickly 😄

marcaaron (changes requested):

lib/storage/providers/SQLiteStorage.js
tgolen: Oh, good call.

tgolen:
OK, updated. Here is the output of `stringifiedPairs`:

```
[["session", "{\"loading\":false}"], ["account", "{\"errors\":null,\"success\":\"\",\"isLoading\":false}"], ["network", "{\"isOffline\":false}"], ["iou", "{\"loading\":false,\"error\":false,\"creatingIOUTransaction\":false,\"isRetrievingCurrency\":false}"], ["isSidebarLoaded", "false"], ["shouldShowComposeInput", "true"], ["preferredLocale", "\"es\""], ["activeClients", "null"]]
```

dangrous (approved):
Code makes sense to me! Will leave to @marcaaron to push the button

marcaaron (approved):
LGTM 👍 thanks for catching this one!

PR #237, opened 2023-02-28
Fix issue #15321: The app crashes when the user is logged into multiple tabs and logs out of one of the tabs

tienifr:
### Details

This PR changes the logic of the clear() function which cause issue https://github.com/Expensify/App/issues/15321 so that instead of removing all the keys then adding the default value for some important keys back, it only removes the keys that are not needed and retains the important keys.
It also address the performance concern of https://github.com/Expensify/App/issues/15321#issuecomment-1443948184 since the key removals are execute asynchronously. The total execution time has almost no difference compared to the old implementation.

### Fixed Issues

$ https://github.com/Expensify/App/issues/15321
$ https://github.com/Expensify/App/issues/15321#issuecomment-1441602988

### Tests

1. Open the app in the browser
2. Open another tab in the same browser
3. Log out from one of the tabs 
4. Verify that in other tab the app is not crashed and logged out to the login page.

### QA Steps

1. Open the app in the browser
2. Open another tab in the same browser
3. Log out from one of the tabs 
4. Verify that in other tab the app is not crashed and logged out to the login page.

#### Web

https://user-images.githubusercontent.com/113963320/224473159-5d1da53f-696e-466e-88df-07c0c37c2f0b.mp4

https://user-images.githubusercontent.com/113963320/224472049-4c273651-9365-49ee-b44d-1ca0f96bcb4f.mov

#### Mobile Web - Chrome
https://user-images.githubusercontent.com/113963320/224473167-bbf6ec5e-a5d1-448b-9eb2-f1fd5ea5baf4.mp4

#### Mobile Web - Safari
https://user-images.githubusercontent.com/113963320/224472235-d5bba632-c7c4-4667-adc5-3696e9b45cd8.mov

#### Desktop

https://user-images.githubusercontent.com/113963320/224473704-b90d659a-3c32-41b8-8374-0c90f40bdd0a.mov

#### iOS

https://user-images.githubusercontent.com/113963320/224472664-72caec3a-dc51-437c-8099-246baf56db7b.mov

#### Android

https://user-images.githubusercontent.com/113963320/224473179-662b014c-1f4d-4c13-835d-278b3be2ea5c.mp4

### DIFF
--- a/__mocks__/localforage-removeitems.js
+++ b/__mocks__/localforage-removeitems.js
@@ -1,0 +1,14 @@
+import _ from 'underscore';
+
+function extendPrototype(localforage) {
+    const newLocalforage = localforage;
+    newLocalforage.removeItems = keys => new Promise((resolve) => {
+        _.each(keys, (key) => {
+            delete newLocalforage.storageMap[key];
+        });
+        resolve();
+    });
+}
+
+// eslint-disable-next-line import/prefer-default-export
+export {extendPrototype};
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -1047,55 +1047,46 @@ function initializeWithDefaultKeyStates() {
 function clear(keysToPreserve = []) {
     return getAllKeys()
         .then((keys) => {
-            const keyValuesToReset = [];
-            const defaultKeys = _.keys(defaultKeyStates);
-
-            // Get all the values for the keys that need to be preserved. These key/value pairs will be set
-            // in Onyx after the database is cleared().
-            const keyValuesToPreserve = _.map(keysToPreserve, key => [key, cache.getValue(key)]);
+            const keysToBeClearedFromStorage = [];
+            const keyValuesToResetAsCollection = {};
+            const keyValuesToResetIndividually = {};
 
             // The only keys that should not be cleared are:
             // 1. Anything specifically passed in keysToPreserve (because some keys like language preferences, offline
             //      status, or activeClients need to remain in Onyx even when signed out)
             // 2. Any keys with a default state (because they need to remain in Onyx as their default, and setting them
             //      to null would cause unknown behavior)
-            const keysToClear = _.difference(keys, keysToPreserve, defaultKeys);
-            keyValuesToReset.push(..._.map(keysToClear, key => [key, null]));
+            _.each(keys, (key) => {
+                const isKeyToPreserve = _.contains(keysToPreserve, key);
+                const isDefaultKey = _.has(defaultKeyStates, key);
 
-            // Remove any keysToPreserve from the defaultKeyStates because if they are passed in it has been explicitly
-            // called out to preserve those values instead of resetting them back
-            // to the default.
-            const defaultKeyValuePairs = _.pairs(_.omit(defaultKeyStates, ...keysToPreserve));
-
-            // Add the default key value pairs to the keyValuesToReset so that they get set back to their default values
-            // when we clear Onyx
-            keyValuesToReset.push(...defaultKeyValuePairs);
-
-            // We now have all the key/values that need to be reset, but we're not done yet!
-            // There will be two groups of key/values and they each need to be updated a little bit differently.
-            // Collection keys need to be notified differently than non collection keys
-            const keyValuesToResetAsCollection = {};
-            const keyValuesToResetIndividually = {};
-
-            // Make sure that we also reset the cache values before clearing the values from storage.
-            // We do this before clearing Storage so that any call to clear() followed by merge() on a key with a
-            // default state results in the merged value getting saved, since the update from the merge() call would
-            // happen on the tick after the update from this clear()
-            _.each(keyValuesToReset, (keyValue) => {
-                const key = keyValue[0];
-                const value = keyValue[1];
-                cache.set(key, value);
-
-                const collectionKey = key.substring(0, key.indexOf('_') + 1);
-                if (collectionKey) {
-                    if (!keyValuesToResetAsCollection[collectionKey]) {
-                        keyValuesToResetAsCollection[collectionKey] = {};
+                // If the key is being removed or reset to default:
+                // 1. Update it in the cache
+                // 2. Figure out whether it is a collection key or not,
+                //      since collection key subscribers need to be updated differently
+                if (!isKeyToPreserve) {
+                    const oldValue = cache.getValue(key);
+                    const newValue = _.get(defaultKeyStates, key, null);
+                    if (newValue !== oldValue) {
+                        cache.set(key, newValue);
+                        const collectionKey = key.substring(0, key.indexOf('_') + 1);
+                        if (collectionKey) {
+                            if (!keyValuesToResetAsCollection[collectionKey]) {
+                                keyValuesToResetAsCollection[collectionKey] = {};
+                            }
+                            keyValuesToResetAsCollection[collectionKey][key] = newValue;
+                        } else {
+                            keyValuesToResetIndividually[key] = newValue;
+                        }
                     }
-                    keyValuesToResetAsCollection[collectionKey][key] = value;
+                }
+
+                if (isKeyToPreserve || isDefaultKey) {
                     return;
                 }
 
-                keyValuesToResetIndividually[key] = value;
+                // If it isn't preserved and doesn't have a default, we'll remove it
+                keysToBeClearedFromStorage.push(key);
             });
 
             // Notify the subscribers for each key/value group so they can receive the new values
@@ -1106,11 +1097,10 @@ function clear(keysToPreserve = []) {
                 notifyCollectionSubscribersOnNextTick(key, value);
             });
 
-            // Call clear() and make sure that the default key/values and the key/values from the parameter
-            // are preserved in storage. This makes sure to always leave storage in a state that contains
-            // all the default values and any additional values that we want to remain after the database is cleared.
-            return Storage.clear()
-                .then(() => Storage.multiSet([...defaultKeyValuePairs, ...keyValuesToPreserve]));
+            const defaultKeyValuePairs = _.pairs(_.omit(defaultKeyStates, keysToPreserve));
+
+            // Remove only the items that we want cleared from storage, and reset others to default
+            return Storage.removeItems(keysToBeClearedFromStorage).then(() => Storage.multiSet(defaultKeyValuePairs));
         });
 }
 
--- a/lib/storage/WebStorage.js
+++ b/lib/storage/WebStorage.js
@@ -17,6 +17,12 @@ function raiseStorageSyncEvent(onyxKey) {
     global.localStorage.removeItem(SYNC_ONYX, onyxKey);
 }
 
+function raiseStorageSyncManyKeysEvent(onyxKeys) {
+    _.each(onyxKeys, (onyxKey) => {
+        raiseStorageSyncEvent(onyxKey);
+    });
+}
+
 const webStorage = {
     ...Storage,
 
@@ -30,6 +36,9 @@ const webStorage = {
 
         this.removeItem = key => Storage.removeItem(key)
             .then(() => raiseStorageSyncEvent(key));
+
+        this.removeItems = keys => Storage.removeItems(keys)
+            .then(() => raiseStorageSyncManyKeysEvent(keys));
 
         // If we just call Storage.clear other tabs will have no idea which keys were available previously
         // so that they can call keysChanged for them. That's why we iterate over every key and raise a storage sync
--- a/lib/storage/providers/AsyncStorage.js
+++ b/lib/storage/providers/AsyncStorage.js
@@ -74,6 +74,13 @@ const provider = {
     removeItem: AsyncStorage.removeItem,
 
     /**
+     * Remove given keys and their values from storage
+     * @param {Array} keys
+     * @returns {Promise}
+     */
+    removeItems: keys => Promise.all(_.map(keys, key => AsyncStorage.removeItem(key))),
+
+    /**
      * Clear everything from storage
      * @returns {Promise<void>}
      */
--- a/lib/storage/providers/LocalForage.js
+++ b/lib/storage/providers/LocalForage.js
@@ -6,8 +6,11 @@
 
 import localforage from 'localforage';
 import _ from 'underscore';
+import {extendPrototype} from 'localforage-removeitems';
 import SyncQueue from '../../SyncQueue';
 import fastMerge from '../../fastMerge';
+
+extendPrototype(localforage);
 
 localforage.config({
     name: 'OnyxDB',
@@ -102,6 +105,16 @@ const provider = {
      * @returns {Promise<void>}
      */
     removeItem: localforage.removeItem,
+
+    /**
+     * Remove given keys and their values from storage
+     *
+     * @param {Array} keys
+     * @returns {Promise}
+     */
+    removeItems(keys) {
+        return localforage.removeItems(keys);
+    },
 
     /**
      * Sets the value for a given key. The only requirement is that the value should be serializable to JSON string
--- a/package.json
+++ b/package.json
@@ -79,6 +79,7 @@
     "@react-native-async-storage/async-storage": "^1.17.11",
     "expensify-common": ">=1",
     "localforage": "^1.10.0",
+    "localforage-removeitems": "^1.4.0",
     "react": ">=18.1.0",
     "react-native-performance": "^4.0.0",
     "react-native-quick-sqlite": "^8.0.0-beta.2"
@@ -94,6 +95,9 @@
       "optional": true
     },
     "localforage": {
+      "optional": true
+    },
+    "localforage-removeitems": {
       "optional": true
     }
   },
--- a/lib/storage/providers/SQLiteStorage.js
+++ b/lib/storage/providers/SQLiteStorage.js
@@ -112,6 +112,18 @@ const provider = {
     removeItem: key => db.executeAsync('DELETE FROM keyvaluepairs WHERE record_key = ?;', [key]),
 
     /**
+     * Removes given keys and their values from storage
+     *
+     * @param {Array<String>} keys
+     * @returns {Promise<void>}
+     */
+    removeItems: (keys) => {
+        const placeholders = _.map(keys, () => '?').join(',');
+        const query = `DELETE FROM keyvaluepairs WHERE record_key IN (${placeholders});`;
+        return db.executeAsync(query, keys);
+    },
+
+    /**
     * Clears absolutely everything from storage
     * @returns {Promise<void>}
     */
--- a/package-lock.json
+++ b/package-lock.json
@@ -53,6 +53,7 @@
         "@react-native-async-storage/async-storage": "^1.17.11",
         "expensify-common": ">=1",
         "localforage": "^1.10.0",
+        "localforage-removeitems": "^1.4.0",
         "react": ">=18.1.0",
         "react-native-performance": "^4.0.0",
         "react-native-quick-sqlite": "^8.0.0-beta.2"
@@ -62,6 +63,9 @@
           "optional": true
         },
         "localforage": {
+          "optional": true
+        },
+        "localforage-removeitems": {
           "optional": true
         },
         "react-native-performance": {
@@ -8702,7 +8706,7 @@
       "version": "3.0.6",
       "resolved": "https://registry.npmjs.org/immediate/-/immediate-3.0.6.tgz",
       "integrity": "sha1-nbHb0Pr43m++D13V5Wu2BigN5ps=",
-      "dev": true
+      "devOptional": true
     },
     "node_modules/import-fresh": {
       "version": "3.2.2",
@@ -11175,7 +11179,7 @@
       "version": "3.1.1",
       "resolved": "https://registry.npmjs.org/lie/-/lie-3.1.1.tgz",
       "integrity": "sha1-mkNrLMd0bKWd56QfpGmz77dr2H4=",
-      "dev": true,
+      "devOptional": true,
       "dependencies": {
         "immediate": "~3.0.5"
       }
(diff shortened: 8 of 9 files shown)


tienifr:

lib/Onyx.js
```diff
@@ -1106,11 +1106,14 @@ function clear(keysToPreserve = []) {
                 notifyCollectionSubscribersOnNextTick(key, value);
             });
 
-            // Call clear() and make sure that the default key/values and the key/values from the parameter
-            // are preserved in storage. This makes sure to always leave storage in a state that contains
-            // all the default values and any additional values that we want to remain after the database is cleared.
```
madmax330, line 1111: Let's keep a comment here to explain what we're doing and why

tienifr: @madmax330 Thanks for pointing that out. I have updated, please help to check again

roryabraham (changes requested):
I think this function could use some additional cleanup. I found it pretty hard to follow, and for the most part we just need one main loop.

```js
function clear(keysToPreserve = []) {
    return getAllKeys()
        .then((keys) => {
            const keysToBeClearedFromStorage = [];
            const keyValuesToResetAsCollection = {};
            const keyValuesToResetIndividually = {};

            // The only keys that should not be cleared are:
            // 1. Anything specifically passed in keysToPreserve (because some keys like language preferences, offline
            //      status, or activeClients need to remain in Onyx even when signed out)
            // 2. Any keys with a default state (because they need to remain in Onyx as their default, and setting them
            //      to null would cause unknown behavior)
            _.each(keys, (key) => {
                const isKeyToPreserve = _.contains(keysToPreserve, key);
                const isDefaultKey = _.has(defaultKeyStates, key);

                // If the key is being removed or reset to default:
                // 1. Update it in the cache
                // 2. Figure out whether it is a collection key or not,
                //      since collection key subscribers need to be updated differently
                if (!isKeyToPreserve) {
                    const newValue = defaultKeyStates[key] || null;
                    cache.set(key, newValue);
                    const collectionKey = key.substring(0, key.indexOf('_') + 1);
                    if (collectionKey) {
                        if (!keyValuesToResetAsCollection[collectionKey]) {
                            keyValuesToResetAsCollection[collectionKey] = {};
                        }
                        keyValuesToResetAsCollection[collectionKey][key] = newValue;
                    } else {
                        keyValuesToResetIndividually[key] = newValue;
                    }
                }

                if (isKeyToPreserve || isDefaultKey) {
                    return;
                }

                // If it isn't preserved and doesn't have a default, we'll remove it
                keysToBeClearedFromStorage.push(key);
            });

            // Notify the subscribers for each key/value group so they can receive the new values
            _.each(keyValuesToResetIndividually, (value, key) => {
                notifySubscribersOnNextTick(key, value);
            });
            _.each(keyValuesToResetAsCollection, (value, key) => {
                notifyCollectionSubscribersOnNextTick(key, value);
            });

            // Remove only the items that we want cleared from storage
            return Storage.removeItems(keysToBeClearedFromStorage);
        });
```

lib/Onyx.js
```diff
@@ -1106,11 +1106,18 @@ function clear(keysToPreserve = []) {
...
             // all the default values and any additional values that we want to remain after the database is cleared.
-            return Storage.clear()
-                .then(() => Storage.multiSet([...defaultKeyValuePairs, ...keyValuesToPreserve]));
+
+            const retainedKeyValuePairs = [...defaultKeyValuePairs, ...keyValuesToPreserve];
+
+            return Storage.getAllKeys().then((allKeys) => {
+                const retainedKeys = _.map(retainedKeyValuePairs, ([key]) => key);
+                const keysToBeCleared = _.difference(allKeys, retainedKeys);
+                return Promise.all(_.map(keysToBeCleared, key => Storage.removeItem(key)))
```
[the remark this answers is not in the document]
roryabraham: Actually, it looks like we can use https://github.com/localForage/localForage-removeItems to optimize `removeItems` on web, so let's do that

sobitneupane (commented):
@tienifr I am not getting the excepted results. Can you please rerun the tests with latest changes?

https://user-images.githubusercontent.com/25876548/222147585-dfee60c8-672a-4ea2-b7fa-fa23bfb8048d.mov

lib/storage/WebStorage.js
```diff
@@ -17,6 +17,12 @@ function raiseStorageSyncEvent(onyxKey) {
     global.localStorage.removeItem(SYNC_ONYX, onyxKey);
 }
 
+function raiseStorageSyncManyKeysEvent(onyxKeys) {
+    onyxKeys.forEach((onyxKey) => {
```
roryabraham, line 21: Use `_.each` (just by convention)

lib/storage/providers/LocalForage.js
```diff
@@ -6,9 +6,13 @@
 
 import localforage from 'localforage';
 import _ from 'underscore';
+
```
roryabraham, line 9:
```suggestion
```

roryabraham (changes requested):

lib/storage/providers/LocalForage.js
```diff
@@ -103,6 +107,14 @@ const provider = {
      */
     removeItem: localforage.removeItem,
 
+    /**
+     * Remove given keys and their values from storage
+     *
+     * @param {Array} keys
+     * @returns {Promise}
+     */
+    removeItems: keys => Promise.all(_.map(keys, key => localforage.removeItem(key))),
```
roryabraham, line 116: Because you added the `removeItems` plugin, you should be able to use it here:

```suggestion
    removeItems: localforage.removeItems,
```

roryabraham (changes requested):

tienifr:
@roryabraham Thanks for your comment, but the PR isn't completed, I still need to change a few things to make it work

tienifr:
@roryabraham @sobitneupane I just fixed all comments, also tested again and it works correctly. Please help to check again. Thanks 

https://user-images.githubusercontent.com/113963320/222215886-6b621d90-1241-41c5-8e8b-7eddb7519bda.mov

lib/Onyx.js
```diff
@@ -1106,11 +1096,9 @@ function clear(keysToPreserve = []) {
...
             });
 
-            // Call clear() and make sure that the default key/values and the key/values from the parameter
-            // are preserved in storage. This makes sure to always leave storage in a state that contains
-            // all the default values and any additional values that we want to remain after the database is cleared.
-            return Storage.clear()
-                .then(() => Storage.multiSet([...defaultKeyValuePairs, ...keyValuesToPreserve]));
+            // Remove only the items that we want cleared from storage
+            const retainedKeyValuePairs = [...defaultKeyValuePairs, ...keyValuesToPreserve];
+            return Storage.removeItems(keysToBeClearedFromStorage).then(() => Storage.multiSet(retainedKeyValuePairs));
```
roryabraham, line 1101: Do we still need to do a `multiSet`, given that we didn't remove any key/value pairs for the retained keys?

```suggestion
            return Storage.removeItems(keysToBeClearedFromStorage);
```

roryabraham (changes requested):

lib/Onyx.js
roryabraham: Oh, I guess we do want to reset the default keys back to their default

tienifr: > Oh, I guess we do want to reset the default keys back to their default

We still reset the default keys back to default value

roryabraham: Okay, so that means we should make changes like this:

```diff
diff --git a/lib/Onyx.js b/lib/Onyx.js
index ed93f15..9322e69 100644
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -1050,8 +1050,6 @@ function clear(keysToPreserve = []) {
             const keysToBeClearedFromStorage = [];
             const keyValuesToResetAsCollection = {};
             const keyValuesToResetIndividually = {};
-            const keyValuesToPreserve = _.map(keysToPreserve, key => [key, cache.getValue(key)]);
-            const defaultKeyValuePairs = _.pairs(_.omit(defaultKeyStates, ...keysToPreserve));
 
             // The only keys that should not be cleared are:
             // 1. Anything specifically passed in keysToPreserve (because some keys like language preferences, offline
@@ -1096,9 +1094,10 @@ function clear(keysToPreserve = []) {
                 notifyCollectionSubscribersOnNextTick(key, value);
             });
 
-            // Remove only the items that we want cleared from storage
-            const retainedKeyValuePairs = [...defaultKeyValuePairs, ...keyValuesToPreserve];
-            return Storage.removeItems(keysToBeClearedFromStorage).then(() => Storage.multiSet(retainedKeyValuePairs));
+            // Remove only the items that we want cleared from storage, and reset others to default
+            const defaultKeyValuePairs = _.pairs(_.omit(defaultKeyStates, keysToPreserve));
+            return Storage.removeItems(keysToBeClearedFromStorage)
+                .then(() => Storage.multiSet(defaultKeyValuePairs));
         });
 }
```

roryabraham (changes requested):

tienifr:
@roryabraham I agree with you that we should only mutilSet the default keys. I just fixed that, please check again.

sobitneupane (commented):
I am getting following console errors.
Screenshot 2023-03-02 at 18 48 24

Why does the other tab updates only when the tab is active? You can notice in the video that tab updates only when the user switches to the tab. It indicates that the onyx is synced only when the tab is active which can create some serious issues.

https://user-images.githubusercontent.com/25876548/222438276-acbfa09e-a63e-4371-bc7b-ff6ea294f3d5.mov

lib/Onyx.js
```diff
@@ -1106,11 +1094,10 @@ function clear(keysToPreserve = []) {
                 notifyCollectionSubscribersOnNextTick(key, value);
             });
 
-            // Call clear() and make sure that the default key/values and the key/values from the parameter
-            // are preserved in storage. This makes sure to always leave storage in a state that contains
-            // all the default values and any additional values that we want to remain after the database is cleared.
-            return Storage.clear()
-                .then(() => Storage.multiSet([...defaultKeyValuePairs, ...keyValuesToPreserve]));
+            const defaultKeyValuePairs = _.pairs(_.omit(defaultKeyStates, ...keysToPreserve));
```
roryabraham, line 1097: Unnecessary spread:

```suggestion
            const defaultKeyValuePairs = _.pairs(_.omit(defaultKeyStates, keysToPreserve));
```

image

roryabraham (changes requested):

roryabraham:
> I am getting following console errors.

@sobitneupane Did you remember to run `npm install localforage-removeitems` before copying over this change and running E/App?

package.json
```diff
@@ -41,6 +41,7 @@
   "dependencies": {
     "ascii-table": "0.0.9",
     "fast-equals": "^4.0.3",
+    "localforage-removeitems": "^1.4.0",
```
roryabraham, line 44: I think this should be a `peerDependency` instead of a `dependency`

roryabraham (changes requested):

package.json
```diff
@@ -95,6 +96,9 @@
     },
     "localforage": {
       "optional": true
+    },
+    "localforage-removeitems": {
```
christianwen, line 100: this peer dependency is optional, similar to `localforage`, reference https://github.com/Expensify/react-native-onyx/pull/132/files#r873998291

package.json
```diff
@@ -95,6 +96,9 @@
...
+      "optional": true
```
tienifr, line 101: this peer dependency is optional, similar to `localforage`. Reference: https://github.com/Expensify/react-native-onyx/pull/132/files#r873998291

lib/Onyx.js
tienifr: Fixed, thanks

package.json
tienifr: Fixed, thanks

tienifr:
@roryabraham I've fixed your comments, pls help review again
Thanks!

tienifr:
> Why does the other tab updates only when the tab is active? You can notice in the video that tab updates only when the user switches to the tab. It indicates that the onyx is synced only when the tab is active which can create some serious issues.

@sobitneupane I think this is not related to the scope of this PR (or this issue), since even if we apply only [this patch](https://github.com/Expensify/App/pull/15556) on `App` without this `react-native-onyx` root cause fix, we can still see that delay behavior.

roryabraham (approved):
This PR is looking good to me 👍🏼

lib/storage/providers/AsyncStorage.js
```diff
@@ -73,6 +73,13 @@ const provider = {
      */
     removeItem: AsyncStorage.removeItem,
 
+    /**
+     * Remove given keys and their values from storage
+     * @param {Array} keys
+     * @returns {Promise}
+     */
+    removeItems: keys => Promise.all(_.map(keys, key => AsyncStorage.removeItem(key))),
```
tienifr, line 81: This is required because we're using `AsyncStorage` as the mock provider for `NativeStorage` in unit tests, so `removeItems` needs to be defined here as well

tienifr:
@roryabraham, I've added `localforage-removeitems.js` file to mock removeItems function on test environment. I have also updated the logic in `clear` function to add `defaultKeys` to `keys` like the old logic. Please help take a look, thanks.

sobitneupane:
> @sobitneupane I think this is not related to the scope of this PR (or this issue), since even if we apply only [this patch](https://github.com/Expensify/App/pull/15556) on `App` without this `react-native-onyx` root cause fix, we can still see that delay behavior.

I don't agree with you @tienifr. I tested the [PR](https://github.com/Expensify/react-native-onyx/pull/237#pullrequestreview-1321796687). This is the result in the PR.

https://user-images.githubusercontent.com/25876548/224104762-86ba6df9-1f72-4c9f-bffe-49050bc70acb.mov



And following is the result with the changes in this PR.

https://user-images.githubusercontent.com/25876548/224104397-bb5d9fe1-dd29-45d3-89c1-9d4119b7421e.mov


Please notice the unread indicators in both the videos. With the change in this PR, unread indicator only goes away after user switches to the tab indicating onyx is cleared only after user switches to the tab.

tienifr:
> @tienifr Many changes are made since PR is created. So, can you please update the screen recordings in PR body. Also, please do include screen recordings for heavy traffic accounts.

@sobitneupane 

Thanks for the heads up! I have updated the screen recordings with HTA.I also made some small changes so the code could pass the lint test. Please have a look.

lib/Onyx.js
```diff
@@ -1047,55 +1047,46 @@ function initializeWithDefaultKeyStates() {
...
-            const keyValuesToReset = [];
-            const defaultKeys = _.keys(defaultKeyStates);
-
-            // Get all the values for the keys that need to be preserved. These key/value pairs will be set
-            // in Onyx after the database is cleared().
-            const keyValuesToPreserve = _.map(keysToPreserve, key => [key, cache.getValue(key)]);
+            const keysToBeClearedFromStorage = [];
+            const keyValuesToResetAsCollection = {};
+            const keyValuesToResetIndividually = {};
+            const defaultKeys = _.keys(_.omit(defaultKeyStates, ...keysToPreserve));
```
roryabraham, line 1053: Why do we need the `keysToReset` here? It seems to me like this logic should be covered in the main loop right?

roryabraham: Also, if this _is_ needed for some reason I'm not seeing, there's an unnecessary spread:

```suggestion
            const defaultKeys = _.keys(_.omit(defaultKeyStates, keysToPreserve));
```

tienifr: @roryabraham Thanks for your comment. I think we should notify subscribers when data has been cleared as I can see this test on `onyxTest.js`. After checking the old logic I see we push the `defaultKeyValuePairs` to `keyValuesToReset`, that why I decide to have `keysToReset`. You can check the test https://github.com/Expensify/react-native-onyx/blob/4887af4b350de5febe3084e6d13ac07d8f617021/tests/unit/onyxTest.js#L95 
the keys = [ 'test' ] and defaultKeyStates = { otherTest: 42 }, if we don't push `otherTest` to keys, we'll miss this value.
 
And yes, you're right, having spread is unnecessary. I updated my PR to remove the spread.   
https://github.com/Expensify/react-native-onyx/blob/4887af4b350de5febe3084e6d13ac07d8f617021/lib/Onyx.js#L1070-L1072

sobitneupane (commented):
### Screenshots/Videos
<details>
<summary>Web</summary>

https://user-images.githubusercontent.com/25876548/224715558-13186e3b-117f-4e12-9cee-92b2875d4404.mov



</details>

<details>
<summary>Mobile Web - Chrome</summary>

https://user-images.githubusercontent.com/25876548/224715597-1a546356-2572-4f5f-966c-55c36a2f7730.mov


</details>

<details>
<summary>Mobile Web - Safari</summary>

https://user-images.githubusercontent.com/25876548/224716055-2629f90c-9759-4203-b7a3-ee5075742b5f.mov




</details>

<details>
<summary>Desktop</summary>

https://user-images.githubusercontent.com/25876548/224976282-ee504e7e-0800-421e-920a-f2769dd2f25b.mov

</details>

<details>
<summary>iOS</summary>

https://user-images.githubusercontent.com/25876548/224715858-b1628103-7e81-459f-852a-110cb4054fff.mov

</details>

<details>
<summary>Android</summary>

https://user-images.githubusercontent.com/25876548/224716005-f777298a-05ec-4d82-a22b-d1e163edf75e.mov




</details>

lib/Onyx.js
roryabraham: Thanks for explaining. I've spent some time testing, and I still don't think this code is needed. I modified the `onyxTest` a bit to test that default keys are reset:

```diff
diff --git a/tests/unit/onyxTest.js b/tests/unit/onyxTest.js
index 50b8608..e9a955e 100644
--- a/tests/unit/onyxTest.js
+++ b/tests/unit/onyxTest.js
@@ -112,9 +112,10 @@ describe('Onyx', () => {
 
         return waitForPromisesToResolve()
             .then(() => Onyx.set(ONYX_KEYS.TEST_KEY, 'test'))
+            .then(() => Onyx.set(ONYX_KEYS.OTHER_TEST, 420))
             .then(() => {
                 expect(testKeyValue).toBe('test');
-                expect(otherTestValue).toBe(42);
+                expect(otherTestValue).toBe(420);
                 return Onyx.clear().then(waitForPromisesToResolve);
             })
             .then(() => {
```

Then I reset the code to how it was before:

```diff
diff --git a/lib/Onyx.js b/lib/Onyx.js
index 910b759..2667754 100644
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -1050,16 +1050,13 @@ function clear(keysToPreserve = []) {
             const keysToBeClearedFromStorage = [];
             const keyValuesToResetAsCollection = {};
             const keyValuesToResetIndividually = {};
-            const defaultKeys = _.keys(_.omit(defaultKeyStates, keysToPreserve));
-            const keysToReset = _.difference(keys, defaultKeys);
-            keysToReset.push(...defaultKeys);
 
             // The only keys that should not be cleared are:
             // 1. Anything specifically passed in keysToPreserve (because some keys like language preferences, offline
             //      status, or activeClients need to remain in Onyx even when signed out)
             // 2. Any keys with a default state (because they need to remain in Onyx as their default, and setting them
             //      to null would cause unknown behavior)
-            _.each(keysToReset, (key) => {
+            _.each(keys, (key) => {
                 const isKeyToPreserve = _.contains(keysToPreserve, key);
                 const isDefaultKey = _.has(defaultKeyStates, key);

```

And everything worked as expected. Since this is simpler, let's change it back.

roryabraham (commented):
Also, not going to _require_ this since it's probably _fine_ if we don't, but I think it might be good to prevent updating subscribers unless the data actually changes:

```diff
 
@@ -1068,16 +1065,19 @@ function clear(keysToPreserve = []) {
                 // 2. Figure out whether it is a collection key or not,
                 //      since collection key subscribers need to be updated differently
                 if (!isKeyToPreserve) {
+                    const oldValue = cache.getValue(key);
                     const newValue = _.get(defaultKeyStates, key, null);
-                    cache.set(key, newValue);
-                    const collectionKey = key.substring(0, key.indexOf('_') + 1);
-                    if (collectionKey) {
-                        if (!keyValuesToResetAsCollection[collectionKey]) {
-                            keyValuesToResetAsCollection[collectionKey] = {};
+                    if (newValue !== oldValue) {
+                        cache.set(key, newValue);
+                        const collectionKey = key.substring(0, key.indexOf('_') + 1);
+                        if (collectionKey) {
+                            if (!keyValuesToResetAsCollection[collectionKey]) {
+                                keyValuesToResetAsCollection[collectionKey] = {};
+                            }
+                            keyValuesToResetAsCollection[collectionKey][key] = newValue;
+                        } else {
+                            keyValuesToResetIndividually[key] = newValue;
                         }
-                        keyValuesToResetAsCollection[collectionKey][key] = newValue;
-                    } else {
-                        keyValuesToResetIndividually[key] = newValue;
                     }
                 }
 

```

tienifr:
@roryabraham Yes, you're right. So I updated my PR as your suggestion, I also removed unnecessary test case. Please help review again. Thanks.

@sobitneupane Here is my result on desktop

https://user-images.githubusercontent.com/113963320/224788160-ee3afe00-8453-4882-9213-2bcdf4fb6509.mov

sobitneupane (dismissed):

sobitneupane:
@roryabraham I think we are good to merge this PR.

tienifr:
@roryabraham 

Are we good to go for this PR to be merged?

tienifr:
@roryabraham Could you check again and merged it if there are no more problems?

roryabraham (dismissed):

roryabraham:
Shoot, was just about to merge this, but I can't because it contains unsigned commits. @tienifr Please either retroactively sign all these commits then force-push to this branch, or create a new PR with 100% signed commits and tag us again.

image

tienifr:
@roryabraham My bad, I just squashed commit and forced push. Please check again and merge this PR

roryabraham (approved):

__mocks__/localforage-removeitems.js
```diff
@@ -0,0 +1,14 @@
+import _ from 'underscore';
+
+function extendPrototype(localforage) {
+    const newLocalforage = localforage;
+    newLocalforage.removeItems = keys => new Promise((resolve) => {
+        _.each(keys, (key) => {
+            delete newLocalforage.storageMap[key];
```
marcaaron, line 7: I'm confused by this. Does this actually delete the stuff from IndexedDB? It looks like it would not...

marcaaron: Ohhh nvm it's a mock...

roryabraham: Well this is a mock. This code has changed a bit since this PR was created, but the mock isn't actually writing anything to IndexedDB. `storageMapInternal` is just an object so I think `delete` does work here?

lib/Onyx.js
marcaaron: updated it to only say _what_ and not why...

PR #238, opened 2023-03-02
Implement JSON_PATCH() in Onyx.merge() in react-native-onyx(native)

chrispader:
Please review: @marcaaron 

### Related Issues

https://github.com/Expensify/App/issues/13972

### DIFF
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -296,7 +296,7 @@ function getCachedCollection(collectionKey) {
 function keysChanged(collectionKey, partialCollection) {
     // We are iterating over all subscribers similar to keyChanged(). However, we are looking for subscribers who are subscribing to either a collection key or
     // individual collection key member for the collection that is being updated. It is important to note that the collection parameter cane be a PARTIAL collection
-    // and does not represent all of the combined keys and values for a collection key. It is just the "new" data that was merged in via mergeCollection().
+    // and does not represent all of the combined keys and values for a collection key. It is just the "new" data that was qd in via mergeCollection().
     const stateMappingKeys = _.keys(callbackToStateMapping);
     for (let i = 0; i < stateMappingKeys.length; i++) {
         const subscriber = callbackToStateMapping[stateMappingKeys[i]];
@@ -905,58 +905,13 @@ function multiSet(data) {
         .catch(error => evictStorageAndRetry(error, multiSet, data));
 }
 
-// Key/value store of Onyx key and arrays of values to merge
-const mergeQueue = {};
-
 /**
  * @private
  * @param {String} key
  * @returns {Boolean}
  */
 function hasPendingMergeForKey(key) {
-    return Boolean(mergeQueue[key]);
-}
-
-/**
- * Given an Onyx key and value this method will combine all queued
- * value updates and return a single value. Merge attempts are
- * batched. They must occur after a single call to get() so we
- * can avoid race conditions.
- *
- * @private
- * @param {String} key
- * @param {*} data
- *
- * @returns {*}
- */
-function applyMerge(key, data) {
-    const mergeValues = mergeQueue[key];
-    if (_.isArray(data) || _.every(mergeValues, _.isArray)) {
-        // Array values will always just concatenate
-        // more items onto the end of the array
-        return _.reduce(mergeValues, (modifiedData, mergeValue) => [
-            ...modifiedData,
-            ...mergeValue,
-        ], data || []);
-    }
-
-    if (_.isObject(data) || _.every(mergeValues, _.isObject)) {
-        // Object values are merged one after the other
-        return _.reduce(mergeValues, (modifiedData, mergeValue) => {
-            // lodash adds a small overhead so we don't use it here
-            // eslint-disable-next-line prefer-object-spread, rulesdir/prefer-underscore-method
-            const newData = Object.assign({}, fastMerge(modifiedData, mergeValue));
-
-            // We will also delete any object keys that are undefined or null.
-            // Deleting keys is not supported by AsyncStorage so we do it this way.
-            // Remove all first level keys that are explicitly set to null.
-            return _.omit(newData, (value, finalObjectKey) => _.isNull(mergeValue[finalObjectKey]));
-        }, data || {});
-    }
-
-    // If we have anything else we can't merge it so we'll
-    // simply return the last value that was queued
-    return _.last(mergeValues);
+    return Storage.hasPendingMergeForKey(key);
 }
 
 /**
@@ -982,28 +937,7 @@ function applyMerge(key, data) {
  * @returns {Promise}
  */
 function merge(key, value) {
-    if (mergeQueue[key]) {
-        mergeQueue[key].push(value);
-        return Promise.resolve();
-    }
-
-    mergeQueue[key] = [value];
-    return get(key)
-        .then((data) => {
-            try {
-                const modifiedData = applyMerge(key, data);
-
-                // Clean up the write queue so we
-                // don't apply these changes again
-                delete mergeQueue[key];
-
-                return set(key, modifiedData);
-            } catch (error) {
-                Logger.logAlert(`An error occurred while applying merge for key: ${key}, Error: ${error}`);
-            }
-
-            return Promise.resolve();
-        });
+    return Storage.mergeItem(key, value);
 }
 
 /**
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -838,22 +838,15 @@ function evictStorageAndRetry(error, onyxMethod, ...args) {
 }
 
 /**
- * Write a value to our store with the given key
- *
- * @param {String} key ONYXKEY to set
- * @param {*} value value to store
- *
+ * @param {String} key
+ * @param {*} value
+ * @param {Function} writeMethod
  * @returns {Promise}
  */
-function set(key, value) {
-    if (_.isNull(value)) {
-        return remove(key);
-    }
-
-    // eslint-disable-next-line no-use-before-define
-    if (hasPendingMergeForKey(key)) {
-        Logger.logAlert(`Onyx.set() called after Onyx.merge() for key: ${key}. It is recommended to use set() or merge() not both.`);
-    }
+// eslint-disable-next-line rulesdir/no-negated-variables
+function notifyAndWriteToStorage(key, value, writeMethod = Storage.setItem) {
+    // Logging properties only since values could be sensitive things we don't want to log
+    Logger.logInfo(`${writeMethod.name}() called for key: ${key}${_.isObject(value) ? ` properties: ${_.keys(value).join(',')}` : ''}`);
 
     // If the value in the cache is the same as what we have then do not update subscribers unless they
     // have initWithStoredValues: false then they MUST get all updates even if nothing has changed.
@@ -867,9 +860,21 @@ function set(key, value) {
     cache.set(key, value);
     notifySubscribersOnNextTick(key, value);
 
-    // Write the thing to persistent storage, which will trigger a storage event for any other tabs open on this domain
-    return Storage.setItem(key, value)
-        .catch(error => evictStorageAndRetry(error, set, key, value));
+    return writeMethod(key, value)
+        // eslint-disable-next-line no-use-before-define
+        .catch(error => evictStorageAndRetry(error, writeMethod === Storage.setItem ? set : merge, key, value));
+}
+
+/**
+ * Write a value to our store with the given key
+ *
+ * @param {String} key ONYXKEY to set
+ * @param {*} value value to store
+ *
+ * @returns {Promise}
+ */
+function set(key, value) {
+    return notifyAndWriteToStorage(key, value);
 }
 
 /**
@@ -943,7 +948,9 @@ function merge(key, value) {
         return set(key, value);
     }
 
-    return Storage.mergeItem(key, value);
+    return new Promise((resolve) => {
+        get(key).then(currentValue => Storage.mergeItem(key, value, currentValue).then(updatedValue => resolve(notifyAndWriteToStorage(key, updatedValue))));
+    });
 }
 
 /**
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -937,6 +937,12 @@ function hasPendingMergeForKey(key) {
  * @returns {Promise}
  */
 function merge(key, value) {
+    // Array values are always set never merged because the MergePatch algorithm in SQLite will always replace arrays.
+    // Old concatenating of array values behavior is deprecated.
+    if (_.isArray(value)) {
+        return set(key, value);
+    }
+
     return Storage.mergeItem(key, value);
 }
 
--- a/lib/storage/providers/AsyncStorage.js
+++ b/lib/storage/providers/AsyncStorage.js
@@ -5,6 +5,11 @@
 
 import _ from 'underscore';
 import AsyncStorage from '@react-native-async-storage/async-storage';
+import applyMerge from '../../applyMerge';
+import * as Logger from '../../Logger';
+
+// Key/value store of Onyx key and arrays of values to merge
+const mergeQueue = {};
 
 const provider = {
     /**
@@ -56,13 +61,35 @@ const provider = {
      * @return {Promise<void>}
      */
     multiMerge(pairs) {
-        const stringPairs = _.map(pairs, ([key, value]) => [key, JSON.stringify(value)]);
+        const tasks = _.map(pairs, ([key, value, currentValue]) => {
+            if (mergeQueue[key]) {
+                mergeQueue[key].push(value);
+                return Promise.resolve();
+            }
 
-        return AsyncStorage.multiMerge(stringPairs);
+            mergeQueue[key] = [value];
+
+            try {
+                const modifiedData = applyMerge(key, currentValue, mergeQueue);
+
+                // Clean up the write queue so we
+                // don't apply these changes again
+                delete mergeQueue[key];
+
+                return this.setItem(key, modifiedData).then(() => Promise.resolve());
+            } catch (error) {
+                Logger.logAlert(`An error occurred while applying merge for key: ${key}, Error: ${error}`);
+            }
+
+            return Promise.resolve();
+        });
+
+        // We're returning Promise.resolve, otherwise the array of task results will be returned to the caller
+        return Promise.all(tasks).then(() => Promise.resolve());
     },
 
-    mergeItem(key, value) {
-        return this.multiMerge([[key, value]]);
+    mergeItem(key, value, currentValue) {
+        return this.multiMerge([[key, value, currentValue]]);
     },
 
     /**
@@ -83,6 +110,10 @@ const provider = {
      * @returns {Promise<void>}
      */
     clear: AsyncStorage.clear,
+
+    hasPendingMergeForQueue(key) {
+        return Boolean(mergeQueue[key]);
+    },
 };
 
 export default provider;
--- a/lib/storage/providers/AsyncStorage.js
+++ b/lib/storage/providers/AsyncStorage.js
@@ -57,7 +57,12 @@ const provider = {
      */
     multiMerge(pairs) {
         const stringPairs = _.map(pairs, ([key, value]) => [key, JSON.stringify(value)]);
+
         return AsyncStorage.multiMerge(stringPairs);
+    },
+
+    mergeItem(key, value) {
+        return this.multiMerge([[key, value]]);
     },
 
     /**
--- a/lib/storage/providers/LocalForage.js
+++ b/lib/storage/providers/LocalForage.js
@@ -8,6 +8,7 @@ import localforage from 'localforage';
 import _ from 'underscore';
 import SyncQueue from '../../SyncQueue';
 import fastMerge from '../../fastMerge';
+import applyMerge from '../../applyMerge';
 import * as Logger from '../../Logger';
 
 localforage.config({
@@ -16,40 +17,6 @@ localforage.config({
 
 // Key/value store of Onyx key and arrays of values to merge
 const mergeQueue = {};
-
-/**
- * Given an Onyx key and value this method will combine all queued
- * value updates and return a single value. Merge attempts are
- * batched. They must occur after a single call to get() so we
- * can avoid race conditions.
- *
- * @private
- * @param {String} key
- * @param {*} data
- *
- * @returns {*}
- */
-function applyMerge(key, data) {
-    const mergeValues = mergeQueue[key];
-
-    if (_.isObject(data) || _.every(mergeValues, _.isObject)) {
-        // Object values are merged one after the other
-        return _.reduce(mergeValues, (modifiedData, mergeValue) => {
-            // lodash adds a small overhead so we don't use it here
-            // eslint-disable-next-line prefer-object-spread, rulesdir/prefer-underscore-method
-            const newData = Object.assign({}, fastMerge(modifiedData, mergeValue));
-
-            // We will also delete any object keys that are undefined or null.
-            // Deleting keys is not supported by AsyncStorage so we do it this way.
-            // Remove all first level keys that are explicitly set to null.
-            return _.omit(newData, (value, finalObjectKey) => _.isNull(mergeValue[finalObjectKey]));
-        }, data || {});
-    }
-
-    // If we have anything else we can't merge it so we'll
-    // simply return the last value that was queued
-    return _.last(mergeValues);
-}
 
 const provider = {
     /**
@@ -95,36 +62,35 @@ const provider = {
      * @return {Promise<void>}
      */
     multiMerge(pairs) {
-        const tasks = _.map(pairs, ([key, value]) => {
+        const tasks = _.map(pairs, ([key, value, currentValue]) => {
             if (mergeQueue[key]) {
                 mergeQueue[key].push(value);
                 return Promise.resolve();
             }
 
             mergeQueue[key] = [value];
-            Storage.getItem(key).then((data) => {
-                try {
-                    const modifiedData = applyMerge(key, data);
 
-                    // Clean up the write queue so we
-                    // don't apply these changes again
-                    delete mergeQueue[key];
+            try {
+                const modifiedData = applyMerge(key, currentValue, mergeQueue);
 
-                    this.setItemQueue.push({key, modifiedData, shouldMerge: true}).then(() => Promise.resolve());
-                } catch (error) {
-                    Logger.logAlert(`An error occurred while applying merge for key: ${key}, Error: ${error}`);
-                }
+                // Clean up the write queue so we
+                // don't apply these changes again
+                delete mergeQueue[key];
 
-                return Promise.resolve();
-            });
+                this.setItemQueue.push({key, modifiedData, shouldMerge: true}).then(() => Promise.resolve());
+            } catch (error) {
+                Logger.logAlert(`An error occurred while applying merge for key: ${key}, Error: ${error}`);
+            }
+
+            return Promise.resolve();
         });
 
         // We're returning Promise.resolve, otherwise the array of task results will be returned to the caller
         return Promise.all(tasks).then(() => Promise.resolve());
     },
 
-    mergeItem(key, value) {
-        return this.multiMerge([[key, value]]);
+    mergeItem(key, value, currentValue) {
+        return this.multiMerge([[key, value, currentValue]]);
     },
 
     /**
--- a/lib/storage/providers/LocalForage.js
+++ b/lib/storage/providers/LocalForage.js
@@ -8,10 +8,56 @@ import localforage from 'localforage';
 import _ from 'underscore';
 import SyncQueue from '../../SyncQueue';
 import fastMerge from '../../fastMerge';
+import * as Logger from '../../Logger';
 
 localforage.config({
     name: 'OnyxDB',
 });
+
+// Key/value store of Onyx key and arrays of values to merge
+const mergeQueue = {};
+
+/**
+ * Given an Onyx key and value this method will combine all queued
+ * value updates and return a single value. Merge attempts are
+ * batched. They must occur after a single call to get() so we
+ * can avoid race conditions.
+ *
+ * @private
+ * @param {String} key
+ * @param {*} data
+ *
+ * @returns {*}
+ */
+function applyMerge(key, data) {
+    const mergeValues = mergeQueue[key];
+    if (_.isArray(data) || _.every(mergeValues, _.isArray)) {
+        // Array values will always just concatenate
+        // more items onto the end of the array
+        return _.reduce(mergeValues, (modifiedData, mergeValue) => [
+            ...modifiedData,
+            ...mergeValue,
+        ], data || []);
+    }
+
+    if (_.isObject(data) || _.every(mergeValues, _.isObject)) {
+        // Object values are merged one after the other
+        return _.reduce(mergeValues, (modifiedData, mergeValue) => {
+            // lodash adds a small overhead so we don't use it here
+            // eslint-disable-next-line prefer-object-spread, rulesdir/prefer-underscore-method
+            const newData = Object.assign({}, fastMerge(modifiedData, mergeValue));
+
+            // We will also delete any object keys that are undefined or null.
+            // Deleting keys is not supported by AsyncStorage so we do it this way.
+            // Remove all first level keys that are explicitly set to null.
+            return _.omit(newData, (value, finalObjectKey) => _.isNull(mergeValue[finalObjectKey]));
+        }, data || {});
+    }
+
+    // If we have anything else we can't merge it so we'll
+    // simply return the last value that was queued
+    return _.last(mergeValues);
+}
 
 const provider = {
     /**
@@ -57,10 +103,36 @@ const provider = {
      * @return {Promise<void>}
      */
     multiMerge(pairs) {
-        const tasks = _.map(pairs, ([key, value]) => this.setItemQueue.push({key, value, shouldMerge: true}));
+        const tasks = _.map(pairs, ([key, value]) => {
+            if (mergeQueue[key]) {
+                mergeQueue[key].push(value);
+                return Promise.resolve();
+            }
+
+            mergeQueue[key] = [value];
+            Storage.getItem(key).then((data) => {
+                try {
+                    const modifiedData = applyMerge(key, data);
+
+                    // Clean up the write queue so we
+                    // don't apply these changes again
+                    delete mergeQueue[key];
+
+                    this.setItemQueue.push({key, modifiedData, shouldMerge: true}).then(() => Promise.resolve());
+                } catch (error) {
+                    Logger.logAlert(`An error occurred while applying merge for key: ${key}, Error: ${error}`);
+                }
+
+                return Promise.resolve();
+            });
+        });
 
         // We're returning Promise.resolve, otherwise the array of task results will be returned to the caller
         return Promise.all(tasks).then(() => Promise.resolve());
+    },
+
+    mergeItem(key, value) {
+        return this.multiMerge([[key, value]]);
     },
 
     /**
@@ -111,6 +183,10 @@ const provider = {
      */
     setItem(key, value) {
         return this.setItemQueue.push({key, value});
+    },
+
+    hasPendingMergeForQueue(key) {
+        return Boolean(mergeQueue[key]);
     },
 };
 
--- a/lib/storage/providers/LocalForage.js
+++ b/lib/storage/providers/LocalForage.js
@@ -31,14 +31,6 @@ const mergeQueue = {};
  */
 function applyMerge(key, data) {
     const mergeValues = mergeQueue[key];
-    if (_.isArray(data) || _.every(mergeValues, _.isArray)) {
-        // Array values will always just concatenate
-        // more items onto the end of the array
-        return _.reduce(mergeValues, (modifiedData, mergeValue) => [
-            ...modifiedData,
-            ...mergeValue,
-        ], data || []);
-    }
 
     if (_.isObject(data) || _.every(mergeValues, _.isObject)) {
         // Object values are merged one after the other
--- a/lib/storage/providers/SQLiteStorage.js
+++ b/lib/storage/providers/SQLiteStorage.js
@@ -120,6 +120,10 @@ const provider = {
     * @returns {Promise<void>}
     */
     clear: () => db.executeAsync('DELETE FROM keyvaluepairs;', []),
+
+    hasPendingMergeForQueue() {
+        return false;
+    },
 };
 
 export default provider;
--- a/lib/storage/providers/SQLiteStorage.js
+++ b/lib/storage/providers/SQLiteStorage.js
@@ -82,16 +82,20 @@ const provider = {
     multiMerge(pairs) {
         // Note: We use `ON CONFLICT DO UPDATE` here instead of `INSERT OR REPLACE INTO`
         // so the new JSON value is merged into the old one if there's an existing value
-        const query = `INSERT INTO keyvaluepairs (record_key, valueJSON) 
-             VALUES (:key, JSON(:value)) 
-             ON CONFLICT DO UPDATE 
-             SET valueJSON = JSON_PATCH(valueJSON, JSON(:value)); 
+        const query = `INSERT INTO keyvaluepairs (record_key, valueJSON)
+             VALUES (:key, JSON(:value))
+             ON CONFLICT DO UPDATE
+             SET valueJSON = JSON_PATCH(valueJSON, JSON(:value));
         `;
         const queryArguments = _.map(pairs, (pair) => {
             const value = JSON.stringify(pair[1]);
             return [pair[0], value];
         });
         return db.executeBatchAsync([[query, queryArguments]]);
+    },
+
+    mergeItem(key, value) {
+        return this.multiMerge([[key, value]]);
     },
 
     /**
--- a/lib/applyMerge.js
+++ b/lib/applyMerge.js
@@ -1,0 +1,38 @@
+import _ from 'underscore';
+import fastMerge from './fastMerge';
+
+/**
+ * Given an Onyx key and value this method will combine all queued
+ * value updates and return a single value. Merge attempts are
+ * batched. They must occur after a single call to get() so we
+ * can avoid race conditions.
+ *
+ * @private
+ * @param {String} key
+ * @param {*} data
+ * @param {Array} mergeQueue
+ * @returns {*}
+ */
+function applyMerge(key, data, mergeQueue) {
+    const mergeValues = mergeQueue[key];
+
+    if (_.isObject(data) || _.every(mergeValues, _.isObject)) {
+        // Object values are merged one after the other
+        return _.reduce(mergeValues, (modifiedData, mergeValue) => {
+            // lodash adds a small overhead so we don't use it here
+            // eslint-disable-next-line prefer-object-spread, rulesdir/prefer-underscore-method
+            const newData = Object.assign({}, fastMerge(modifiedData, mergeValue));
+
+            // We will also delete any object keys that are undefined or null.
+            // Deleting keys is not supported by AsyncStorage so we do it this way.
+            // Remove all first level keys that are explicitly set to null.
+            return _.omit(newData, (value, finalObjectKey) => _.isNull(mergeValue[finalObjectKey]));
+        }, data || {});
+    }
+
+    // If we have anything else we can't merge it so we'll
+    // simply return the last value that was queued
+    return _.last(mergeValues);
+}
+
+export default applyMerge;
--- a/tests/unit/onyxMultiMergeWebStorageTest.js
+++ b/tests/unit/onyxMultiMergeWebStorageTest.js
@@ -17,7 +17,7 @@ const initialData = {
 
 localforageMock.setInitialMockData(initialData);
 
-describe('Onyx.mergeCollection() amd WebStorage', () => {
+describe('Onyx.mergeCollection() and WebStorage', () => {
     let Onyx;
 
     beforeAll(() => {
--- a/tests/unit/onyxTest.js
+++ b/tests/unit/onyxTest.js
@@ -146,7 +146,7 @@ describe('Onyx', () => {
             });
     });
 
-    it('should merge arrays by appending new items to the end of a value', () => {
+    it('should merge arrays by replacing previous value with new value', () => {
         let testKeyValue;
         connectionID = Onyx.connect({
             key: ONYX_KEYS.TEST_KEY,
@@ -162,7 +162,7 @@ describe('Onyx', () => {
                 return Onyx.merge(ONYX_KEYS.TEST_KEY, ['test2', 'test3', 'test4']);
             })
             .then(() => {
-                expect(testKeyValue).toStrictEqual(['test1', 'test2', 'test3', 'test4']);
+                expect(testKeyValue).toStrictEqual(['test2', 'test3', 'test4']);
             });
     });
 
@@ -198,7 +198,7 @@ describe('Onyx', () => {
         Onyx.merge(ONYX_KEYS.TEST_KEY, ['test2']);
         return waitForPromisesToResolve()
             .then(() => {
-                expect(testKeyValue).toEqual(['test1', 'test2']);
+                expect(testKeyValue).toEqual(['test2']);
             });
     });
 


lib/Onyx.js
```diff
@@ -867,9 +857,25 @@ function set(key, value) {
...
+ * @returns {Promise}
+ */
+function set(key, value) {
+    // Logging properties only since values could be sensitive things we don't want to log
+    Logger.logInfo(`set() called for key: ${key}${_.isObject(value) ? ` properties: ${_.keys(value).join(',')}` : ''}`);
+
+    return new Promise(resolve => Storage.setItem(key, value)
+        .then(updatedValue => resolve(
+            notifyAfterWrite(key, updatedValue, Storage.mergeItem),
+        )).catch(error => evictStorageAndRetry(error, set, key, value)));
```
marcaaron, line 878: Not sure if I am reading this correctly - but we are using the returned promise value of `Storage.setItem()` and resolving this outer promise with the result of calling `notifyAfterWrite()`. What is the overall thought process behind this change?

marcaaron: Doesn't this mean that subscribers will be notified only after we save to storage? My assumption just based on looking at this is that it will be slower to notify subscribers and maybe break some other expectations about the order of operations. I also generally feel like the order of operations here should by "cache", "notify" then "store". That ordering just _feels_ correct to me. Doesn't matter when the writes happen and we are optimizing updating the cache and subscribers.

marcaaron (commented):
Just leaving a couple comments. I think it would be good to lay out this approach somewhere and get more feedback. 

@tgolen or @kidroca might also be able to help validate the correct approach. 

The existing code (merge queue stuff) is far from perfect and could likely be improved - but something about the `notifyAfterWrite()` direction is making me uneasy and I would appreciate some more eyes on this problem.

lib/Onyx.js
kidroca: From what you're correct @marcaaron, this would only notify subscribers after writes complete
This is something that we don't want, not just because of slow performance, but also because the write might not complete at all 

We shouldn't wait to notify, but preserve the existing functionality and notify then write

_One of the big performance hits on Android was waiting for the write to complete and then notifying connections_

lib/storage/providers/LocalForage.js
```diff
@@ -57,12 +62,37 @@ const provider = {
...
+                // Clean up the write queue so we
+                // don't apply these changes again
+                delete mergeQueue[key];
+
+                this.setItemQueue.push({key, modifiedData, shouldMerge: true}).then(() => Promise.resolve());
+            } catch (error) {
+                Logger.logAlert(`An error occurred while applying merge for key: ${key}, Error: ${error}`);
+            }
+
+            return Promise.resolve();
```
kidroca, line 85: This method is not waiting for the write operation to complete, but resolves immediately - the method should resolve only after the write is over, same as the other storage provider implementations

lib/storage/providers/LocalForage.js
```diff
@@ -57,12 +62,37 @@ const provider = {
      * @return {Promise<void>}
      */
     multiMerge(pairs) {
-        const tasks = _.map(pairs, ([key, value]) => this.setItemQueue.push({key, value, shouldMerge: true}));
+        const tasks = _.map(pairs, ([key, value, currentValue]) => {
```
kidroca, line 65: If the parameter is a no longer a key/value pair but also contains `currentValue`, the variable should no longer be called `pairs` 

---

Have we searched for ways to not pass the `currentValue` for each key?

Since this is being queued it's possible that `currentValue` would no longer be current at the time of the write

This is the only storage provider that needs `currentValue` to work - the rest don't have such parameter

lib/storage/providers/LocalForage.js
```diff
@@ -57,12 +62,37 @@ const provider = {
...
+            if (mergeQueue[key]) {
+                mergeQueue[key].push(value);
+                return Promise.resolve();
```
kidroca, line 68: Another place where the method resolve before the write is over 

This might swallow errors, like prevents us from running the storage eviction logic to cleanup space 

There should be a way to wait and resolve after `mergeQueue[key]` is done

lib/Onyx.js
```diff
@@ -982,28 +934,20 @@ function applyMerge(key, data) {
...
-    return get(key)
-        .then((data) => {
-            try {
-                const modifiedData = applyMerge(key, data);
-
-                // Clean up the write queue so we
-                // don't apply these changes again
-                delete mergeQueue[key];
+    // Logging properties only since values could be sensitive things we don't want to log
+    Logger.logInfo(`emrge() called for key: ${key}${_.isObject(value) ? ` properties: ${_.keys(value).join(',')}` : ''}`);
```
kidroca, line 938:
```suggestion
    Logger.logInfo(`merge() called for key: ${key}${_.isObject(value) ? ` properties: ${_.keys(value).join(',')}` : ''}`);
```

lib/storage/providers/AsyncStorage.js
```diff
@@ -56,8 +61,35 @@ const provider = {
      * @return {Promise<void>}
      */
     multiMerge(pairs) {
-        const stringPairs = _.map(pairs, ([key, value]) => [key, JSON.stringify(value)]);
-        return AsyncStorage.multiMerge(stringPairs);
+        const tasks = _.map(pairs, ([key, value, currentValue]) => {
+            if (mergeQueue[key]) {
+                mergeQueue[key].push(value);
+                return Promise.resolve();
```
kidroca, line 67: This method sometimes resolve early, when the write is not yet complete

---

I think we can skip the changes here and delete this storage provider 
IIRC it's no longer used
- web/desktop uses LocalForage
- native uses SQLite

lib/Onyx.js
```diff
@@ -982,28 +934,20 @@ function applyMerge(key, data) {
...
+    if (_.isArray(value)) {
+        return set(key, value);
+    }
 
-            return Promise.resolve();
-        });
+    return new Promise(resolve => get(key)
+        .then(currentValue => Storage.mergeItem(key, value, currentValue)
+            .then(updatedValue => resolve(
+                notifyAfterWrite(key, updatedValue, Storage.mergeItem),
```
kidroca, line 949: `mergeItem` does not resolve  with `updatedValue` it resolves with `Promise.resolve()` - nothing

kidroca (commented):
Left some notes 

---

I have a question:
- We need to perform the merge in order to update memory cache
- Then why do we move the merge logic to storage providers

If we need to extract the merge logic we can extract it to a lib and import the storage provider there (this would import the Storage provider we're currently using)

lib/storage/providers/AsyncStorage.js
chrispader: It's still being used for tests though.

lib/Onyx.js
kidroca: > If i understand the current approach correctly this would essentially mean that we effectively can't use `JSON_PATCH()` on native without manually merging large JS objects too. (because we need to notify before write)

Do we have a proven problem with merging large objects in JS 
Letting native do the merge, would mean we still get one hell of a chunk back - the whole merged object communicated between native and JS 
I think ideally we'd just pass the delta to native, while JS should be performant enough to merge the result to cache

Another way to look at this might be - If large JS objects are a problem - don't have large JS objects - structure App in smaller chunks (paging for report actions)

lib/storage/providers/AsyncStorage.js
chrispader: I'm currently trying to remove all the remaining dependencies on async-storage in the tests. After that, we should be ready to remove it

lib/Onyx.js
marcaaron: I'm not sure if I follow the discussion completely, but my thinking is:

1. if we are calling `Onyx.set()` there is no advantage to JSON_PATCH()
2. if we are calling `Onyx.merge()` then we should still "merge the complete object" in the JS cache. But when we do the hand off to storage just pass the partial object to SQLite JSON_PATCH() if we can (native) otherwise (web) set the complete object.

marcaaron: >if we explicitly want sqlite to handle merging directly on native, instead of merging big objects in onyx before passing the new data to the storage provide

said another way, I don't see a clear way for this to work atm without major changes so I would suggest not trying to avoid it and just focus on implementing JSON_PATCH() for partial objects. The "cache merge" and the "storage merge" behavior should match so it seems fine to pass partial objects to SQLite.

marcaaron:
I think maybe this one is on hold pending the AsyncStorage removal?

marcaaron:
Still holding

chrispader:
Should be ready for review!

lib/Onyx.js
```diff
@@ -837,6 +841,32 @@ function evictStorageAndRetry(error, onyxMethod, ...args) {
...
+ */
+// eslint-disable-next-line rulesdir/no-negated-variables
+function notify(key, modifiedData, isMerge = false) {
+    // Logging properties only since values could be sensitive things we don't want to log
+    Logger.logInfo(`${isMerge ? 'merge' : 'set'}() called for key: ${key}${_.isObject(modifiedData) ? ` properties: ${_.keys(modifiedData).join(',')}` : ''}`);
+
+    // If the value in the cache is the same as what we have then do not update subscribers unless they
+    // have initWithStoredValues: false then they MUST get all updates even if nothing has changed.
+    if (!cache.hasValueChanged(key, modifiedData)) {
+        cache.addToAccessedKeys(key);
```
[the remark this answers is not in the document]
kidroca: I see this code was originally part of `set` and was moved here, so I'll ask @marcaaron to chime in

lib/Onyx.js
```diff
@@ -855,21 +885,11 @@ function set(key, value) {
...
-    cache.set(key, value);
-    notifySubscribersOnNextTick(key, value);
+    // We don't wait until the new data is written to storage.
+    // Instead, we immediately notify subscribers and then return the new data
+    const setPromise = Storage.setItem(key, value);
 
-    // Write the thing to persistent storage, which will trigger a storage event for any other tabs open on this domain
-    return Storage.setItem(key, value)
-        .catch(error => evictStorageAndRetry(error, set, key, value));
+    return notify(key, value).then(() => setPromise).catch(error => evictStorageAndRetry(error, set, key, value));
```
kidroca, line 892: It seems unnecessary to have `notify` return a promise. We know `notify` always resolve immediately 

Why not just call `notify(key, value)` and leave the rest of the code unmodified?

```js
    notify(key, value, 'set');

    // Write the thing to persistent storage, which will trigger a storage event for any other tabs open on this domain
    return Storage.setItem(key, value)
        .catch(error => evictStorageAndRetry(error, set, key, value));
```

lib/Onyx.js
```diff
@@ -926,19 +934,12 @@ function hasPendingMergeForKey(key) {
...
-    if (_.isArray(data) || _.every(mergeValues, _.isArray)) {
-        // Array values will always just concatenate
-        // more items onto the end of the array
-        return _.reduce(mergeValues, (modifiedData, mergeValue) => [
-            ...modifiedData,
-            ...mergeValue,
-        ], data || []);
-    }
+
+    if (_.isArray(data) || _.every(mergeValues, _.isArray)) { return _.last(mergeValues); }
```
kidroca, line 942: It'll probably help if we leave a comment about why we do this

lib/Onyx.js
```diff
@@ -997,7 +998,11 @@ function merge(key, value) {
                 // don't apply these changes again
                 delete mergeQueue[key];
 
-                return set(key, modifiedData);
+                // We don't wait until the merge is written to storage.
+                // Instead, we immediately notify subscribers and then return the merge promise
+                const mergePromise = OnyxMerge.merge(key, changes, modifiedData);
+
+                return notify(key, modifiedData, true).then(() => mergePromise);
```
kidroca, line 1005: Similar to the comment in `set` - we don't need `notify` to be a promise returning method 

We can just go for:

```js
// We relay changes to subscribers optimistically to avoid lag (think UI updates)
notify(key, modifiedData, 'merge');
return OnyxMerge.merge(key, changes, modifiedData);
```

We don't need a comment explaining "what" and "how" but "why"

lib/Onyx.js
```diff
@@ -997,7 +998,11 @@ function merge(key, value) {
                 // don't apply these changes again
                 delete mergeQueue[key];
 
-                return set(key, modifiedData);
+                // We don't wait until the merge is written to storage.
+                // Instead, we immediately notify subscribers and then return the merge promise
+                const mergePromise = OnyxMerge.merge(key, changes, modifiedData);
```
kidroca, line 1003: Seeing what `OnyxMerge.native` and `OnyxMerge.web` files do, why not just call `Storage.mege` here

e.g.
- make `Storage.merge` provider for native have the interface of `merge(key, changes)`
- make `Storage.merge` for web have the an interface like `merge(key, changes, modifiedData)` 

So we don't have 2 new files for this 
Also OnyxMerge.merge is a bit weird name/concept

kidroca (changes requested):
Added a few notes on potential improvements

lib/Onyx.js
marcaaron: Am I seeing this correctly that we will only call `OnyxMerge.merge()` with the final `changes`? The way the merge queue thing works is that it will batch them all together and then do a `set()`. Here we are doing a `merge()` not with the complete partial changes, but only the very last merge call?

lib/Onyx.js
marcaaron: We also have to be careful as this does change some expectations about how we handle array merging. But last time I looked we got rid of most logic that concatenated arrays.

This comment needs to be updated as well:

https://github.com/Expensify/react-native-onyx/blob/e5ac3187565e4e2fad0504da6a121d8419cb01dc/lib/Onyx.js#L966-L969

marcaaron (changes requested):

jestSetup.js
```diff
@@ -2,3 +2,7 @@ jest.mock('./lib/storage');
 jest.mock('./lib/storage/NativeStorage', () => require('./lib/storage/__mocks__'));
 jest.mock('./lib/storage/WebStorage', () => require('./lib/storage/__mocks__'));
 jest.mock('./lib/storage/providers/LocalForage', () => require('./lib/storage/__mocks__'));
+
+// Jest doesn't recognize native file extensions such as `.web.js` or `.native.js`.
+// This is why we need to default one of the platforms implementations
```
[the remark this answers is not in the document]
chrispader: The only reason why this is needed, is because jest will not recognise these native file extension. In the end, we can only use the localForageMock in `./src/storage/__mocks__/index.js` for testing..

lib/Onyx.js
chrispader: I think i understand what you guys are saying now.

I already simplified `OnyxMerge.merge()` to `Storage.merge()`, as @kidroca suggested btw...

It's definitely wrong, that we only pass the "current" `changes` from a `Onyx.merge()` call to the `OnyxMerge.merge()` function. From my understanding though, this would mean, that it's not the latest, but the first changes that are passed to `OnyxMerge.merge()`, since the function will early-return if there is already a value in the `mergeQueue`, and thus the `OnyxMerge.merge()` call with the "latest" changes will never happen.

I also went for the approach of removing the `mergeQueue` completely. The only problem in this case is, that we need to make sure, that a previous' `Onyx.merge()` promise has already finished. Otherwise, the second `Onyx.merge()` call will simply overwrite the previous one.

In the tests, this would mean to add a lot of `waitForPromisesToResolve()` calls and in production, we would need to adapt to this behaviour too.

jestSetup.js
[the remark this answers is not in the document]
chrispader: The need for a platform (storage provider) specific is the essential thing for this issue. We need to be able to pass only the delta to the `SQLiteStorage` and pass the (already) modified/merged data to `LocalForage`.

The `Storage.merge(key, changes, modifiedData)` signature with the additional `changes` paremeter solves this, as we can use the changes on native platforms and ignore the `modifiedData`. On web, we would just ignore the `changes`.

lib/Onyx.js
chrispader: I created a separate branch with the `mergeQueue` removed [here](https://github.com/chrispader/react-native-onyx/tree/%40chrispader/feat/use-json-patch-for-merge+remove-merge-queue)

lib/Onyx.js
chrispader: Updated the comment, though one of you should probably look over it, maybe the explanation can be formed more easily..

lib/Onyx.js
```diff
@@ -978,16 +972,16 @@ function applyMerge(key, data) {
...
+ * @param {(Object|Array)} changes Object or Array value to merge
  * @returns {Promise}
  */
-function merge(key, value) {
+function merge(key, changes) {
     if (mergeQueue[key]) {
-        mergeQueue[key].push(value);
+        mergeQueue[key].push(changes);
         return Promise.resolve();
     }
```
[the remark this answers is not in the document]
kidroca: To avoid race conditions something like the following might be used (using the SyncQueue would be similar, but blocking for all keys)

```js
// pending merges by key
const mergePromises = {};

function merge(key, changes) {
  // wait any prior updates by chaining after an existing promise or set up a new chain
  const megePromise = mergePromises[key] || Promise.resolve();

  mergePromises[key] = mergePromise
    .then(() => get(key))
    .then(data => {
      const modifiedData = applyMerge(key, data);      

      // We relay changes to subscribers optimistically to avoid lag (think UI updates)
      notify(key, modifiedData, 'merge');
      return Storage.merge(key, changes, modifiedData);
    })
    .catch((error) => Logger.logAlert(`An error occurred while applying merge for key: ${key}, Error: ${error}`));

  return mergePromises[key];
}
```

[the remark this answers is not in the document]
chrispader: I like that idea! Let's do that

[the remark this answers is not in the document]
chrispader: With this approach, we'll still need to explicitly wait for a merge to complete, for example if we call `Onyx.clear`/`Onyx.set` right after...

Ofc, multiple merges will wait for the the previous merge to complete, but we would also need to add waiting to any other Onyx functions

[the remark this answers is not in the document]
kidroca: I think keys are only used in a certain way - we either update a key using only `merge` or only `set` - they are rarely mixed 
It's unlikely that a merge and a set for the same key would overlap - updates are primarily merged 
The initial value might be defined with set though I'm not sure 

It's true that nothing should run after Onyx.clear, but what do we do about it right now - I think it's covered by existing logic, and it would remain covered by the same logic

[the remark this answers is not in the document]
chrispader: Ok, got it. We'd still need to update tests, so those cover only the real scenarios, that we support. Because atm there are also some tests, where set and merge are used successively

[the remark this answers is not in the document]
chrispader: I'm mostly referring to [this thread on Slack](https://expensify.slack.com/archives/C01GTK53T8Q/p1683300135739449)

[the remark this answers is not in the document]
kidroca: To clarify: my proposal is to promisify the `merge` func so it waits any running merge for the same key: https://github.com/Expensify/react-native-onyx/pull/238#discussion_r1184585616

We would also like to `logAlert` when either `set` is called for a key we're still merging or vice versa
I wouldn't worry much about clear, because it very unlikely to overlap with a merge and IIRC calling `clear` aborts any pending merges in the queue

Back to `logAlert` - to make this work we'll need to temporarily store which key is currently being set/merged 
Instead of doing that we can just as easily put a promise in `mergePromises` (let's rename that to `keyOperations` or `keyWrites`) 
Now whenever something is being written to a key, other methods that want to write to the same key can wait

```js
// pending set or merges by key
const keyOperations = {};

function merge(key, changes) {
  // wait any prior updates by chaining after an existing promise or set up a new chain
  const writePromise = keyOperations[key] || Promise.resolve();

  keyOperations[key] = writePromise
    .then(() => get(key))
    .then(data => { ... })
    .catch((error) => Logger.logAlert(`An error occurred while applying merge for key: ${key}, Error: ${error}`));

  return keyOperations[key];
}
```
`function set` would be doing something similar 

```js
function set(key, value) {
  // ... existing logic

  // wait any prior updates by chaining after an existing promise or set up a new chain
  const writePromise = keyOperations[key] || Promise.resolve();

   keyOperations[key]  = writePromise
     .then(() => Storage.setItem(key, value))
     .catch(error => evictStorageAndRetry(error, set, key, value))

  return keyOperations[key];
}
```

Adding logging would have brought similar logic in order to track overlaps, but we can just solve the problem, by making key writes wait each other (when they overlap)

The same concept can be applied to `function remove` so it locks the `key` by setting a promise in `keyOperations[key]` - any writes that started after `remove()` would wait and proceed after

---

Whatever we do we also need to find a way to verify any changes to `function merge` did not introduce regressions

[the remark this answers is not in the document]
chrispader: Thanks, that makes sense!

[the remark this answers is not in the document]
chrispader: `Onyx.clear` is not aborting current merge's, so we can go for one of these 3 options:

1. Use the same concept (as above) in `clear` and wait for `merge`/`set`/`remove` to be done, before clearing
2. Abort current promise, e.g. with an `AbortController`
3. Ignore this case, and remove tests that therefore currently fail, such as this one:

```js
    it('should replace the value of Onyx.merge with the default key state in the cache', () => {
        expect.assertions(3);

        // Given that Onyx is completely clear
        // When merge then clear is called on a key with a default key state
        Onyx.merge(ONYX_KEYS.DEFAULT_KEY, MERGED_VALUE);
        Onyx.clear();
            .then(() => {
                // Then the value in Onyx, the cache, and Storage is the default key state
                expect(onyxValue).toBe(DEFAULT_VALUE);
                const cachedValue = cache.getValue(ONYX_KEYS.DEFAULT_KEY);
                expect(cachedValue).toBe(DEFAULT_VALUE);
                const storedValue = StorageMock.getItem(ONYX_KEYS.DEFAULT_KEY);
                return expect(storedValue).resolves.toBe(DEFAULT_VALUE);
            });
    });
```

cc @kidroca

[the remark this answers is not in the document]
kidroca: Since `clear` does not wait `merge` I find it interesting how this test was passing before?

This is what I meant, when I said `clear` aborts write operations:
https://github.com/Expensify/react-native-onyx/blob/1d496b001f327c509131597faf260a81c002be77/lib/storage/providers/LocalForage.js#L84-L87

I think that was only needed on Web, because write operations are queued

I think you can make `clear` wait all the `keyOperations` and then run, but I don't think this would be necessary for App to function correctly
Aborting would be optimal, but it would introduce more complexity

lib/Onyx.js
```diff
@@ -804,7 +807,14 @@ function notifyCollectionSubscribersOnNextTick(key, value) {
 function remove(key) {
     cache.drop(key);
     notifySubscribersOnNextTick(key, null);
-    return Storage.removeItem(key);
+
+    const writePromise = keyOperations[key] || Promise.resolve();
```
kidroca, line 811: Since now many functions are using this logic, and it might not be clear to other what we're doing here, it would be best to extract it to a function like `queueModification` .
This would make it clearer and the function itself might provide further documentation on its usage / use case 

```js
function queueModification(key, runModification) {
  const currentTask = keyOperations[key] || Promise.resolve();
  keyOperations[key] = currentTask.then(runModification);
  return keyOperations[key];
}
```

```js
function remove(key) {
    cache.drop(key);
    notifySubscribersOnNextTick(key, null);

   const nextModification = () => 
       Storage.removeItem(key)
              .catch(error => Logger.logAlert(`An error occurred while removing key: ${key}, Error: ${error}`)));

    return queueModification(key, nextModification);
}
```

Later `queueModification` can serve as optimization hatch, where it can dedupe or skip redundant ops

kidroca (changes requested):
This is looking good for me, though I noted a few places for improvement

chrispader:
@kidroca @marcaaron All done and ready for review!

chrispader:
@kidroca `Onyx.clear` now also waits for prior promises to resolve. Aborting current merge/set would have been to complex

lib/storage/providers/SQLiteStorage.js
```diff
@@ -94,6 +94,10 @@ const provider = {
         return db.executeBatchAsync([[query, queryArguments]]);
     },
 
+    mergeItem(key, changes) {
```
marcaaron, line 97: Should have a doc (for now - one day typescript hopefully 😄)

lib/Onyx.js
```diff
@@ -981,32 +1008,27 @@ function applyMerge(key, data) {
...
-                const modifiedData = applyMerge(key, data);
+                const modifiedData = applyMerge(existingValue, changes);
 
-                // Clean up the write queue so we
-                // don't apply these changes again
-                delete mergeQueue[key];
-
-                return set(key, modifiedData);
+                // We relay changes to subscribers optimistically to avoid lag (think UI updates)
+                notify(key, modifiedData, 'merge');
```
marcaaron, line 1021: Am I wrong in thinking that this logic will now get called n times per concurrent key we try to update the value for?

lib/Onyx.js
```diff
@@ -981,32 +1008,27 @@ function applyMerge(key, data) {
...
-    if (mergeQueue[key]) {
-        mergeQueue[key].push(value);
-        return Promise.resolve();
-    }
-
-    mergeQueue[key] = [value];
-    return get(key)
-        .then((data) => {
+function merge(key, changes) {
+    const nextModification = () => get(key)
```
marcaaron, line 1015: I think I am right in thinking that this logic is "fine" to duplicate because concurrent calls to `get()` are already hooked into a single request?

lib/Onyx.js
```diff
@@ -908,67 +955,47 @@ function multiSet(data) {
...
  *
  * The types of values that can be merged are `Object` and `Array`. To set another type of value use `Onyx.set()`. Merge
- * behavior uses lodash/merge under the hood for `Object` and simple concatenation for `Array`. However, it's important
- * to note that if you have an array value property on an `Object` that the default behavior of lodash/merge is not to
- * concatenate. See here: https://github.com/lodash/lodash/issues/2872
+ * behavior uses lodash/merge under the hood for `Object` and
+ * for `Array`'s we simply replace the current value with the new one.
  *
  * Calls to `Onyx.merge()` are batched so that any calls performed in a single tick will stack in a queue and get
  * applied in the order they were called. Note: `Onyx.set()` calls do not work this way so use caution when mixing
```
marcaaron, line 1001: Is this comment still true?

marcaaron (changes requested):
I am a little skeptical about putting all the various operations into a queue. I can't really say with confidence whether that will have any unexpected effects. But I'm also still a little confused about why we need it 😬

It does seem like a good idea to make these things all "atomic" and easier to reason about which order they happen - I have always hated the caveat that "set" calls are vaguely "independent" from "merges" as far as batching in a single tick of the event loop goes... not sure if we actually solved that or not in this PR as `merge()` still notifies after a `get() + one tick`  and `set()` notifies after `one tick`.

I think my two main concerns are: 

- queue logic is a bit hard to follow and harder still to understand why it's necessary
- we are adding multiple "notify" calls for concurrent merges (not sure if this will actually make any difference, but it feels like a change from what we had before)

I'm going to tag in @neil-marcellini to get a second opinion on this as he has had his nose in some Onyx concurrency issues before and I might be able to keep this conversation alive while I am OOO this week.

lib/Onyx.js
kidroca: Instead of `queueModification`, we might have `waitPendingWrites(key)` which should make flow/logic more obvious

```
waitPendingWrites(key)
  .then(() => get(key))
```

kidroca:
Some more insights on the storage operation queues 
In one of the previous iterations we discovered that the old `mergeQueue[key]` wasn't more performant than simply queueing the merges based on promises, which would make it possible to handle concurrency with other overlapping operations

Queueing merges also helps with keeping a simple JSON_PATCH logic, where we try to merge the delta ASAP 
_(Yes we still wait for `get`, but most if not always it comes from cache right away)
(Also the old logic did the same - wait for `get` and then notify subscribers)_

lib/Onyx.js
```diff
@@ -18,6 +20,12 @@ const METHOD = {
     CLEAR: 'clear',
 };
 
+// A map controlling concurrent storage modification.
+// Each key is associated with a promise of an ongoing storage operation.
+// This allows sequential chaining of overlapping writes to the same key,
+// effectively preventing concurrent modification conflicts.
```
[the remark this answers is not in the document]
chrispader: Wouldn't it technically be possible then, that the earlier `set` call wins, because e.g. the later one takes longer due to a lot of data being written?

I think we can't really control how long it really takes to write data to storage and without the queuing we cannot 100% assure, that the value from latest `set` call will be written in the end..

lib/Onyx.js
```diff
@@ -794,6 +802,27 @@ function notifyCollectionSubscribersOnNextTick(key, value) {
...
 
+/**
+ * This method waits for prior modifications to complete before running the next modification.
+ *
+ * @param {Function} runModification
+ * @param {String} key If "key" is not set (or set to ALL_KEYS_OPERATION) then the modification will be queued
+ * until all prior modifications complete. After that, a new modification with key -1 will be queued.
+ * @returns {Promise}
+ */
+function queueModification(runModification, key = ALL_KEYS_OPERATION) {
```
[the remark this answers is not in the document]
chrispader: Agree, this is a more obvious approach. The reason why i added this, is because we want to be able to

1. wait with `clear` to execute until all previous write operations are complete (because e.g. `merge` could overwrite `clear` if it takes longer)
2. wait with all write operations for `clear` to finish, because otherwise `clear` might remove certain keys afterwards again

lib/Onyx.js
```diff
@@ -853,26 +909,17 @@ function set(key, value) {
...
-        return Promise.resolve();
-    }
-
-    // Adds the key to cache when it's not available
-    cache.set(key, value);
-    notifySubscribersOnNextTick(key, value);
+    const nextModification = () => Storage.setItem(key, value)
+        .catch((error) => {
+            // We have to resolve the current promise for the key, so that the next set doesn't block
+            keyOperations[key] = Promise.resolve();
```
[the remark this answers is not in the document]
chrispader: Yes, as @kidroca wrote, this is because the "inner" `set` call in `evictStorageAndRetry` would be blocked by the "outer" `set` call, because there will already be a promise in the `keyOperations`.

I don't know what the expected behaviour is, if `Storage.setItem` rejects after evicting. Should it try to evict again or just abort? If eviction should happen again, we will need to pass `set` rather than `Storage.setItem`

lib/Onyx.js
chrispader: Agree, this naming would seem more obvious/easy

lib/Onyx.js
chrispader: hmm, well it differs between web and native now. We could just remove the part where it says, "uses lodash/merge under the hood". It's not important for using `merge` anyway

lib/Onyx.js
chrispader: Should we change this behaviour?

lib/Onyx.js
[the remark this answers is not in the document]
kidroca: Regardless of the duration it takes - in the internal (native) structure, write operations occur sequentially. Hence, it's improbable for a later `set` to complete before an earlier `set`. Instead, it would hold off until the prior `set` operation is done, a process that happens at a lower level. 

The queue we're constructing serves to facilitate the particular Onyx `merge` behavior which require access to the most recent data. However, reading the latest data isn't feasible if a `set` is concurrently modifying the data key in question.

chrispader:
On all of these latest questions, should we invite some more devs to clarify which approach were gonna go for?

After that i can finish up the PR and get this done! 🚀

cc @kidroca @neil-marcellini @marcaaron (i know you're OOO so don't mind this ping)

lib/Onyx.js
neil-marcellini: FYI, we have some logic in App now which expects that all of the updates in a list are applied before notifying the subscribers. I believe it works for most cases, but it's not perfect. I've thought about adding a `batchUpdate` function to Onyx, but Marc explained that we probably don't need it since calls to merge are effectively batched from the perspective of the subscriber. We also situations where there are multiple merge updates sent for the same object so I think it would be good to keep the existing behavior.

lib/Onyx.js
```diff
@@ -908,67 +965,46 @@ function multiSet(data) {
...
-            const newData = Object.assign({}, fastMerge(modifiedData, mergeValue));
-
-            // We will also delete any object keys that are undefined or null.
-            // Deleting keys is not supported by AsyncStorage so we do it this way.
-            // Remove all first level keys that are explicitly set to null.
-            return _.omit(newData, (value, finalObjectKey) => _.isNull(mergeValue[finalObjectKey]));
-        }, data || {});
+    if (_.isArray(existingValue) || _.isArray(changes)) {
+        return changes;
+    }
```
neil-marcellini, line 986: Why are we no longer concatenating arrays? It doesn't look like we do that in LocalForage so has the behavior changed or am I missing something?

lib/Onyx.js
```diff
@@ -804,7 +843,11 @@ function notifyCollectionSubscribersOnNextTick(key, value) {
 function remove(key) {
     cache.drop(key);
     notifySubscribersOnNextTick(key, null);
-    return Storage.removeItem(key);
+
+    const nextModification = () => Storage.removeItem(key)
+        .catch(error => Logger.logAlert(`An error occurred while removing key: ${key}, Error: ${error}`));
+
+    return waitPendingWrites(nextModification, key);
```
neil-marcellini, line 850: Should we notifySubscribersOnNextTick after waiting for the pending writes instead of before? What happens if one of the pending writes sets the key? With the current changes would we notify that it was null, then non null, then actually set it to null?

neil-marcellini (commented):
I looked at all the changes and asked some questions where things were unclear to me. I'll review this more tomorrow to catch up on the existing conversations.

lib/Onyx.js
```diff
@@ -794,6 +802,27 @@ function notifyCollectionSubscribersOnNextTick(key, value) {
...
+ * until all prior modifications complete. After that, a new modification with key -1 will be queued.
+ * @returns {Promise}
+ */
+function queueModification(runModification, key = ALL_KEYS_OPERATION) {
+    const currentTask = key === ALL_KEYS_OPERATION
+        ? Promise.all(_.values(keyOperations))
+        : Promise.all([
+            keyOperations[ALL_KEYS_OPERATION],
+            keyOperations[key],
+        ]);
```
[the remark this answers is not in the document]
neil-marcellini: > From past experience, we've had situations where `clear` was executed after a logout action. Some pending data merges, which were outdated, were mistakenly combined into the storage that was just cleared. This resulted in a bug. A potential issue arises if we queue write operations to occur after `clear` – it could reintroduce this bug.

This is very important. In my opinion Onyx should work in such a way that clear is absolute, it should do what it says. If I make a bunch of calls in App to Onyx.merge with some massive objects, and then the user is signed out and we clear Onyx, we don't want those pending writes to happen at all, or at least they should not happen after clearing. I think the current implementation achieves that by getting all keys before clearing.

It would be a performance optimization to cancel the pending writes and clear right away, but probably not necessary right now. I think I'm happy with these changes to push a clear operation after pending writes.

lib/Onyx.js
neil-marcellini: BTW [here](https://github.com/Expensify/App/blob/36102d192643703cd83ce0f620f6639ac877a382/src/libs/Network/SequentialQueue.js#L86) is where we are expecting that all updates are applied before notifying subscribers.

I've observed that this doesn't always work as expected and I would appreciate some input on [this comment](https://github.com/Expensify/App/issues/15550#issuecomment-1513933360) if you have time.

Anyways, I think we should keep the current behavior where multiple merges in the span of a single tick would accumulate and then notify.

neil-marcellini (commented):
I read through some of the existing outstanding conversations. I think we should make a few changes to keep this more in line with the existing behavior, but other than that it looks quite good to me.

Question, are our tests adequate for these changes? Do they effectively test the behavior for concurrent storage modifications? It would make me feel better if we also tested some flows in App with these changes, such as transitioning from OldDot as a different user (tests calls to merge and clear), and also maybe a flow that does a lot of calls to merge, like creating a policy, changing the name, inviting two members and deleting one.

lib/Onyx.js
chrispader: Answer in the next question

lib/Onyx.js
```diff
@@ -908,67 +965,46 @@ function multiSet(data) {
...
 
 /**
  * Merge a new value into an existing value at a key.
  *
- * The types of values that can be merged are `Object` and `Array`. To set another type of value use `Onyx.set()`. Merge
- * behavior uses lodash/merge under the hood for `Object` and simple concatenation for `Array`. However, it's important
- * to note that if you have an array value property on an `Object` that the default behavior of lodash/merge is not to
- * concatenate. See here: https://github.com/lodash/lodash/issues/2872
+ * The types of values that can be merged are `Object` and `Array`. To set another type of value use `Onyx.set()`.
+ * Values of type `Object` get merged with the old value, whilst for `Array`'s we simply replace the current value with the new one.
```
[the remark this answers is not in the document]
chrispader: We did that, because there were no cases in the app, were we actually rely on this kind of merging. Also it doesn't really make sense to do that anyway.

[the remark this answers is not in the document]
chrispader: A different scenario that @perunt recently came up with was if we should check for top-level arrays and see if we can merge the contained items. For example we could replace onyx entries like `reportActions_<id>` with just `reportActions`

lib/Onyx.js
chrispader: i think we generally want to notify before, because we're also setting the cache in `notify`. But in this case i agree that we probably shouldn't notify 3x times..

chrispader: With the current implementation, we're also notifying before actually setting the value, right?

Like [here](https://github.com/Expensify/react-native-onyx/blob/5a97bf5282e614b9e9357fb8d620815f0399cef5/lib/Onyx.js#LL871C45-L871C45)

lib/Onyx.js
chrispader: This would bring us back to the original `mergeQueue`.

The only way i could think of to make `JSON_PATCH` work with this type of queuing, would be to only do a native merge (JSON_PATCH) if we have a single item in the queue, a.k.a. only one delta that doesn't need to be combined before actually sending it to Storage.

If multiple merge values exist in `mergeQueue`, we rather merge the value via `applyMerge` and use `Storage.setItem` instead of `Storage.mergeItem`

chrispader: Like @neil-marcellini 's [proposal](https://github.com/Expensify/react-native-onyx/pull/238#discussion_r1204618185), would it be good idea to wait with notifying until after the write operation is done and extract setting the cache out of the `notify` function?

This way, we could spare the `applyMerge` call for single merge items in the queue, since we just notify with what `JSON_PATCH` returns to us as the merge result.

For multiple items in the queue, we'd still need `applyMerge`

lib/Onyx.js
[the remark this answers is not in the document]
neil-marcellini: Thanks for explaining why we decided to change the array merge behavior.

> A different scenario that @perunt recently came up with was if we should check for top-level arrays and see if we can merge the contained items.

So to provide an example it would be something like this right?
```javascript
const reportActions = [
    {
        id: 1,
        message: 'Hi',
        accountID: 1,
    },
    {
        id: 2,
        message: 'How are you?',
        accountID: 2,
    },
    {
        id: 1,
        message: 'Hi',
        accountID: 1,
    },
];

Ony.merge('reportActionsList', reportActions);

const newReportActions = [
    {
        id: 1,
        message: 'Hi',
        accountID: 1,
    },
    {
        id: 2,
        message: 'How are you?',
        accountID: 2,
    },
    {
        id: 3,
        message: 'I'm good thanks',
        accountID: 1,
    },
];

Ony.merge('reportActionsList', newReportActions);
```
The resulting reportActionsList in Onyx would contain exactly what we see in newReportActions, but it would only add the third object to the list since the other two already exist? Also, if we updated a field of an existing object then it would get merged in. If that's what you mean then I think that would be a really nice way for it to work.

lib/Onyx.js
neil-marcellini: > With the current implementation, we're also notifying before actually setting the value, right?

Yes that's true. My concern is that previously we notify and then remove it from storage right away, now we remove it from storage after the pending writes have occurred.

But I suppose that's in line with how Onyx.set works in this implementation. I'm not sure about this honestly. I'm curious to hear what others think.

lib/Onyx.js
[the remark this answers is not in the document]
chrispader: Yes, that's correct! Onyx should then not only compare top-level - in this case an array - but also every items by an identified by some primary key field.

P.S. i think the second item with `id: 1` in `reportActions` is a not intended, is it?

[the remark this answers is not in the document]
marcaaron: >We did that, because there were no cases in the app, were we actually rely on this kind of merging. Also it doesn't really make sense to do that anyway.

We also do it this way because that's how SQLite MergePatch algorithm works by default and the array merging behavior we had was totally confusing and barely used.

marcaaron:
>It would make me feel better if we also tested some flows in App with these changes, such as transitioning from OldDot as a different user (tests calls to merge and clear), and also maybe a flow that does a lot of calls to merge, like creating a policy, changing the name, inviting two members and deleting one.

I agree. I think we need some manual tests here. These changes look OK to me in the abstract (maybe still a bit hard to fully grasp for the uninitiated). But my main concern (and why I have been procrastinating on the review here) is that I'm not entirely confident we aren't missing things and would feel better if we tested some flows. Sign in/out also comes to mind as an important flow to test for regressions.

lib/Onyx.js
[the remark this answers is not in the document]
marcaaron: Formally requesting here that we write a clear explanation in the code as a comment so that anyone who will read this can understand why we have this queue and how it works. Providing an example would also help visualize things. The current comment requires more detail.

lib/Onyx.js
```diff
@@ -18,6 +18,14 @@ const METHOD = {
     CLEAR: 'clear',
 };
 
+// A map controlling concurrent storage modification.
+// Each key is associated with a promise of an ongoing storage operation.
+// This allows sequential chaining of overlapping writes to the same key,
+// effectively preventing concurrent modification conflicts.
+const pendingWriteOperations = {};
+
+let pendingClearOperation = Promise.resolve();
```
marcaaron, line 27: Please add a code comment here to explain what this is.

lib/Onyx.js
```diff
@@ -794,6 +802,36 @@ function notifyCollectionSubscribersOnNextTick(key, value) {
     Promise.resolve().then(() => keysChanged(key, value));
 }
 
+/**
+ * This method waits for prior modifications to complete before running the next modification.
```
marcaaron, line 806: Please define what a "modification" is in this code comment with an example prior to explaining what the function will do with them.

lib/Onyx.js
```diff
@@ -794,6 +802,36 @@ function notifyCollectionSubscribersOnNextTick(key, value) {
...
+ *
+ * @param {Function} nextModification
+ * @param {String} key If "key" is not set then the modification will be queued after all prior modifications completed.
```
marcaaron, line 809: If the key might be null this should be a union type I think like `(string|null)`.

lib/Onyx.js
```diff
@@ -794,6 +802,36 @@ function notifyCollectionSubscribersOnNextTick(key, value) {
...
+ * @returns {Promise}
+ */
+function waitPendingWrites(nextModification, key) {
```
marcaaron, line 812: Maybe default to `null` if we are going to have a check for it? Otherwise, I would not use `null` and would just check to see if we have a key at all.

lib/Onyx.js
```diff
@@ -794,6 +802,36 @@ function notifyCollectionSubscribersOnNextTick(key, value) {
...
+    // If key is not set, then we are doing a modification that affects all keys. We must wait for all prior modifications to complete.
```
marcaaron, line 813: There is a _very specific_ single case where we do this... we are making things harder to understand by alluding to the "modification that affects all keys".

lib/Onyx.js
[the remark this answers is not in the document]
marcaaron: >An earlier suggestion proposed that the clear command should abort any pending operations. However, this could be quite complex to implement, and it might be best to address this in a separate Pull Request.

I guess I'm just not sure how we got to this point or why we started worrying about the clear method and queuing things at all in this PR. How does it relate to implementing JSON_PATCH()? Are we still focused on fixing a single problem here? Can someone TL;DR?

marcaaron (changes requested):
Before I review anymore I want to propose we step back and outline what problems (besides the original one of implementing JSON_PATCH()) we are attempting to solve here and how it all fits together. I am a bit concerned we could be introducing complexity for not much value, but that largely stems from confusion about what we are doing and why we are doing it.

chrispader:
🧵 **Trying to sum up the above discussion about all of the various topics in this PR.** 🧵

`JSON_PATCH` is an algorithm in `SQLite` we can use on iOS and Android to merge delta changes natively (in the storage provider) instead of manually "pre-merging" in the "Onyx layer .

The original goal of this PR was to utilise `JSON_PATCH` on native platforms, whilst still retaining the current merge logic on web.

> (By "pre-merging" i mean using the current implementation of [`applyMerge`](https://github.com/Expensify/react-native-onyx/blob/5a97bf5282e614b9e9357fb8d620815f0399cef5/lib/Onyx.js#L935-L963) which uses a mergeQueue to batch and merge delta changes, after which data will be written to storage with `Storage.setItem` instead of `Storage.mergeItem`)   

Due to the way we have to send delta data to the native storage provider for `JSON_PATCH` to merge we cannot keep the queuing mechanism we had before. (sending only the delta instead the whole pre-merged object)

**Why?** Because with `JSON_PATCH` we want to send (new) deltas to the storage provider immediately, without waiting for other `Onyx.merge` calls to be batched together and pre-merged. The "old" merging logic would initially trigger the logic in `Onyx.merge` once ([on the first call](https://github.com/Expensify/react-native-onyx/blob/5a97bf5282e614b9e9357fb8d620815f0399cef5/lib/Onyx.js#L988-L995)) and then add new deltas to a `mergeQueue` without actually starting any storage operations. The additional values from `mergeQueue` [would then be batched together](https://github.com/Expensify/react-native-onyx/blob/5a97bf5282e614b9e9357fb8d620815f0399cef5/lib/Onyx.js#L935-L963) and merged in `applyMerge`.

This approach isn't compatible with the way we want to use `JSON_PATCH`, because we want to prevent delta data from being combined before actually sending it to the storage provider, so that `JSON_PATCH` can take over merging.

**Why do we need to apply the merge before writing to storage?** Mostly because we don't use `JSON_PATCH` on web and we also have to set the cache, which is custom-built and therefore needs the fully merged data too. 

**So, what can we do?**

The first thing that came to mind is to wait for prior write operations to complete, before starting new write operations (`set`, `merge`, `remove`, `clear`). The downside of that, is that this slows down operations and disables batching of Onyx methods.

**We still want to be able batch `Onyx.merge` calls for the user though, so that the user doesn't encounter multiple unnecessary re-renders in the app.**

**How can we keep the "old" merging mechanism, while still utilising `JSON_PATCH` directly?**

Another possibility is to move notifying (updating the subscriber) until after the write operations. We could then check, if there was another `Onyx.merge` call in the meantime, and if so, start another write operation. After all write operations are completed, we batch all the changes and update the subscriber.

We could still update the cache before writing to storage, by applying the same delta merging for the cache, by calling `applyMerge` with only the delta changes and updating the cache more often. The user would still only be notified once after all `Onyx.merge` operations 

Initially explained some of these thoughts [here](https://github.com/Expensify/react-native-onyx/pull/238#discussion_r1213098537) and [here](https://github.com/Expensify/react-native-onyx/pull/238#discussion_r1213105084)...

**What do you guys think?** Does the last approach seem logical?

cc @kidroca @marcaaron @neil-marcellini

lib/Onyx.js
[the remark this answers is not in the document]
chrispader: Added a more detailed explanation [in this comment](https://github.com/Expensify/react-native-onyx/pull/238#issuecomment-1584173684)

lib/Onyx.js
chrispader: doing `value == null` instead of `value === null` actually means, that we check for undefined too. So we are already checking if the key is defined or not

lib/Onyx.js
chrispader: Do you mean, that we should extract this logic (when key is not defined) to the `clear` function itself? Or rather just change the documentation?

lib/Onyx.js
marcaaron: better to be strict/explicit here

lib/Onyx.js
marcaaron: I personally would extract it. Or add something like `isClearingStorage: bool` argument to be explicit. Then the code documents itself.

marcaaron:
>Because with JSON_PATCH we want to send (new) deltas to the storage provider immediately, without waiting for other Onyx.merge calls to be batched together and pre-merged.

Can you expand on why we want to do this immediately?

>This approach isn't compatible with the way we want to use JSON_PATCH, because we want to prevent delta data from being combined before actually sending it to the storage provider, so that JSON_PATCH can take over merging.

I think same question here. Why do we want it to work this way and prevent the data from getting combined? `JSON_PATCH` can still merge the combined delta. We just need to pass the delta of the result of the existing merge queue.

kidroca:
@marcaaron @neil-marcellini 
The root cause of our issue apply the JSON PATCH originates from the `merge` code in this section: 

https://github.com/Expensify/react-native-onyx/blob/5a97bf5282e614b9e9357fb8d620815f0399cef5/lib/Onyx.js#L997-L1003

Our goal was to:
- ✅ Extract the `Storage.merge` method, native would use JSON PATCH, while web would use the old logic.
- ⚠️ Substitute the `set` call with `Storage.merge`.

A complication arose with the `applyMerge` function as it combined the batched deltas (incremental changes) with the current storage value. As a result, `modifiedData` ends up as the fully merged result. Therefore, calling `Storage.merge(key, modifiedData)` is practically the same as calling `Storage.set`.

Implementing a promise-based queue seemed to be a balanced solution that could work across all platforms, though at the expense of `Onyx.merge` potentially triggering more notifications for subscribers. Initially, we intended to establish this queue only for `Onyx.merge`. However, since the old merge logic is tightly linked with `Onyx.set`, and both can alter the current value, it seemed best to queue all modification operations (anything other than `get`). This strategy ensures that a modification like `merge` that needs to `get` the latest value will wait for anything that is currently modifying the key.

---

To address the original issue, a potentially simpler solution could be introducing a `mergeBatchedDeltas(key)` function. This function would streamline some of the logic from `applyMerge`. Consequently, we could call `Storage.merge` with `key`, `delta` (the merged delta), and `mergedData`. In this setup, the native `Storage.merge` would only use the merged `delta`, while the web would continue to utilize the `mergedData`.

```js
const deltaChanges = mergeBatchedChanges(key);
const modifiedData = applyMerge(deltaChanges, currentValue);

delete mergeQueue[key];

return Storage.merge(key, deltaChanges, modifiedData);
```

marcaaron:
@chrispader is this ready for review yet?

chrispader:
Sorry for the delay! @marcaaron 

Just read the proposal by @kidroca which seems like the optimal solution for now. Gonna start implementing this `mergeBatchedChanges` logic now 👍

neil-marcellini:
Cool! Marking as a draft for now.

chrispader:
Updated the logic with a slightly adapted version of [@kidroca's approach](https://github.com/Expensify/react-native-onyx/pull/238#issuecomment-1584904643).

No promises anymore and we only utilise the `applyMerge` function for both batched changes as well as merging the value for web and notifying, as `applyMerge` and `mergeBatchedChanges` would have done almost the same thing.

Feel free to review and give feedback, but i'm optimistic that this is the final solution! 🥳

cc @kidroca @marcaaron @neil-marcellini

lib/Onyx.js
```diff
@@ -840,6 +843,31 @@ function evictStorageAndRetry(error, onyxMethod, ...args) {
...
 
+/**
+ * Notifys subscribers and writes current value to cache
+ *
+ * @param {String} key
+ * @param {*} value
+ * @param {String} method
+ */
+// eslint-disable-next-line rulesdir/no-negated-variables
+function notifyAndWriteToCache(key, value, method) {
```
kidroca, line 854: This looks like a false positive, due to the function name starting with `not` 

Maybe it's a good time to change the name to something more generic, that doesn't explain details like cache, AFAIK this is some common logic that runs for merges and sets, so a name like `broadcastUpdate` or `propagateUpdate` might be suitable 

Perhaps it's an opportunity to change the function name into something more abstract, effectively encapsulating the internal details, such as cache manipulation. 

Given that this function is running common logic shared by `merge` and `set`, we could consider a more generic name like `broadcastUpdate` or `propagateUpdate`. These names are not only descriptive of the function's primary task (i.e., notifying the necessary components of updates), but also abstract enough to hide the internal specifics of how this task is accomplished.

kidroca:
@chrispader can you resolve outdated threads as they only gets in the way of new reviews - I'm not able to resolve them myself. The old threads around `merge` and `applyMerge` are examples of noise for the latest review

chrispader:
> @chrispader can you resolve outdated threads as they only gets in the way of new reviews - I'm not able to resolve them myself. The old threads around `merge` and `applyMerge` are examples of noise for the latest review

done 👍

lib/Onyx.js
```diff
@@ -908,67 +921,43 @@ function multiSet(data) {
...
-const mergeQueue = {};
-
 /**
- * @private
- * @param {String} key
- * @returns {Boolean}
- */
-function hasPendingMergeForKey(key) {
-    return Boolean(mergeQueue[key]);
-}
```
kidroca, line 921: Why are we removing this functionality?

kidroca (commented):
The merge change looks good

I've added some tips regarding some of the support logic

lib/Onyx.js
chrispader: oh, it just got moved down [here](https://github.com/chrispader/react-native-onyx/blob/d0f261ed20f4dd6295c870404a8edcfe46e34326/lib/Onyx.js#L1009-L1016)

I can move it back, just though it would fit more where the other merge-related functions are

kidroca: Never mind it seems it was moved further down

lib/Onyx.js
chrispader: Already fixed this rule in the `eslint-config-expensify` repo [here](https://github.com/Expensify/eslint-config-expensify/pull/68), but forgot to update the package version.

I'll change this to `broadcastUpdate` anway 👍

chrispader:
@kidroca everything resolved 👍

lib/Onyx.js
```diff
@@ -981,32 +1008,27 @@ function applyMerge(key, data) {
...
             try {
-                const modifiedData = applyMerge(key, data);
+                const modifiedData = applyMerge(existingValue, changes);
 
-                // Clean up the write queue so we
-                // don't apply these changes again
-                delete mergeQueue[key];
-
-                return set(key, modifiedData);
+                // We relay changes to subscribers optimistically to avoid lag (think UI updates)
```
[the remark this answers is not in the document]
marcaaron: Kind of, but I don't feel too passionately. I'd also be satisfied by something like:

>This approach prioritizes fast UI changes without waiting for data to be stored in device storage.

marcaaron (commented):
❤️  👏 a few comments about comments, but overall, LGTM

And there are some lint issues.

chrispader:
@marcaaron @kidroca change requests resolved 👍

marcaaron:
Can we fix the lint errors?

chrispader:
> Can we fix the lint errors?

Done, though i'm not sure if this eslint rule should trigger here. Basically it gave errors because the `withOnyx` HOC calls in the tests don't have `propTypes`, although the component that it wraps has those defined...

marcaaron:
>Basically it gave errors because the withOnyx HOC calls in the tests don't have propTypes, although the component that it wraps has those defined...

I think it's ok to disable it in this case.

chrispader:
> > Basically it gave errors because the withOnyx HOC calls in the tests don't have propTypes, although the component that it wraps has those defined...
> 
> I think it's ok to disable it in this case.

already done 👍

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "react-native-onyx",
-  "version": "1.0.50",
+  "version": "1.0.53",
```
marcaaron, line 3: confused why this shows up in the diff it should be the same as `main` 🤷‍♂️

lib/Onyx.js
```diff
@@ -1222,6 +1233,9 @@ function update(data) {
  */
 function setMemoryOnlyKeys(keyList) {
     Storage.setMemoryOnlyKeys(keyList);
+
+    // When in memory only mode for certain keys we do not want to ever drop items from the cache as the user will have no way to recover them again via storage.
+    cache.setRecentKeysLimit(Infinity);
```
marcaaron, line 1238: Same here. This should be the same in `main` 🤔

lib/Onyx.js
```diff
@@ -853,24 +874,9 @@ function set(key, value) {
...
 
-    // eslint-disable-next-line no-use-before-define
-    if (hasPendingMergeForKey(key)) {
-        Logger.logAlert(`Onyx.set() called after Onyx.merge() for key: ${key}. It is recommended to use set() or merge() not both.`);
-    }
-
-    // If the value in the cache is the same as what we have then do not update subscribers unless they
-    // have initWithStoredValues: false then they MUST get all updates even if nothing has changed.
-    if (!cache.hasValueChanged(key, value)) {
-        cache.addToAccessedKeys(key);
```
marcaaron, line 864: We are removing this, but not adding it back anywhere. Is that intentional?

marcaaron (changes requested):
Can we clean up the diff? There are a bunch of what look like new changes, but are actually recent changes to `main`.

lib/Onyx.js
chrispader: I simplified [this section](https://github.com/chrispader/react-native-onyx/blob/29cc4a31132fb7143d68f10c6b44608bc20ec659/lib/Onyx.js#L857-L861) based on [a suggestion](https://github.com/Expensify/react-native-onyx/pull/238#discussion_r1260997522) by @kidroca

chrispader:
@marcaaron just updated the diff and replied to your questions :)

lib/Onyx.js
marcaaron: Ok, I see. so it happens when we call `cache.set()` now.

marcaaron (approved):
LGTM 👍

marcaaron:
@chrispader Let's raise a PR in `App` to bump the version of Onyx there and do any final testing? Thanks so much for your patience and continued effort on this one. And thank you @kidroca for the additional thoughts and review 🙇

chrispader:
Some issues that arose from this PR are fixed [in this PR](https://github.com/Expensify/react-native-onyx/pull/284)

PR #252, opened 2023-06-10
Use GitHub native concurrency in react-native-onyx

roryabraham:
### Details
This PR is a bit of an experiment so that we can get some hands-on experience using GitHub's native concurrency feature.

To test this PR, we need to:

1. Merge it
1. Merge another two test PRs in quick succession.

and we should expect that:

1. A publishing workflow starts when this PR is merged
1. When the first test PR is merged, another publishing workflow is queued but does not start.
1. When the second test PR is merged, the previously queued publishing workflow is cancelled and a new publishing workflow is queued.

### Related Issues
n/a

### Automated Tests
None

### Linked PRs
None.

PR #256, opened 2023-06-12
Checkout main branch instead of ref that triggered workflow

roryabraham:
Following-up on https://github.com/Expensify/react-native-onyx/pull/252, we need to checkout the main branch, not the ref that triggered the workflow, for this to work as expected.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -18,6 +18,8 @@ jobs:
 
         steps:
             - uses: actions/checkout@v3
+              with:
+                ref: main
 
             - name: Decrypt & Import OSBotify GPG key
               run: |


roryabraham:
This is valuable lesson to learn w/ much faster feedback than we would have in E/App, sorry for the noise react-native-onyx

dangrous (approved):
This seems good (assuming `ref:` is correct, which the internet suggested it was)! I'll let you merge for the testing.

roryabraham:
- https://github.com/Expensify/react-native-onyx/actions/runs/5247955421
- https://github.com/Expensify/react-native-onyx/actions/runs/5247956234

roryabraham:
kk this worked, we are all set here

PR #261, opened 2023-06-19
Don't call selector with selector results

hannojg:
### Details

While investigating an issue in the Expensify app, I found a case where a onyx selector seemed to get called with the previous selector's return value, instead with the value from onyx.
I added a test and was able to verify the buggy behaviour. After some debugging I found that we put the state of the component into the selector function (which feels just wrong, because the selector is supposed to receive the real onyx data). This explained why I was seeing calls of the selector with the previous computed selector value.

This PR fixes this behaviour and thus the `deepEqual` check can run correctly.

### Related Issues

https://github.com/Expensify/App/issues/21022

### Automated Tests

I updated the tests to check for that case.

### Linked PRs

I don't understand this one?

### DIFF
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -364,7 +364,7 @@ function keysChanged(collectionKey, partialCollection) {
                 // returned by the selector.
                 if (subscriber.selector) {
                     subscriber.withOnyxInstance.setState((prevState) => {
-                        const previousData = reduceCollectionWithSelector(prevState[subscriber.statePropertyName], subscriber.selector, subscriber.withOnyxInstance.state);
+                        const previousData = prevState[subscriber.statePropertyName];
                         const newData = reduceCollectionWithSelector(cachedCollection, subscriber.selector, subscriber.withOnyxInstance.state);
 
                         if (!deepEqual(previousData, newData)) {
--- a/tests/unit/subscribeToPropertiesTest.js
+++ b/tests/unit/subscribeToPropertiesTest.js
@@ -190,6 +190,14 @@ describe('Only the specific property changes when using withOnyx() and ', () => 
         })(ViewWithObject);
         return runAllAssertionsForCollection(TestComponentWithOnyx)
             .then(() => {
+                // Expect that the selector always gets called with the full object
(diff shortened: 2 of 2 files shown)


mountiny (approved):
changes look good to me but I will also tag @marcaaron who has the best context for Onyx

luacmartins (commented):
Makes sense to me. What issue were you investigating in App? How can we test this in App to make sure things are not breaking?

mountiny:
@hannojg yep would be great to link the issue for reference

luacmartins:
@hannojg bump on adding the link

hannojg:
This was when working on:

 - https://github.com/Expensify/App/issues/21022

In general I'd say you can test the change by opening the app and check that it doesn't break immediately 😅 (no crash warning on app start or similar).

The `SidebarLinks` are already using the `selector` function in onyx, and this change e.g. doesn't break it.

When using the react performance profiler you can see that you have less re-renders for onyx connected components using the `selector` feature, as the selector won't produce different results (because before this change we were passing to it once the onyx state data, and once the previous selector data, which caused the output to be different, which caused the component to get updated).

lib/Onyx.js
```diff
@@ -364,7 +364,7 @@ function keysChanged(collectionKey, partialCollection) {
                 // returned by the selector.
                 if (subscriber.selector) {
                     subscriber.withOnyxInstance.setState((prevState) => {
-                        const previousData = reduceCollectionWithSelector(prevState[subscriber.statePropertyName], subscriber.selector, subscriber.withOnyxInstance.state);
+                        const previousData = prevState[subscriber.statePropertyName];
```
marcaaron, line 367: Ah hmm so we were "selecting" the already selected data in state? And it would always be different. Makes sense. Great catch.

marcaaron (approved):

marcaaron:
Nice work @hannojg, thanks for catching this one. 🙇

lib/Onyx.js
hannojg: Yes, exactly

marcaaron: Awesome! You are a gem! 💎

PR #262, opened 2023-06-19
Fix unnecessary re-computation

hannojg:
### Details

We already calculated the result earlier, no need to calculate it a second time.

### Related Issues

GH_LINK


MariaHCD (commented):
LGTM, are there any tests (manual or automated) that we can do to verify this doesn't break anything?

hannojg:
@MariaHCD I just relied on the unit tests, and tested the changes locally in the new Expensify app. The tests are all green and the app still works, so I assume we are good 😄 👍

MariaHCD (approved):
Nice, thanks for confirming!

Issue #260, opened 2023-06-12
Prepare react-native-onyx for TypeScript migration


hayata-suenaga:
Part of [the TypeScript migration](https://github.com/Expensify/App/issues/16665) issues.

- [Design doc](https://docs.google.com/document/d/1LNcgYeYyovQ88gNjLIgBqBReLtbLAbxeAbna9QKxeAQ/edit#bookmark=id.f26etjel5h0v)
- Remove `expensify-common` package.
  - Copy methods imported from `expensify-commmon`. There are only two functions imported. ([Stack discussion](https://expensify.slack.com/archives/C0593T50UHG/p1684971334416899?thread_ts=1684958481.278379&cid=C0593T50UHG))
- Create type definition files for exported APIs

neil-marcellini:
I'm also assigning to get credit for all the time I put into the review.

hayata-suenaga:
[The PR](https://github.com/Expensify/react-native-onyx/pull/267) for creating type declaration files are merged and Onyx is ready to be used in App for migration. Closing this 🥳

PR #265, opened 2023-06-27
Remove expensify-common library from project

fabioh8010:
### Details

This PR extracts the two methods from `expensify-common` library that are being used in this project, and remove `expensify-common` as it's not necessary anymore.

### Related Issues

https://github.com/Expensify/react-native-onyx/issues/260

### Automated Tests

Not applicable as this PR is only extracting some methods from `expensify-common`.

### Linked PRs

Not applicable.

### DIFF
--- a/lib/Str.js
+++ b/lib/Str.js
@@ -1,0 +1,30 @@
+import _ from 'underscore';
+
+const Str = {
+    /**
+     * Returns true if the haystack begins with the needle
+     *
+     * @param {String} haystack  The full string to be searched
+     * @param {String} needle    The case-sensitive string to search for
+     * @return {Boolean} Retruns true if the haystack starts with the needle.
+     */
+    startsWith(haystack, needle) {
+        return _.isString(haystack)
+                && _.isString(needle)
+                && haystack.substring(0, needle.length) === needle;
+    },
+
+    /**
+     * Checks if parameter is a string or function
+     * if it is a function then we will call it with
+     * any additional arguments.
+     *
+     * @param {String|Function} parameter
+     * @returns {String}
+     */
+    result(parameter, ...args) {
+        return _.isFunction(parameter) ? parameter(...args) : parameter;
+    },
+};
+
+export default Str;
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -1,14 +1,14 @@
 /* eslint-disable no-continue */
-import _ from 'underscore';
-import Str from 'expensify-common/lib/str';
 import {deepEqual} from 'fast-equals';
 import lodashGet from 'lodash/get';
-import Storage from './storage';
+import _ from 'underscore';
 import * as Logger from './Logger';
 import cache from './OnyxCache';
+import Str from './Str';
 import createDeferredTask from './createDeferredTask';
 import fastMerge from './fastMerge';
 import * as PerformanceUtils from './metrics/PerformanceUtils';
+import Storage from './storage';
 
 // Method constants
 const METHOD = {
--- a/lib/withOnyx.js
+++ b/lib/withOnyx.js
@@ -3,11 +3,11 @@
  * something in Onyx (a key/value store). That way, as soon as data in Onyx changes, the state will be set and the view
  * will automatically change to reflect the new data.
  */
+import PropTypes from 'prop-types';
 import React from 'react';
 import _ from 'underscore';
-import PropTypes from 'prop-types';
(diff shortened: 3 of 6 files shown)


fabioh8010:

lib/Str.js
```diff
@@ -0,0 +1,30 @@
...
+    startsWith(haystack, needle) {
+        return _.isString(haystack)
+                && _.isString(needle)
+                && haystack.substring(0, needle.length) === needle;
+    },
+
+    /**
+     * Checks if parameter is a string or function
+     * if it is a function then we will call it with
+     * any additional arguments.
```
amyevans, line 20:
```suggestion
     * Checks if parameter is a string or function.
     * If it is a function, then we will call it with
     * any additional arguments.
```

lib/Str.js
```diff
@@ -0,0 +1,30 @@
...
+     *
+     * @param {String|Function} parameter
+     * @returns {String}
```
amyevans, line 23: This doesn't necessarily return a string, can we update it?

amyevans (commented):
Some nit picks but otherwise 👍

blazejkustra (approved):

lib/Str.js
hayata-suenaga: I think if a function is passed to the first argument (instead of string), that function should return a string.

I don't know how to provide type information of callbacks, but hopefully when we migrate onyx, we can type that.

hayata-suenaga (approved):

neil-marcellini (dismissed):
Looks good! Why do we need to do this? Is it because expensify-common is going to be using Typescript now? We are only adding type declarations there so I'm still unsure 🤔 

cc @hayata-suenaga

hayata-suenaga:
@neil-marcellini you're right. We don't necessarily have to do this.

However, while examining `onyx`, we discovered there are only two functions used from `expensify-common` in `react-native-onyx` and we decided just remove the library completely.

Eventually, we will migrate `onyx` to TypeScript (vs. just creating definition files). When we do that, it makes things earlier if we have one less internal library.

Relevant slack discussion: https://expensify.slack.com/archives/C0593T50UHG/p1684971334416899?thread_ts=1684958481.278379&cid=C0593T50UHG

Relevant design doc section: https://docs.google.com/document/d/1LNcgYeYyovQ88gNjLIgBqBReLtbLAbxeAbna9QKxeAQ/edit#bookmark=id.f26etjel5h0v

lib/Str.js
```diff
@@ -0,0 +1,30 @@
...
+     * Returns true if the haystack begins with the needle
+     *
+     * @param {String} haystack  The full string to be searched
+     * @param {String} needle    The case-sensitive string to search for
+     * @return {Boolean} Retruns true if the haystack starts with the needle.
+     */
+    startsWith(haystack, needle) {
+        return _.isString(haystack)
+                && _.isString(needle)
+                && haystack.substring(0, needle.length) === needle;
```
roryabraham, line 14: NAB:

```suggestion
                && haystack.startsWith(needle);
```

lib/Str.js
roryabraham: This function is pretty weird 🙃

roryabraham (approved):
LGTM

roryabraham:
Nothing glaring here but I think there's enough little things that I'll wait for @fabioh8010 to make a few of the requested changes before merging. If anyone else is feeling 🤠 feel free to merge.

lib/Str.js
fabioh8010: This function returns `string` if we pass a `string` to `parameter`, or returns the result of `parameter` if it's a function. So, I will change to `{*}` as if we pass a function it can return any type. WDYT?

fabioh8010: @hayata-suenaga In this [example](https://github.com/Expensify/react-native-onyx/blob/main/lib/withOnyx.js#L121) the function is returning a `boolean` instead of a `string`.

fabioh8010:
I've addressed the comments and improved `result` function's JSDoc to be more clearer. @hayata-suenaga As I just copied these two functions from `expensify-common`, should we address those changes there as well in the PR Blazej created?

amyevans (approved):

lib/Str.js
hayata-suenaga: I just don't know why this function exists 😭

hayata-suenaga:
>  should we address those changes there as well in the PR Blazej created?

which PR are you talking about???

PR #277, opened 2023-07-13
Throw in dev if collection keys don't belong to the same parent

luacmartins:
### Details
Coming from [this comment](https://github.com/Expensify/react-native-onyx/pull/272/files#r1261673744), we should make the error easier to catch on dev by throwing it, instead of logging.

cc @marcaaron @mountiny 

### Related Issues
https://github.com/Expensify/react-native-onyx/pull/272/files#r1261673744

### Automated Tests
N/A

### Linked PRs
TBD


mountiny:
I think we can go ahead with this one as its late for Vivek and Marc has proposed this change before

PR #279, opened 2023-07-14
Revert back to `Storage.clear()` in `Onyx.clear()` instead of `removeItems()`. Abort the `setItem()` queue.

marcaaron:
### Details

This is a kind of soft revert of the changes here -> https://github.com/Expensify/react-native-onyx/pull/237

I am not removing the `removeItems()` code yet until we know why it doesn't work.

For now, this will re-introduce the bug here: https://github.com/Expensify/App/issues/15321

I'm not sure why the issue is open and haven't really caught up on it so maybe @roryabraham has more context.

Regardless, having Guides be able to cleanly log out of their accounts seems higher value.

### Related Issues

https://github.com/Expensify/App/issues/21766

### Automated Tests
❌ 

### Linked PRs
TBD

### DIFF
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -1105,7 +1105,7 @@ function clear(keysToPreserve = []) {
 
             // Remove only the items that we want cleared from storage, and reset others to default
             _.each(keysToBeClearedFromStorage, key => cache.drop(key));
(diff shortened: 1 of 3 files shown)


roryabraham:
Agree with the evaluation that guides signing in + out is higher priority than multi-tab edge cases.

marcaaron:
I'm going to close this and leave all the context I have in an issue.

PR #280, opened 2023-07-17
[Fix #269] use cached data

### LINKED ISSUE #269: Proposal: Use cached data directly, skip loading state when possible
If you haven’t already, check out our [contributing guidelines](https://github.com/Expensify/ReactNativeChat/blob/main/contributingGuides/CONTRIBUTING.md) for onboarding and email contributors@expensify.com to request to join our Slack channel!
___

## What performance issue do we need to solve?

[[Slack discussion](https://expensify.slack.com/archives/C01GTK53T8Q/p1688580518417739)]

By default `withOnyx` works like this:
Wrap a component, and set a loading state, while Onyx gets the requested data from the storage (this is async)
When we have the data from the storage we update the state with the real data and set loading to `false`.
In general this is fine, however, this causes an extra render. And we want to avoid extra re-renders as much as possible to have a good performance.
We have the case that we have data already in the memory, which could be read synchronously at the first render of a onyx connected component. However, we always run though the async connect which causes the extra re-render, switching from loading state to the state with the final data.

## What is the impact of this on end-users?

- We often start with our components in a loading state, although we could show the content right away (improved UX)
- Opening screens often feels jumpy / not-smooth as we have a lot of re-renders

This is for example the current state of opening a chat screen in the new dot app:

https://github.com/Expensify/react-native-onyx/assets/16821682/2911ff16-8d85-4ff7-99b6-aeb88dfec56b

## List any benchmarks that show the severity of the issue

This is a react-devtoosl profiler recording of when the app opens. Roughly 66 commits and 1s until the last commit is made (the SidebarLinks becomes visible)

Screenshot 2023-06-28 at 12 36 48

Also you can measure the time it takes until the report screen is open and ready, which is in our testing around ~800ms.

## Proposed solution (if any)

Onyx internally already has a cache map. When a component connects, and the data is already in the cache map we return the data like this:
```js
function get(key) {
    // When we already have the value in cache - resolve right away
    if (cache.hasCacheForKey(key)) {
        return Promise.resolve(cache.getValue(key));
    }
```
This is good, however, it still runs through the async/callback and thus will cause an extra re-render.
I propose that we, in the constructor of the withOnyx wrapper, access the cache map directly and set the data immediately when we have them. We only set the loading state to true if we don’t have all the data right away.
This way, a withOnyx connected component will most of the times only render initially once, instead of twice.

(Note: once we do that we will notice that there are a few places in the `Onyx.js` code where we don't compare the data before sending it to the component. As a extension to this proposal we should wrap all those places with equal checks to avoid updating components with the same data twice).

## List any benchmarks after implementing the changes to show impacts of the proposed solution (if any)

Profiling again we get 44 commits, and just 0.6s for the initial render of SidebarLinks to finish!

Screenshot 2023-06-28 at 12 36 57

When it comes to opening the report screen we see 4x improvements, as it only takes 100-200ms to show the whole page:

https://github.com/Expensify/react-native-onyx/assets/16821682/8868bd11-3048-4625-a456-313631061ae7

## Platforms:

Which of our officially supported platforms is this issue occurring on?
- [x] Android / native
- [x] Android / Chrome
- [x] iOS / native
- [x] iOS / Safari
- [x] MacOS / Chrome / Safari
- [x] MacOS / Desktop

**Version Number: latest**
**Reproducible in staging?: yes**
**Reproducible in production?: yes**
**Email or phone of affected tester (no customers):**
**Logs:** https://stackoverflow.com/c/expensify/questions/4856
**Notes/Photos/Videos:** Any additional supporting documentation
**Expensify/Expensify Issue URL:**
**Issue reported by:**
**Slack conversation: https://expensify.slack.com/archives/C01GTK53T8Q/p1688580518417739**

hannojg:
### Details

This PR introduces support for `withOnyx` to load data from cache immediately and synchronously. This greatly increases the chance when rending a `withOnyx` wrapped component that it can render right away, instead of being blocked by a loading state at first.

Overview of the changes made: 

- **`tryGetCachedValue`**: New method in `Onyx.js` that is used by `withOnyx` to get values from cache
- **`withOnyx.js`**: In the constructor of the wrapper class try to load all data from cache. Only activate loading state, if not all data was available from cache.
- **`Onyx.js`**: 
  - Cleaned up the code a bit by combining certain statements
  - Added a few more equality checks before calling `setState`

### Related Issues

https://github.com/Expensify/react-native-onyx/issues/269

### Automated Tests

The tests have been modified to reflect the changes made. In particular a lot of the `waitForPromiseToResolve` calls were removed. They aren't needed anymore, as a component can render immediately, once we have the data in onyx, as we don't need to wait for the next tick.

### Linked PRs

/

### DIFF
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -154,18 +154,6 @@ function isCollectionMemberKey(collectionKey, key) {
     return Str.startsWith(key, collectionKey) && key.length > collectionKey.length;
 }
 
-function isCollectionKeyAndNoMember(key) {
-    const keys = _.keys(onyxKeys.COLLECTION);
-    // eslint-disable-next-line no-restricted-syntax
-    for (const collectionKeyCandidate in keys) {
-        if (Str.contains(key, collectionKeyCandidate)) {
-            return !isCollectionMemberKey(collectionKeyCandidate, key);
-        }
-    }
-
-    return false;
-}
-
 /**
  * Checks to see if a provided key is the exact configured key of our connected subscriber
  * or if the provided key is a collection member key (in case our configured key is a "collection key")
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -154,6 +154,18 @@ function isCollectionMemberKey(collectionKey, key) {
     return Str.startsWith(key, collectionKey) && key.length > collectionKey.length;
 }
 
+function isCollectionKeyAndNoMember(key) {
+    const keys = _.keys(onyxKeys.COLLECTION);
+    // eslint-disable-next-line no-restricted-syntax
+    for (const collectionKeyCandidate in keys) {
+        if (Str.contains(key, collectionKeyCandidate)) {
+            return !isCollectionMemberKey(collectionKeyCandidate, key);
+        }
+    }
+
+    return false;
+}
+
 /**
  * Checks to see if a provided key is the exact configured key of our connected subscriber
  * or if the provided key is a collection member key (in case our configured key is a "collection key")
@@ -193,25 +205,25 @@ function isSafeEvictionKey(testKey) {
  * @returns {string}
  */
 function tryGetCachedValue(key, mapping = {}) {
+    let val = cache.hasCacheForKey(key) ? cache.getValue(key) : defaultKeyStates[key];
+
     if (isCollectionKey(key)) {
         const allKeys = cache.getAllKeys();
         const matchingKeys = _.filter(allKeys, k => k.startsWith(key));
         const values = _.reduce(matchingKeys, (finalObject, matchedKey) => {
-            const val = cache.getValue(matchedKey);
-            if (val) {
+            const cachedValue = cache.getValue(matchedKey);
+            if (cachedValue) {
                 // This is permissible because we're in the process of constructing the final object in a reduce function.
                 // eslint-disable-next-line no-param-reassign
-                finalObject[matchedKey] = val;
+                finalObject[matchedKey] = cachedValue;
             }
             return finalObject;
         }, {});
         if (_.keys(values).length === 0) {
             return undefined;
         }
-        return values;
+        val = values;
     }
-
-    const val = cache.hasCacheForKey(key) ? cache.getValue(key) : defaultKeyStates[key];
 
     if (mapping.selector) {
         const state = mapping.withOnyxInstance ? mapping.withOnyxInstance.state : undefined;
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -410,13 +410,12 @@ function keysChanged(collectionKey, partialCollection) {
                         const previousData = prevState[subscriber.statePropertyName];
                         const newData = reduceCollectionWithSelector(cachedCollection, subscriber.selector, subscriber.withOnyxInstance.state);
 
-                        if (deepEqual(previousData, newData)) {
-                            return null;
+                        if (!deepEqual(previousData, newData)) {
+                            return {
+                                [subscriber.statePropertyName]: newData,
+                            };
                         }
-
-                        return {
-                            [subscriber.statePropertyName]: newData,
-                        };
+                        return null;
                     });
                     continue;
                 }
@@ -454,18 +453,32 @@ function keysChanged(collectionKey, partialCollection) {
                 // If the subscriber has a selector, then the component's state must only be updated with the data
                 // returned by the selector and the state should only change when the subset of data changes from what
                 // it was previously.
-                subscriber.withOnyxInstance.setState((prevState) => {
-                    const prevData = prevState[subscriber.statePropertyName];
-                    const data = cachedCollection[subscriber.key];
-                    const newData = subscriber.selector ? getSubsetOfData(data, subscriber.selector, subscriber.withOnyxInstance.state) : data;
+                if (subscriber.selector) {
+                    subscriber.withOnyxInstance.setState((prevState) => {
+                        const prevData = prevState[subscriber.statePropertyName];
+                        const newData = getSubsetOfData(cachedCollection[subscriber.key], subscriber.selector, subscriber.withOnyxInstance.state);
+                        if (!deepEqual(prevData, newData)) {
+                            PerformanceUtils.logSetStateCall(subscriber, prevData, newData, 'keysChanged', collectionKey);
+                            return {
+                                [subscriber.statePropertyName]: newData,
+                            };
+                        }
 
-                    if (deepEqual(prevData, newData)) {
+                        return null;
+                    });
+                    continue;
+                }
+
+                subscriber.withOnyxInstance.setState((prevState) => {
+                    const data = cachedCollection[subscriber.key];
+                    const previousData = prevState[subscriber.statePropertyName];
+                    if (deepEqual(data, previousData)) {
                         return null;
                     }
 
-                    PerformanceUtils.logSetStateCall(subscriber, prevData, newData, 'keysChanged', collectionKey);
+                    PerformanceUtils.logSetStateCall(subscriber, previousData, data, 'keysChanged', collectionKey);
                     return {
-                        [subscriber.statePropertyName]: newData,
+                        [subscriber.statePropertyName]: data,
                     };
                 });
             }
@@ -521,40 +534,70 @@ function keyChanged(key, data, canUpdateSubscriber) {
             if (isCollectionKey(subscriber.key)) {
                 // If the subscriber has a selector, then the consumer of this data must only be given the data
                 // returned by the selector and only when the selected data has changed.
-                subscriber.withOnyxInstance.setState((prevState) => {
-                    const prevData = prevState[subscriber.statePropertyName] || {};
-                    const newData = {
-                        [key]: subscriber.selector ? getSubsetOfData(data, subscriber.selector, subscriber.withOnyxInstance.state) : data,
-                    };
-                    const newCollection = {
-                        ...prevData,
-                        ...newData,
-                    };
-
-                    if (deepEqual(prevData, newCollection)) {
+                if (subscriber.selector) {
+                    subscriber.withOnyxInstance.setState((prevState) => {
+                        const prevData = prevState[subscriber.statePropertyName];
+                        const newData = {
+                            [key]: getSubsetOfData(data, subscriber.selector, subscriber.withOnyxInstance.state),
+                        };
+                        const prevDataWithNewData = {
+                            ...prevData,
+                            ...newData,
+                        };
+                        if (!deepEqual(prevData, prevDataWithNewData)) {
+                            PerformanceUtils.logSetStateCall(subscriber, prevData, newData, 'keyChanged', key);
+                            return {
+                                [subscriber.statePropertyName]: prevDataWithNewData,
+                            };
+                        }
                         return null;
-                    }
+                    });
+                    continue;
+                }
 
-                    PerformanceUtils.logSetStateCall(subscriber, prevData, newCollection, 'keyChanged', key);
-                    return {
-                        [subscriber.statePropertyName]: newCollection,
+                subscriber.withOnyxInstance.setState((prevState) => {
+                    const collection = prevState[subscriber.statePropertyName] || {};
+                    const newCollection = {
+                        ...collection,
+                        [key]: data,
                     };
+                    if (!deepEqual(collection, newCollection)) {
+                        PerformanceUtils.logSetStateCall(subscriber, collection, newCollection, 'keyChanged', key);
+                        return {
+                            [subscriber.statePropertyName]: newCollection,
+                        };
+                    }
+                    return null;
                 });
                 continue;
             }
 
             // If the subscriber has a selector, then the component's state must only be updated with the data
             // returned by the selector and only if the selected data has changed.
+            if (subscriber.selector) {
+                subscriber.withOnyxInstance.setState((prevState) => {
+                    const previousValue = getSubsetOfData(prevState[subscriber.statePropertyName], subscriber.selector, subscriber.withOnyxInstance.state);
+                    const newValue = getSubsetOfData(data, subscriber.selector, subscriber.withOnyxInstance.state);
+                    if (!deepEqual(previousValue, newValue)) {
+                        return {
+                            [subscriber.statePropertyName]: newValue,
+                        };
+                    }
+                    return null;
+                });
+                continue;
+            }
+
+            // If we did not match on a collection key then we just set the new data to the state property
             subscriber.withOnyxInstance.setState((prevState) => {
-                const previousValue = prevState[subscriber.statePropertyName];
-                const newValue = subscriber.selector ? getSubsetOfData(data, subscriber.selector, subscriber.withOnyxInstance.state) : data;
-                if (deepEqual(previousValue, newValue)) {
+                const previousData = prevState[subscriber.statePropertyName];
+                if (previousData === data) {
                     return null;
                 }
 
-                PerformanceUtils.logSetStateCall(subscriber, previousValue, newValue, 'keyChanged', key);
+                PerformanceUtils.logSetStateCall(subscriber, previousData, data, 'keyChanged', key);
                 return {
-                    [subscriber.statePropertyName]: newValue,
+                    [subscriber.statePropertyName]: data,
                 };
             });
             continue;
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -193,7 +193,7 @@ function isSafeEvictionKey(testKey) {
  * @returns {Mixed}
  */
 function tryGetCachedValue(key, mapping = {}) {
-    let val = cache.hasCacheForKey(key) ? cache.getValue(key) : defaultKeyStates[key];
+    let val = cache.getValue(key);
 
     if (isCollectionKey(key)) {
         const allKeys = cache.getAllKeys();
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -561,13 +561,10 @@ function keyChanged(key, data, canUpdateSubscriber) {
                         ...collection,
                         [key]: data,
                     };
-                    if (!deepEqual(collection, newCollection)) {
-                        PerformanceUtils.logSetStateCall(subscriber, collection, newCollection, 'keyChanged', key);
-                        return {
-                            [subscriber.statePropertyName]: newCollection,
-                        };
-                    }
-                    return null;
+                    PerformanceUtils.logSetStateCall(subscriber, collection, newCollection, 'keyChanged', key);
+                    return {
+                        [subscriber.statePropertyName]: newCollection,
+                    };
                 });
                 continue;
             }
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -213,8 +213,6 @@ function tryGetCachedValue(key, mapping = {}) {
 
     const val = cache.hasCacheForKey(key) ? cache.getValue(key) : defaultKeyStates[key];
 
-    // If the mapping has a selector, then the component's state must only be updated with the data
-    // returned by the selector.
     if (mapping.selector) {
         const state = mapping.withOnyxInstance ? mapping.withOnyxInstance.state : undefined;
         if (isCollectionKey(key)) {
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -184,12 +184,24 @@ function isSafeEvictionKey(testKey) {
 /**
  * Tries to return all keys
  * from the current cache.
+ * (internal)
  * @returns {Array<string>}
  */
 function getAllKeysSync() {
     return cache.getAllKeys();
 }
 
+/**
+ * Tries to get a value from the cache.
+ * If the value is not present in cache it will
+ * return the default value or undefined.
+ * If the requested key is a collection, it will
+ * return an object with all the collection members.
+ *
+ * @param {string} key
+ * @param {Object} mapping
+ * @returns {string}
+ */
 function tryGetCachedValue(key, mapping = {}) {
     if (isCollectionKey(key)) {
         const allKeys = getAllKeysSync();
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -210,11 +210,7 @@ function tryGetCachedValue(key, mapping = {}) {
         }
     }
 
-    if (!cache.hasCacheForKey(key)) {
-        return defaultKeyStates[key]; // We might have the value in the default key state
-    }
-
-    const val = cache.getValue(key);
+    const val = cache.hasCacheForKey(key) ? cache.getValue(key) : defaultKeyStates[key];
 
     // If the mapping has a selector, then the component's state must only be updated with the data
     // returned by the selector.
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -185,15 +185,12 @@ function isSafeEvictionKey(testKey) {
 }
 
 /**
- * Tries to get a value from the cache.
- * If the value is not present in cache it will
- * return the default value or undefined.
- * If the requested key is a collection, it will
- * return an object with all the collection members.
+ * Tries to get a value from the cache. If the value is not present in cache it will return the default value or undefined.
+ * If the requested key is a collection, it will return an object with all the collection members.
  *
- * @param {string} key
+ * @param {String} key
  * @param {Object} mapping
- * @returns {string}
+ * @returns {Mixed}
  */
 function tryGetCachedValue(key, mapping = {}) {
     let val = cache.hasCacheForKey(key) ? cache.getValue(key) : defaultKeyStates[key];
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -210,7 +210,7 @@ function tryGetCachedValue(key, mapping = {}) {
             }
             return finalObject;
         }, {});
-        if (_.keys(values).length === 0) {
+        if (_.size(values) === 0) {
             return;
         }
         val = values;
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -643,7 +643,7 @@ function sendDataToConnection(mapping, val, matchedKey) {
             }
         }
 
-        // sendDataToConnection only get's called in the connect code.
+        // sendDataToConnection only gets called in the connect code.
         // withOnyx is already setting data from cache initially. If we now
         // want to call sendDataToConnection, we want to make sure, we
         // are not sending again the same data from cache, if the component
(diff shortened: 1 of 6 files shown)


kidroca (changes requested):
Thank you for your efforts on this pull request. Overall, the changes look promising and I appreciate the thoughtfulness behind them. 

I've shared some specific comments for your consideration to improve the overall quality and maintainability of the code. 

One particular observation is that there seem to be some changes in this PR that aren't directly tied to the referenced issue. While these changes, which appear to be performance optimizations and refactoring, could indeed enhance our codebase, they also introduce extra complexity to the review process.

To streamline the review, I suggest we separate these optimizations and refactoring changes into a subsequent PR. This approach would allow us to maintain focus on addressing the primary objective of the current PR, and may even speed up its approval. 

Looking forward to your thoughts. Thanks again for your contributions!

lib/Onyx.js
```diff
@@ -181,6 +181,64 @@ function isSafeEvictionKey(testKey) {
...
+function tryGetCachedValue(key, mapping = {}) {
+    if (isCollectionKey(key)) {
+        const allKeys = getAllKeysSync();
+        const matchingKeys = _.filter(allKeys, k => k.startsWith(key));
+        const values = _.reduce(matchingKeys, (finalObject, matchedKey) => {
+            const val = cache.getValue(matchedKey);
+            if (val) {
+                // We know a 100% where the object is coming from, so its fine
+                // eslint-disable-next-line no-param-reassign
+                finalObject[matchedKey] = val;
```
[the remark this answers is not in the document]
Julesssss: +1 to the above comment change, thanks @kidroca!

lib/Onyx.js
```diff
@@ -181,6 +181,64 @@ function isSafeEvictionKey(testKey) {
...
+    }
+
+    if (!cache.hasCacheForKey(key)) {
+        return defaultKeyStates[key]; // We might have the value in the default key state
+    }
+
+    const val = cache.getValue(key);
+
+    // If the mapping has a selector, then the component's state must only be updated with the data
+    // returned by the selector.
```
Julesssss, line 230: What were we doing previously? Returning additional data from outside the selector?

Julesssss (commented):
Great work, would love to see a bit more thorough comments for people like me who don't fully understand the logic.

lib/Onyx.js
```diff
@@ -181,6 +181,64 @@ function isSafeEvictionKey(testKey) {
...
+            return finalObject;
+        }, {});
+        if (_.keys(values).length > 0) {
+            return values;
+        }
+    }
+
+    if (!cache.hasCacheForKey(key)) {
+        return defaultKeyStates[key]; // We might have the value in the default key state
+    }
```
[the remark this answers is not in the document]
hannojg: very good catch!

lib/Onyx.js
```diff
@@ -181,6 +181,64 @@ function isSafeEvictionKey(testKey) {
...
+            if (val) {
+                // We know a 100% where the object is coming from, so its fine
+                // eslint-disable-next-line no-param-reassign
+                finalObject[matchedKey] = val;
+            }
+            return finalObject;
+        }, {});
+        if (_.keys(values).length > 0) {
+            return values;
+        }
```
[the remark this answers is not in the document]
hannojg: Hey, good point, I updated the logic. When we didn't find any values for the collection key we shouldn't continue with the functions behaviour. So I added an early return statement for that.

lib/Onyx.js
hannojg: I removed the comment as it seems to be ambiguous. As this logic is new there is no previous behaviour, and the comment was just trying to say that we need to use the selector which is redundant

hannojg:
Good for review again. I was also experimenting with removing the `if(loading) return null` logic, but that needs a bit more thorough testing.

hannojg:
Sorry, fixed!

lib/Onyx.js
```diff
@@ -181,6 +181,49 @@ function isSafeEvictionKey(testKey) {
...
+            const cachedValue = cache.getValue(matchedKey);
+            if (cachedValue) {
+                // This is permissible because we're in the process of constructing the final object in a reduce function.
+                // eslint-disable-next-line no-param-reassign
+                finalObject[matchedKey] = cachedValue;
+            }
+            return finalObject;
+        }, {});
+        if (_.keys(values).length === 0) {
+            return undefined;
```
marcaaron, line 211:
```suggestion
            return;
```

lib/withOnyx.js
```diff
@@ -37,12 +37,27 @@ export default function (mapOnyxToState) {
...
+
                 // Object holding the temporary initial state for the component while we load the various Onyx keys
-                this.tempState = {};
+                this.tempState = cachedState;
 
                 this.state = {
-                    // If there are no required keys for init then we can render the wrapped component immediately
-                    loading: requiredKeysForInit.length > 0,
+                    loading,
+                    ...(loading ? {} : cachedState),
```
marcaaron, line 60: nab, can avoid the spread with something like:

```js
cachedState.loading = loading;
this.state = cachedState;
```

lib/withOnyx.js
```diff
@@ -88,7 +103,7 @@ export default function (mapOnyxToState) {
              * @param {*} val
              */
             setWithOnyxState(statePropertyName, val) {
-                if (!this.state.loading) {
+                if (!this.state.loading && this.state[statePropertyName] !== val) {
```
marcaaron, line 106: Why this part?

```js
this.state[statePropertyName] !== val
```

Like, say we were able to set up the component in the constructor with all cache values. Loading is `false`. But if the value has changed we will start messing with `tempState` again? My expectation is that once we have all the values then we should be done with the other part of this method.

lib/withOnyx.js
```diff
@@ -37,12 +37,27 @@ export default function (mapOnyxToState) {
...
+                    const key = Str.result(mapping.key, props);
+                    const value = Onyx.tryGetCachedValue(key, mapping);
+
+                    if (value !== undefined) {
+                        cachedState[propertyName] = value;
+                    }
+                });
+
+                // If we have all the data we need, then we can render the component immediately
+                const loading = _.keys(cachedState).length < requiredKeysForInit.length;
```
marcaaron, line 53: I think we only need to set the `cachedState` if we are still loading stuff.

lib/Onyx.js
```diff
@@ -590,6 +643,15 @@ function sendDataToConnection(mapping, val, matchedKey) {
             }
         }
 
+        // sendDataToConnection only get's called in the connect code.
```
marcaaron, line 646:
```suggestion
        // sendDataToConnection only gets called in the connect code.
```

lib/Onyx.js
```diff
@@ -561,6 +612,8 @@ function keyChanged(key, data, canUpdateSubscriber) {
  *     - sets state on the withOnyxInstances
  *     - triggers the callback function
  *
+ * Note: This is only invoked from Onyx.connect
```
marcaaron, line 615: while true, I'd maybe remove this as comments like this can become of date and confuse people.

lib/Onyx.js
```diff
@@ -590,6 +643,15 @@ function sendDataToConnection(mapping, val, matchedKey) {
...
         }
 
+        // sendDataToConnection only get's called in the connect code.
+        // withOnyx is already setting data from cache initially. If we now
+        // want to call sendDataToConnection, we want to make sure, we
+        // are not sending again the same data from cache, if the component
+        // already has the cached data.
+        if (deepEqual(mapping.withOnyxInstance.state[mapping.statePropertyName], newData)) {
+            return;
+        }
```
marcaaron, line 653: Do we need a `deepEqual()` here? Wouldn't it be enough to see that it has the property name at all? It must have been set in the constructor? Or maybe I am missing some situation where these could actually be different.

lib/Onyx.js
```diff
@@ -513,10 +561,13 @@ function keyChanged(key, data, canUpdateSubscriber) {
...
-                    return {
-                        [subscriber.statePropertyName]: newCollection,
-                    };
+                    if (!deepEqual(collection, newCollection)) {
+                        PerformanceUtils.logSetStateCall(subscriber, collection, newCollection, 'keyChanged', key);
+                        return {
+                            [subscriber.statePropertyName]: newCollection,
+                        };
+                    }
+                    return null;
```
marcaaron, line 570: Is this change related to the "init with cache data" fix?

lib/Onyx.js
```diff
@@ -424,7 +472,7 @@ function keysChanged(collectionKey, partialCollection) {
                 subscriber.withOnyxInstance.setState((prevState) => {
                     const data = cachedCollection[subscriber.key];
                     const previousData = prevState[subscriber.statePropertyName];
-                    if (data === previousData) {
+                    if (deepEqual(data, previousData)) {
```
marcaaron, line 475: Same here. Is this related or just something that seems more correct?

marcaaron (changes requested):

lib/Onyx.js
```diff
@@ -378,13 +421,18 @@ function keysChanged(collectionKey, partialCollection) {
...
-                    const finalCollection = _.clone(prevState[subscriber.statePropertyName] || {});
+                    const prevCollection = prevState[subscriber.statePropertyName];
+                    const finalCollection = _.clone(prevCollection || {});
                     const dataKeys = _.keys(partialCollection);
                     for (let j = 0; j < dataKeys.length; j++) {
                         const dataKey = dataKeys[j];
                         finalCollection[dataKey] = cachedCollection[dataKey];
                     }
 
+                    if (deepEqual(prevCollection, finalCollection)) {
```
arosiclair, line 432: Are there any performance implications on adding another deep equality check here?

marcaaron: Good point and - there might be. Guides e.g. can have 10,000 reports that would exist in a "collection"

lib/withOnyx.js
hannojg: you are right, i double checked it, and this part isn't necessary anymore!

lib/withOnyx.js
hannojg: Hm, not sure i understand you correctly. With the current changes we 

- always set all the cached state we could load for the keys to the component state 
- if not all keys have been retrievable from cache the loading value will be true

lib/Onyx.js
hannojg: Hm, you are right in a way. I think deepEqual is overkill, and a simple ref comparison should be enough. However we should do a check here. There is the situation where a prop provided to the component using withOnyx changes, which changes the key requested. In this case we connect the component again, and this code gets called again. However, at this point in time there will already be a state value. Thats why we can't return simply when the value seems to be in state>

hannojg: ah lol no, that wouldn't be enough, because we also have selectors, and their return value is always a different ref, even if the input data have been the same. Hmmm ...

hannojg: Okay, so i optimised it to use deepEqual for selectors only. Please check again

lib/Onyx.js
hannojg: Yeah again, you are right, thats just something thats belonging more to a different set of performance improvements. Will remove it from here and do a follow up issue/PR.

lib/Onyx.js
hannojg: removed it for now, lets discuss this in a separate issue / PR!

hannojg:
Okay, lgtm now!

lib/Onyx.js
marcaaron: >prop provided to the component using withOnyx changes, which changes the key requested. In this case we connect the component again, and this code gets called again. However, at this point in time there will already be a state value. Thats why we can't return simply when the value seems to be in state.

Whoaaa! Did we see this happen? 🙃 

Let me make sure I understand the order of operations:

- Connect to a key with a prop (`withOnyx()` `constructor()` called)
- Component mounts and there is no value for something in cache so we try get the stored value (async operation)
- _Maybe_ we changed the prop influencing this key while we are getting the data from storage
- Component updates and we `disconnect()` (but don't cancel the reading from storage that is happening)
- We finish reading the first thing from storage and call `sendDataToConnection()`
- It will return early due to the logic on line 629

**Question:** What happens to the `loading` state in this scenario since we are never calling `setWithOnyxState()` and the value of `requiredKeysForInit` is never updated when the other key is disconnected? Does the component never render?

FWIW I am still not getting why we need the `deepEqual()`. Maybe clearer way to look at this is...

If we have the cached data then why are we calling `sendDataToConnection()` at all? Maybe we pass something to [this method](https://github.com/Expensify/react-native-onyx/blob/d4fc53cc3229808e9c1c542c13ce0089e5893d96/lib/withOnyx.js#L146C13-L157) and just skip it during the `connect()` method.

lib/withOnyx.js
marcaaron: Sorry that was really unclear haha. 

I meant that setting up the `this.tempState` variable is not needed if we got the entire `cachedState`? Because we would never use it for anything.

(not sure why my brain said the other thing)

lib/withOnyx.js
marcaaron: Great, let's remove it 😄

marcaaron: wow it is gone GitHub trolled me 😂

lib/withOnyx.js
```diff
@@ -37,15 +37,28 @@ export default function (mapOnyxToState) {
...
+                this.tempState = cachedState;
 
-                this.state = {
-                    // If there are no required keys for init then we can render the wrapped component immediately
-                    loading: requiredKeysForInit.length > 0,
-                };
+                this.state = cachedState;
             }
 
+            // TODO: I think we don't have to wait until mount to subscribe to Onyx keys. We can do it in the constructor as well.
```
marcaaron, line 61: Random (impossible to know) Expensify rule incoming, but we never put "TODO" in the code anywhere. something something it is better to make an issue so that we actually do whatever the "TODO" asks us to do.

marcaaron: FWIW, I can't remember but this might _possibly_ be how we had it at some point. The data can be returned very fast and the component could takes a while to mount - then we end up setting state on an unmounted component (which throws the ugly warning). Honestly though - can't remember.

lib/withOnyx.js
```diff
@@ -58,9 +71,9 @@ export default function (mapOnyxToState) {
                 // If any of the mappings use data from the props, then when the props change, all the
                 // connections need to be reconnected with the new props
                 _.each(mapOnyxToState, (mapping, propertyName) => {
+                    // Check if we can update from cache already
```
marcaaron, line 74: This comment looks like maybe it should be removed?

lib/withOnyx.js
```diff
@@ -88,6 +101,14 @@ export default function (mapOnyxToState) {
              * @param {*} val
              */
             setWithOnyxState(statePropertyName, val) {
+                // This method gets called by Onyx.connect no matter if the cached
+                // data is already there or not. Thus we have to check whether
```
marcaaron, line 105: NAB, but maybe we should not call this at all in the case the cached data is there. This makes more sense to me vs. letting it get called but "cancelling" it. I also think the code we have here works no NAB.

marcaaron (commented):
LGTM just need to remove the TODO comment and we are 🍒

lib/withOnyx.js
hannojg: Oh, we continue using the tempState later on in `setStateWithOnyx`

arosiclair (approved):
LGTM other than Marc's comments

j-piasecki:
I've tried your changes on the new architecture and they improve the performance greatly ❤️! One problem I've observed is that nesting reports sometimes results in the `ReportScreen` falling into an infinite render loop.

https://github.com/Expensify/react-native-onyx/assets/21055725/adabc7a7-f648-4ce0-b161-0e0d6d4b5944

I doubt it's directly tied to your changes, most probably it simply enabled a problem elsewhere but I wanted you to know just in case.

lib/withOnyx.js
marcaaron: Only if `loading: true` ?

marcaaron:
>I doubt it's directly tied to your changes, most probably it simply enabled a problem elsewhere but I wanted you to know just in case.

Oh, nice catch! Does it happen in the previous version? Sounds like something we should try to address when merging the update PR into `App`.

hannojg:
Hey 👋 I think I'd like to test for that infinite render behaviour before we merge this, so we know that once we merge this PR it will work in the app.

Will try to reproduce and fix it today!

j-piasecki:
> Will try to reproduce and fix it today!

Thanks! Here's a branch with the new arch setup: https://github.com/WoLewicki/App/tree/andrew-fabric. Let me know if you find anything, if not then I can look into it more tomorrow.

lib/withOnyx.js
hannojg: yeah, I think this change would be a separate PR as well where we can look at and discuss all those cases (however, I might propose a bit of a architectural change to onyx in the following days which would handle this point as well)

lib/withOnyx.js
hannojg: the condition is paired with the loading state, which is something the `withOnyx` is aware about, while `Onyx.js` isn't really. Hence I think it makes sense to leave the condition in this place 🤔

hannojg:
Thank you! Checking now (would be great to get this PR merged today!)

lib/withOnyx.js
hannojg: yeah, as `loading: true` is like "we didn't have all the data in cache and need to wait for storage or maybe even wait until a network request puts the data into onyx".

j-piasecki:
@hannojg The render loop seems to be some weird interaction with `react-freeze` - setting `freeze` to `false` [here](https://github.com/Expensify/App/blob/88fb2ada376ce103224e30a028c48350db29f78d/src/libs/Navigation/FreezeWrapper.js#L46) seems to stop the render loop. Adding `initWithStoredValues: false` [here](https://github.com/Expensify/App/blob/main/src/pages/home/ReportScreen.js#L449-L451) also looks like it’s helping.

hannojg:
Hey Thanks for the insights @j-piasecki! Will double check what's wrong with the usage of these!

hannojg:
@j-piasecki and me just looked into the issue and agreed that the issue with the fabric architecture enabled is not a problem with the changes from this PR, but rather a problem with react-freeze. That means the issue isn't blocking this PR.

Green light! :shipit:

Julesssss (dismissed):
Fantastic work. 

I would just suggest that we introduce a test specifically to protect us from a future regression that nulls the perf improvement. Specifically, it would be great if we can ensure [this condition](https://github.com/Expensify/react-native-onyx/pull/280/files#diff-364fd7349a1af7b52091427c5a0734da8ff56057fbc7fe2be10734a06c97264cR52-R53) is always met.

hannojg:
@Julesssss I already added a test earlier that checks if the component is rendered immediately with the data from cache. I extended this test to make sure the component doesn't re-render with the same data (e.g. after the connection was created and `Onyx.js` tries to send the data with `sendDataToConnection`)

Julesssss (dismissed):

tests/unit/withOnyxTest.js
```diff
@@ -276,16 +272,12 @@ describe('withOnyx', () => {
...
-            })
-            .then(() => {
-                // When a single item in the collection is updated with mergeCollect()
-                Onyx.mergeCollection(ONYX_KEYS.COLLECTION.TEST_KEY, {
-                    test_1: {ID: 1, newProperty: 'yay'},
-                });
-                return waitForPromisesToResolve();
             })
+
+            // When a single item in the collection is updated with mergeCollect()
```
mountiny, line 277: I know this was there but
```suggestion
            // When a single item in the collection is updated with mergeCollection()
```

mountiny (approved):
Thanks! I think would be great to get approvals of @marcaaron and @tgolen before merging

lib/Onyx.js
```diff
@@ -184,6 +184,49 @@ function isSafeEvictionKey(testKey) {
...
+        const values = _.reduce(matchingKeys, (finalObject, matchedKey) => {
+            const cachedValue = cache.getValue(matchedKey);
+            if (cachedValue) {
+                // This is permissible because we're in the process of constructing the final object in a reduce function.
+                // eslint-disable-next-line no-param-reassign
+                finalObject[matchedKey] = cachedValue;
+            }
+            return finalObject;
+        }, {});
+        if (_.keys(values).length === 0) {
```
tgolen, line 213: NAB 
```suggestion
        if (_.size(values) === 0) {
```

lib/withOnyx.js
```diff
@@ -37,13 +37,25 @@ export default function (mapOnyxToState) {
...
+
+                // Check if we can get all the data synchronously from cache
+                _.each(mapOnyxToState, (mapping, propertyName) => {
+                    const key = Str.result(mapping.key, props);
+                    const value = Onyx.tryGetCachedValue(key, mapping);
+
+                    if (value !== undefined) {
+                        cachedState[propertyName] = value;
+                    }
+                });
```
tgolen, line 50: NAB: This could be a `_.reduce()` instead.

lib/withOnyx.js
```diff
@@ -37,13 +37,25 @@ export default function (mapOnyxToState) {
...
+
+                // If we have all the data we need, then we can render the component immediately
+                cachedState.loading = _.keys(cachedState).length < requiredKeysForInit.length;
```
tgolen, line 53:
```suggestion
                cachedState.loading = _.size(cachedState) < requiredKeysForInit.length;
```

lib/withOnyx.js
```diff
@@ -88,6 +99,14 @@ export default function (mapOnyxToState) {
              * @param {*} val
              */
             setWithOnyxState(statePropertyName, val) {
+                // This method gets called by Onyx.connect no matter if the cached
+                // data is already there or not. Thus we have to check whether
+                // the data has changed or not before we set the state.
```
tgolen, line 104: Could you please update this comment to answer the question "Why do we have to check whether the data has changed or not before setting the state?" Right now, the comment only explains what is happening and not _why_ it is happening, and I'm not entirely sure about the reason.

lib/withOnyx.js
```diff
@@ -100,7 +119,17 @@ export default function (mapOnyxToState) {
...
 
-                this.setState({...this.tempState, loading: false});
+                const stateUpdate = {...this.tempState, loading: false};
+
+                // Why are we setting the state here manually, instead of just using setState?
+                // Explanation: setState is a async operation, meaning it might execute on the next task,
+                // or to a later point in the microtask queue. That means, that there is a chance that
+                // setWithOnyxState is called again before the state is actually set. This causes issues
+                // with the above checks for the loading state.
+                // Note: This behaviour has been mainly observed on fabric.
```
tgolen, line 129:
```suggestion
                // The state is set here manually, instead of using setState because setState is an async operation, meaning it might execute on the next task,
                // or at a later point in the microtask queue. That can lead to a race condition where
                // setWithOnyxState is called before the state is actually set. This results in unreliable behavior when checking the loading state and has been mainly observed on fabric.
```

tgolen (changes requested):
Just a few small things I noticed that can be buttoned up a little.

lib/withOnyx.js
hannojg: I updated it, I hope its more obvious now?

tgolen (dismissed):

luacmartins (commented):
Looking good. Left a couple of small comments

lib/Onyx.js
```diff
@@ -184,6 +184,46 @@ function isSafeEvictionKey(testKey) {
...
+/**
+ * Tries to get a value from the cache. If the value is not present in cache it will return the default value or undefined.
+ * If the requested key is a collection, it will return an object with all the collection members.
+ *
+ * @param {String} key
+ * @param {Object} mapping
+ * @returns {Mixed}
+ */
+function tryGetCachedValue(key, mapping = {}) {
+    let val = cache.hasCacheForKey(key) ? cache.getValue(key) : defaultKeyStates[key];
```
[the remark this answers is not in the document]
marcaaron: Thinking more on it. If I delete a key by setting it to `null` - it should not go back to the default key state (unless I call `clear()` or `init()` again). But maybe that behavior is disputed.

marcaaron (changes requested):
Looks good, but I have some concern about using the `defaultKeyStates` as a cache value fallback. I think this behavior is wrong.

lib/Onyx.js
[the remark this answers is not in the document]
hannojg: I think you have a point here. The defaultKeyStates should be set to the cache on init. Thus it's introducing a wrong behaviour when falling back to the default here. Will change that, thank you!

lib/withOnyx.js
```diff
@@ -37,13 +37,25 @@ export default function (mapOnyxToState) {
...
+
+                    return resultObj;
+                }, {});
+
+                // If we have all the data we need, then we can render the component immediately
+                cachedState.loading = _.size(cachedState) < requiredKeysForInit.length;
+
                 // Object holding the temporary initial state for the component while we load the various Onyx keys
-                this.tempState = {};
+                this.tempState = cachedState;
```
mountiny, line 56: I am not sure why do we set the tempState to the same cachedState as the normal state. I might be missing something but could you help me understand this? Should we update the comment above?

mountiny (approved):
Thanks for pushing this forwards!

luacmartins (approved):

Julesssss (approved):

tgolen (approved):

mountiny:
Per the slack discussion we got enough approvals here and we want to get this change out to test how it improves performance so I am going to merge this https://expensify.slack.com/archives/C035J5C9FAP/p1690473404613159?thread_ts=1690290738.145189&cid=C035J5C9FAP

lib/withOnyx.js
marcaaron: The intention here is that `tempState` is still used in the same way as it was before. 

If we did not get all the values from the cache then the values are held in the `tempState` until all are "loaded". This way we avoid calling `setState()` each time a value comes back from storage. That was the original optimization added here and is a _good thing_ because otherwise you may have like several calls to setState() happening in a kind of burst before the first render (at least that's what we saw and why we changed how this works).

Though the question did make me wonder about something... if `this.tempState` and `this.state` are initialized with the same object reference then will we be mutating the state directly on this line:

https://github.com/Expensify/react-native-onyx/blob/80c6f001e27e798ed56cc4f3a39ee27c5efae375/lib/withOnyx.js#L117

Seems like that would be undesirable in the case where you did not initialize all the values in the constructor.

marcaaron:
haven't tested the theory, but this might be a problem -> https://github.com/Expensify/react-native-onyx/pull/280#discussion_r1276988396

lib/withOnyx.js
mountiny: Thanks for the write up. @hannojg what do you think of the concern Marc raised above?

PR #283, opened 2023-07-24
fix: state management flow to avoid data loss

pasyukevich:
### Related Issues

https://github.com/Expensify/App/issues/22122
https://github.com/Expensify/App/issues/21881

### DIFF
--- a/lib/withOnyx.js
+++ b/lib/withOnyx.js
@@ -121,14 +121,15 @@ export default function (mapOnyxToState) {
                     return;
                 }
 
-                const stateUpdate = {...this.tempState, loading: false};
+                // Leave untouched previous state to avoid data loss during pre-load updates.
+                // This handles case when setState was called before the setWithOnyxState.
+                // For example, when an Onyx property was updated by keyChanged before the call of the setWithOnyxState.
+                this.setState((prevState) => {
+                    const remainingTempState = _.omit(this.tempState, _.keys(prevState));
 
-                // The state is set here manually, instead of using setState because setState is an async operation, meaning it might execute on the next tick,
-                // or at a later point in the microtask queue. That can lead to a race condition where
-                // setWithOnyxState is called before the state is actually set. This results in unreliable behavior when checking the loading state and has been mainly observed on fabric.
-                this.state = stateUpdate;
+                    return ({...remainingTempState, loading: false});
+                });
 
-                this.setState(stateUpdate); // Trigger a render
                 delete this.tempState;
             }
 


pasyukevich:

Julesssss (commented):
@pasyukevich can you add some manual test cases to the PR description, please? So reviewers know how to verify the fix.

Julesssss (commented):
Hmm, this doesn't seem to resolve the Android splash bug for me. To test I manually changed the code in the App node modules file. Am I missing something?: 

Screenshot_20230726_094039

pasyukevich:
@Julesssss could you provide details on how you managed to reproduce the issue?

I will take a look

Julesssss:
Yeah, so internally we use Ngrok for dev testing. A lot of us have noticed that NewDot is bricked on Android emulators with the chrome app: https://github.com/Expensify/App/issues/22122

But perhaps that is a separate issue after all... oh well. 

We still need some tests that we can use to verify this fix though.

pasyukevich:
@Julesssss I'm in the process of preparing the instruction to verify the fix. However, it has become more challenging to do so with my previous flow, after updating to the latest version of the app. Now, I'm observing that in most instances, 'setState' is not invoked before 'setWithOnyxState' when using the same input.

When it comes to testing it locally, I've simply modified 'webOnyx.js' and 'web.development.js' by incorporating this fix for both the mobile and web platforms.

pasyukevich:
Hi @Julesssss 

I'm not able to reproduce the issue on the latest app version, so I prepared steps on how to reproduce it on the old app but with the latest Onyx library

Step-by-step on how to reproduce the broken flow for Onyx 

1. Log into the app locally using the main branch.
2. Shut down the current environment.
3. Fetch this branch from the forked repo - https://github.com/pasyukevich/Expensify-app/tree/test-report-expanded (This is the older app version with the port updated to 8082, Onyx version updated, and the Test component prepared).
4. Install the necessary modules and run the app.
5. Open any report, type a few lines of text, and expand the composer.
6. Retrieve the report ID from the URL.
7. Navigate to Test.js in the `'src/libs/Navigation/AppNavigator'`.
8. Replace `'PLACE_HERE_ID_OF_REPORT'` with the actual report ID.
9. Open CentralPaneNavigator.js
10. Replace ReportScreenWrapper with the Test component (use ['import Test from '../Test'](url)).
11. Refresh the page.
12. Observe that `Expensify` logs as true, while `Test` logs as false (in both instances, this.props.isSidebarLoaded is logged).
13.  Apply the Onyx changes in `web.development.js located` in the `node_modules`. (to apply the same changes for mobile - `withOnyx.js`)
14. Refresh the page again and note that 'Test' is now true.

pasyukevich:
@aimane-chnaif Sure 

This change should fix the flow when the Onyx variable is not updated to the latest

  1. SetState is called for the Onyx instance and adding an updated flag to the state
  2. setWithOnyxState is  called and tempState is set, the previous value erased
  3. As a result we have the initial value in the current state

trjExpensify:
This PR has stalled for a couple of weeks. Can we keep it moving @aimane-chnaif?

pasyukevich:
@aimane-chnaif PR is updated

pasyukevich:
I see that the update was applied to the Onyx library that was aimed to fix the same problem, but during my test, it did not work well in the described flow 

With [this](https://github.com/pasyukevich/Expensify-app/tree/test-report-expanded ) branch after cleaning node modules and reinstalling them I was able to reproduce the composer issue. - https://github.com/Expensify/App/issues/21881

I found that the approach with function in the setState fixes this issue,  but the manual assign is not (as it is now done in the wIthOnyx)

pasyukevich:
@aimane-chnaif this flow is not reproducible on the current app version due to the page commits

best choice is to use [this](https://github.com/pasyukevich/Expensify-app/tree/test-report-expanded) branch of app with updated Onyx library  version 

I can prepare an updated branch and flow with newer Onyx and recheck

pasyukevich:
I created a new test branch with updated Onyx [test-report-expanded-latest-onyx](https://github.com/pasyukevich/Expensify-app/tree/test-report-expanded-latest-onyx)

Instruction updated - https://github.com/Expensify/react-native-onyx/pull/283#issuecomment-1653441265 

This behavior is still reproducible

mountiny:
@hannojg @Szymon20000 Could you help review this PR and issue its attempting to solve, your experience with Onyx might be handy, thanks!

lib/withOnyx.js
```diff
@@ -121,14 +121,9 @@ export default function (mapOnyxToState) {
...
-
-                // The state is set here manually, instead of using setState because setState is an async operation, meaning it might execute on the next tick,
-                // or at a later point in the microtask queue. That can lead to a race condition where
-                // setWithOnyxState is called before the state is actually set. This results in unreliable behavior when checking the loading state and has been mainly observed on fabric.
-                this.state = stateUpdate;
-
-                this.setState(stateUpdate); // Trigger a render
+                // Retain previous state to avoid data loss during pre-load updates.
+                // This handles race conditions where setWithOnyxState might be called after setState
+                this.setState(prevState => ({...this.tempState, ...prevState, loading: false}));
```
marcaaron, line 126: What is the race condition exactly? Is there any example? Can we break this problem down a bit further in the comment? Who is calling `setState()` and why should we ever prefer the previous state of the `withOnyx()` component over the initial value being set here?

lib/withOnyx.js
```diff
@@ -121,14 +121,9 @@ export default function (mapOnyxToState) {
                     return;
                 }
 
-                const stateUpdate = {...this.tempState, loading: false};
-
-                // The state is set here manually, instead of using setState because setState is an async operation, meaning it might execute on the next tick,
-                // or at a later point in the microtask queue. That can lead to a race condition where
-                // setWithOnyxState is called before the state is actually set. This results in unreliable behavior when checking the loading state and has been mainly observed on fabric.
```
marcaaron, line 128: Is all this stuff no longer true? Can't say this old comment really makes any sense to me tbh.

lib/withOnyx.js
pasyukevich: Exact flow described [here](https://github.com/Expensify/App/issues/21881#issuecomment-1643878246)

lib/withOnyx.js
pasyukevich: This stuff is still true. I observed the same behavior and just explained it in different words

This pr was created before the commit, that's why it led to the merge conflict

I found that comment a bit confusing and decided to choose a different version

lib/withOnyx.js
marcaaron: That's pretty interesting.

So ultimately, sounds like we have made the bad assumption here that `setState()` can never be called before `setWithOnyxState()`.

If we got here and the `state` already has the key then we could 
 in theory remove it from `tempState`? That approach might be a little more obvious vs. spreading the `prevState` after the `tempState`. Though they would achieve the same result. Thoughts?

pasyukevich: Yeah, I agree with you that it would be more straightforward to remove properties from the  `tempState` that already exist in `prevState`

I'd assume that you mean something like `omit`

I will  recheck the flow again with the update and apply changes

pasyukevich: @marcaaron PR updated with omit approach

marcaaron (approved):
LGTM thanks for the change

marcaaron:
All yours @Julesssss !

Julesssss (approved):
Okay, we'll have to be careful to test this in the App PR that bumps the Onyx lib.

Julesssss:
Sweet. I created a bump PR [here](https://github.com/Expensify/App/pull/26274) but I'm struggling to build Desktop now...

Julesssss:
Actually, the Desktop error seems unrelated

PR #284, opened 2023-07-25
Fix minor issues after introducing incremental JSON_PATCH merge

chrispader:
@marcaaron

### Details

Fixing some issues i encountered while testing the [`JSON_PATCH` merge changes](https://github.com/Expensify/react-native-onyx/pull/238) in the `Expensify/App` repo [in this PR](https://github.com/Expensify/App/pull/23151)

When batching the changes from `mergeQueue`, top-level `null` values may not be omitted. Only when merging the final value, we can remove these `null` values.

Also added back a log line, that was removed in the other PR

### DIFF
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -885,7 +885,7 @@ function set(key, value) {
     // This approach prioritizes fast UI changes without waiting for data to be stored in device storage.
     const hasChanged = broadcastUpdate(key, value, 'set');
 
-    if (hasChanged) { return Promise.resolve(); }
+    if (!hasChanged) { return Promise.resolve(); }
 
     return Storage.setItem(key, value)
         .catch(error => evictStorageAndRetry(error, set, key, value));
@@ -1001,7 +1001,7 @@ function merge(key, changes) {
                 // This approach prioritizes fast UI changes without waiting for data to be stored in device storage.
                 const hasChanged = broadcastUpdate(key, modifiedData, 'merge');
 
-                if (hasChanged) { return Promise.resolve(); }
+                if (!hasChanged) { return Promise.resolve(); }
 
                 return Storage.mergeItem(key, batchedChanges, modifiedData);
             } catch (error) {
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -948,7 +948,6 @@ function multiSet(data) {
  * @private
  * @param {*} existingValue
  * @param {Array<*>} changes Array of changes that should be applied to the existing value
- * @param {bool} omitNull Whether null top-level null values should be omitted
  * @returns {*}
  */
 function applyMerge(existingValue, changes) {
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -848,16 +848,16 @@ function evictStorageAndRetry(error, onyxMethod, ...args) {
  *
  * @param {String} key
  * @param {*} value
+ * @param {Boolean} hasChanged
  * @param {String} method
  * @returns {Boolean} whether the value for the given key has changed
  */
-function broadcastUpdate(key, value, method) {
+function broadcastUpdate(key, value, hasChanged, method) {
     // Logging properties only since values could be sensitive things we don't want to log
     Logger.logInfo(`${method}() called for key: ${key}${_.isObject(value) ? ` properties: ${_.keys(value).join(',')}` : ''}`);
 
     // Update subscribers if the cached value has changed, or when the subscriber specifically requires
     // all updates regardless of value changes (indicated by initWithStoredValues set to false).
-    const hasChanged = cache.hasValueChanged(key, value);
     if (hasChanged) {
         cache.set(key, value);
     } else {
@@ -895,8 +895,10 @@ function set(key, value) {
         Logger.logAlert(`Onyx.set() called after Onyx.merge() for key: ${key}. It is recommended to use set() or merge() not both.`);
     }
 
+    const hasChanged = cache.hasValueChanged(key, value);
+
     // This approach prioritizes fast UI changes without waiting for data to be stored in device storage.
-    const hasChanged = broadcastUpdate(key, value, 'set');
+    broadcastUpdate(key, value, hasChanged, 'set');
 
     // If the value has not changed then we don't need to write it to storage
     if (!hasChanged) {
@@ -1015,8 +1017,10 @@ function merge(key, changes) {
                 // After that we merge the batched changes with the existing value
                 const modifiedData = applyMerge(existingValue, [batchedChanges]);
 
+                const hasChanged = cache.hasValueChanged(key, modifiedData);
+
                 // This approach prioritizes fast UI changes without waiting for data to be stored in device storage.
-                const hasChanged = broadcastUpdate(key, modifiedData, 'merge');
+                broadcastUpdate(key, modifiedData, hasChanged, 'merge');
 
                 // If the value has not changed then we don't need to write it to storage
                 if (!hasChanged) {
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -1012,10 +1012,6 @@ function merge(key, changes) {
                 // After that we merge the batched changes with the existing value
                 const modifiedData = applyMerge(existingValue, [batchedChanges]);
 
-                console.log({
-                    key, existingValue, batchedChanges, modifiedData,
-                });
-
                 // This approach prioritizes fast UI changes without waiting for data to be stored in device storage.
                 const hasChanged = broadcastUpdate(key, modifiedData, 'merge');
 
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -941,11 +941,12 @@ function multiSet(data) {
  * Merges an array of changes with an existing value
  *
  * @private
- * @param {Array<*>} changes Array of changes that should be applied to the existing value
  * @param {*} existingValue
+ * @param {Array<*>} changes Array of changes that should be applied to the existing value
+ * @param {bool} omitNull Whether null top-level null values should be omitted
  * @returns {*}
  */
-function applyMerge(changes, existingValue) {
+function applyMerge(existingValue, changes, omitNull = true) {
     const lastChange = _.last(changes);
 
     if (_.isArray(existingValue) || _.isArray(lastChange)) {
@@ -960,7 +961,7 @@ function applyMerge(changes, existingValue) {
             const newData = Object.assign({}, fastMerge(modifiedData, change));
 
             // Remove all first level keys that are explicitly set to null.
-            return _.omit(newData, value => _.isNull(value));
+            return omitNull ? _.omit(newData, value => _.isNull(value)) : newData;
         }, existingValue || {});
     }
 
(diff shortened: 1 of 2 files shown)


lib/Onyx.js
```diff
@@ -874,8 +891,14 @@ function set(key, value) {
...
 
+    if (hasPendingMergeForKey(key)) {
+        Logger.logAlert(`Onyx.set() called after Onyx.merge() for key: ${key}. It is recommended to use set() or merge() not both.`);
+    }
+
     // This approach prioritizes fast UI changes without waiting for data to be stored in device storage.
-    broadcastUpdate(key, value, 'set');
+    const hasChanged = broadcastUpdate(key, value, 'set');
+
+    if (!hasChanged) { return Promise.resolve(); }
```
tgolen, line 901:
```suggestion
    if (!hasChanged) { 
        return Promise.resolve();
     }
```
Can you also please add a comment about _why_ this early return should be here?

lib/Onyx.js
```diff
@@ -979,17 +1003,19 @@ function merge(key, changes) {
...
 
                 // After that we merge the batched changes with the existing value
-                const modifiedData = applyMerge([batchedChanges], existingValue);
+                const modifiedData = applyMerge(existingValue, [batchedChanges]);
 
                 // This approach prioritizes fast UI changes without waiting for data to be stored in device storage.
-                broadcastUpdate(key, modifiedData, 'merge');
+                const hasChanged = broadcastUpdate(key, modifiedData, 'merge');
+
+                if (!hasChanged) { return Promise.resolve(); }
```
tgolen, line 1018: same

tgolen (changes requested):

lib/Onyx.js
marcaaron: +1 to the comment

lib/Onyx.js
```diff
@@ -849,6 +849,7 @@ function evictStorageAndRetry(error, onyxMethod, ...args) {
  * @param {String} key
  * @param {*} value
  * @param {String} method
+ * @returns {Boolean} whether the value for the given key has changed
```
marcaaron, line 852: Returning `hasChanged` from `broadcastUpdate()` changes it's responsibility a bit. 

Could we maybe pass `hasChanged` as an argument instead? That way the callers already have the value?

lib/Onyx.js
```diff
@@ -918,11 +941,12 @@ function multiSet(data) {
...
  *
  * @private
- * @param {Array<*>} changes Array of changes that should be applied to the existing value
  * @param {*} existingValue
+ * @param {Array<*>} changes Array of changes that should be applied to the existing value
+ * @param {bool} omitNull Whether null top-level null values should be omitted
  * @returns {*}
  */
-function applyMerge(changes, existingValue) {
+function applyMerge(existingValue, changes, omitNull = true) {
```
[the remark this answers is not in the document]
marcaaron: This is a tricky one. If we are using this method to build the "batch of changes" then don't we actually want to _drop_ those keys from storage and not just set them to "null" or is it the same thing as far as SQLite is concerned?

The only reason to "omit" them from the final object is so they get removed from storage.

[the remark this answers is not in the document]
marcaaron: Maybe it would be clearer to explicitly remove the top level null keys with some brief explanation _after_ we get the "result" (i.e. `modifiedData`) and basically make this method not responsible for modifying the object in that way?

marcaaron (changes requested):

lib/Onyx.js
chrispader: Added comments! We returned a resolved promise before too, when the value hasn't changed. That's why i added this back, same for `merge`

chrispader:
@marcaaron @tgolen all review comments adressed! 👍

lib/Onyx.js
```diff
@@ -874,8 +891,19 @@ function set(key, value) {
...
+        Logger.logAlert(`Onyx.set() called after Onyx.merge() for key: ${key}. It is recommended to use set() or merge() not both.`);
+    }
+
+    const hasChanged = cache.hasValueChanged(key, value);
+
     // This approach prioritizes fast UI changes without waiting for data to be stored in device storage.
-    broadcastUpdate(key, value, 'set');
+    broadcastUpdate(key, value, hasChanged, 'set');
+
+    // If the value has not changed then we don't need to write it to storage
```
tgolen, line 903: This comment should explain _why_ and it currently only explains _what_. Why doesn't it need to be written to storage?

> If the value has not changed, calling Storage.setItem() would be redundant and a waste of performance, so return early instead.

lib/Onyx.js
```diff
@@ -931,14 +960,9 @@ function applyMerge(changes, existingValue) {
...
         // Object values are merged one after the other
-        return _.reduce(changes, (modifiedData, change) => {
-            // lodash adds a small overhead so we don't use it here
-            // eslint-disable-next-line prefer-object-spread, rulesdir/prefer-underscore-method
-            const newData = Object.assign({}, fastMerge(modifiedData, change));
-
-            // Remove all first level keys that are explicitly set to null.
-            return _.omit(newData, value => _.isNull(value));
-        }, existingValue || {});
+        // lodash adds a small overhead so we don't use it here
```
tgolen, line 963: This comment is no longer valid. It also makes me think that we should still be using `Object.assign()` instead of the spread operator since performance is a very big consideration in this code.

lib/Onyx.js
```diff
@@ -979,17 +1003,29 @@ function merge(key, changes) {
...
 
                 // Clean up the write queue so we
                 // don't apply these changes again
                 delete mergeQueue[key];
 
                 // After that we merge the batched changes with the existing value
-                const modifiedData = applyMerge([batchedChanges], existingValue);
+                let modifiedData = applyMerge(existingValue, [batchedChanges]);
+
+                // For objects, we want to remove all top level keys that are null
```
tgolen, line 1015: Again, please modify the comment to explain _why_ we want to remove all top level keys that are null.

> For objects, the key for null values needs to be removed from the object to ensure the value will get removed from storage completely.

lib/Onyx.js
```diff
@@ -979,17 +1003,29 @@ function merge(key, changes) {
...
+                    modifiedData = _.omit(modifiedData, value => _.isNull(value));
+                }
+
+                const hasChanged = cache.hasValueChanged(key, modifiedData);
 
                 // This approach prioritizes fast UI changes without waiting for data to be stored in device storage.
-                broadcastUpdate(key, modifiedData, 'merge');
+                broadcastUpdate(key, modifiedData, hasChanged, 'merge');
+
+                // If the value has not changed then we don't need to write it to storage
```
tgolen, line 1025: Same as above

marcaaron:
+1 to @tgolen's comments about comments :)

>don't we actually want to drop those keys from storage and not just set them to "null" or is it the same thing as far as SQLite is concerned?

Also, I think this is a part that could be better explained in the comments as well. `JSON_PATCH()` behavior will delete a top level key when the value is `null` like in this example...

chrispader:
> Also, I think this is a part that could be better explained in the comments as well. `JSON_PATCH()` behavior will delete a top level key when the value is `null` like in this example...

Ah i understand, didn't know, that that's why we added this manually for web. Gonna further explain this in the comments then!

chrispader:
> If we are using this method to build the "batch of changes" then don't we actually want to drop those keys from storage and not just set them to "null" or is it the same thing as far as SQLite is concerned?

When building the batch of changes, we don't want to omit the null keys, because then this would happen:

Changes like `[{key1: null, key2: null}, {key3: null}]` turn into `modifiedData`: `{}`. Merging the existing object value with `{}` will not update it though, because merging `{}` is the same as merging nothing.

For creating the `modifiedData` we'll want to omit the keys, so they get removed from storage.

lib/Onyx.js
chrispader: Oh i thought the lodash comment was referring to that we the `fastMerge` instead of `_.merge()`.

According to [this source](https://thecodebarbarian.com/object-assign-vs-object-spread.html), object spreading seems to be the faster choice, especially when we pass an empty object as the target to `Object.assign`, as we do right now.

tgolen: I'd be careful referring to that article which is about Node.js. I've done a lot of performance analyzing in the browser and found the spread operator to be much slower than `Object.assign` (which could have just as much to do with our transpiler).

lib/Onyx.js
```diff
@@ -1012,7 +1013,8 @@ function merge(key, changes) {
                 // After that we merge the batched changes with the existing value
                 let modifiedData = applyMerge(existingValue, [batchedChanges]);
 
-                // For objects, we want to remove all top level keys that are null
+                // For objects, the key for null values needs to be removed from the object to ensure the value will get removed from storage completely.
+                // On native, SQLite will remove keys top-level keys that are null. To be consistent, we remove them on web too.
```
tgolen, line 1017:
```suggestion
                // On native, SQLite will remove top-level keys that are null. To be consistent, we remove them on web too.
```

tgolen (changes requested):
Getting close!

lib/Onyx.js
chrispader: Alright! Let's switch back to `Object.assign` then. 👍

chrispader: Technically, we don't need the `Object.assign` after all, because fast merge is already copying into a new object:

https://github.com/Expensify/react-native-onyx/blob/d4fc53cc3229808e9c1c542c13ce0089e5893d96/lib/fastMerge.js#L20

tgolen: I'm all for removing it entirely!

tgolen (dismissed):

lib/Onyx.js
```diff
@@ -918,11 +946,12 @@ function multiSet(data) {
  * Merges an array of changes with an existing value
  *
  * @private
- * @param {Array<*>} changes Array of changes that should be applied to the existing value
  * @param {*} existingValue
+ * @param {Array<*>} changes Array of changes that should be applied to the existing value
+ * @param {bool} omitNull Whether null top-level null values should be omitted
```
marcaaron, line 951: The `omitNull` here needs to be removed.

marcaaron (changes requested):
One small thing that needs to be addressed. Looks good - thanks for the changes!

lib/Onyx.js
chrispader: right, sorry for that 🥴

lib/Onyx.js
```diff
@@ -848,17 +848,34 @@ function evictStorageAndRetry(error, onyxMethod, ...args) {
...
-    cache.set(key, value);
+    if (hasChanged) {
+        cache.set(key, value);
+    } else {
+        cache.addToAccessedKeys(key);
+    }
+
     notifySubscribersOnNextTick(key, value, subscriber => hasChanged || subscriber.initWithStoredValues === false);
+
+    return hasChanged;
```
marcaaron, line 869:
```suggestion
```

lib/Onyx.js
```diff
@@ -848,17 +848,34 @@ function evictStorageAndRetry(error, onyxMethod, ...args) {
  *
  * @param {String} key
  * @param {*} value
+ * @param {Boolean} hasChanged
  * @param {String} method
+ * @returns {Boolean} whether the value for the given key has changed
```
marcaaron, line 853:
```suggestion
```

marcaaron (dismissed):

marcaaron:
One last thing 🥲

chrispader:
> One last thing 🥲

fixed!

mountiny (approved):
Thanks @tgolen and @marcaaron for pushing this forwards.

To unblock the other onyx caching PR, I am going to approve and merge this, given Tim already gave approval and the last changes requested by Marc has been addressed and they were just comments.

PR #292, opened 2023-08-02
Fix: Remove nullish values on set and on merge when no value is in storage

chrispader:
@tgolen @marcaaron @hannojg 

In the native implementation with SQLite, the first call to `merge` on a key where no value is in storage will act like a regular `set` since the values are inserted into storage directly.

This is due to this SQLite operation in `SQLiteStorage.mergeItem`:

```sqlite
INSERT INTO keyvaluepairs (record_key, valueJSON)
VALUES (:key, JSON(:value))
ON CONFLICT DO UPDATE
```

 In order for this to be consistent with how later `merge` with `JSON_PATCH` are working, we need to remove nullish values from the `batchedChanged` in the first merge operation.

Additionally, we'll want to remove nullish values on `set` as well, so that the values in cache and storage are always identical.

### DIFF
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -910,6 +910,21 @@ function hasPendingMergeForKey(key) {
 }
 
 /**
+ * For objects, the key for null values needs to be removed from the object to ensure the value will get removed from storage completely.
+ * On native, SQLite will remove top-level keys that are null. To be consistent, we remove them on web too.
+ * @private
+ * @param {*} value
+ * @returns {*}
+ */
+function cleanObject(value) {
+    if (!_.isArray(value) && _.isObject(value)) {
+        return _.omit(value, objectValue => _.isNull(objectValue));
+    }
+
+    return value;
+}
+
+/**
  * Write a value to our store with the given key
  *
  * @param {String} key ONYXKEY to set
@@ -918,7 +933,9 @@ function hasPendingMergeForKey(key) {
  * @returns {Promise}
  */
 function set(key, value) {
-    if (_.isNull(value)) {
+    let valueCopy = value;
+
+    if (_.isNull(valueCopy)) {
         return remove(key);
     }
 
@@ -926,18 +943,20 @@ function set(key, value) {
         Logger.logAlert(`Onyx.set() called after Onyx.merge() for key: ${key}. It is recommended to use set() or merge() not both.`);
     }
 
-    const hasChanged = cache.hasValueChanged(key, value);
+    valueCopy = cleanObject(valueCopy);
+
+    const hasChanged = cache.hasValueChanged(key, valueCopy);
 
     // This approach prioritizes fast UI changes without waiting for data to be stored in device storage.
-    broadcastUpdate(key, value, hasChanged, 'set');
+    broadcastUpdate(key, valueCopy, hasChanged, 'set');
 
     // If the value has not changed, calling Storage.setItem() would be redundant and a waste of performance, so return early instead.
     if (!hasChanged) {
         return Promise.resolve();
     }
 
-    return Storage.setItem(key, value)
-        .catch(error => evictStorageAndRetry(error, set, key, value));
+    return Storage.setItem(key, valueCopy)
+        .catch(error => evictStorageAndRetry(error, set, key, valueCopy));
 }
 
 /**
@@ -1034,7 +1053,7 @@ function merge(key, changes) {
         .then((existingValue) => {
             try {
                 // We first only merge the changes, so we can provide these to the native implementation (SQLite uses only delta changes in "JSON_PATCH" to merge)
-                const batchedChanges = applyMerge(undefined, mergeQueue[key]);
+                let batchedChanges = applyMerge(undefined, mergeQueue[key]);
 
                 // Clean up the write queue so we
                 // don't apply these changes again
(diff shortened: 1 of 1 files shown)


hannojg:
To give another explanation to the issue at hand:

**Calling `Onyx.merge` for a new key:**

- `Onyx.merge(newKey, { prop1: 12, prop2: null})`
- Calls `INSERT INTO` as there is no conflict
- In storage we now have: `{ prop1: 12, prop2: null}`

Where as, when...

**Calling `Onyx.merge` for an existing key:**

- `Onyx.merge(key, { prop1: 12, prop2: null})`
- Calls `JSON_PATCH` as we have a conflict
- In storage we now have: `{ prop1: 12 }`

First of all, this is inconsistent behaviour, and secondly in the app code we don't expect the `null` values to be persisted, so it creates issues like this one: 

- https://github.com/Expensify/App/issues/24006

lib/Onyx.js
```diff
@@ -1376,7 +1397,6 @@ const Onyx = {
     multiSet,
     merge,
     mergeCollection,
-    hasPendingMergeForKey,
```
mountiny, line 1379: Is this not used anywhere? Was this removed on purpose?

mountiny (approved):
Looking good but also tagging Marc and Tim for review

lib/Onyx.js
chrispader: Yes, this was removed on purpose. This is actually a private function and is not meant to be exported, yet i added it in the last [this PR](https://github.com/Expensify/react-native-onyx/pull/238), which was a mistake

lib/Onyx.js
```diff
@@ -909,6 +909,21 @@ function hasPendingMergeForKey(key) {
     return Boolean(mergeQueue[key]);
 }
 
+/**
+ * For objects, the key for null values needs to be removed from the object to ensure the value will get removed from storage completely.
+ * On native, SQLite will remove top-level keys that are null. To be consistent, we remove them on web too.
```
tgolen, line 914: It might be better to expand this comment a little more to explain:

1. What the default behavior is on Web since it's not using SQLite
2. Why that default behavior isn't consistent with what SQLite does
3. Why the behavior needs to be consistent for all platforms

It could be something like this (though I don't know that these answers are 100% correct).

>  On native, SQLite removes top-level keys that have null values. This is desired because it decreases the amount of data stored in memory and on disk. On web, IndexedDB will keep the top-level keys along with a null value and this uses up storage and memory. This method will ensure that keys for null values are removed before an object is persisted to disk so that all platforms are storing the data in the same efficient way.

lib/Onyx.js
```diff
@@ -909,6 +909,21 @@ function hasPendingMergeForKey(key) {
...
+ * @private
+ * @param {*} value
+ * @returns {*}
+ */
+function cleanObject(value) {
+    if (!_.isArray(value) && _.isObject(value)) {
+        return _.omit(value, objectValue => _.isNull(objectValue));
+    }
+
+    return value;
```
tgolen, line 924: Early return (suggestion is untested)
```suggestion
    if (_.isArray(value) || !_.isObject(value)) {
        return value;
    }

    return _.omit(value, objectValue => _.isNull(objectValue));
```

lib/Onyx.js
```diff
@@ -918,26 +933,30 @@ function hasPendingMergeForKey(key) {
  * @returns {Promise}
  */
 function set(key, value) {
-    if (_.isNull(value)) {
+    let valueCopy = value;
```
tgolen, line 936: This variable name is misleading because it's not a copy yet, but still a reference to the same value. It doesn't become an actual copy until `cleanObject(valueCopy)` is ran. My suggestion is to remove this line, change `if (_.isNull(value)) {` and then `const valueCopy = cleanObject(value);`

lib/Onyx.js
```diff
@@ -1043,10 +1062,12 @@ function merge(key, changes) {
                 // After that we merge the batched changes with the existing value
                 let modifiedData = applyMerge(existingValue, [batchedChanges]);
```
tgolen, line 1063:
```suggestion
                const modifiedData = cleanObject(applyMerge(existingValue, [batchedChanges]));
```

Then you can remove the line below.

lib/Onyx.js
```diff
@@ -1043,10 +1062,12 @@ function merge(key, changes) {
...
 
-                // For objects, the key for null values needs to be removed from the object to ensure the value will get removed from storage completely.
-                // On native, SQLite will remove top-level keys that are null. To be consistent, we remove them on web too.
-                if (!_.isArray(modifiedData) && _.isObject(modifiedData)) {
-                    modifiedData = _.omit(modifiedData, value => _.isNull(value));
+                modifiedData = cleanObject(modifiedData);
+
+                // On the first write SQLite doesn't use JSON_PATCH, which means, nullish values won't be omitted.
```
tgolen, line 1067: How does this code know whether or not this is the first write or not? Also, why doesn't it use JSON_PATCH on the first write? Also, why does only JSON_PATCH remove null values?

I know that's a lot of questions, but answering them in the code comment will help other engineers that see this.

lib/Onyx.js
```diff
@@ -1043,10 +1062,12 @@ function merge(key, changes) {
...
+                // Therefore we remove null values from the `batchedChanges` which are sent to the native implementation, if no existing value is present.
```
tgolen, line 1068: > which are sent to the native implementation, if no existing value is present.

This doesn't make sense to me. What is "the native implementation"? The native implementation of what? And why does it only get sent  if no existing value is present?

tgolen (changes requested):
Left some comments to hopefully help the comments to be more clear to others.

lib/Onyx.js
```diff
@@ -909,6 +909,21 @@ function hasPendingMergeForKey(key) {
...
 }
 
+/**
+ * For objects, the key for null values needs to be removed from the object to ensure the value will get removed from storage completely.
+ * On native, SQLite will remove top-level keys that are null. To be consistent, we remove them on web too.
+ * @private
+ * @param {*} value
+ * @returns {*}
+ */
+function cleanObject(value) {
```
marcaaron, line 919: NAB, I would prefer to call this something like `removeNullObjectValues()` #notASecretJavaDeveloper

lib/Onyx.js
marcaaron: +1 

With a NAB, I still don't love `valueCopy`. it's not really a copy at all since we modify the object and remove the null properties. Maybe `valueWithNullRemoved`. Sorry if that's too Clean Code™️ I swear I am not a Java developer 😄

marcaaron (dismissed):
Code LGTM. Agree with most of Tim's comments and think we could explain this edge case a bit better with some inline comments. @hannojg's description of the problem was very clear to me so maybe it can be incorporated somehow.

lib/Onyx.js
chrispader: I actually prefer the two-line approach more, since it's easier for readability imo. But i can also change this to your version 👍

hannojg:
Awaiting another round of review, are we good to continue? This is blocking:

- https://github.com/Expensify/App/pull/24041

@mountiny @marcaaron @tgolen

tgolen (approved):
Love the comments. Very clear now. Thank you!

Issue #281, opened 2023-07-17
Investigate flushing Onyx cache to disk on idle


marcaaron:
We have observed that IndexedDB can be super slow with writes and cause major lag and jank in the app when there is a huge amount of data to store on App init. 

Our solution for now has been to just completely cease storing certain keys entirely. But, another potentially viable solution is to keep the data in memory for some time and then persist it to disk when the user is "idling".

cc @hannojg @kidroca curious if you guys have considered this one yet.

h/t @quinthar for the idea

kidroca:
Hi @marcaaron, 

Technically, it's not so much that `IndexedDB` is inherently slow, but more so that it's implemented inefficiently by `localforage` when subjected to frequent write operations.

To illustrate, as noted in this thread: https://expensify.slack.com/archives/C01GTK53T8Q/p1689267792230089?thread_ts=1689036681.307159&cid=C01GTK53T8Q:

> "Reviewing the `localForage` [source code for setItem](https://github.com/localForage/localForage/blob/c5b66157252988dc4e74a3eab5226b59be376a2f/src/drivers/indexeddb.js#L640C49-L640C49), we can observe that a new transaction is indeed opened each time the function is invoked. This could be a significant contributing factor to the noticeable slowdown experienced when executing `multiSet` or `multiMerge` operations. Currently, both these operations use `setItem` to process each item individually, which could potentially lead to high resource contention and decreased performance."

In light of this, I propose that we consider implementing a storage provider that circumvents the use of `localforage`. Instead, this provider would interface directly with the IndexedDB API, enabling it to more effectively strategize and optimize transactions. For instance, it could perform a single transaction to update multiple keys as opposed to sequential transactions. This would decrease the likelihood of lock contention and should, in theory, enhance speed significantly.
_(The change should also let us drop the `SyncQueue` as we should be able to write in one go, not having to schedule writes one by one)_

Alternatively, we could explore other storage driver libraries that might be a better fit for our needs. There exist libraries that streamline the IndexedDB interface, offering ease of use without forfeiting granular control.

I like the idea of batching a few changes and applying them at the best time. It appears that the app does not require immediate data writes. In the worst-case scenario, we might "skip of a beat," and some information might need to be fetched remotely due to delayed persistence.

marcaaron:
Great thoughts @kidroca. So it seems like our options are:

- Investigate writing our own IndexedDB implementation (Feels potentially high complexity, but I lack experience working with the API directly so not sure)
[...]

PR #293, opened 2023-08-03
Perf: use idb-keyval

hannojg:
### Details

While investigating performance issues on large accounts we found an issue where storing large amounts of data in onyx takes a very long time (E.g. merging the data from a OpenApp command).
Largely this was due to the underlying storage provider opening a new database transaction for every operation in a  multiMerge/multiSet operation.
The [idb-keyval](https://www.npmjs.com/package/idb-keyval) package optimises this, as it can handle multiple operations in one database transaction.
The good thing is, that `idb-keyval` also uses indexedDB as storage provider, so no migration is needed.

I created a performance comparison tool, and it showed how much better `idb-keyval` performs:

> Each operation has been executed with 10,000 items (depth 1), which has a size of ~2.3MB

| Operation | Onyx w/ Localforage | Onyx w/ idb-keyval | Improvement in %
|--------|--------|--------|--------|
| Single Set | 42ms | 36ms | **-14%**
| Get (collection) | 26ms | 11ms | **-58%**
| MergeCollection | 57926ms | 775ms | **-99%**
| Single Merge | 48ms | 30ms | **-38%**

<details><summary><b>How to run the performance tests yourself:<b></summary>
<p>

1. Hop on [this branch](https://github.com/margelo/react-native-onyx/tree/poc/other-databases) (its on our fork, sorry)
2. `cd example`
3. `npm i`
4. `npm start`
5. Open in your web browser `http://localhost:8080`

Screenshot 2023-08-03 at 16 45 41

</p>
</details> 

For the product the results can be seen in this video:

https://expensify.slack.com/archives/C035J5C9FAP/p1691063111819169

### Related Issues

https://github.com/Expensify/react-native-onyx/issues/281


lib/storage/WebStorage.js
```diff
@@ -4,7 +4,7 @@
  * data changes and then stay up-to-date with everything happening in Onyx.
  */
 import _ from 'underscore';
-import Storage from './providers/LocalForage';
+import Storage from './providers/IDBKeyVal';
```
MonilBhavsar, line 7: We can update the comment above this to use idb-keyval

MonilBhavsar (dismissed):
Looks great! Any idea how can we test this in product? 🚀

marcaaron:
>Looks great! Any idea how can we test this in product?

I would try this:

1. Run `npm build` from inside `react-native-onyx`
2. Replace the `/dist` folder in `App/node_modules/react-native-onyx`
3. Use the production API proxy in the `.env` file
4. Sign in with an account that is completely buggered and has tons of reports + policies
5. See if it improves performance

lib/storage/providers/IDBKeyVal.js
```diff
@@ -0,0 +1,111 @@
...
+    del,
+    delMany,
+    createStore,
+    promisifyRequest,
+} from 'idb-keyval';
+import _ from 'underscore';
+import fastMerge from '../../fastMerge';
+
+// Same config as localforage, so we can swap the providers easily
+const customStore = createStore('OnyxDB', 'keyvaluepairs');
```
marcaaron, line 17: **confirming:** This will have the existing data from the current DB previously set up by `localforage`?

marcaaron (commented):
Maybe we should just remove all traces of `localforage` if nothing will use it? 🤔

lib/storage/providers/IDBKeyVal.js
hannojg: Yes exactly 😊 IT can just be replaced in place

hannojg:
@marcaaron You mean to delete the provider file?

marcaaron:
>You mean to delete the provider file?

👍

hannojg:
Alright, I removed all traces of local forage 😊

roryabraham (commented):
NAB @hannojg maybe we can benchmark `Onyx.clear` as well since I know there are some race conditions around it and it's unclear how well the `localforage-removeItems` plugin ever actually worked

marcaaron:
oh yeah great thought @roryabraham! adding to that point, I think the place where I see this slowing things down the most is when someone signs out of an account with a ton of data. so we should be sure to A/B test that flow and see if things improved.

danieldoglas:
> maybe we can benchmark Onyx.clear as well since I know there are some race conditions around it and it's unclear how well the localforage-removeItems plugin ever actually worked

I don't think this is blocker for merging this PR since we don't know if the old one actually worked, right? 

It would be great to have this merged Today so we can see the benefits for this mostly in Guides and AM accounts

danieldoglas:
Both of them return 0ms when running `Clear`

image

image

hannojg:
Let me fix the benchmark for clearing and give numbers for that 😊  With this implementation deleting / clearing should definitely work

hannojg:
Clearing seems to be a bit slower with idb-keyval, however the impact doesn't look too dramatic. I think we should be able to optimise it to be on par with local forage. However, I don't think that has too much priority at the moment / blocking this from getting merged?

localforage

idb-keyval

(Note: for a lower item count there is almost no difference between the two implementations)

MonilBhavsar (approved):

marcaaron:
>While investigating performance issues on large accounts we found an issue where storing large amounts of data in onyx takes a very long time (E.g. merging the data from a OpenApp command).
>Largely this was due to the underlying storage provider opening a new database transaction for every operation in a multiMerge/multiSet operation.

Are we sure this is faster than the previous solution of setting one value at a time synchronously? Did we verify that things improved on a large account? Which account did we test with?

>The [idb-keyval](https://www.npmjs.com/package/idb-keyval) package optimises this, as it can handle multiple operations in one database transaction.

Is is possible we can ask it to do too many at once? I am seeing various weird logs like:


```
Error: TimeoutError: Transaction timed out due to inactivity.. onyxMethod: Ae ~~ parameters: '' userAgent: 'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/116.0.0.0 Safari/537.36 Edg/116.0.1938.54'
```

mountiny:
I believe @hannojg is on vacation @Szymon20000 Would you be able to look into the above?

Szymon20000:
I will try but my focus this week is on improving chat switching and batching updates.

Szymon20000:
Added to my todo list!

PR #300, opened 2023-08-16
Add better handling when users run out of storage.

marcaaron:
### Details

When a user runs out of device storage and doesn't have anything they can evict apps will get caught in an infinite loop.

The solution I came up with:

When we run out of evictable keys we will

- Try to see how much storage is left on the device + how much we used (for logging purposes)
- Skip the write we are trying to do if we have no evictable keys
- Resolve the promise that is currently throwing (this gets people stuck in a loop)

### Related Issues

https://github.com/Expensify/Expensify/issues/308389

### Automated Tests

Maybe we can add some in the future. I am not prioritizing them right now.

### Manual Tests

**Web**
1. Test against `App/main`
2. Artificially put a limit on Chrome storage like this:

3. Signed in with an account that is totally messed up with thousands of policies.
4. Saw these logs:


marcochavezf:
Oops conflicts

marcaaron:
Updated!

marcochavezf (approved):
LGTM

PR #305, opened 2023-08-18
Add multiSet to update

yuwenmemon:
@marcaaron please review

### Details
Adds `multiset` to onyx's update.

### Related Issues
https://github.com/Expensify/Expensify/issues/309134

### Automated Tests
Added 2 tests

### DIFF
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -14,6 +14,7 @@ const METHOD = {
     SET: 'set',
     MERGE: 'merge',
     MERGE_COLLECTION: 'mergecollection',
+    MULTI_SET: 'multiset',
     CLEAR: 'clear',
 };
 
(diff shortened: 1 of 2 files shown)


lib/Onyx.js
```diff
@@ -1268,16 +1269,28 @@ function mergeCollection(collectionKey, collection) {
...
+                throw new Error('Invalid value provided in Onyx multiSet. Onyx multiSet value must be of type object.');
+            }
+
+            _.each(_.keys(value), (multiSetKey) => {
+                if (_.isString(multiSetKey)) {
+                    return;
+                }
+
+                throw new Error(`Invalid ${typeof key} key provided in Onyx multiSet. Onyx key must be of type string.`);
+            });
```
marcaaron, line 1292: This seems not logically possible because all keys of an object must be strings?

tests/unit/onyxTest.js
```diff
@@ -494,15 +494,15 @@ describe('Onyx', () => {
...
+            {onyxMethod: 'murge', key: ONYX_KEYS.OTHER_TEST, value: {test2: 'test2'}},
         ];
 
         try {
             // When we pass it to Onyx.update
             Onyx.update(data);
         } catch (error) {
             // Then we should expect the error message below
-            expect(error.message).toEqual('Invalid onyxMethod multiSet in Onyx update.');
+            expect(error.message).toEqual('Invalid onyxMethod murge in Onyx update.');
```
marcaaron, line 505: murge?! 😂

marcaaron (changes requested):
Changes looks good. Maybe a good follow up eventually is to update the `multiSet()` logic so that it can group which of the keys belong to a collection and then update subscribers in a more optimized way.

https://github.com/Expensify/react-native-onyx/blob/f96fa2dc38b6abd1f99512af59647162fa11c88e/lib/Onyx.js#L985-L992

That _might_ bite us in the future (sudden bad performance) if someone tries to use `multiSet()` to update a collection and also when there is a subscriber for the collection key.

lib/Onyx.js
yuwenmemon: They can be other types as well. At least, what I'm looking to do if prevent number-type keys at the very minimum. You can see this in my unit test.: https://github.com/Expensify/react-native-onyx/pull/305/files#diff-fb75d1d7a8480f8429b43b89f94665322ee4576f95220815a5f7f3275e99fb40R605

tests/unit/onyxTest.js
yuwenmemon: Haha - yeah open to ideas for better names for a "wrong" onyxMethod.

yuwenmemon:
> Maybe a good follow up eventually is to update the multiSet() logic so that it can group which of the keys belong to a collection and then update subscribers in a more optimized way.

Yeah I thought about that. Seems like a good issue for a follow-up since there isn't an immediate need for it.

lib/Onyx.js
marcaaron: sorry, not trying to split hairs. I thought they would be cast to string. At least when iterating over them with `Object.keys()`, but maybe numbers are unique 🤷‍♂️ 



NAB, I am not sure why should we not use a number as a key for Onyx. We would never do it. But doesn't seem like it should be explicitly prevented for any reason.

yuwenmemon: Ah nice, okay. I can remove those checks then.

marcaaron (approved):

yuwenmemon:
Follow-up issue is here: https://github.com/Expensify/react-native-onyx/issues/313

PR #307, opened 2023-08-21
Fix idb keyval provider usage in electron environment

hannojg:
### Details

Follow up on as there is a crash in the expensify desktop app:

- #293 

Don't initialise the store at bundle load time, but at runtime. At load time in electron we are on the main thread, which doesn't has access to the browser environment.

### Related Issues

Related to https://github.com/Expensify/react-native-onyx/issues/281

### DIFF
--- a/lib/storage/providers/IDBKeyVal.js
+++ b/lib/storage/providers/IDBKeyVal.js
@@ -13,7 +13,15 @@ import {
 import _ from 'underscore';
 import fastMerge from '../../fastMerge';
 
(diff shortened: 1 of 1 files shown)


mountiny:
Related to this Slack thread.

mountiny (approved):
thanks for a quick follow up

mountiny:
I will go ahead and merge this without waiting for Florent as its a quick follow up to previous pr to unblock testing of this perf improvement

Issue #310, opened 2023-08-22
Onyx collection values sometimes appear as null to subscribers.


roryabraham:
Coming from https://github.com/Expensify/App/issues/25481#issuecomment-1687321203

### Problem

Generally, to delete stuff from Onyx you set it to null. However:

- Given a subscriber to a collection key with waitForCollectionCallback: true
- When a collection member key is set to null
- Then the subscriber should get the new complete collection with any null values removed

However, the subscriber is getting the new collection with a null value for the item that was meant to be deleted.

### Solution

TBD. According to a [contributor](https://github.com/Expensify/App/issues/25481#issuecomment-1686741391):

> it is due to notify the subscribers with the changed values before updating the storage.
https://github.com/Expensify/react-native-onyx/blob/79bfb5ef713e288fea8827cb4efe0d677177a525/lib/Onyx.js#L508-L515

roryabraham:
Created a quick fix: https://github.com/Expensify/react-native-onyx/pull/311

roryabraham:
Whoa, how is this issue still open>

PR #311, opened 2023-08-22
[HOLD] Don't include null in collection callbacks

roryabraham:
### Details
Coming from https://github.com/Expensify/App/issues/25481#issuecomment-1687321203, we are including null in `Onyx.connect` subscriber callbacks if it's a collection key w/ `waitForCollectionCallback: true` 🐛 

This PR adds a test to cover this edge case and then prevents it such that the full collection w/o null is given to the subscriber.

### Related Issues
Fixes https://github.com/Expensify/react-native-onyx/issues/310

### Automated Tests
Added automated test to cover this.

### Linked PRs
Not created yet

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -15,4 +15,13 @@ module.exports = {
         },
     },
     ignorePatterns: 'dist',
+    overrides: [
+        {
+            files: ['tests/**/*.{js,jsx,ts,tsx}'],
+            rules: {
+                '@lwc/lwc/no-async-await': 'off',
+                'no-await-in-loop': 'off',
+            },
+        },
+    ],
 };
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -1259,18 +1259,33 @@ function mergeCollection(collectionKey, collection) {
 
     return getAllKeys()
         .then((persistedKeys) => {
-            // Split to keys that exist in storage and keys that don't
-            const [existingKeys, newKeys] = _.chain(collection)
-                .keys()
-                .partition(key => persistedKeys.includes(key))
-                .value();
+            const existingKeysToKeep = [];
+            const keysToRemove = [];
+            const collectionWithoutKeysToRemove = collection;
+            const keyValuePairsForExistingCollection = [];
+            const keyValuePairsForNewCollection = [];
+            _.each(collection, (value, key) => {
+                if (value === null) {
+                    keysToRemove.push(key);
+                    delete collectionWithoutKeysToRemove[key];
+                    return;
+                }
 
-            const existingKeyCollection = _.pick(collection, existingKeys);
-            const newCollection = _.pick(collection, newKeys);
-            const keyValuePairsForExistingCollection = prepareKeyValuePairsForStorage(existingKeyCollection);
-            const keyValuePairsForNewCollection = prepareKeyValuePairsForStorage(newCollection);
+                if (persistedKeys.includes(key)) {
+                    existingKeysToKeep.push(key);
+                    keyValuePairsForExistingCollection.push([key, removeNullObjectValues(value)]);
+                    return;
+                }
+
+                keyValuePairsForNewCollection.push([key, removeNullObjectValues(value)]);
+            });
 
             const promises = [];
+
+            if (keysToRemove.length > 0) {
+                _.each(keysToRemove, (key) => cache.drop(key));
+                promises.push(Storage.removeItems(keysToRemove));
+            }
 
             // New keys will be added via multiSet while existing keys will be updated using multiMerge
             // This is because setting a key that doesn't exist yet with multiMerge will throw errors
@@ -1284,9 +1299,9 @@ function mergeCollection(collectionKey, collection) {
 
             // Prefill cache if necessary by calling get() on any existing keys and then merge original data to cache
             // and update all subscribers
-            Promise.all(_.map(existingKeys, get)).then(() => {
-                cache.merge(collection);
-                keysChanged(collectionKey, collection);
+            Promise.all(_.map(existingKeysToKeep, get)).then(() => {
+                cache.merge(collectionWithoutKeysToRemove);
+                keysChanged(collectionKey, collectionWithoutKeysToRemove);
             });
 
             return Promise.all(promises)
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -636,7 +636,7 @@ function sendDataToConnection(mapping, val, matchedKey) {
         return;
     }
 
-    if (_.isFunction(mapping.callback)) {
+    if (_.isFunction(mapping.callback) && (isCollectionKey(mapping.key) || matchedKey)) {
         mapping.callback(val, matchedKey);
     }
 }
@@ -1021,6 +1021,10 @@ function multiSet(data) {
  * @returns {*}
  */
 function applyMerge(existingValue, changes) {
+    if (_.some(changes, change => _.isNull(change))) {
+        return null;
+    }
+
     const lastChange = _.last(changes);
 
     if (_.isArray(existingValue) || _.isArray(lastChange)) {
@@ -1081,6 +1085,11 @@ function merge(key, changes) {
 
                 // After that we merge the batched changes with the existing value
                 const modifiedData = removeNullObjectValues(applyMerge(existingValue, [batchedChanges]));
+
+                // If the data is null, remove it from cache and store
+                if (_.isNull(modifiedData)) {
+                    return remove(key);
+                }
 
                 // On native platforms we use SQLite which utilises JSON_PATCH to merge changes.
                 // JSON_PATCH generally removes top-level nullish values from the stored object.
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -1021,21 +1021,21 @@ function multiSet(data) {
  * @returns {*}
  */
 function applyMerge(existingValue, changes) {
-    if (_.some(changes, change => _.isNull(change))) {
-        return null;
-    }
-
     const lastChange = _.last(changes);
+    if (_.isNull(lastChange)) {
+        return lastChange;
+    }
 
     if (_.isArray(existingValue) || _.isArray(lastChange)) {
         return lastChange;
     }
 
-    if (_.isObject(existingValue) || _.every(changes, _.isObject)) {
+    const changesToMake = _.filter(changes, change => Boolean(change));
+    if (_.isObject(existingValue) || _.every(changesToMake, _.isObject)) {
         // Object values are merged one after the other
         // lodash adds a small overhead so we don't use it here
         // eslint-disable-next-line prefer-object-spread, rulesdir/prefer-underscore-method
-        return _.reduce(changes, (modifiedData, change) => fastMerge(modifiedData, change),
+        return _.reduce(changesToMake, (modifiedData, change) => fastMerge(modifiedData, change),
             existingValue || {});
     }
 
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -488,7 +488,7 @@ function keysChanged(collectionKey, partialCollection) {
  * @param {*} data
  * @param {Function} [canUpdateSubscriber] only subscribers that pass this truth test will be updated
  */
-function keyChanged(key, data, canUpdateSubscriber) {
+function keyChanged(key, data, canUpdateSubscriber = () => true) {
     // Add or remove this key from the recentlyAccessedKeys lists
     if (!_.isNull(data)) {
         addLastAccessedKey(key);
@@ -502,7 +502,7 @@ function keyChanged(key, data, canUpdateSubscriber) {
     const stateMappingKeys = _.keys(callbackToStateMapping);
     for (let i = 0; i < stateMappingKeys.length; i++) {
         const subscriber = callbackToStateMapping[stateMappingKeys[i]];
-        if (!subscriber || !isKeyMatch(subscriber.key, key) || (_.isFunction(canUpdateSubscriber) && !canUpdateSubscriber(subscriber))) {
+        if (!subscriber || !isKeyMatch(subscriber.key, key) || !canUpdateSubscriber(subscriber)) {
             continue;
         }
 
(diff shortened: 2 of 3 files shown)


lib/Onyx.js
```diff
@@ -509,7 +509,9 @@ function keyChanged(key, data, canUpdateSubscriber) {
         if (_.isFunction(subscriber.callback)) {
             if (isCollectionKey(subscriber.key) && subscriber.waitForCollectionCallback) {
                 const cachedCollection = getCachedCollection(subscriber.key);
-                cachedCollection[key] = data;
+                if (data !== null) {
+                    cachedCollection[key] = data;
+                }
```
Pujan92, line 514: For data `null`, don't we need to remove that key from the `cachedCollection` as we are not updating with any other value?

marcaaron: I agree. Gonna ask the question another way:

Why do we need to set `data` to the `key` for the `cachedCollection`? Shouldn't the cache _already_ be correct by this point? If we are adding a key it should be there. If we are removing a key it should be there.

tests/unit/onyxTest.js
```diff
@@ -53,6 +53,22 @@ describe('Onyx', () => {
             expect(keys.includes(ONYX_KEYS.OTHER_TEST)).toBe(false);
         }));
 
+    it('should not provide null values to subscriber callback when set is called with null value', () => {
```
marcaaron, line 56:
```suggestion
    it('should not provide null values to subscriber callback using waitForCollectionCallback when set is called with null value', () => {
```

tests/unit/onyxTest.js
```diff
@@ -53,6 +53,22 @@ describe('Onyx', () => {
...
+        const subscriberCallback = jest.fn();
+        Onyx.connect({
+            key: ONYX_KEYS.COLLECTION.TEST_CONNECT_COLLECTION,
+            waitForCollectionCallback: true,
+            callback: subscriberCallback,
+        });
+        const itemID = 'abcd';
+        const testKey = `${ONYX_KEYS.COLLECTION.TEST_CONNECT_COLLECTION}${itemID}`;
+        return Onyx.set(testKey, 42)
```
marcaaron, line 65: Should we add tests for the `merge()` and `mergeCollection()` cases?

marcaaron (changes requested):

tests/unit/onyxTest.js
roryabraham: Done! This exercise uncovered that `merge` and `mergeCollection` weren't really behaving how I would expect

tests/unit/onyxTest.js
roryabraham: done

lib/Onyx.js
roryabraham: Added assertions to ensure that keys that are being deleted due to being null are dropped from both cache and storage.

lib/Onyx.js
```diff
@@ -634,7 +636,7 @@ function sendDataToConnection(mapping, val, matchedKey) {
         return;
     }
 
-    if (_.isFunction(mapping.callback)) {
+    if (_.isFunction(mapping.callback) && (isCollectionKey(mapping.key) || matchedKey)) {
```
roryabraham, line 639: This change prevents extra calls to `Onyx.connect` subscribers with value `null` and key `undefined`. `withOnyx` waits for all connection callbacks to have been called once before rendering, so that's why we added this type of callback. However, I don't think that this type of callback is helpful to other library code using `Onyx.connect` directly, so I prevented that here. `withOnyx` callbacks are handled in the block above which terminates with an early return.

tests/unit/onyxTest.js
```diff
@@ -671,14 +802,11 @@ describe('Onyx', () => {
                 return waitForPromisesToResolve();
             })
             .then(() => {
-                // Then we expect the callback to have called twice, once for the initial connect call + once for the collection update
-                expect(mockCallback).toHaveBeenCalledTimes(2);
-
-                // AND the value for the first call should be null since the collection was not initialized at that point
-                expect(mockCallback).toHaveBeenNthCalledWith(1, null, undefined);
```
roryabraham, line 678: This is an instance of the `null, undefined` callback mentioned above that I removed.

.eslintrc.js
```diff
@@ -15,4 +15,13 @@ module.exports = {
         },
     },
     ignorePatterns: 'dist',
+    overrides: [
```
roryabraham, line 18: Discussion:

- https://expensify.slack.com/archives/C01GTK53T8Q/p1692889359641599
- https://expensify.slack.com/archives/C03TQ48KC/p1692889446895489

Also, this same change is also present in E/App: https://github.com/Expensify/App/blob/f74b3325b3622f8afdee44ce8de1a40b8895c1bc/.eslintrc.js#L166-L173

lib/Onyx.js
marcaaron: Can you clarify which call to `sendDataToConnection()` we are trying to prevent and why? There are only a few of them - so I worry that we are adding this conditional exception in the wrong place. And maybe don't really need this change at all.

lib/Onyx.js
```diff
@@ -1019,6 +1021,10 @@ function multiSet(data) {
  * @returns {*}
  */
 function applyMerge(existingValue, changes) {
+    if (_.some(changes, change => _.isNull(change))) {
+        return null;
+    }
```
marcaaron, line 1026: This seems wrong to me. If _any_ of the changes in the merge queue array are `null` then we `null` out the resulting value?

lib/Onyx.js
```diff
@@ -1080,6 +1086,11 @@ function merge(key, changes) {
                 // After that we merge the batched changes with the existing value
                 const modifiedData = removeNullObjectValues(applyMerge(existingValue, [batchedChanges]));
 
+                // If the data is null, remove it from cache and store
+                if (_.isNull(modifiedData)) {
+                    return remove(key);
+                }
```
marcaaron, line 1092: I think this can happen sooner. If the `batchedChanges` are `null` then we don't need to concern ourselves with the `existingValue` because we are deleting everything anyway.

lib/Onyx.js
```diff
@@ -1248,19 +1259,34 @@ function mergeCollection(collectionKey, collection) {
...
-                .value();
+            const existingKeysToKeep = [];
+            const keysToRemove = [];
+            const collectionWithoutKeysToRemove = collection;
+            const keyValuePairsForExistingCollection = [];
+            const keyValuePairsForNewCollection = [];
+            _.each(collection, (value, key) => {
+                if (value === null) {
+                    keysToRemove.push(key);
+                    delete collectionWithoutKeysToRemove[key];
```
marcaaron, line 1270: Rather than delete the keys it might be safer to add the ones we are keeping to a new object and avoid mutating the object passed to `mergeCollection()` (I think maybe we are mutating it).

lib/Onyx.js
```diff
@@ -1248,19 +1259,34 @@ function mergeCollection(collectionKey, collection) {
...
+                }
+
+                keyValuePairsForNewCollection.push([key, removeNullObjectValues(value)]);
+            });
 
             const promises = [];
 
+            if (keysToRemove.length > 0) {
+                _.each(keysToRemove, key => cache.drop(key));
+                promises.push(Storage.removeItems(keysToRemove));
```
marcaaron, line 1287: Why not use `Onyx.remove()` here like we did in `Onyx.merge()`?

lib/Onyx.js
```diff
@@ -1248,19 +1259,34 @@ function mergeCollection(collectionKey, collection) {
...
+                    return;
+                }
 
-            const existingKeyCollection = _.pick(collection, existingKeys);
-            const newCollection = _.pick(collection, newKeys);
-            const keyValuePairsForExistingCollection = prepareKeyValuePairsForStorage(existingKeyCollection);
-            const keyValuePairsForNewCollection = prepareKeyValuePairsForStorage(newCollection);
+                if (persistedKeys.includes(key)) {
+                    existingKeysToKeep.push(key);
+                    keyValuePairsForExistingCollection.push([key, removeNullObjectValues(value)]);
```
marcaaron, line 1276: Hmm using `removeNullObjectValues()` seems like a bug here. Wouldn't that mean that when we merge an object and properties that are set to `null` (presumably to remove those top-level values) they will keep existing?

marcaaron (changes requested):

lib/Onyx.js
roryabraham: Yes, if you do this:

```js
const myCallback = () => {};
Onyx.connect({
   key: 'someKey',
   callback: myCallback,
});
```

then `myCallback` will _always_ first be called with `null, undefined` for the value and key. I think this is counter-intuitive and unhelpful in most cases.

The reason we do that is because `withOnyx` waits for each `Onyx.connect` callback to be called at least once before rendering the child component. So even if the data is missing, we call it with `null, undefined`.

My thinking is that by doing this we're exposing an internal implementation detail of `withOnyx` unnecessarily, and in the general case `Onyx.connect` should only execute the callback when there's some data in the store with the given key.

lib/Onyx.js
```diff
@@ -488,7 +488,7 @@ function keysChanged(collectionKey, partialCollection) {
  * @param {*} data
  * @param {Function} [canUpdateSubscriber] only subscribers that pass this truth test will be updated
  */
-function keyChanged(key, data, canUpdateSubscriber) {
+function keyChanged(key, data, canUpdateSubscriber = () => true) {
```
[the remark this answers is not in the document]
roryabraham: I agree this wasn't really needed for this PR, but I disagree that it was more obvious before...

Before:
- param can be function or undefined
- no default for an optional param. in order to understand the default behavior, you need to read the function and see how it's implemented

After:
- param is always a function
- obvious default for an optional param. can be understood just by looking at the function definition

lib/Onyx.js
roryabraham: Just because we're bulk-removing items with `Storage.removeItems` instead of multiple calls to `Storage.removeItem`

lib/Onyx.js
roryabraham: I am not 100% sure about this one, but I _think_ it makes sense? i.e: if I'm trying to merge all these changes into an empty object:

```js
[
    {a: 1},
    {b: 2},
    null,
    {c: 3},
]
```

then what should the result be? This code makes it `null`, but maybe it should actually be applied in order such such that the final result is `{c: 3}`?

lib/Onyx.js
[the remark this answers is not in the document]
dangrous: I'd lean towards this new version as well, as I agree it's more understandable at a glance!

lib/Onyx.js
```diff
@@ -1020,16 +1022,20 @@ function multiSet(data) {
  */
 function applyMerge(existingValue, changes) {
     const lastChange = _.last(changes);
+    if (_.isNull(lastChange)) {
```
dangrous, line 1025: Can you combine this with the if() on new line 1029? Has the same result

lib/Onyx.js
```diff
@@ -1020,16 +1022,20 @@ function multiSet(data) {
...
+        return lastChange;
+    }
 
     if (_.isArray(existingValue) || _.isArray(lastChange)) {
         return lastChange;
     }
 
-    if (_.isObject(existingValue) || _.every(changes, _.isObject)) {
+    const changesToMake = _.filter(changes, change => Boolean(change));
```
dangrous, line 1033: NAB - might be good to leave a comment here explaining the Boolean thing (i.e. I *think* I understand this - that it's filtering out any null/undefineds? - but I'm not 100% sure so others might not be either)

lib/Onyx.js
dangrous: This makes sense to me

lib/Onyx.js
dangrous: Different comment, same line - can we just have `existingKeysToKeep.push(key);` in this if, and then use the `keyValuePairsForExistingCollection.push` on line 1280, instead of doubling it here?

Assuming this is still what we want to do, after @marcaaron's comment

dangrous (changes requested):
some additional comments outside of the main discussions

lib/Onyx.js
marcaaron: Sounds like the main premise in your argument is that "no Onyx.connect() subscribers need this". 

**thought:** passing `null` also provides a way for an Onyx.connect() subscriber to see that it initialized _without a value from storage_. 

Is it helpful to know this? Do we have existing code that depends on this information? I think it's safer to assume we do than don't, but also don't want to audit the 446 usages 😄

lib/Onyx.js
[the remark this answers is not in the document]
marcaaron: it's not a hill I will die on. I will simply change it back to what _I prefer_ in my next Onyx PR :trollface:

lib/Onyx.js
marcaaron: >then what should the result be?

It should be the last value in the array

marcaaron: >If null is anywhere else in the array, only the last item in the array will be applied.

Not anywhere else in the array. Specifically, it should clear out the items that came before it. The rest should merge normally.

marcaaron: >If null is anywhere else in the array, it is ignored.

I think this logic is wrong. It should be clearing the previous values called in the same tick of the event loop. 

If we do this:

```js
Onyx.merge('test', {a: 1});
Onyx.merge('test', null);
Onyx.merge('test', {b: 2});
Onyx.merge('test', {c: 3});
```
Then the expected result would be `{b:2, c: 3}`?

roryabraham:
Thanks for the reviews here everyone. While I think this is a valuable improvement, as far as I know there are no active bugs associated with the current behavior. So I'm going to reprioritize a bit and probably not pick this PR back up until next week. Thanks for understanding 🙇🏼

marcaaron:
@roryabraham still focusing on this one? or should we close it and reopen later?

roryabraham:
Haven't focused on it for a bit, but I think it could be a good one to pick back up. Maybe just a holistic look at how deleting stuff from Onyx works today and how we expect everything to work would be a good start

tgolen:

roryabraham:
Closing this as not planned for now. I believe there are separate efforts to make merging / setting null more consistent

PR #312, opened 2023-08-22
fix callback key param when not waiting for collection callback

hannojg:
### Details

When `waitForCollectionCallback` is undefined or false, the callback will always be called with a key.

### Related Issues

GH_LINK: /

### DIFF
--- a/lib/Onyx.d.ts
+++ b/lib/Onyx.d.ts
@@ -50,7 +50,7 @@ type ConnectOptions<TKey extends OnyxKey> = BaseConnectOptions &
           }
         | {
               key: TKey;
-              callback?: (value: OnyxEntry<KeyValueMapping[TKey]>, key?: TKey) => void;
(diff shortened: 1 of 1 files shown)


fabioh8010 (approved):
I've checked the code and makes sense @hannojg, when `waitForCollectionCallback` is `false` or `undefined` the `key` is always passed in `callback` function.

cc @hayata-suenaga it's not a blocker for starting migration though

fabioh8010:
@hannojg I think we can remove `key` parameter if `waitForCollectionCallback` is `true`, from what I checked here it will never pass the key if this flag is enabled. Could you confirm?

grgia (dismissed):

grgia:
will hold merging on above comment

hannojg:
@fabioh8010 Yes, agree, I changed it! 

I think we can merge @grgia 😊

grgia (approved):

PR #315, opened 2023-08-23
Batch updates second try

Szymon20000:
### The Problem this pull-request solves
One of the main problems with react-native-onyx is that it renders every component in a separate micro-task. Because right now we don't use concurrent rendering every setState not within the same event listener will cause a separate render. Every react render has certain overhead so if it's possible to render 2 state changes together we will only pay for the overhead once. Additionally we introduce additional overhead in render side-effects. For instance in SidebarLinks component we compute which report should be visible. That method is heavy and we run it almost on every render. If we batch 2 setStates together that would save us one such computation. This pull-requests batches all setStates caused by onyx changes that occurs within the same microtask cycle. That saves a ton of time. 

#### Examples of the problem
1) Imagine there are 10 components that are listening to a current report changes. Currently all those components will be notified separately 
2) Imagine we have one component that listens to 4 different onyx props that are changed one by one (SidebarLinks is such component) right now we will render is as many times as we have changes. With this pr we will merge it only once per micotask cycle what will avoid a lot of recomputation

> Note that is change won't be needed anymore once we use concurrent rendering. 

### Details
We use  `unstable_batchedUpdates` method from react (That is currently used in even listeners so it's pretty safe). 
This method triggers single react render for all the setState changes passed to it. 

#### How it works 
Screenshot 2023-08-28 at 14 46 16
For every notifySubscribers or collection subscribers we don't queue the update on micro task queue instead we add that change to pending changes list that is flushed at the end of the current micro task cycle. We do it by setting macro task which is `setTimeout(() => {}, 0)`.
### Related Issues

GH_LINK

### DIFF
--- a/jestSetup.js
+++ b/jestSetup.js
@@ -9,3 +9,7 @@ jest.mock('react-native-quick-sqlite', () => ({
 }));
 
 jest.useRealTimers();
+
+const unstable_batchedUpdates_jest = require('react-test-renderer').unstable_batchedUpdates;
+require('./lib/batch.native').default = unstable_batchedUpdates_jest;
+
--- a/jestSetup.js
+++ b/jestSetup.js
@@ -2,3 +2,5 @@ jest.mock('./lib/storage');
 jest.mock('./lib/storage/NativeStorage', () => require('./lib/storage/__mocks__'));
 jest.mock('./lib/storage/WebStorage', () => require('./lib/storage/__mocks__'));
 jest.mock('./lib/storage/providers/IDBKeyVal', () => require('./lib/storage/__mocks__'));
+
+jest.useRealTimers();
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -803,10 +803,11 @@ function disconnect(connectionID, keyToRemoveFromEvictionBlocklist) {
     delete callbackToStateMapping[connectionID];
 }
 
-// [key, value, canUpdateSubscriber]
-let pendingUpdates = [];
 let scheduledPromise;
-let pendingCollectionUpdates = [];
+
+// [key, value, canUpdateSubscriber]
+let pendingSubscriberUpdates = [];
+let pendingCollectionSubscriberUpdates = [];
 
 /**
  * We are batching together onyx updates. This helps with use cases where we schedule onyx updates after each other.
@@ -815,7 +816,7 @@ let pendingCollectionUpdates = [];
  * cause react to schedule the updates at once instead of after each other. This is mainly a performance optimization.
  * @returns {Promise}
  */
-function scheduleOnyxFlushIfNeeded() {
+function scheduleSubscriberUpdatesFlushIfNeeded() {
     if (scheduledPromise) {
         return scheduledPromise;
     }
@@ -823,10 +824,10 @@ function scheduleOnyxFlushIfNeeded() {
     scheduledPromise = new Promise((resolve) => {
         setTimeout(() => {
             scheduledPromise = null;
-            const updatesCopy = pendingUpdates;
-            pendingUpdates = [];
-            const updatesCollectionCopy = pendingCollectionUpdates;
-            pendingCollectionUpdates = [];
+            const updatesCopy = pendingSubscriberUpdates;
+            pendingSubscriberUpdates = [];
+            const updatesCollectionCopy = pendingCollectionSubscriberUpdates;
+            pendingCollectionSubscriberUpdates = [];
             unstable_batchedUpdates(() => {
                 for (let i = 0; i < updatesCopy.length; ++i) {
                     keyChanged(updatesCopy[i][0], updatesCopy[i][1], updatesCopy[i][2]);
@@ -845,16 +846,16 @@ function scheduleOnyxFlushIfNeeded() {
  * Schedules an update that will be appended to the macro task queue (so it doesn't update the subscribers immediately).
  *
  * @example
- * scheduleNotifySubscribers(key, value, subscriber => subscriber.initWithStoredValues === false)
+ * scheduleSubscriberUpdate(key, value, subscriber => subscriber.initWithStoredValues === false)
  *
  * @param {String} key
  * @param {*} value
  * @param {Function} [canUpdateSubscriber] only subscribers that pass this truth test will be updated
  * @returns {Promise}
  */
-function scheduleNotifySubscribers(key, value, canUpdateSubscriber) {
-    pendingUpdates.push([key, value, canUpdateSubscriber]);
-    return scheduleOnyxFlushIfNeeded();
+function scheduleSubscriberUpdate(key, value, canUpdateSubscriber) {
+    pendingSubscriberUpdates.push([key, value, canUpdateSubscriber]);
+    return scheduleSubscriberUpdatesFlushIfNeeded();
 }
 
 /**
@@ -867,8 +868,8 @@ function scheduleNotifySubscribers(key, value, canUpdateSubscriber) {
  * @returns {Promise}
  */
 function scheduleNotifyCollectionSubscribers(key, value) {
-    pendingCollectionUpdates.push([key, value]);
-    return scheduleOnyxFlushIfNeeded();
+    pendingCollectionSubscriberUpdates.push([key, value]);
+    return scheduleSubscriberUpdatesFlushIfNeeded();
 }
 
 /**
@@ -880,7 +881,7 @@ function scheduleNotifyCollectionSubscribers(key, value) {
  */
 function remove(key) {
     cache.drop(key);
-    scheduleNotifySubscribers(key, null);
+    scheduleSubscriberUpdate(key, null);
     return Storage.removeItem(key);
 }
 
@@ -955,7 +956,7 @@ function broadcastUpdate(key, value, hasChanged, method) {
         cache.addToAccessedKeys(key);
     }
 
-    return scheduleNotifySubscribers(key, value, subscriber => hasChanged || subscriber.initWithStoredValues === false);
+    return scheduleSubscriberUpdate(key, value, subscriber => hasChanged || subscriber.initWithStoredValues === false);
 }
 
 /**
@@ -1044,7 +1045,7 @@ function multiSet(data) {
     const updatePromises = _.map(data, (val, key) => {
         // Update cache and optimistically inform subscribers on the next tick
         cache.set(key, val);
-        return scheduleNotifySubscribers(key, val);
+        return scheduleSubscriberUpdate(key, val);
     });
 
     return Storage.multiSet(keyValuePairs)
@@ -1238,7 +1239,7 @@ function clear(keysToPreserve = []) {
 
             // Notify the subscribers for each key/value group so they can receive the new values
             _.each(keyValuesToResetIndividually, (value, key) => {
-                updatePromises.push(scheduleNotifySubscribers(key, value));
+                updatePromises.push(scheduleSubscriberUpdate(key, value));
             });
             _.each(keyValuesToResetAsCollection, (value, key) => {
                 updatePromises.push(scheduleNotifyCollectionSubscribers(key, value));
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -21,6 +21,7 @@ const METHOD = {
 
 // Key/value store of Onyx key and arrays of values to merge
 const mergeQueue = {};
+const mergeQueuePromise = {};
 
 // Keeps track of the last connectionID that was used so we can keep incrementing it
 let lastConnectionID = 0;
@@ -804,7 +805,7 @@ function disconnect(connectionID, keyToRemoveFromEvictionBlocklist) {
 
 // [key, value, canUpdateSubscriber]
 let pendingUpdates = [];
-let scheduled = false;
+let scheduledPromise;
 let pendingCollectionUpdates = [];
 
 /**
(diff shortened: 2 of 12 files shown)


Li357:
Is this ready for review?

lib/Onyx.js
```diff
@@ -801,22 +803,58 @@ function disconnect(connectionID, keyToRemoveFromEvictionBlocklist) {
...
+    scheduledPromise = new Promise((resolve) => {
+        setTimeout(() => {
+            scheduledPromise = null;
+            const updatesCopy = pendingUpdates;
+            pendingUpdates = [];
+            const updatesCollectionCopy = pendingCollectionUpdates;
+            pendingCollectionUpdates = [];
+            unstable_batchedUpdates(() => {
+                for (let i = 0; i < updatesCopy.length; ++i) {
+                    keyChanged(updatesCopy[i][0], updatesCopy[i][1], updatesCopy[i][2]);
```
Julesssss, line 832: I couldn't find anything that confirms that the `pendingUpdate|updatesCopy` is has a 3rd child. Presumably, this is guaranteed as that's how our Onyx updates are written?

Julesssss (approved):
While this looks good, it would be great to test the component renders to prove the updates are now batched.

jestSetup.js
```diff
@@ -2,7 +2,10 @@ jest.mock('./lib/storage');
 jest.mock('./lib/storage/NativeStorage', () => require('./lib/storage/__mocks__'));
 jest.mock('./lib/storage/WebStorage', () => require('./lib/storage/__mocks__'));
 jest.mock('./lib/storage/providers/IDBKeyVal', () => require('./lib/storage/__mocks__'));
+
 jest.mock('react-native-device-info', () => ({getFreeDiskStorage: () => {}}));
 jest.mock('react-native-quick-sqlite', () => ({
     open: () => ({execute: () => {}}),
 }));
+
+jest.useRealTimers();
```
tgolen, line 11: Why is this added here? Do you mind adding a code comment to explain?

lib/Onyx.js
```diff
@@ -801,22 +803,58 @@ function disconnect(connectionID, keyToRemoveFromEvictionBlocklist) {
     delete callbackToStateMapping[connectionID];
 }
 
+// [key, value, canUpdateSubscriber]
+let pendingUpdates = [];
+let scheduledPromise;
+let pendingCollectionUpdates = [];
```
tgolen, line 809: Could you also please add a doc comment for this (like you did for `pendingUpdates`)?

lib/Onyx.js
```diff
@@ -801,22 +803,58 @@ function disconnect(connectionID, keyToRemoveFromEvictionBlocklist) {
...
+ * cause react to schedule the updates at once instead of after each other. This is mainly a performance optimization.
+ * @returns {Promise}
+ */
+function scheduleOnyxFlushIfNeeded() {
+    if (scheduledPromise) {
+        return scheduledPromise;
+    }
+
+    scheduledPromise = new Promise((resolve) => {
+        setTimeout(() => {
```
tgolen, line 824: Please add a code comment explaining why it's necessary to have a `setTimout(0)` here.

lib/Onyx.js
```diff
@@ -20,6 +21,7 @@ const METHOD = {
 
 // Key/value store of Onyx key and arrays of values to merge
 const mergeQueue = {};
+const mergeQueuePromise = {};
```
tgolen, line 24: I followed most of the changes in this PR except for where it came to `mergeQueuePromise`. Could you add some code comments to give a general idea around what it's for and how it's used?

tgolen (changes requested):
Is there any way to write some unit tests to verify that the updates were batched as expected? Something like... have a mocked function that is triggered on render, and updating multiple properties that the component is listening to triggers the render callback the right number of times.

This is an example test of what I'm thinking of: https://github.com/Expensify/react-native-onyx/blob/ad5533e199b348713f7ff9dd4b75894af7a1a597/tests/unit/withOnyxTest.js#L59-L77

lib/Onyx.js
```diff
@@ -966,15 +1006,16 @@ function set(key, value) {
     const hasChanged = cache.hasValueChanged(key, valueWithNullRemoved);
 
     // This approach prioritizes fast UI changes without waiting for data to be stored in device storage.
-    broadcastUpdate(key, valueWithNullRemoved, hasChanged, 'set');
+    const updatePromise = broadcastUpdate(key, valueWithNullRemoved, hasChanged, 'set');
 
     // If the value has not changed, calling Storage.setItem() would be redundant and a waste of performance, so return early instead.
     if (!hasChanged) {
-        return Promise.resolve();
+        return updatePromise;
```
marcaaron, line 1013: NAB, I am a _little_ skeptical of this change mainly because it is now harder to reason about when this promise will resolve. 

Before we can easily see it will either resolve right away or after the value is saved to storage. Now it's less clear... it's still both of those, but maybe also however long it takes the queue to flush? Ultimately, I don't know if this concern is valid or if it makes any difference to the existing code we have.

lib/Onyx.js
```diff
@@ -1091,20 +1133,22 @@ function merge(key, changes) {
...
-                    return Promise.resolve();
+                    return updatePromise;
                 }
 
-                return Storage.mergeItem(key, batchedChanges, modifiedData);
+                return Storage.mergeItem(key, batchedChanges, modifiedData)
+                    .then(() => updatePromise);
             } catch (error) {
                 Logger.logAlert(`An error occurred while applying merge for key: ${key}, Error: ${error}`);
+                return Promise.resolve();
```
marcaaron, line 1147: well, this change actually will have some interesting consequences (I _think_ probably good ones) when a storage error happens.

e.g. if something breaks like not able to save the merge to storage because you are out of space then we will resolve the promise (which would allow anything depending on it to complete to continue as well - which sounds good).

But, if I am not mistaken - I think in this case... we can and should still wait for the `updatePromise` to resolve? 🤔 

Unless that throws an error too. And in that case we should return the resolved promise 😄

marcaaron (changes requested):

marcaaron:
>One of the main problems with react-native-onyx is that it renders every component in a seperate promise

I am a little unsure about what “renders every component in a separate promise” means exactly. Could you expand on that a bit?

>right now we don't use concurrent rendering every setState not within the same event listener will cause a separate render

Is this referring to what happens in the `notifySubscribersOnNextTick()` method?

https://github.com/Expensify/react-native-onyx/blob/7f23ec8a279c93616b9e2d43ccb1574ccd6be9c9/lib/Onyx.js#L805-L820

One thought I had while reading the comment is if we still really _need_ that promise. It seems unnecessary and I wonder if it really is dangerous to remove - maybe it just needs to be a “breaking change” and we fix whatever might depend on this logic in App.

Though, thinking more on that - we will always have the `get()` in `Onyx.merge()` that would create a promise. So, this observation doesn't affect the solution you are proposing much.

jestSetup.js
Szymon20000: Hanno added that because otherwise setTimeout is not executed unless we call jest.advanceTime(). So to avoid calling that everywhere we used real times.

lib/Onyx.js
Szymon20000: I added a comment explaining when the batch will be flushed. Explaining how timers and batching works in react native is not possible in a single comment so I simplified it a bit.

lib/Onyx.js
Szymon20000: This change is mostly for adjusting tests from what I remember. 
Looks like we use it in tests for making sure all the pending merges has been applied.

lib/Onyx.js
```diff
@@ -801,22 +803,58 @@ function disconnect(connectionID, keyToRemoveFromEvictionBlocklist) {
...
+let pendingUpdates = [];
+let scheduledPromise;
+let pendingCollectionUpdates = [];
+
+/**
+ * We are batching together onyx updates. This helps with use cases where we schedule onyx updates after each other.
+ * This happens for example in the Onyx.update function, where we process API responses that might contain a lot of
+ * update operations. Instead of calling the subscribers for each update operation, we batch them together which will
+ * cause react to schedule the updates at once instead of after each other. This is mainly a performance optimization.
+ * @returns {Promise}
```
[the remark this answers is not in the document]
Szymon20000: Hanno already renamed all variables and functions so they use "subscriberUpdates".

lib/Onyx.js
Szymon20000: Screenshot 2023-09-05 at 09 39 03
The keyChanged method has 3 arguments so we only delay the execution.

lib/Onyx.js
```diff
@@ -801,22 +803,58 @@ function disconnect(connectionID, keyToRemoveFromEvictionBlocklist) {
...
+        }, 0);
+    });
+    return scheduledPromise;
+}
+
 /**
- * This method mostly exists for historical reasons as this library was initially designed without a memory cache and one was added later.
- * For this reason, Onyx works more similar to what you might expect from a native AsyncStorage with reads, writes, etc all becoming
- * available async. Since we have code in our main applications that might expect things to work this way it's not safe to change this
- * behavior just yet.
```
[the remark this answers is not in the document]
Szymon20000: Changed!

lib/Onyx.js
Szymon20000: I don't see the difference to be honest :) Looks like Promise.resolve was also returned in this case just right below catch block.

Szymon20000:
@marcaaron As long as we don't use fabric we are not able to use concurrent rendering. As a result whenever you call setState on a component it will right away cause a synchronous render (apart from few edge cases like events or active render). Because onyx notifies withOnyx components in separate micro-tasks it causes all of them to start synchronous render.  By “renders every component in a separate promise” I meant separate micro tasks :)

lib/Onyx.js
```diff
@@ -801,22 +803,58 @@ function disconnect(connectionID, keyToRemoveFromEvictionBlocklist) {
     delete callbackToStateMapping[connectionID];
 }
 
+// [key, value, canUpdateSubscriber]
```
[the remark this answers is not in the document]
Szymon20000: It will have the same types as keyChange and keysChanged method and will be typed once we rewrite the code to typescript. Can we leave it for now?

Szymon20000:
Found how it can be done

tests/unit/withOnyxTest.js
```diff
@@ -76,6 +77,34 @@ describe('withOnyx', () => {
...
+        render(<TestComponentWithOnyx onRender={onRender} />);
+
+        return waitForPromisesToResolve()
+            .then(() => {
+                Onyx.mergeCollection(ONYX_KEYS.COLLECTION.TEST_KEY, {test_1: {ID: 123}});
+                Onyx.merge(ONYX_KEYS.SIMPLE_KEY, 'string');
+                return Onyx.merge(ONYX_KEYS.SIMPLE_KEY_2, 'string2');
+            })
+            .then(() => {
+                expect(onRender).toHaveBeenCalledTimes(2);
```
tgolen, line 104: Could you please add a comment to explain why it's expected to be called exactly two times?

Szymon20000: Sure, just a sec

Szymon20000: done

lib/Onyx.js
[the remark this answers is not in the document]
marcaaron: Cool thanks!

lib/Onyx.js
```diff
@@ -801,22 +803,65 @@ function disconnect(connectionID, keyToRemoveFromEvictionBlocklist) {
...
+ * cause react to schedule the updates at once instead of after each other. This is mainly a performance optimization.
+ * @returns {Promise}
+ */
+function scheduleSubscriberUpdatesFlushIfNeeded() {
+    if (scheduledPromise) {
+        return scheduledPromise;
+    }
+
+    scheduledPromise = new Promise((resolve) => {
+        /* We use (setTimout, 0) here which should be called once native module calles are flushed (usually at the end of the frame)
```
marcaaron, line 827:
```suggestion
        /* We use (setTimeout, 0) here which should be called once native module calls are flushed (usually at the end of the frame)
```

lib/Onyx.js
```diff
@@ -801,22 +803,65 @@ function disconnect(connectionID, keyToRemoveFromEvictionBlocklist) {
...
+         * We may investigate if (setTimout, 1) (which in React Native is equal to requestAnimationFrame) works even better
```
marcaaron, line 828:
```suggestion
         * We may investigate if (setTimeout, 1) (which in React Native is equal to requestAnimationFrame) works even better
```

lib/Onyx.js
marcaaron: Oh duh, ok I missed that line I guess 😄

marcaaron (changes requested):
Just small typos to fix now. I will resolve my NABs.

lib/Onyx.js
marcaaron: we will find out I guess 😄

Szymon20000:
The App tests are failing with this pr. I will let you know when they are fixed. 
I just realised that we only want to batch withOnyx subscribers as regular ones should be notified right away.

Szymon20000:
@tgolen @marcaaron Are there any places in the app when we trigger Onyx update in Onyx.connect subscriber? 
If that's the case then I think we shouldn't batch them. 
I can image that we have places like that when once onyx.update triggers another one. 
So for examples. update A -> update B -> update C would require 3 batches now. 
If you think we may have such case in the app I create a pr that fixes that problem. https://github.com/margelo/react-native-onyx/pull/1

tgolen:
I just looked at all the usages of `Onyx.update` and there are none that happen inside a `connect()` subscriber.

marcaaron:
>I just realised that we only want to batch withOnyx subscribers as regular ones should be notified right away

Oh, right, because they do not call `setState()`. Makes sense. 

>Are there any places in the app when we trigger Onyx update in Onyx.connect subscriber?
>If that's the case then I think we shouldn't batch them.

Sorry, I did not follow the reasoning. Is there some kind of critical problem that we will see if we batch those updates?

Szymon20000:
Ah if we don't change state in onyx connect then probably it doesn't make any difference. I did it because I thought you may have some reactive code that is triggered once we merge something.

Szymon20000:
Screenshot 2023-09-07 at 16 03 19
@tgolen is may be the case with network calls

tgolen:
Ah, you're right. I'm on a branch where we've refactored that specific code to be a little different and it's not clear anymore that the `Onyx.update()` is inside the `connect()`.

https://github.com/Expensify/App/blob/5373041c5776fd8a51ea3646710ed9c7ccc50658/src/libs/Network/SequentialQueue.js#L126

Szymon20000:
I was able to fix most of failing tests in the app repo. There are 6 that left. 
I think we need the pr for only batching withOnyx subscribers as networking flushing would be delayed otherwise.

tgolen (approved):

marcaaron (approved):

tgolen (approved):

mountiny:
To get this going ahead I will merge this now given there are two approvals here

marcaaron:
Should we make an announcement not to update the Onyx version until that testing is done? These feel like pretty significant changes.

PR #316, opened 2023-08-24
Use collection key value map instead of recomputing everytime

hannojg:
### Details

During a performance audit of a large account we noticed, that the `isKeyMatch` time was taking a lot of time. After further investigation we found that this code is problematic:

https://github.com/Expensify/react-native-onyx/blob/680e3cb10dcfeba9cf40f7f32fd0aa1bdd8b01e0/lib/Onyx.js#L143-L145

We recompute an array again and again and then use includes on an array to check for a match. Do this thousand of times and it becomes slow. Optimising this code reduced the startup time by ~600ms (on a powerful emulator, effects might be stronger on physical device)

### Related Issues

GH_LINK: https://expensify.slack.com/archives/C035J5C9FAP/p1692808883217069

### DIFF
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -27,8 +27,8 @@ let lastConnectionID = 0;
 // Holds a mapping of all the react components that want their state subscribed to a store key
 const callbackToStateMapping = {};
 
-// Stores all of the keys that Onyx can use. Must be defined in init().
-let onyxKeys = {};
+// Keeps a copy of the values of the onyx collection keys as a a map for faster lookups
+let onyxCollectionKeyMap = {};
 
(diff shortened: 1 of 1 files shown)


mountiny (approved):

mountiny (approved):
Clean simple change, thanks!

lib/Onyx.js
```diff
@@ -141,7 +141,7 @@ function getAllKeys() {
  * @returns {Boolean}
  */
 function isCollectionKey(key) {
-    return _.contains(_.values(onyxKeys.COLLECTION), key);
+    return onyxCollectionKeyMap.has(key);
```
tgolen, line 144: This is causing the jest tests to fail: https://github.com/Expensify/App/actions/runs/6027173662/job/16353221477

PR #329, opened 2023-09-05
Add initialValue, delayUpdates to allow components to eagerly render.

ospfranco:
### Details

Margelo is currently with improvements to the ReportScreen speed. After thorough analysis a big bottleneck came from react-native-onyx itself. Components with a lot of connections can cause frequent re-renders or delay the rendering of the component itself until all data has been fetched/hydrated.

We had to implement several changes in order to get the rendering speed down:
- Add `initialValue` to a key. Allows to pass a initial value while the data is being fetched, thus allowing the component to mount/render ASAP.
- Added `delayUpdates` to a withOnyx connection. Allows to delay setting updates/state of the `withOnyx` hook to allow the component to mount first and then update in a single batch at a different time. The child component gets passed a `allowUpdates` function that it can use to tell the `withOnyx` HOC to start applying the batched updates.
- [Another PR](https://github.com/Expensify/react-native-onyx/pull/330) has been created that brings back a reverted improvement made by margelo on collection keys. This also has a big impact on this performance improvements.

### Related Issues

Part of https://github.com/Expensify/App/issues/26087


ospfranco:

lib/Onyx.js
```diff
@@ -1,21 +1,22 @@
...
-import {deepEqual} from 'fast-equals';
-import _ from 'underscore';
-import * as Logger from './Logger';
-import cache from './OnyxCache';
-import * as Str from './Str';
-import createDeferredTask from './createDeferredTask';
-import fastMerge from './fastMerge';
-import * as PerformanceUtils from './metrics/PerformanceUtils';
-import Storage from './storage';
+import { deepEqual } from "fast-equals";
```
yuwenmemon, line 2: A lot of updates here that deviate from our style guide. Can we re-update to get these back to that standard?

lib/Onyx.js
```diff
@@ -54,7 +55,8 @@ const deferredInitTask = createDeferredTask();
  * @param {Object} [withOnyxInstanceState]
  * @returns {Mixed}
  */
-const getSubsetOfData = (sourceData, selector, withOnyxInstanceState) => selector(sourceData, withOnyxInstanceState);
+const getSubsetOfData = (sourceData, selector, withOnyxInstanceState) =>
+    selector(sourceData, withOnyxInstanceState);
```
yuwenmemon, line 59: I also question if updates like this adding new lines for lines that weren't already that long are necessary. They're adding a lot of noise to this PR IMO.

lib/utils.js
```diff
@@ -0,0 +1,3 @@
+export function areObjectsEmpty(a, b) {
+  return typeof data === 'object' && typeof previousData === 'object' && Object.values(a).length === 0 && Object.values(b).length === 0
```
yuwenmemon, line 2: Where do `data` and `previous` data come from and what is this method achieving exactly?

yuwenmemon (changes requested):
Can we undo the whitespace and quotation changes so this PR is a lot less noisy?

also cc @marcaaron and @tgolen who might be interested in having a peek at this.

marcaaron:
Sorry I have been reviewing a lot of Onyx PRs and can't prioritize a review today as these changes looks quite involved. Also +1 to @yuwenmemon's comment about formatting. It looks like prettier ran with the wrong settings maybe.

First, thanks for the work @ospfranco. Excited to discuss everything with you. I will leave some questions and suggestions regarding the PR description as there are some things I'm not really clear on and wouldn't really be able to review the changes without understanding them first.

>After thorough analysis a big bottleneck came from react-native-onyx itself

Where exactly was the bottleneck and what did the analysis show? Please add this to the description.

>Components with a lot of connections can cause frequent re-renders 

Yes, if you have a lot of subscribers then they will re-render when there is a change. Is there some specific example of this that showcases the problem better? Did we discuss any alternatives?

>or delay the rendering of the component itself until all data has been fetched/hydrated.

@hannojg's PR to use the cached values was helping with this part. Is it not enough? Where does this delay happen? What does it feel like?

>Add initialValue to a key. Allows to pass a initial value while the data is being fetched, thus allowing the component to mount/render ASAP.

Any discussion to link? I am not sure if it's clear that we want or need anything like an "initial value" to be added to the Onyx API.

>Added delayUpdates to a withOnyx connection. Allows to delay setting updates/state of the withOnyx hook to allow the component to mount first and then update in a single batch at a different time. The child component gets passed a allowUpdates function that it can use to tell the withOnyx HOC to start applying the batched updates.

Is the hook something we already have or are adding here? I didn't see it in the PR but only looked briefly. Overall, this idea sounds kind of neat, but also sounds like added complexity. Also, this might be semantics, but I think `delayUpdates` is ambiguous as the rendering of the component is currently "delayed" (we wait for all the values to be retrieved). Maybe we can find clearer language to use around that - I'll think about it.

>https://github.com/Expensify/react-native-onyx/pull/330 has been created that brings back a reverted improvement made by margelo on collection keys. This also has a big impact on this performance improvements.

Is this related to the current PR?

lib/Onyx.js
ospfranco: Ah, I just left prettier do it's thing, what is the configuration I should use?

ospfranco: I cleaned all the formatting changes, should be a lot let noisy now

lib/Onyx.js
ospfranco: I cleaned all the formatting changes, should be a lot let noisy now

lib/utils.js
ospfranco: You are right, this was a copy and paste mistake from my other PR, the correct values are `a` and `b`

ospfranco:
@marcaaron


> > After thorough analysis a big bottleneck came from react-native-onyx itself

> Where exactly was the bottleneck and what did the analysis show? Please add this to the description.

Components with a lot of connections are too slow, e.g. ReportScreen, this container grew and is now connected to a lot of entities. As onyx updates the component state via HOC, more updates are fetched from the DB or API responses come back, this causes the component to be constantly re-render, sometimes it delays the initial rendering and almost always causes subsequent re-renders. In the case of the ReportScreen being this a container on top of the tree, each re-render is very expensive and delays everything, from screen transition animation, to the list of actions rendering.

> > Components with a lot of connections can cause frequent re-renders

> Yes, if you have a lot of subscribers then they will re-render when there is a change. Is there some specific example of this that showcases the problem better? Did we discuss any alternatives?

All the changes were done when testing the ReportScreen and that is the best example. You can see on [this line on the PR on expensify](https://github.com/Expensify/App/pull/26772/files#diff-dd4bfa50713397aca8cb40145317936ab0affe19abc2a9c24d6c8849ed75dc9bR312), one solution we had to come up to keep the animation running smoothly is to delay the update of the component, after is has mounted and has been layed out, then allow onyx to start the many queued state changes of all the data is has pending.

The only other solution we considered might have an impact was the batching of updates, as it avoid frequent re-renders by batching the updates, but that PR/discussion is still pending.

> > or delay the rendering of the component itself until all data has been fetched/hydrated.

> @hannojg's PR to use the cached values was helping with this part. Is it not enough? Where does this delay happen? What does it feel like?

No, it is not enough, once values are in cache it surely does help, but on first mount/render and on a large-subtree with lots connections it causes delays in many components, in total delaying the transition/first render. By adding the initialValue prop, we can allow the component to render as fast as possible, while Onyx is fetching stuff from DB. The state might not be always consistent with the data, but it is rather not an issue for most components. Again, the best example is the entire ReportScreen rendering that hang-ups while the many re-renders ocurr.

> > Add initialValue to a key. Allows to pass a initial value while the data is being fetched, thus allowing the component to mount/render ASAP.

> Any discussion to link? I am not sure if it's clear that we want or need anything like an "initial value" to be added to the Onyx API.

No, all the discussions were internal. Here is an example of a Performance trace in flipper:

CleanShot 2023-09-07 at 09 44 51

It's hard to explain, but basically, the first delay, plus frequent sub-sequent state updates simply do not allow React to flush and therefore delay everything until everything settles down.

> > Added delayUpdates to a withOnyx connection. Allows to delay setting updates/state of the withOnyx hook to allow the component to mount first and then update in a single batch at a different time. The child component gets passed a allowUpdates function that it can use to tell the withOnyx HOC to start applying the batched updates.

> Is the hook something we already have or are adding here? I didn't see it in the PR but only looked briefly. Overall, this idea sounds kind of neat, but also sounds like added complexity. Also, this might be semantics, but I think delayUpdates is ambiguous as the rendering of the component is currently "delayed" (we wait for all the values to be retrieved). Maybe we can find clearer language to use around that - I'll think about it.

It is something added. I cleaned the PR, it should be easier to spot now. You can remove it, but then you will always have the same problem as components grow over time. Fetching from DB/API will not allow them to mount on first render and screen transitions/mounting will be delayed until onyx reaches a stable state.

> > https://github.com/Expensify/react-native-onyx/pull/330 has been created that brings back a reverted improvement made by margelo on collection keys. This also has a big impact on this performance improvements.

> Is this related to the current PR?

Yes, it does have a big impact on screen transitions.

Just to further make the point, here is the ReportScreen transition, before and after hour changes:

Most of the performance gains came from adding initialValues to many smaller components and the delayUpdate of the big container:


https://github.com/Expensify/react-native-onyx/assets/1634213/c4e65122-524b-43fb-889d-b21b5dc7e982


https://github.com/Expensify/react-native-onyx/assets/1634213/7f8be845-9683-4179-91ba-34733cf15f1e

tgolen:
Thanks for all that information.

> Here is an example of a Performance trace in flipper:

Can you please add before/after traces so it can be directly compared? Another thing that would be useful is specific timings for what "delayed" means. Is it seconds? milliseconds? microseconds?

When we do anything related to performance there should always be measurements of before and after to quantify and demonstrate the improvement.

> Most of the performance gains came from adding initialValues to many smaller components

Could you please link to a PR (it's fine if it's only a draft) of App with these changes so I can see what the `initialValues` change actually looks like?

lib/utils.js
```diff
@@ -0,0 +1,12 @@
+import * as _ from 'underscore';
+
+function areObjectsEmpty(a, b) {
+    return (
+        typeof a === 'object'
+        && typeof b === 'object'
+        && _.values(a).length === 0
```
tgolen, line 7: What's the difference between this and using `_. isEmpty_()` or `_.size() === 0`?

lib/utils.js
```diff
@@ -0,0 +1,12 @@
+import * as _ from 'underscore';
+
+function areObjectsEmpty(a, b) {
```
tgolen, line 3: I think this implementation is too specific in that it requires two objects. I think there are already existing methods to check if a single object is empty, and then it's just as easy as doing `if (_.isEmpty(a) && _.isEmpty(b))` and then it makes this utility method go away.

lib/withOnyx.js
```diff
@@ -196,7 +231,19 @@ export default function (mapOnyxToState) {
...
+
+                this.pendingSetStates.forEach((modifier) => {
+                    this.setState(modifier);
+                });
+                this.pendingSetStates = [];
+            }
+
             render() {
+                // Remove any null values so that React replaces them with default props
+                const propsToPass = _.omit(this.props, value => _.isNull(value));
```
tgolen, line 245: Does this work?
```suggestion
                const propsToPass = _.omit(this.props, _.isNullvalue);
```

tgolen (changes requested):

lib/withOnyx.js
```diff
@@ -28,37 +29,51 @@ export default function (mapOnyxToState) {
...
-                    if (value !== undefined && !Onyx.hasPendingMergeForKey(key)) {
-                        // eslint-disable-next-line no-param-reassign
-                        resultObj[propertyName] = value;
-                    }
+                const cachedState = _.reduce(
+                    mapOnyxToState,
+                    (resultObj, mapping, propertyName) => {
+                        const key = Str.result(mapping.key, props);
+                        let value = Onyx.tryGetCachedValue(key, mapping);
+                        if (!value && mapping.initialValue) {
```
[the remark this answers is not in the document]
marcaaron: reminder for @ospfranco (because he is new to the repo) that these docs are generated:

https://github.com/Expensify/react-native-onyx/blob/bf3ffa5f934e0197590997f945f40ab6ddf704e1/package.json#L40

marcaaron:
Thanks for the writeup @ospfranco !

This might be a naive interpretation of the problem, but we can reduce it to:

_it takes time to rehydrate state from storage and this blocks rendering_

It looks like these delays are the cost of loading from storage into memory lazily. So, one thought that comes to mind is whether we can solve this by prioritizing the work to hydrate the state earlier?

If the app was burdened with less data requirements and could rehydrate the critical state faster - would we still need this solution?

I am curious about proceeding here without an answer to that question though not trying to block on this if @yuwenmemon and @tgolen feel more confident than me. 

From what I can tell the interactions will be faster, but the UX also janky because we load an initial value -> stored value later -> maybe API stuff later. Probably we will want to smooth that out with a loading state of some kind, but having faster interactions is a net positive change.

lib/Onyx.js
```diff
@@ -408,9 +409,13 @@ function keysChanged(collectionKey, partialCollection) {
...
                 if (subscriber.selector) {
-                    subscriber.withOnyxInstance.setState((prevState) => {
+                    subscriber.withOnyxInstance.setStateProxy((prevState) => {
                         const previousData = prevState[subscriber.statePropertyName];
-                        const newData = reduceCollectionWithSelector(cachedCollection, subscriber.selector, subscriber.withOnyxInstance.state);
+                        const newData = reduceCollectionWithSelector(
+                            cachedCollection,
+                            subscriber.selector,
+                            subscriber.withOnyxInstance.state,
+                        );
```
yuwenmemon, line 418: This can be reverted back to one line

lib/Onyx.js
```diff
@@ -422,8 +427,10 @@ function keysChanged(collectionKey, partialCollection) {
                     continue;
                 }
 
-                subscriber.withOnyxInstance.setState((prevState) => {
-                    const finalCollection = _.clone(prevState[subscriber.statePropertyName] || {});
+                subscriber.withOnyxInstance.setStateProxy((prevState) => {
+                    const finalCollection = _.clone(
+                        prevState[subscriber.statePropertyName] || {},
+                    );
```
yuwenmemon, line 433: This can be reverted back to one line

lib/Onyx.js
```diff
@@ -451,11 +458,21 @@ function keysChanged(collectionKey, partialCollection) {
...
                 if (subscriber.selector) {
-                    subscriber.withOnyxInstance.setState((prevState) => {
+                    subscriber.withOnyxInstance.setStateProxy((prevState) => {
                         const prevData = prevState[subscriber.statePropertyName];
-                        const newData = getSubsetOfData(cachedCollection[subscriber.key], subscriber.selector, subscriber.withOnyxInstance.state);
+                        const newData = getSubsetOfData(
+                            cachedCollection[subscriber.key],
+                            subscriber.selector,
+                            subscriber.withOnyxInstance.state,
+                        );
```
yuwenmemon, line 467: This can be reverted back to one line

lib/Onyx.js
```diff
@@ -451,11 +458,21 @@ function keysChanged(collectionKey, partialCollection) {
...
                         if (!deepEqual(prevData, newData)) {
-                            PerformanceUtils.logSetStateCall(subscriber, prevData, newData, 'keysChanged', collectionKey);
+                            PerformanceUtils.logSetStateCall(
+                                subscriber,
+                                prevData,
+                                newData,
+                                'keysChanged',
+                                collectionKey,
+                            );
```
yuwenmemon, line 475: This can be reverted back to one line

lib/Onyx.js
```diff
@@ -532,10 +552,14 @@ function keyChanged(key, data, canUpdateSubscriber) {
...
-                    subscriber.withOnyxInstance.setState((prevState) => {
+                    subscriber.withOnyxInstance.setStateProxy((prevState) => {
                         const prevData = prevState[subscriber.statePropertyName];
                         const newData = {
-                            [key]: getSubsetOfData(data, subscriber.selector, subscriber.withOnyxInstance.state),
+                            [key]: getSubsetOfData(
+                                data,
+                                subscriber.selector,
+                                subscriber.withOnyxInstance.state,
+                            ),
```
yuwenmemon, line 562: This can be reverted back to one line

lib/Onyx.js
```diff
@@ -569,9 +593,17 @@ function keyChanged(key, data, canUpdateSubscriber) {
...
+                    const previousValue = getSubsetOfData(
+                        prevState[subscriber.statePropertyName],
+                        subscriber.selector,
+                        subscriber.withOnyxInstance.state,
+                    );
+                    const newValue = getSubsetOfData(
+                        data,
+                        subscriber.selector,
+                        subscriber.withOnyxInstance.state,
+                    );
```
yuwenmemon, line 606: These can be reverted back to one line

lib/withOnyx.js
```diff
@@ -84,7 +99,10 @@ export default function (mapOnyxToState) {
                     const previousKey = Str.result(mapping.key, prevProps);
                     const newKey = Str.result(mapping.key, this.props);
                     if (previousKey !== newKey) {
-                        Onyx.disconnect(this.activeConnectionIDs[previousKey], previousKey);
+                        Onyx.disconnect(
+                            this.activeConnectionIDs[previousKey],
+                            previousKey,
+                        );
```
yuwenmemon, line 105: This can be reverted back to one line

lib/withOnyx.js
```diff
@@ -116,31 +142,35 @@ export default function (mapOnyxToState) {
                 // already loaded the values from cache. Thus we have to check whether the value has changed
                 // before we set the state to prevent unnecessary renders.
                 const prevValue = this.state[statePropertyName];
-                if (!this.state.loading && prevValue === val) {
+                if (
+                    !this.state.loading
+                    && (prevValue === val || utils.areObjectsEmpty(prevValue, val))
+                ) {
```
yuwenmemon, line 148: This can be reverted back to one line

lib/withOnyx.js
```diff
@@ -116,31 +142,35 @@ export default function (mapOnyxToState) {
...
                     return;
                 }
 
                 this.tempState[statePropertyName] = val;
 
                 // All state keys should exist and at least have a value of null
-                if (_.some(requiredKeysForInit, key => _.isUndefined(this.tempState[key]))) {
+                if (
+                    _.some(requiredKeysForInit, key => _.isUndefined(this.tempState[key]))
+                ) {
```
yuwenmemon, line 162: This can be reverted back to one line

lib/withOnyx.js
```diff
@@ -162,11 +192,16 @@ export default function (mapOnyxToState) {
                     const key = Str.result(mapping.key, this.props);
 
                     if (!Onyx.isSafeEvictionKey(key)) {
-                        throw new Error(`canEvict can't be used on key '${key}'. This key must explicitly be flagged as safe for removal by adding it to Onyx.init({safeEvictionKeys: []}).`);
+                        throw new Error(
+                            `canEvict can't be used on key '${key}'. This key must explicitly be flagged as safe for removal by adding it to Onyx.init({safeEvictionKeys: []}).`,
+                        );
```
yuwenmemon, line 197: This can be reverted back to one line

lib/withOnyx.js
```diff
@@ -162,11 +192,16 @@ export default function (mapOnyxToState) {
...
                     }
 
                     if (canEvict) {
-                        Onyx.removeFromEvictionBlockList(key, mapping.connectionID);
+                        Onyx.removeFromEvictionBlockList(
+                            key,
+                            mapping.connectionID,
+                        );
```
yuwenmemon, line 204: This can be reverted back to one line

yuwenmemon (changes requested):
A few more notes to reduce noise even more

lib/utils.js
ospfranco: Initial implementation was Object.values(a).length, linter barked at me I should use underscore, no other reason. I will change it to `_.isEmpty`

lib/utils.js
```diff
@@ -0,0 +1,12 @@
+import * as _ from 'underscore';
```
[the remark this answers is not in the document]
ospfranco: Yeah, makes sense.

lib/utils.js
ospfranco: `_.isEmpty(8)` is giving me `true`. Because we don't know what Onyx is sending to the connection, we only really want to do this check to avoid unnecessary re-renders on objects (and maybe arrays? I didn't see any array while debugging)

lib/withOnyx.js
ospfranco: only passing _.isNull should work

lib/withOnyx.js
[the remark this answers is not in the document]
ospfranco: Will do! thanks!

lib/withOnyx.js
```diff
@@ -28,37 +29,51 @@ export default function (mapOnyxToState) {
...
-                        // eslint-disable-next-line no-param-reassign
-                        resultObj[propertyName] = value;
-                    }
+                const cachedState = _.reduce(
+                    mapOnyxToState,
+                    (resultObj, mapping, propertyName) => {
+                        const key = Str.result(mapping.key, props);
+                        let value = Onyx.tryGetCachedValue(key, mapping);
+                        if (!value && mapping.initialValue) {
+                            value = mapping.initialValue;
```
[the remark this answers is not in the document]
ospfranco: That's a good idea, I'm not sure, but I will take a look

ospfranco:
> 

@marcaaron pre-loading/pre-fetching is a solution, but it is not always applicable or practical. For example, action reactions are not really linked to anything and it is not possible to know before hand which ones can be pre-loaded. Other entities might require loading the entire db into memory to be really effective. In any case, this is rather a techinique for further optimization and the current implemented solution allows for a more generic solution, which is allowing the components to render while state is being fetched.

ospfranco:
@tgolen 

> Thanks for all that information.
> 
> > Here is an example of a Performance trace in flipper:
> 
> Can you please add before/after traces so it can be directly compared? Another thing that would be useful is specific timings for what "delayed" means. Is it seconds? milliseconds? microseconds?

Here is the trace with all the performance optimizations:

CleanShot 2023-09-08 at 18 15 00

Everything is in milliseconds

> When we do anything related to performance there should always be measurements of before and after to quantify and demonstrate the improvement.
> 
> > Most of the performance gains came from adding initialValues to many smaller components
> 
> Could you please link to a PR (it's fine if it's only a draft) of App with these changes so I can see what the `initialValues` change actually looks like?

Here is the related PR

https://github.com/Expensify/App/pull/26772

marcaaron:
Thanks for continuing the discussion with me @ospfranco! 

>pre-loading/pre-fetching is a solution, but it is not always applicable or practical. 
>For example, action reactions are not really linked to anything and it is not possible to know before hand which ones can be pre-loaded. 

The reactions are something we could prefetch. They are linked to the actions via `reportActionID`, the actions are linked to the report via `reportID`, the reports are shown in the sidebar (and sometimes other places). We can prefetch the needed data on layout, scroll, hover, etc.

We could also layer something like this over your proposed changes later to improve the UX so I'd love to hear your thoughts on this idea either way.

Which data are we trying to load specifically that is causing a delay? Is it all stuff we need? Why is it slow? Can we do anything to load it faster?

>Other entities might require loading the entire db into memory to be really effective. 

Which entity? I feel like we are talking about some hypothetical mega component that for some reason needs the entire database in memory and actually can't think of what that would be.

>In any case, this is rather a techinique for further optimization and the current implemented solution allows for a more generic solution, which is allowing the components to render while state is being fetched.

Overall, this looks like a good idea that we _probably_ need. I'm just trying to understand if we need it right now. It being a "generic solution" is part of my concern. Generic solutions sometimes add complexity or cause regressions for not much obvious benefit.

It does also feel like we may end up with 99% of cases where we're not gonna need `initialValue`. Introducing an "Onyx hook" will force everyone to provide that (I think). And to provide it in cases where they don't actually need it at all because it's logically impossible for the data _not_ to have been loaded in the cache already.

Thanks!

ospfranco:
> > pre-loading/pre-fetching is a solution, but it is not always applicable or practical.
> > For example, action reactions are not really linked to anything and it is not possible to know before hand which ones can be pre-loaded.
> 
> The reactions are something we could prefetch. They are linked to the actions via `reportActionID`, the actions are linked to the report via `reportID`, the reports are shown in the sidebar (and sometimes other places). We can prefetch the needed data on layout, scroll, hover, etc.
> 
> We could also layer something like this over your proposed changes later to improve the UX so I'd love to hear your thoughts on this idea either way.
> 
> Which data are we trying to load specifically that is causing a delay? Is it all stuff we need? Why is it slow? Can we do anything to load it faster?

Ah sorry, I thought reactions where only based on the actionID.

The point I'm trying to make is rather: before the user clicks on the report you don't really know what to pre-fetch. In order to make it really effective we would have to load the entire DB into memory.

Once the report is opened we basically load everything, it's done via different components ofc, but it ends up being the entire sub-tree of data linked to the clicked Report. Report, ReportMetadata, Actions, Reactions to those Actions, IOUReports, etc. Each one causes a delay at different components and times. And yes, it is all needed, we got rid of a couple of unneeded connections on the corresponding Expensify PR but nothing major. As far as I can see there is nothing to be done to load it faster.

> > Other entities might require loading the entire db into memory to be really effective.
> 
> Which entity? I feel like we are talking about some hypothetical mega component that for some reason needs the entire database in memory and actually can't think of what that would be.

Right now the ReportScreen is the biggest one, but as far as I'm told, into only grew in complexity somewhat recently. However, as time passes it is the natural progression more and more components will get connected to more entities.

The problem is not only one single component, but the entire sub-tree that gets rendered. It's on a case to case basis that sub-trees get expensive to render or cause a lot of re-renders. It's not about one entity, but rather how onyx holds rendering/pushes states and the interaction between components and React trying to flush/render components.

> > In any case, this is rather a techinique for further optimization and the current implemented solution allows for a more generic solution, which is allowing the components to render while state is being fetched.
> 
> Overall, this looks like a good idea that we _probably_ need. I'm just trying to understand if we need it right now. It being a "generic solution" is part of my concern. Generic solutions sometimes add complexity or cause regressions for not much obvious benefit.

As stated, we could not apply pre-fetching to the ReportScreen and it's entire sub-tree of components that mounted and caused delays in rendering. The only solution we found was to allow the components to eagerly render. If we do not merge this change, we only apply superficial optimizations but the entire sub-tree will block as onyx pushes updates to the components.

> It does also feel like we may end up with 99% of cases where we're not gonna need `initialValue`. Introducing an "Onyx hook" will force everyone to provide that (I think). And to provide it in cases where they don't actually need it at all because it's logically impossible for the data _not_ to have been loaded in the cache already.

Unless you load the entire DB into cache, it is impossible to predict a cache miss. As `initialValue` is also optional, it doesn't really affect how the current app is developed, in the future a hook can be introduced, but the current performance really makes the app unusable and we were tasked to find a solution for now :)

Let me know if you still have questions, or maybe we can jump into a call to have a faster feedback loop, we can even play around with your ideas if that is what you need, and hopefully the ideas will become clearer.

lib/utils.js
marcaaron: There are some arrays, but they are less frequently used.

This one comes to mind: https://github.com/Expensify/App/blob/48437acb34dbc38b9b7a60dae4c272727b8ce3ab/src/libs/actions/ActiveClients.js#L9

marcaaron (changes requested):
Couple small things / questions. I am more ok with the idea as a whole now after reviewing more thoroughly. Though, still kind of wonder about the behavior of `initialValue` and if it should maybe replace `defaultValue`.

lib/Onyx.js
```diff
@@ -466,9 +466,12 @@ function keysChanged(collectionKey, partialCollection) {
                     continue;
                 }
 
-                subscriber.withOnyxInstance.setState((prevState) => {
+                subscriber.withOnyxInstance.setStateProxy((prevState) => {
                     const data = cachedCollection[subscriber.key];
                     const previousData = prevState[subscriber.statePropertyName];
+                    if (utils.areObjectsEmpty(data, previousData)) {
```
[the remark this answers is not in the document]
ospfranco: done!

lib/withOnyx.js
```diff
@@ -196,22 +218,32 @@ export default function (mapOnyxToState) {
                 });
             }
 
+            flushPendingSetStates() {
```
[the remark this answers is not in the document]
ospfranco: The pending queue gets emptied at the end, but it will not hurt. It's added.

README.md
```diff
@@ -135,7 +135,35 @@ export default withOnyx({
 })(App);
 ```
 
-It is preferable to use the HOC over `Onyx.connect()` in React code as `withOnyx()` will delay the rendering of the wrapped component until all keys have been accessed and made available.
+It is preferable to use the HOC over `Onyx.connect()` in React code as `withOnyx()` will delay the rendering of the wrapped component until all keys/entities have been fetched and passed to the component. This however, can really delay your application if many entities are connected to the same component, you can pass an `initialValue` to each key to allow Onyx to eagerly render your component with this value.
```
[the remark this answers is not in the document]
ospfranco: Changed the wording a bit

README.md
```diff
@@ -135,7 +135,35 @@ export default withOnyx({
...
+```javascript
+export default withOnyx({
+    session: {
+        key: ONYXKEYS.SESSION,
+        initialValue: {}
+    },
+})(App);
+```
+
+Additionally, if your component has many keys/entities you might also trigger frequent re-renders on the initial mounting. You can workaround this by passing an additional object with the `delayUpdates` property set to true. Onyx will then put all the updates in a queue until you decide when then should be applied, the component will receive a function `markReadyForHydration`. A good place to call this function is on the `onLayout` method, which gets triggered after your component has been rendered.
```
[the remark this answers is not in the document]
ospfranco: It all the depends on the product. On the ReportScreen we do care, as there are many API calls that get triggered and is preferable to mount an skeleton while data is fetched/hydrated/parsed/etc.

lib/withOnyx.js
```diff
@@ -132,15 +155,14 @@ export default function (mapOnyxToState) {
                     return;
                 }
 
-                // Leave untouched previous state to avoid data loss during pre-load updates.
-                // This handles case when setState was called before the setWithOnyxState.
-                // For example, when an Onyx property was updated by keyChanged before the call of the setWithOnyxState.
-                this.setState((prevState) => {
-                    const remainingTempState = _.omit(this.tempState, _.keys(prevState));
```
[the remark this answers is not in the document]
ospfranco: I don't quite understand how is this change correct, what if a new value is set for a key in the prevState? Then it will be ommitted?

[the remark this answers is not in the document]
marcaaron: I _think_ the idea is that there can be a race between a new value calling `setState()` on `withOnyx()` and the time it takes to get this initial value from storage. 

And if that happens then we should use the most recent value (the one that was set and not the stale value that we read). 

Tbh - not in love with how this was implemented. I would prefer find some way to mark the data as stale if something changed it's value, but I guess we decided the correct place to do that would be in the `setState()` callback.

marcaaron (commented):
Latest changes LGTM. I am only unsure about this one change [here](https://github.com/Expensify/react-native-onyx/pull/329/files#r1323486211). If we don't need to make the change maybe we can reserve it for a follow up.

lib/withOnyx.js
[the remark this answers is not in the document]
ospfranco: Ah, now I got it. Well, this is no longer correct if `initialValue` is passed, it will never update because the component starts with a value on that key.

I need to think about how to solve this.

ospfranco:
@marcaaron I updated the merging algorithm between prevState and the new state on `setWithOnyxState` unfortunately it is far more complex. I also tried to update the comment to explain the situation. Could you review it one more time and carefully to make sure I did not miss anything?

lib/withOnyx.js
```diff
@@ -19,7 +20,7 @@ function getDisplayName(component) {
     return component.displayName || component.name || 'Component';
 }
 
-export default function (mapOnyxToState) {
+export default function (mapOnyxToState, options = {}) {
```
tgolen, line 23: Until there are other options, a simple param is fine.
```suggestion
export default function (mapOnyxToState, shouldDelayUpdates = false) {
```

tgolen (changes requested):

tgolen (changes requested):
Can you please add unit tests to ensure that `initialValue` and `shouldDelayUpdates` work as expected?

lib/withOnyx.js
ospfranco: I strongly argue against it, boolean params are confusing because they are not labeled. An options object future proofs while keeping things explicit.

tgolen: While I don't disagree with the points of your logic, I disagree with the results of that logic. If that logic was followed, then no method anywhere would ever have single params and they would all have one big arguments object.

What about moving it to `mapOnyxToState`?

ospfranco: Robert C. Martin says it himself:

https://www.informit.com/articles/article.aspx?p=1392524

But Szymon didn't want the key inside mapOnyxToState because that would change how the necessary keys for rendering are discovered. I will change it as a single boolean for now.

tests/unit/withOnyxTest.js
```diff
@@ -362,4 +381,77 @@ describe('withOnyx', () => {
                 expect(onRender.mock.calls[1][0].simple).toBe('long_string');
             });
     });
+
+    it('initialValue should be feed into component', () => {
+        const onRender = jest.fn();
+        const markReadyForHydration = jest.fn();
+
+        // Given there is a simple key that is not an array or object value
+        // Onyx.merge(ONYX_KEYS.SIMPLE_KEY, 'string');
```
tgolen, line 390: Should this be commented out, or can it just be removed?

tests/unit/withOnyxTest.js
```diff
@@ -362,4 +381,77 @@ describe('withOnyx', () => {
...
+                    simple: {
+                        key: ONYX_KEYS.SIMPLE_KEY,
+                        initialValue: 'initialValue',
+                    },
+                })(ViewWithCollections);
+                render(<TestComponentWithOnyx markReadyForHydration={markReadyForHydration} onRender={onRender} />);
+            })
+
+            // And we set the value to the same value it was before
+            // .then(() => Onyx.merge(ONYX_KEYS.SIMPLE_KEY, 'string'))
```
tgolen, line 405: Same

tests/unit/withOnyxTest.js
```diff
@@ -362,4 +381,77 @@ describe('withOnyx', () => {
...
+
+                // just to test we change the value
+                return Onyx.merge(ONYX_KEYS.SIMPLE_KEY, 'long_string');
+            })
+            .then(() => {
+                // Component still has not updated
+                expect(onRender).toHaveBeenCalledTimes(1);
+
+                // We can now tell component to update
+                // console.log(componentInstance);
```
tgolen, line 447: Remove

tgolen (changes requested):
Thanks for making the change to the boolean param and for the tests.

API.md
```diff
@@ -131,6 +131,7 @@ Subscribes a react component's state directly to a store key
...
 | [mapping.selector] | <code>String</code> \| <code>function</code> | THIS PARAM IS ONLY USED WITH withOnyx(). If included, this will be used to subscribe to a subset of an Onyx key's data.       If the selector is a string, the selector is passed to lodashGet on the sourceData. If the selector is a function, the sourceData and withOnyx state are       passed to the selector and should return the simplified data. Using this setting on `withOnyx` can have very positive performance benefits because the component       will only re-render when the subset of data changes. Otherwise, any change of data on any property would normally cause the component to re-render (and that can       be expensive from a performance standpoint). |
+| [mapping.initialValue] | <code>Any</code> | THIS PARAM IS ONLY USED WITH withOnyx(). If included, this will be passed as the initial hydrated value for the mapping. It will allow the component to render eagerly while data is being fetched from the DB. Note that it will not cause the component to have the `loading` prop set to true. |
```
tgolen, line 134: Is it expected that only `initialValue` is added to this API doc? It seems like `shouldDelayUpdates` and  `markReadyForHydration` should show up somewhere here. Maybe try re-running the generator?

tgolen, line 134: > the initial hydrated value for the mapping

What exactly does "hydrated" mean? 😅 

Is there a better term we can use for this that has the same meaning? I can only think of someone drinking a lot of water.

tgolen (changes requested):

API.md
ospfranco: Ah, this document is generated? I modified manually

ospfranco: Standard [React lingo](https://react.dev/reference/react-dom/hydrate). But open to suggestions.

ospfranco: I re-ran the generator. However `shouldDelayUpdates` and `markReadyForHidration` are part of the HOC and not of Onyx, so not sure if I should put them here. That's why I also updated the README

tgolen: Heh, yeah, but what does it mean to you? Like, let's say you were explaining this to a 7-year-old.

tgolen: Ah, OK... I guess that's fine if the generator didn't pick them up. I didn't think about it only being a part of the HOC. I think the stuff you added in the README is good enough for the HOC.

tgolen: I was talking with another co-worker about this and ChatGPT for the win!

Screenshot_2023-09-15_at_10 44 15_AM-2023-09-15 17_46_10 411

So, I think it would be helpful to change this to say:

> If included, this will be passed to the component so that something can be rendered while data is being fetched from the DB. Note that it will not cause the component to have the `loading` prop set to true.

lib/withOnyx.js
marcaaron: Uncle Bob rules only apply if it's Java code

ospfranco: Why would it? it's just code and boolean functions. They get messy over time, there is plenty of arguments against them and I have also lived through long-lived codebases where they make everything hard to understand.

https://medium.com/@amlcurran/clean-code-the-curse-of-a-boolean-parameter-c237a830b7a3

https://understandlegacycode.com/blog/what-is-wrong-with-boolean-parameters/

In any case, not going to push this on you, just sharing my experience.

ospfranco:
PR is updated with latest comments

tgolen (approved):
Sorry, this is so close! A few conflicts snuck in from the last PR I merged.

ospfranco:
Updated

tgolen (approved):

lib/withOnyx.js
```diff
@@ -101,47 +116,86 @@ export default function (mapOnyxToState) {
...
                     return;
                 }
 
-                // Leave untouched previous state to avoid data loss during pre-load updates.
-                // This handles case when setState was called before the setWithOnyxState.
-                // For example, when an Onyx property was updated by keyChanged before the call of the setWithOnyxState.
+                const stateUpdate = {...this.tempState};
+                delete this.tempState;
+
+                // Full of hacky workarounds to prevent the race condition described above.
```
marcaaron, line 170: NAB / FWIW, I am looking at this code and it looks reasonable since the edge case is well described and this clearly shows the few cases we will run into. Maybe I am a hacker though 😄 Nice work either way.

marcaaron (changes requested):
One small comment about the mapping.config variable.

marcaaron:
Looks like a test is failing as well.

lib/withOnyx.js
```diff
@@ -28,37 +29,51 @@ export default function (mapOnyxToState) {
...
+                         * Onyx.merge('report_123', value);
+                         * Navigation.navigate(route); // Where "route" expects the "value" to be available immediately once rendered.
+                         *
+                         * In reality, Onyx.merge() will only update the subscriber after all merges have been batched and the previous value is retrieved via a get() (returns a promise).
+                         * So, we won't use the cache optimization here as it will lead us to arbitrarily defer various actions in the application code.
+                         */
+                        if (
+                            (value !== undefined
+                                && !Onyx.hasPendingMergeForKey(key))
+                            || mapping.allowStaleData
```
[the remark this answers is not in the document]
ospfranco: tbh, I don't know. @Szymon20000 you knew better how it behaves.

ospfranco:
Looks like a missed a change when I did the merge. Tests should pass now.

lib/withOnyx.js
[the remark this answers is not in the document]
ospfranco: Looks, like this is the only place where it is used. Seems like it allows the component to be feed undefined data or stale data when there is pending merges. There is no documentation since it is not part of `onyx.js` which is the only file the generator looks at. That's why I manually added a section to the README.

marcaaron (approved):

marcaaron:
@yuwenmemon did you want to give this any final review?

yuwenmemon (approved):

PR #333, opened 2023-09-11
Fix: nullish keys not removed/merged in IDB-Keyval (web)

chrispader:
@tgolen @marcaaron 

### Details

On web in the new IDB-Keyval implementation, when merging `null` or `undefined`, the keys are not removed from storage, like it's done on native with SQLite.

Additionally, top-level nullish keys in an object should also be removed when merging. E.g. when merging an existing value `{a: 1, b: 2}` with `{a: null}` or `{a: undefined}` the resulting value should be `{b: 2}`.

This PR aims to fix caused problems like [this one](https://github.com/Expensify/App/issues/26627#issuecomment-1705443948).

### Related Issues

https://github.com/Expensify/App/issues/26627


lib/utils/removeNullObjectValues.js
```diff
@@ -0,0 +1,19 @@
+import _ from 'underscore';
```
tgolen, line 1: This is great moving it here. I swear I saw some other PR also adding a `utils.js` file, but I don't see it in `main`, so maybe it isn't merged yet... Would you mind seeing if you can find the other PR and then making sure that the utils are all being organized consistently?

tgolen (changes requested):

lib/utils/removeNullObjectValues.js
chrispader: You're right, this was added in [this PR](https://github.com/Expensify/react-native-onyx/pull/329).

I also changed this in the [SQLite crash PR](https://github.com/Expensify/react-native-onyx/pull/334)

lib/storage/providers/IDBKeyVal.js
```diff
@@ -54,9 +55,17 @@ const provider = {
 
         return getValues.then((values) => {
             const upsertMany = _.map(pairs, ([key, value], index) => {
+                // If the value is null, we want to delete the key from storage,
+                // to be consistent with the native implementation with SQLite.
+                // SQLite by default removes keys from storage that are set to nullish values
+                if (value == null) {
```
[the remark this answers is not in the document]
chrispader: Same as here: https://github.com/Expensify/react-native-onyx/pull/334#discussion_r1321818427

I changed this to a more strict null check with lodash

lib/fastMerge.js
```diff
@@ -56,8 +56,8 @@ function mergeObject(target, source) {
 function fastMerge(target, source) {
     // lodash adds a small overhead so we don't use it here
     // eslint-disable-next-line rulesdir/prefer-underscore-method
-    const array = Array.isArray(source);
-    if (array) {
+    const isArray = Array.isArray(source);
+    if (isArray || source == null) {
```
tgolen, line 60: `_.isNull()` 😁

lib/storage/providers/IDBKeyVal.js
```diff
@@ -54,9 +55,17 @@ const provider = {
 
         return getValues.then((values) => {
             const upsertMany = _.map(pairs, ([key, value], index) => {
+                // If the value is null, we want to delete the key from storage,
+                // to be consistent with the native implementation with SQLite.
+                // SQLite by default removes keys from storage that are set to nullish values
+                if (_.isUndefined(value) || _.isNull(value)) {
```
tgolen, line 61: When would it be possible for `value` to be undefined? Wouldn't that mean that one of the `pairs` has only a key?

tgolen (changes requested):

lib/DevTools.js
```diff
@@ -1,6 +1,12 @@
...
 import _ from 'underscore';
 
+const {connectViaExtension} = process.env.NODE_ENV === 'production' ? require('remotedev') : {
+    connectViaExtension: () => ({
+        init: () => {},
+        send: () => {},
+    }),
+};
+
 class DevTools {
```
marcaaron, line 10: nab, Where did this file come from and what does it do? 😂 Is there documentation for how to use it available?

lib/storage/providers/IDBKeyVal.js
marcaaron: The `value` part of the pair could also be `undefined` like `['something', undefined]`.

marcaaron: I am sort of wondering how/why that happened though. It seems inconsistent with the merge behavior. Setting something to `null` has a special significance (i.e. delete me). Setting it to `undefined` does not mean anything. 

I would guess that we would just throw these values out vs. allow them to overwrite / delete any keys.

marcaaron (changes requested):

lib/fastMerge.js
```diff
@@ -56,8 +56,8 @@ function mergeObject(target, source) {
 function fastMerge(target, source) {
     // lodash adds a small overhead so we don't use it here
     // eslint-disable-next-line rulesdir/prefer-underscore-method
-    const array = Array.isArray(source);
-    if (array) {
+    const isArray = Array.isArray(source);
```
Gonals, line 59: We don't need this to be a variable

Gonals: And we can use `_.isArray` 😁

Gonals (changes requested):

mountiny:
@chrispader Can we make sure the tests cover this scenario and make sure that setting key to null or undefined will remove it with either storage? I am surprised this was not caught with automated tests

lib/storage/providers/IDBKeyVal.js
chrispader: Just added a `[HOLD on #334]` tag to this PR. The  [SQLite crash PR](https://github.com/Expensify/react-native-onyx/pull/334) actually fixes two things that are expected for this PR as well.

- Allow `undefined` values to be passed to `fastMerge`
- Remove key in `Onyx.merge` if `null` is passed

So we'll first want to merge the other PR in order for this to continue 👍

chrispader: This way, the extra condition in `IDBKeyVal.multiMerge` is not needed anymore

lib/DevTools.js
chrispader: I think the branch was just not up to date with the latest `main`. Outdated now

chrispader:
> @chrispader Can we make sure the tests cover this scenario and make sure that setting key to null or undefined will remove it with either storage? I am surprised this was not caught with automated tests

Yes, i'll update the tests accordingly 👍

chrispader:
@marcaaron another thought... When we merge an object with another object, where every top-level key is set to `null`, it will result in `{}`. Should we instead set the value to `undefined` in this case or leave the empty object in storage as it is?

tgolen (approved):
If it results in an empty object after removing all null values, then I think we should remove the object entirely. There is no sense in storing an empty object.

chrispader:
> If it results in an empty object after removing all null values, then I think we should remove the object entirely. There is no sense in storing an empty object.

Agree. Gonna change it this way then 👍

tgolen (dismissed):

marcaaron:
>When we merge an object with another object, where every top-level key is set to null, it will result in {}. Should we instead set the value to undefined in this case or leave the empty object in storage as it is?

I am not sure why it would make a difference (apart from less space taken up). One thought is that if the "cache" knows the object doesn't exist then on the next attempt to "read" we should not need try and read it from storage. So, there is probably an optimization to consider with that if we would do that vs. having the "cache" know that the value doesn't exist vs. not sure if it exists or not and have to read it again.

chrispader:
> > When we merge an object with another object, where every top-level key is set to null, it will result in {}. Should we instead set the value to undefined in this case or leave the empty object in storage as it is?
> 
> I am not sure why it would make a difference (apart from less space taken up). One thought is that if the "cache" knows the object doesn't exist then on the next attempt to "read" we should not need try and read it from storage. So, there is probably an optimization to consider with that if we would do that vs. having the "cache" know that the value doesn't exist vs. not sure if it exists or not and have to read it again.

I see your point. If we set the value to `undefined` instead of leaving it as `{}`, Onyx will think there is no value in cache and therefore try to read it from storage, right?

In this case i also think it would be better to leave the `{}` in storage. I think we should favor less computing rather than a bit less disk space..

Wdyt @tgolen ?

chrispader:
Agree. Also in favor of leaving `{}` in storage for now.

The current implementation achieves exactly this, so this should be ready for merging 👍🏼

tgolen (dismissed):

Gonals (approved):

tgolen:
I think it's just waiting on @marcaaron for this one. I wasn't sure that all of Marc's concerns were addressed so I didn't want to dismiss his review and merge this yet.

marcaaron:
Heads up, I'm probably not going to get to this today - so if you need to move forward without me, please go ahead.

chrispader:
gonna have to resolve conflicts tomorrow morning anyway, so you can still chime in if you want @marcaaron

chrispader:
> gonna have to resolve conflicts tomorrow morning anyway, so you can still chime in if you want @marcaaron

merge conflicts resolved 👍

chrispader:
@tgolen ill leave it up to you if you want to wait for one more review by @marcaaron

tgolen (approved):

pradeepmdk:
@chrispader This is not yet resolved because when  `const prev = values[index];` has the value means still taking the old value. not remove the key from Onyx when we pass null in the merge command.
https://github.com/margelo/react-native-onyx/blob/cad1b560828d935aff83d3f16c919e125135ff03/lib/storage/providers/IDBKeyVal.js#L57

solution: 
1) you just  we need to check this here  `Utils.removeNullObjectValues` in  `IDBKeyVal.js` place
and need removed from `lib/Onyx.js` on the merge function

pradeepmdk:
adding ref video

https://github.com/Expensify/react-native-onyx/assets/16595846/7694884f-978f-4abc-87fc-3cb1e51ebb9d

chrispader:
> and `removeNullObjectValues` funciton currently we are checking the null only we missed to check undefined case. we need change `isNull => isNil`

This is actually intended as described by @marcaaron in [this GH comment](https://github.com/Expensify/react-native-onyx/pull/339#issuecomment-1718390442). @marcaaron should we maybe overthink this? Currently, we'll leave `undefined` values in storage for top-level object keys

chrispader:
> @chrispader This is not yet resolved because when `const prev = values[index];` has the value means still taking the old value. not remove the key from Onyx when we pass null in the merge command.

I agree, that this isn't optimal yet. This might be part of a bigger change though, because we don't really need to read the existing value from storage in `Onyx.merge` anymore, since we already read it in the "storage layer" for web (SQLite) and native (IDBKeyVal).

chrispader:
> @chrispader This is not yet resolved because when `const prev = values[index];` has the value means still taking the old value. not remove the key from Onyx when we pass null in the merge command. https://github.com/margelo/react-native-onyx/blob/cad1b560828d935aff83d3f16c919e125135ff03/lib/storage/providers/IDBKeyVal.js#L57
> 
> solution: Option 1) just we need to check this here `Utils.removeNullObjectValues` in `IDBKeyVal.js` place and need removed from `lib/Onyx.js` on the merge function Option 2) we need pick the Object.keys from new value and based on keys only we need the select the key from `prev` value
> 
> and `removeNullObjectValues` funciton currently we are checking the null only we missed to check undefined case. we need change `isNull => isNil`

This issue was reference in this PR: https://github.com/Expensify/react-native-onyx/pull/359

PR #334, opened 2023-09-11
Fix: SQLite causes crash when trying to merge undefined

chrispader:
@tgolen @marcaaron 

### Details

When you run `Onyx.merge` with the value `undefined` on native platforms, SQLite will crash the app, because we have a `NOT NULL` constraint for all values in the database.

**Proposal**

When `undefined` is passed as the value in `Onyx.merge` we manually remove the key from storage with `Storage.removeItem` and remove it from the merge delta.

### Related Issues

https://expensify.slack.com/archives/C01GTK53T8Q/p1693404147209609


lib/storage/providers/SQLiteStorage.js
```diff
@@ -88,11 +88,27 @@ const provider = {
              ON CONFLICT DO UPDATE
              SET valueJSON = JSON_PATCH(valueJSON, JSON(:value));
         `;
-        const queryArguments = _.map(pairs, (pair) => {
+
+        const queryArguments = [];
+        const keysToDelete = [];
+        _.forEach(pairs, (pair) => {
+            if (pair[1] == null) {
```
tgolen, line 95: `_.isNull()` or `===`

lib/storage/providers/SQLiteStorage.js
```diff
@@ -88,11 +88,27 @@ const provider = {
...
+            }
+
             const value = JSON.stringify(pair[1]);
-            return [pair[0], value];
+
+            queryArguments.push([pair[0], value]);
         });
-        return db.executeBatchAsync([[query, queryArguments]]);
+
+        const deletePromise = this.removeItems(keysToDelete);
```
tgolen, line 105: Can you add a code comment explaining why there is deleting happening inside of the merge function?

tgolen (changes requested):

lib/storage/providers/SQLiteStorage.js
chrispader: This was intended to check for both `undefined` or `null`. I'll instead check for both `_.isNull` and `_.isUndefined` 👍

chrispader:
@tgolen what's the desired behaviour of Onyx if a key is set to `undefined`? Previously, setting it to/ merging it with `null` will remove the key from storage.

Should we also remove the key when setting/merging `undefined`?

lib/Onyx.js
```diff
@@ -1044,11 +1045,21 @@ function applyMerge(existingValue, changes) {
         return lastChange;
     }
 
-    if (_.isObject(existingValue) || _.every(changes, _.isObject)) {
+    if (_.isObject(existingValue) || _.isObject(lastChange)) {
+        // We want to merge multiple object delates together
```
tgolen, line 1049: What are "delates"?

lib/Onyx.js
```diff
@@ -1044,11 +1045,21 @@ function applyMerge(existingValue, changes) {
...
+        // If all of the changes are objects, we can simply merge all of them together
+        // If the batched changes contain nullish values or arrays, we should only merge the very last objects in the batch
```
tgolen, line 1051: Can you please expand this comment to explain _why_ only the last objects in the batch should be merged? I'm not sure I understand the scenario or what nullish and array values have to do with it.

lib/storage/providers/SQLiteStorage.js
```diff
@@ -88,10 +88,12 @@ const provider = {
              ON CONFLICT DO UPDATE
              SET valueJSON = JSON_PATCH(valueJSON, JSON(:value));
         `;
+
         const queryArguments = _.map(pairs, (pair) => {
-            const value = JSON.stringify(pair[1]);
-            return [pair[0], value];
+            const value = pair[1];
```
tgolen, line 93: Seems like this variable really isn't necessary then, is it?

lib/utils.js
```diff
@@ -0,0 +1,13 @@
+function countObjectsAtEndOfArray(array) {
```
tgolen, line 1: I'm not quite getting this... what about objects at the beginning or the middle of the array?

tgolen (changes requested):

lib/Onyx.js
MonilBhavsar: +1
May be we want to merge the very last value in the batch - irrespective of if it is null or not null

lib/storage/providers/SQLiteStorage.js
MonilBhavsar: +1

marcaaron:
>When undefined is passed as the value in Onyx.merge we manually remove the key from storage with Storage.removeItem and remove it from the merge delta.

I don't think we should do that. Just throw the value out. Probably they wanted to _merge_ something. Not delete it and it just happened to be `undefined`. I think we should just act like it didn't happen. They will figure out quickly that it's the wrong method to use if they really want to remove a key.

marcaaron (changes requested):
I think this new behavior is not desirable. I propose we ignore any merging of `undefined` values. We never said anywhere that you can do this. People might not intentionally be doing this. It is much more likely that a value would be unintentionally `undefined` than unintentionally `null`.

chrispader:
> I think this new behavior is not desirable. I propose we ignore any merging of `undefined` values. We never said anywhere that you can do this. People might not intentionally be doing this. It is much more likely that a value would be unintentionally `undefined` than unintentionally `null`.

Alright 👍 Makes sense! Just wanted to be on the same page on that. Will instead just set the value to `undefined`. Setting/merging `null` will still delete the key though

chrispader:
Please wait with reviews, fixing tests right now...

chrispader:
This PR now only contains the lines of code needed in `SQLiteStorage.js` to fix the crash...

tgolen (approved):
Thanks for decreasing the scope of this!

lib/storage/providers/SQLiteStorage.js
```diff
@@ -88,10 +88,9 @@ const provider = {
              ON CONFLICT DO UPDATE
              SET valueJSON = JSON_PATCH(valueJSON, JSON(:value));
         `;
-        const queryArguments = _.map(pairs, (pair) => {
-            const value = JSON.stringify(pair[1]);
-            return [pair[0], value];
-        });
+
+        const queryArguments = _.map(pairs, pair => [pair[0], _.isUndefined(pair[1]) ? null : pair[1]]);
```
marcaaron, line 92: Can we filter out any pair that has an `undefined` value? I believe this would still delete it from storage.

marcaaron (changes requested):

tgolen (dismissed):

lib/storage/providers/SQLiteStorage.js
```diff
@@ -88,10 +88,9 @@ const provider = {
              ON CONFLICT DO UPDATE
              SET valueJSON = JSON_PATCH(valueJSON, JSON(:value));
         `;
-        const queryArguments = _.map(pairs, (pair) => {
-            const value = JSON.stringify(pair[1]);
-            return [pair[0], value];
-        });
+
+        const queryArguments = _.filter(pairs, pair => !_.isUndefined(pair[1]));
```
marcaaron, line 92: Do we still need to stringify the values?

marcaaron (changes requested):

tgolen (approved):

marcaaron (approved):

PR #335, opened 2023-09-11
fix: cannot find module remotedev in production environment

pac-guerreiro:
### Details

When building react native onyx for `production`,  `node` will throw an error saying that it cannot find `remotedev` module. This is due to the fact that `remotedev` is a `devDependency` which is not included in `production` bundle. To solve this, `remotedev` needs to be dynamically imported.

### Related Issues

https://github.com/Expensify/App/pull/27130#issuecomment-1713473969

### DIFF
--- a/lib/DevTools.js
+++ b/lib/DevTools.js
@@ -1,5 +1,11 @@
-import {connectViaExtension} from 'remotedev';
 import _ from 'underscore';
+
+const {connectViaExtension} = process.env.NODE_ENV === 'production' ? require('remotedev') : {
+    connectViaExtension: () => ({
+        init: () => {},
+        send: () => {},
+    }),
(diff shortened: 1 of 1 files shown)


mountiny:
@ospfranco has been testing if the tests are fixed after this PR

mountiny (approved):
Thanks, Oscar confirmed this works here https://github.com/Expensify/App/pull/27130#issuecomment-1713950505

ospfranco:
Just tested this trying to start the app with metro running and got the same error. Seems on Metro NODE_ENV is indeed set to `production`.

CleanShot 2023-09-12 at 10 36 36

@pac-guerreiro did you try running the app and not just the tests with your change?

Issue #301, opened 2023-08-16
Onyx.merge does not remove a nested object key


thienlnam:
If you set an object key to null, Onyx.merge will remove it 

```
Object: {
	transactionID: 12312312
	comment: {
			waypoints: {
				waypoint0: {}
			}
     }
}
```

This works for first level keys
// Remove all first level keys that are explicitly set to null.
```
Onyx.merge(key, {
	transactionID: null
}
```

Corresponding object: 
```
{
	comment: {
			waypoints: {
				waypoint0: {}
			}
     }
}
```

But if you try to remove a nested key

```
Onyx.merge(key, {
	comment: {
		waypoints: {
			waypoint0: null
		}
	}
}
```

It will just set the value to null
```
{
	comment: {
			waypoints: {
				waypoint0: null
			}
     }
}
```

neil-marcellini:
following

tgolen:
This is very interesting! I didn't believe it at first, but I did a quick test and confirmed exactly what you report. @chrispader you might be interested in checking this one out (our resident expert on `Onyx.merge()` 😁 ).

tgolen:
@thienlnam Could you give a little more context into the underlying thing you're trying to do and why it's preferrable to remove the nested key as opposed to having it remain as `null`?

I could see "consistency" as one argument (and a good one at that), but I kind of feel there needs to be a better reason to change it since it's changing some very low-lying parts of the lib.

chrispader:
Yes this is indeed a bit inconsistent, but until now this was actually the intended output.

I'm open to changing this so all `null` values will be removed when merging

neil-marcellini:
A PR was raised to fix this https://github.com/Expensify/react-native-onyx/pull/350. If top level keys are removed when set to null then we should do the same for nested keys IMO.

thienlnam:
> @thienlnam Could you give a little more context into the underlying thing you're trying to do and why it's preferrable to remove the nested key as opposed to having it remain as null?

Yeah, mostly just an expectation on how merging a null value would work and specifically for waypoints we had the issue because we keyed them like this, and used that to determine the order of the stops / when to add new ones

```
...
{
	waypoint1: {}
    waypoint2: {}
    waypoint3: {}
}
```

It just becomes difficult to determine when to add a new waypoint. So for example, if I remove a waypoint it should be
```
...
{
	waypoint1: {}
    waypoint2: {}
}
```

But since it doesn't null out, it would be
```
...
{
[...]

PR #339, opened 2023-09-12
Ignore undefined values in merge queue when merging objects

chrispader:
@tgolen @marcaaron

Moving code from [this PR](https://github.com/Expensify/react-native-onyx/pull/334) as it became kind of a scope creep.

### Details

This PR intends to improve how we merge a batch of changes which includes nullish values and objects.

**Notice:** This PR does not handle nullish values in nested objects, there's [another PR](https://github.com/Expensify/react-native-onyx/pull/353) for this change.

In general, we have the following possible values that are passed **directly** to `Onyx.merge`:

- `object`: Merges the existing value with the merge changes
- `arrays`: Just replaces the existing value
- `undefined`: Also replaces the existing value
- `null`: Will remove the key from storage

When we call `Onyx.merge` multiple times with different types of data, we can eventually get a batch of changes like this:

```js
const existingValue = {a: 1, b: 2}
const mergeQueue = [{a: 2}, undefined, {c: 3, d: 4}, {d: 5, e: 6}]
const batchedChanges = applyMerge(undefined, mergeQueue) // evaluates to {c: 3, d: 5, e: 6}
const modifiedData = applyMerge(existingValue, [batchedChanges]) // evaluates to {a: 1, b: 2, c: 3, d: 5, e: 6}
```

With the current implementation of `applyMerge`, `batchedChanges` would evaluate to `{a: 1, b: 2, c: 3, d: 5, e: 6}`. This behaviour ignores the fact, that we had an `undefined` change in the merge queue, which should basically reset the existing value.

Therefore, we need to check if the `mergeQueue` contains any nullish value, and if so, we have to ignore the `existingValue` in the process of merging the delta changes.


tgolen:
> an undefined change in the merge queue, which should basically reset the existing value.

Why would it reset the value? I would think it would just not do anything.

chrispader:
> > an undefined change in the merge queue, which should basically reset the existing value.
> 
> Why would it reset the value? I would think it would just not do anything.

Ok, if that's the expected behaviour then we can ignore this change. Still, the current implementation isn't 100% correct then.

I created a new test, which has a lot of `Onyx.merge` calls including some with the value `undefined`:

Screenshot 2023-09-13 at 11 13 28

The expected output should be `{test1: 'test1', test2: 'test2', test3: 'test3', test4: 'test4'}`, is that correct?

chrispader:
The solution to this would be to simply remove all `undefined` values from the merge queue and instead only merge the delta changes when it's an object.

chrispader:
> > an undefined change in the merge queue, which should basically reset the existing value.
> 
> Why would it reset the value? I would think it would just not do anything.

Different question. What if a batch of merge changes contains a `null`. Should it also be ignored, if there are other objects in the merge queue. I suppose so? Because it would be set in storage right afterwards anyway... just for clarification

chrispader:
@tgolen @marcaaron i updated the PR according to my latest proposal

tgolen:
> The solution to this would be to simply remove all undefined values from the merge queue and instead only merge the delta changes when it's an object.

I personally like this. I don't think we should try to do anything special for handling `undefined` values. If the value is undefined, it's a mistake from the engineer who implemented it.

> What if a batch of merge changes contains a null. Should it also be ignored, if there are other objects in the merge queue. I suppose so? Because it would be set in storage right afterwards anyway... just for clarification

`null` is an expected and valid value, so I think it would just need to be careful to apply everything in the proper order. If this is the set of merges: `{a: 1}, {b: 2}, {b: null}`, then it should result in `{a: 1}`.

tgolen:
> Still, the current implementation isn't 100% correct then.

I forgot to touch on this. I'm glad you found this! I definitely don't think we ever considered undefined values, so I think it's totally possible that there would be unexpected behaviors around this.

marcaaron:
I agree with @tgolen. The test you wrote makes sense given the fact that we allow `undefined` values into the merge queue. But I think the implementation is wrong. And the `null` behavior described is what we want. 

That said, I'm not sure if we really _need_ to handle a case like this: 

```
[{a: 1}, {b: 1}, {b: undefined}]
```
It's hard to say what my expectations would be here because I can't really think of _why_ someone would do it. It seems reasonable to maybe end up with `{a: 1, b: undefined}`. But in practice I'd suggest using `null` instead of `undefined` because you goal (if this is intentional) is most likely to end up with `{a: 1}`.

chrispader:
> I agree with @tgolen. The test you wrote makes sense given the fact that we allow `undefined` values into the merge queue. But I think the implementation is wrong. And the `null` behavior described is what we want.

The implementation i wrote is only about handling top-level `undefined` values, which will be ignored. 

For nested objects, nothing changes by this. Actually, @thienlnam recently [created an issue here](https://github.com/Expensify/react-native-onyx/issues/301) that claims, that `null` values in nested objects aren't removed from storage. This is technically not a bug, but we just never implemented it. Possible to do so though. 

> That said, I'm not sure if we really _need_ to handle a case like this:
> 
> ```
> [{a: 1}, {b: 1}, {b: undefined}]
> ```

The use case my PR tries to fix is more for merge queues like this:

```js
[{a: 1}, {b: 1}, undefined, {c: 1}]
```

where we would expect the `undefined` value to be ignored and end up with a delta of `{a: 1, b: 1, c: 1}`. In the current implementation, this would end up as just `{c: 1}`, because `undefined` would overwrite the previous reduced changes.

chrispader:
@tgolen @marcaaron just resolved merge conflicts! 👍 

I think this PR is still valid, as i explained in [this comment](https://github.com/Expensify/react-native-onyx/pull/339#issuecomment-1718989344). This one only handles nullish values that are directly passed to `Onyx.merge(key, <undefined|null>)`.

I opened [another PR](https://github.com/Expensify/react-native-onyx/pull/353), which addresses the nested nullish values as discussed on Slack.

lib/Onyx.js
```diff
@@ -1082,15 +1082,18 @@ function multiSet(data) {
...
     }
 
-    if (_.isObject(existingValue) || _.every(changes, _.isObject)) {
-        // Object values are merged one after the other
-        // lodash adds a small overhead so we don't use it here
-        // eslint-disable-next-line prefer-object-spread, rulesdir/prefer-underscore-method
-        return _.reduce(changes, (modifiedData, change) => utils.fastMerge(modifiedData, change),
+    if (_.some(changes, _.isObject)) {
+        // There might be nullish values in a batch of changes that we want to ignore when merging objects,
+        // because otherwise they would overwrite the previous changes.
```
tgolen, line 1091:
```suggestion
        // because otherwise they overwrite the previous changes in the batch.
```
Just adding a little clarity.

tgolen (approved):

tgolen (dismissed):

tgolen:
All yours @amyevans

amyevans (dismissed):
Looks good to me! @marcaaron would you like to review?

lib/Onyx.js
```diff
@@ -1082,15 +1082,18 @@ function multiSet(data) {
...
     }
 
-    if (_.isObject(existingValue) || _.every(changes, _.isObject)) {
-        // Object values are merged one after the other
-        // lodash adds a small overhead so we don't use it here
-        // eslint-disable-next-line prefer-object-spread, rulesdir/prefer-underscore-method
-        return _.reduce(changes, (modifiedData, change) => utils.fastMerge(modifiedData, change),
+    if (_.some(changes, _.isObject)) {
+        // There might be nullish values in a batch of changes that we want to ignore when merging objects,
+        // because otherwise they overwrite the previous changes in the batch.
```
marcaaron, line 1091: Why would we ignore them? Isn't that expected that they would overwrite the previous changes.

marcaaron (changes requested):

lib/Onyx.js
marcaaron: I think a `null` value would basically cancel the previous ones...

e.g. 

```
Onyx.merge('test', {a: 1});
Onyx.merge('test', null);
Onyx.merge('test', {b: 2});
```
Should give use `{b: 2}`

If we filter out the null values then we are merging the first call with the last and would get `{a: 1, b: 2}`.

marcaaron: I think ignoring `undefined` makes sense. But not `null`.

tgolen: > I think ignoring undefined makes sense. But not null.

I think I buy this. Using `null` is a deliberate action to delete something, so it should not be ignored.

chrispader: @marcaaron you're right, this shouldn't filter for null values. Updated 👍

amyevans (changes requested):
Looks like some unrelated changes snuck in here

chrispader:
> Looks like some unrelated changes snuck in here

Which one do you mean exactly? The added comment was intended actually, also just removed the `fastMerge.js` file, as we moved everything to `utils.js`

chrispader:
@amyevans @marcaaron @tgolen also just added an additional test which checks for `null` values overwriting the existing value. Therefore, i needed to add one extra line in `Onyx.merge`

tests/unit/onyxTest.js
```diff
@@ -848,4 +848,57 @@ describe('Onyx', () => {
...
+            callback: (value) => {
+                testKeyValue = value;
+            },
+        });
+
+        return Onyx.set(ONYX_KEYS.TEST_KEY, {test1: 'test1'})
+            .then(() => {
+                expect(testKeyValue).toEqual({test1: 'test1'});
+                Onyx.merge(ONYX_KEYS.TEST_KEY, null);
+                Onyx.merge(ONYX_KEYS.TEST_KEY, {test2: 'test2'});
```
amyevans, line 895: NAB: Maybe just to make the behavior super obvious, add 1 additional merge after this and demonstrate the result is the merge of the final 2?

amyevans (approved):

tests/unit/onyxTest.js
chrispader: Done!

amyevans (dismissed):

tgolen (dismissed):

lib/Onyx.js
```diff
@@ -1134,12 +1138,20 @@ function merge(key, changes) {
                 // We first only merge the changes, so we can provide these to the native implementation (SQLite uses only delta changes in "JSON_PATCH" to merge)
                 let batchedChanges = applyMerge(undefined, mergeQueue[key]);
 
+                if (_.isNull(batchedChanges)) {
+                    return remove(key);
+                }
+
+                // If the mergeQueue included a null value, we want to overwrite the existing value
+                // In this case, we can't simply merge the batched changes with the existing value, because then the null in the merge queue would have no effect
```
marcaaron, line 1146: Can you think of a cleaner way to do this besides checking for the existence of `null` inside the queued changes? 😄

chrispader: Not really.. 😄

The only other way i could think of is to move this logic to `applyMerge` and return an object like `{mergedChanges, shouldOverwriteKey}` or something. But then, this only makes sense for the `batchedChanges`, because the second call of `applyMerge` will never be `null`.

marcaaron: Ok, I see. Just thinking out loud here - but let's say there is a `null` value in the merge queue... the value is either `null` or everything after a `null`? So wouldn't the "batched changes" essentially be the same as the existing value? Yet we merge them with `undefined`...

Could we also do this? 

```js
// The presence of a `null` in the merge queue instructs us to drop the existing value.
const shouldOverwriteExistingValue = _.includes(mergeQueue[key], null);
const updatedValue = shouldOverwriteExistingValue ? batchedChanges : applyMerge(existingValue, [batchedChanges]);
const modifiedData = utils.removeNullObjectValues(updatedValue);
```

chrispader: Yes, we can do that, you're right!

lib/Onyx.js
```diff
@@ -1082,15 +1082,19 @@ function multiSet(data) {
 function applyMerge(existingValue, changes) {
     const lastChange = _.last(changes);
 
-    if (_.isArray(existingValue) || _.isArray(lastChange)) {
```
marcaaron, line 1085: I think we are missing an edge case where it's possible to:

- Have an array
- Call merge with `undefined` (or the last change is undefined)
- We will set the value to that

Can we ignore all `undefined` by filtering them out first?

marcaaron (changes requested):
If anyone calls `merge()` with `undefined` _ever_ we should just ignore it. We are filtering out these values, but why let them even enter the mergeQueue at all? 🤔 

I think it's a bit simpler to just do:

```diff
diff --git a/lib/Onyx.js b/lib/Onyx.js
index d4d329e..c9f8311 100644
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -1120,6 +1120,10 @@ function applyMerge(existingValue, changes) {
  * @returns {Promise}
  */
 function merge(key, changes) {
+    if (_.isUndefined(changes)) {
+        return Promise.resolve();
+    }
```

chrispader:
> If anyone calls `merge()` with `undefined` _ever_ we should just ignore it. We are filtering out these values, but why let them even enter the mergeQueue at all? 🤔
> 
> I think it's a bit simpler to just do:

Yes, you're right. The only thing i changed is that we need to return the current `mergeQueuePromise[key]` instead of `Promise.resolve()`, in case there's already change in the merge queue

tgolen (approved):

amyevans (approved):

chrispader:
Can we merge this already? 🚀

amyevans:
@marcaaron want a final look?

marcaaron (approved):

PR #342, opened 2023-09-14
Add instructions and best practices to the README for dependent props

tgolen:
Coming from this slack convo: https://expensify.slack.com/archives/C01GTK53T8Q/p1694710295343199

cc @mountiny @marcaaron @allroundexperts

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,7 +161,7 @@ Background info:
 1. The component mounts with a `reportID={1234}` prop
 2. `withOnyx` evaluates the mapping
 3. `withOnyx` connects to the key `reports_1234` because of the prop passed to the component
-3. `withOnyx` connects to the key `policices_undefined` because `report` doesn't exist in the props yet, so the `policyID` defaults to `undefined`. * (see note below)
+3. `withOnyx` connects to the key `policies_undefined` because `report` doesn't exist in the props yet, so the `policyID` defaults to `undefined`. * (see note below)
 4. Onyx reads the data and updates the state of `withOnyx` with:
     - `report={{reportID: 1234, policyID: 1, ... the rest of the object ...}}`
     - `policy={undefined}` (since there is no policy with ID `undefined`)
--- a/README.md
+++ b/README.md
@@ -137,6 +137,46 @@ export default withOnyx({
 
 It is preferable to use the HOC over `Onyx.connect()` in React code as `withOnyx()` will delay the rendering of the wrapped component until all keys have been accessed and made available.
 
+### Dependent Onyx Keys and withOnyx()
+Some components need to subscribe to multiple Onyx keys at once and sometimes, one key might rely on the data from another key. This is similar to a JOIN in SQL.
+
+Example: To get the policy of a report, the `policy` key depends on the `report` key.
+
+```javascript
+export default withOnyx({
+    report: {
+        key: ({reportID) => `${ONYXKEYS.COLLECTION.REPORT}${reportID}`,
+    },
+    policy: {
+        key: ({report}) => `${ONYXKEYS.COLLECTION.POLICY}${report.policyID}`,
+    },
+})(App);
+```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -137,6 +137,46 @@ export default withOnyx({
...
+4. Onyx reads the data and updates the state of `withOnyx` with:
+    - `report={{reportID: 1234, policyID: 1, ... the rest of the object ...}}`
+    - `policy={undefined}` (since there is no policy with ID `undefined`)
+5. There is still an `undefined` key in the mapping, so Onyx reads the data again
+6. This time `withOnyx` connects to the key `policies_1` because the `report` object exists in the component's state and it has a `policyID: 1`
+7. Onyx reads the data and updates the state of withOnyx with:
+    - `policy={{policyID: 1, ... the rest of the object ...}`
+8. Now all mappings have values that are defined (not undefined) and the component is rendered with all necessary data
+  
+* It is VERY important to NOT use empty string default values like `report.policyID || ''`. This results in the key returned to `withOnyx` as `policies_` which subscribes to the ENTIRE POLICY COLLECTION and is most assuredly not what you were intending. You can use a default of `0` (as long as you are reasonably sure that there is never a policyID=0). This allows Onyx to return `undefined` as the value of the policy key, which is handled by `withOnyx` appropriately.
```
hayata-suenaga, line 174: > This allows Onyx to return `undefined` as the value of the policy key

does Onxy return `undefined` or `null` when the key was not found? I rememer Onyx gives you `null` in certain cases

mountiny (dismissed):

mountiny:
Looks good to me, thank you!

hayata-suenaga (approved):

README.md
```diff
@@ -137,6 +137,46 @@ export default withOnyx({
...
+
+Background info:
+- The `key` value can be a function that returns the key that Onyx subscribes to
+- The first argument to the `key` function is the `props` from the component
+
+**Detailed explanation of how this is handled and rendered:**
+1. The component mounts with a `reportID={1234}` prop
+2. `withOnyx` evaluates the mapping
+3. `withOnyx` connects to the key `reports_1234` because of the prop passed to the component
+3. `withOnyx` connects to the key `policices_undefined` because `report` doesn't exist in the props yet, so the `policyID` defaults to `undefined`. * (see note below)
```
BeeMargarida, line 164: Thank you, this makes it clear!
If you don't mind me, a small typo here:

```suggestion
3. `withOnyx` connects to the key `policies_undefined` because `report` doesn't exist in the props yet, so the `policyID` defaults to `undefined`. * (see note below)
```

tgolen:
@parasharrajat It functionally results in the same end result, but with two `withOnyx()` you've now double-wrapped the component which is inefficient and doesn't allow batch rendering to work as well (it's just add more things to the rendering tree). The same thing can be done with a single `withOnyx()` and it works.

README.md
tgolen: A little bit of both depending on the context of your question (eg. whether you're talking about `Onyx` or `withOnyx()`)

Onyx will return `undefined` to `withOnyx`, and you can see `withOnyx` checking [here](https://github.com/Expensify/react-native-onyx/blob/main/lib/withOnyx.js#L130-L133) to make sure all values exist or are at least `null`.

Then down [here](https://github.com/Expensify/react-native-onyx/blob/main/lib/withOnyx.js#L209-L210) `withOnyx` will remove all `null` values, which means that the wrapped component will receive it as `undefined`. This is what allows `defaultProps` to work, because if a prop's value is `undefined` then React gets the value from `defaultProps`.

I hope that helps explain it a little!

tgolen:
@mountiny @hayata-suenaga do one of you want to go ahead and merge this?

hayata-suenaga (approved):

PR #353, opened 2023-09-19
Remove null values in nested objects

chrispader:
@tgolen @blazejkustra @marcaaron @thienlnam 

### Details

At the moment, we only remove nullish values (`undefined` and `null`) for top-level keys in objects when merging.

In this [Slack thread](https://expensify.slack.com/archives/C01GTK53T8Q/p1694688747559989) we agreed to also remove nullish values from nested objects.

### Related Issues

https://github.com/Expensify/react-native-onyx/issues/301


tgolen (approved):
I approve this PR. Let's merge as-is.

marcaaron:
I think it needs some more work, no?

chrispader:
LMAO! 😂 still not sure if it's a joke though.

Yes, it needs more work, just added the TODO comment so far...

tgolen:
All tests are passing and there are no performance regressions. I think we're good here 👍

neil-marcellini:
I would like to review this on Monday since it should fix problems I'm encountering.

lib/Onyx.js
blazejkustra, line 1: @chrispader I prepared a [commit](https://github.com/Expensify/react-native-onyx/commit/a0020c6fe175d437bed071f9bbd84d7605b10387) with update to current Onyx.merge typings. Please add it to this PR. @fabioh8010 please have a look if it looks good to you.

blazejkustra (changes requested):

tests/unit/onyxTest.js
```diff
@@ -234,13 +234,31 @@ describe('Onyx', () => {
...
-                expect(testKeyValue).toEqual({test1: 'test1', test2: 'test2'});
+                expect(testKeyValue).toEqual({
+                    test1: {
+                        test2: 'test2',
+                        test3: 'test3',
+                    },
+                });
+                return Onyx.merge(ONYX_KEYS.TEST_KEY, {
+                    test1: {
+                        test2: null,
```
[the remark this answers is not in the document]
chrispader: @marcaaron similar question about `undefined` values as [here](https://github.com/Expensify/react-native-onyx/pull/339#pullrequestreview-1639274665):

Should we also ignore nested `undefined` values and not overwrite the current value or should the current value be overwritten?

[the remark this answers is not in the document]
chrispader: We have an existing test for OnyxCache, which expects `utils.fastMerge` to do nothing, when a nested `undefined` is passed: https://github.com/Expensify/react-native-onyx/blob/eeebcb34b6a4c7b58b53e3e5ee52b84757560e48/tests/unit/onyxCacheTest.js#L312-L322

lib/Onyx.js
fabioh8010: Left some minor comments @blazejkustra , also would be good if we do some testing here and ensure this new typings won't cause regressions.

tgolen (dismissed):

neil-marcellini (dismissed):
Looks good from what I can tell. Thank you.

lib/Onyx.js
blazejkustra: @fabioh8010 Tested it on `CloseAccount` [PR](https://github.com/Expensify/App/pull/27355). Typecheck runs successfully, I also tested it manually and works well!

tests/unit/onyxTest.js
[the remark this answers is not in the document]
chrispader: cc @marcaaron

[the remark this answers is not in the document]
marcaaron: Yes, I think we would just filter them out? Undefined is the _kind of_ the same as an object with no property (maybe it's arguable these are not the same or there should be some other behavior). 

WRT how Onyx should work - I think these are equivalent and both would leave the source object unchanged

```js
merge({a: '1', b: '2'}, {});
merge({a: '1', b: '2'}, {b: undefined});
```

It can work however we want - this is mainly just my opinion because I'd _prefer_ to see consistency and intentionality applied when deleting values or objects fields. If you pass `undefined` somewhere (intentionally or unintentionally) - I'm not really sure what you're trying to achieve. If you pass `null` I can see it right away. 🤷‍♂️

marcaaron (changes requested):

marcaaron:
Could @Szymon20000 take a look as well? 

IIRC they worked on the original `fastMerge()` implementation so it would be good to get some more 👀

blazejkustra:
I had a second look at the types and noticed that changes in `Onyx.merge` function are also connected to `Onyx.mergeCollection` and `Onyx.update` functions.

I prepared another [commit](https://github.com/Expensify/react-native-onyx/commit/9cd3ef70f0ba5946c03d3bff68180b604d506ea4) with adjusted typings, @chrispader please include it in this PR. @fabioh8010 Could you have a look? (I tested it in the App)

fabioh8010:
Looks good @blazejkustra !

lib/utils.js
```diff
@@ -25,16 +25,24 @@ function isMergeableObject(val) {
...
     if (isMergeableObject(target)) {
         // lodash adds a small overhead so we don't use it here
         // eslint-disable-next-line rulesdir/prefer-underscore-method
         const targetKeys = Object.keys(target);
         for (let i = 0; i < targetKeys.length; ++i) {
             const key = targetKeys[i];
+
+            // If shouldRemoveNullObjectValues is true, we want to remove null values from the merged object
+            // eslint-disable-next-line no-continue
+            if (shouldRemoveNullObjectValues && (target[key] == null || source[key] === null)) { continue; }
```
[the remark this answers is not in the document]
chrispader: updated this to use underscore anyway

lib/utils.js
```diff
@@ -25,16 +25,26 @@ function isMergeableObject(val) {
...
     const destination = {};
     if (isMergeableObject(target)) {
         // lodash adds a small overhead so we don't use it here
         // eslint-disable-next-line rulesdir/prefer-underscore-method
         const targetKeys = Object.keys(target);
         for (let i = 0; i < targetKeys.length; ++i) {
             const key = targetKeys[i];
+
+            // If shouldRemoveNullObjectValues is true, we want to remove null values from the merged object
+            if (shouldRemoveNullObjectValues && (_.isNull(target[key]) || _.isNull(source[key]))) {
```
Szymon20000, line 43: Can we remove lodash from this function? Lodash is slow and this function is used everywhere.

lib/utils.js
```diff
@@ -44,54 +54,48 @@ function mergeObject(target, source) {
     const sourceKeys = Object.keys(source);
     for (let i = 0; i < sourceKeys.length; ++i) {
         const key = sourceKeys[i];
-        if (source[key] === undefined) {
+
+        // If shouldRemoveNullObjectValues is true, we want to remove null values from the merged object
+        if (shouldRemoveNullObjectValues && _.isNull(source[key])) {
```
Szymon20000, line 59: Can we remove lodash from this function? Lodash is slow and this function is used everywhere.

Szymon20000 (approved):
Looks good!
I would remove lodash from the fast merge function. That kills Hermes optimisations and we merge quite frequently.

chrispader:
> Looks good!
> I would remove lodash from the fast merge function. That kills Hermes optimisations and we merge quite frequently.

@marcaaron @tgolen i'll let you guys decide if we want to omit `underscore` here. Sounds like the better approach for me

lib/utils.js
tgolen: I think performance is a good concern, I also agree to remove underscore and lodash for parts like this. I've also found those libraries lead to performance problems in low-lying libs like this.

blazejkustra:
This PR is blocking some of the TS migration issues ([example](https://github.com/Expensify/App/pull/27652)). Any idea when we can expect this PR to be merged? @chrispader

chrispader:
@marcaaron @tgolen feel free to merge this whenever you think it's ready 👍 

Just removed the underscore where possbile, so this should be good to go

marcaaron:
>decide if we want to omit underscore here

Sounds good to me 👍

lib/Onyx.js
```diff
@@ -1172,15 +1180,16 @@ function merge(key, changes) {
                 delete mergeQueuePromise[key];
 
                 // After that we merge the batched changes with the existing value
-                const updatedValue = shouldOverwriteExistingValue ? batchedChanges : applyMerge(existingValue, [batchedChanges]);
-                const modifiedData = utils.removeNullObjectValues(updatedValue);
+                // We can remove null values from the "modifiedData", becasue "null" implicates that the user wants to remove a value from storage.
```
marcaaron, line 1183:
```suggestion
                // We can remove null values from the "modifiedData", because "null" implicates that the user wants to remove a value from storage.
```

marcaaron (changes requested):
noticed a small typo otherwise LGTM 👍

chrispader:
> noticed a small typo otherwise LGTM 👍

fixed! @marcaaron

marcaaron (approved):

ospfranco:
Just tried to merge 1.0.96 which includes this changes and the main app is failing.

Oscar Franco Screen 000021

It also causes an issue on app start:

Screenshot_1695972930
Simulator Screenshot - iPhone SE (3rd generation) - 2023-09-29 at 09 02 10

marcaaron:
Sounds good. I merged the revert.

PR #354, opened 2023-09-20
Batch initial hydration

janicduplessis:
### Details

Batch the initial hydration with the same mechanism introduced in https://github.com/Expensify/react-native-onyx/pull/315. This is useful to make rendering of all connected components happen in the same batch. In this cases this reduces renders when opening a new chat in Expensify from ~100 to ~30.

### Related Issues

N/A

### Automated Tests

This doesn't really add new functionality, I just made sure the existing still pass.

### Linked PRs
N/A

### DIFF
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -679,10 +679,10 @@ function keyChanged(key, data, canUpdateSubscriber, notifyRegularSubscibers = tr
  * @param {Function} [mapping.callback]
  * @param {String} [mapping.selector]
  * @param {*|null} val
- * @param {String} matchedKey
- * @param {Boolean} batched
+ * @param {String|undefined} matchedKey
+ * @param {Boolean} isBatched
  */
-function sendDataToConnection(mapping, val, matchedKey, batched) {
+function sendDataToConnection(mapping, val, matchedKey, isBatched) {
     // If the mapping no longer exists then we should not send any data.
     // This means our subscriber disconnected or withOnyx wrapped component unmounted.
     if (!callbackToStateMapping[mapping.connectionID]) {
@@ -703,7 +703,7 @@ function sendDataToConnection(mapping, val, matchedKey, batched) {
         }
 
         PerformanceUtils.logSetStateCall(mapping, null, newData, 'sendDataToConnection');
-        if (batched) {
+        if (isBatched) {
             batchUpdates(() => {
                 mapping.withOnyxInstance.setWithOnyxState(mapping.statePropertyName, newData);
             });
@@ -814,7 +814,8 @@ function connect(mapping) {
             // since there are none matched. In withOnyx() we wait for all connected keys to return a value before rendering the child
             // component. This null value will be filtered out so that the connected component can utilize defaultProps.
             if (matchingKeys.length === 0) {
-                // Here we cannot use batching because the null value is expected to be set immediately so that default props work.
+                // Here we cannot use batching because the null value is expected to be set immediately for default props
+                // or they will be undefined.
                 sendDataToConnection(mapping, null, undefined, false);
                 return;
             }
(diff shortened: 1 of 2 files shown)


lib/Onyx.js
```diff
@@ -766,7 +814,7 @@ function connect(mapping) {
             // since there are none matched. In withOnyx() we wait for all connected keys to return a value before rendering the child
             // component. This null value will be filtered out so that the connected component can utilize defaultProps.
             if (matchingKeys.length === 0) {
-                sendDataToConnection(mapping, null);
+                sendDataToConnection(mapping, null, undefined, false);
```
janicduplessis, line 817: This update should not be batched because we rely on the value being set immediatly for defaultProps

janicduplessis:

lib/Onyx.js
marcaaron: I think it would be good to add this as a comment. Can you write a brief line or two that also explains:

- why we avoid the batch in this case
- what bad thing will happen if we do batch it
- how it relates to `defaultProps`

marcaaron (changes requested):
The changes LGTM overall. Great stuff @janicduplessis 

Just one small request to aid in understanding the case where we do not batch

tgolen (dismissed):
Looks pretty straight forward

jasperhuangg:
It seems like you're in pretty good hands here so I'm going to unassign myself since I don't have the best context on these changes.

lib/Onyx.js
```diff
@@ -639,8 +680,9 @@ function keyChanged(key, data, canUpdateSubscriber, notifyRegularSubscibers = tr
  * @param {String} [mapping.selector]
  * @param {*|null} val
  * @param {String} matchedKey
+ * @param {Boolean} batched
  */
-function sendDataToConnection(mapping, val, matchedKey) {
+function sendDataToConnection(mapping, val, matchedKey, batched) {
```
mountiny, line 685: For the boolean params it might be better to write them as `isBatched`
```suggestion
function sendDataToConnection(mapping, val, matchedKey, isBatched) {
```

lib/Onyx.js
```diff
@@ -639,8 +680,9 @@ function keyChanged(key, data, canUpdateSubscriber, notifyRegularSubscibers = tr
  * @param {String} [mapping.selector]
  * @param {*|null} val
  * @param {String} matchedKey
```
mountiny, line 682: I can see below the `matchedKey` can be `undefined`, should we update the doc here?

lib/Onyx.js
```diff
@@ -49,6 +49,47 @@ let defaultKeyStates = {};
...
+                });
+            });
+
+            resolve();
+        }, 0);
+    });
+    return batchUpdatesPromise;
+}
+
+function batchUpdates(updates) {
```
danieldoglas, line 88:
```suggestion
function queueUpdates(updates) {
```

I think this is a clearer name for this function?

janicduplessis:
@mountiny @marcaaron @tgolen Thanks for the review! I addressed your feedback.

lib/Onyx.js
danieldoglas: but also it is not only queueing, it's queueing and maybe processing. Not sure. Maybe `batchUpdates` is explicative enough?

janicduplessis: This code was actually just moved and I didn't change it so maybe it would be better to keep it as is?

danieldoglas: Yeah, I didn't review the previous PR so I think it's NAB

mountiny (dismissed):
Thanks @janicduplessis

marcaaron (dismissed):

mountiny:
@janicduplessis seems like tests are not happy

janicduplessis:
@mountiny Strange, the tests are passing locally. I added a potential fix, can you run them again?

mountiny:
Running

janicduplessis:
@mountiny Looks like it passed 🎉

tgolen (approved):

danieldoglas (approved):

mountiny (approved):
I see the App is already on 1.0.87 so I think we can go ahead and merge this

PR #357, opened 2023-09-21
Revert "Allow infinite dependent keys when using withOnyx"

tgolen:
Reverts Expensify/react-native-onyx#355

Discussion in https://expensify.slack.com/archives/C01GTK53T8Q/p1695278625487999

This PR is causing extreme performance problems when switching chats.


mountiny (approved):
Straight revert

PR #360, opened 2023-09-22
Make remove key O(log N) instead of O(N^2)

Szymon20000:
I profiled app start with account provided by @danieldoglas. Turned out that remove key from cache takes O(n^2) where n is number of keys we have in cache which is set to 10k. The reason for that is that `_.difference(storageKeys, recentlyAccessed);` is quadratic. It calls filter method and pass there lambda that calls contains. 
https://github.com/jashkenas/underscore/blob/d12221366eacbb63e37cff328bed475f34ebe083/modules/difference.js#L8
Screenshot 2023-09-22 at 10 55 53

Instead we now make use of BTS structure which set is and nice property that we can access first element in O(log n) time. 
Screenshot 2023-09-22 at 10 58 27
Before ^
Screenshot 2023-09-22 at 10 58 49
After ^
### Related Issues

GH_LINK

### DIFF
--- a/lib/OnyxCache.js
+++ b/lib/OnyxCache.js
@@ -181,9 +181,9 @@ class OnyxCache {
     removeLeastRecentlyUsedKeys() {
         while (this.recentKeys.size > this.maxRecentKeysSize) {
             const iterator = this.recentKeys.values();
-            const value = iterator.next().value
+            const value = iterator.next().value;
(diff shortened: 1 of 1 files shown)


mountiny (approved):
This makes sense to me, thanks!

tgolen (approved):

PR #362, opened 2023-09-22
Fix getAllKeys when keys are removed from cache

janicduplessis:
### Details

When testing an account with a very large amount of chats we hit an issue where onyx will return undefined for an item that actually exists in storage. This is because that item was evicted from Onyx cache.

When items are evicted from cache we should not delete the key in `storageKeys` since this is used to determine if the item exists in storage.

Onyx uses that value to get all keys as soon as it has some items (https://github.com/Expensify/react-native-onyx/blob/main/lib/Onyx.js#L115). This means `getAllKeys` from cache needs to return all keys in Storage, not keys currently cached.

### Related Issues
N/A

### Automated Tests

Updated a test to add a check for getAllKeys after keys are evicted from cache.

### DIFF
--- a/lib/OnyxCache.js
+++ b/lib/OnyxCache.js
@@ -183,7 +183,8 @@ class OnyxCache {
             const iterator = this.recentKeys.values();
             const value = iterator.next().value;
             if (value !== undefined) {
-                this.drop(value);
+                delete this.storageMap[value];
+                this.recentKeys.delete(value);
             }
         }
     }
--- a/tests/unit/onyxCacheTest.js
+++ b/tests/unit/onyxCacheTest.js
@@ -647,6 +647,9 @@ describe('Onyx', () => {
                     expect(cache.hasCacheForKey('key2')).toBe(true);
                     expect(cache.hasCacheForKey('key3')).toBe(true);
                     expect(cache.hasCacheForKey('key4')).toBe(true);
+
+                    // getAllKeys should still return all keys since they exist in storage.
+                    expect(cache.getAllKeys().length).toEqual(9);
                 });
         });
     });


arosiclair:
Is this change coming from slack or an issue somewhere? I don't have any context for this

janicduplessis:
https://expensify.slack.com/archives/C035J5C9FAP/p1695378712978879 Has a little bit of context.

The issue is that when the cache is full keys are removed from the array returned by getAllKeys and then some items will start to return null even if they still exist in the db and could be fetched. This is because we return the keys from cache as soon as we have one.

If you have an account with very large amount of chats, or by reducing cache size to a small value you should be able to reproduce a crash by scrolling in the chats list.

arosiclair:
Thanks for the context

> If you have an account with very large amount of chats, or by reducing cache size to a small value you should be able to reproduce a crash by scrolling in the chats list.

Can you write specific testing steps for this? I'll verify that it's also working on my end

janicduplessis:
@arosiclair The easiest way to repro without an account with a very large amount of data is to change the max cache size [here](https://github.com/Expensify/react-native-onyx/blob/310cd7ea1879424f83afb375dd8ac035058c4c33/lib/OnyxCache.js#L196) to a small value like 30. Then when loading the app you should see the following error.

image

After this patch everything should work.

danieldoglas (approved):

mountiny (approved):
@arosiclair are you happy to merge this? I think this can be tested with the App and the automated test added is handy here

mountiny:
conflicts

arosiclair (approved):
Alright let's just rely on unit tests for now. 

I do feel that we should probably increase the required testing for these Onyx core changes. I've seen regressions slip through from these ([example](https://github.com/Expensify/App/issues/26190)) and they are naturally very hard to trace when they happen. Anyway that's a discussion for another day

mountiny:
> I do feel that we should probably increase the required testing for these Onyx core changes. I've seen regressions slip through from these (https://github.com/Expensify/App/issues/26190) and they are naturally very hard to trace when they happen. Anyway that's a discussion for another day

I agree the current set up makes it hard to test the changes before actaully merging Onyx PR hard and it would be great to streamline this

mountiny:
merged as part of https://github.com/Expensify/react-native-onyx/pull/363

PR #363, opened 2023-09-23
Don't recreate iterator every time

Szymon20000:
Turned out that during chat switching we have cases when we remove several items at the same time. 
In those cases we don't want to recreate iterator. 

### Related Issues

GH_LINK

### DIFF
--- a/lib/OnyxCache.js
+++ b/lib/OnyxCache.js
@@ -179,18 +179,16 @@ class OnyxCache {
      * Remove keys that don't fall into the range of recently used keys
      */
     removeLeastRecentlyUsedKeys() {
-        let toRemove = this.recentKeys.size - this.maxRecentKeysSize;
-        if (toRemove <= 0) {
+        let numKeysToRemove = this.recentKeys.size - this.maxRecentKeysSize;
+        if (numKeysToRemove <= 0) {
             return;
         }
         const iterator = this.recentKeys.values();
         const temp = [];
-        while (toRemove > 0) {
+        while (numKeysToRemove > 0) {
             const value = iterator.next().value;
-            if (value !== undefined) {
-                temp.push(value);
-            }
-            toRemove--;
+            temp.push(value);
+            numKeysToRemove--;
         }
 
         for (let i = 0; i < temp.length; ++i) {
--- a/lib/OnyxCache.js
+++ b/lib/OnyxCache.js
@@ -179,12 +179,23 @@ class OnyxCache {
      * Remove keys that don't fall into the range of recently used keys
      */
     removeLeastRecentlyUsedKeys() {
-        while (this.recentKeys.size > this.maxRecentKeysSize) {
-            const iterator = this.recentKeys.values();
+        let toRemove = this.recentKeys.size - this.maxRecentKeysSize;
+        if (toRemove <= 0) {
+            return;
+        }
+        const iterator = this.recentKeys.values();
+        const temp = [];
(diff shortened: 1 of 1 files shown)


lib/OnyxCache.js
```diff
@@ -179,12 +179,23 @@ class OnyxCache {
...
             if (value !== undefined) {
-                this.drop(value);
+                temp.push(value);
             }
+            toRemove--;
+        }
+
+        for (let i = 0; i < temp.length; ++i) {
+            delete this.storageMap[temp[i]];
+            this.recentKeys.delete(temp[i]);
```
francoisl, line 198: Out of curiosity, why not use `drop()` here?

francoisl (commented):
Makes sense, just one small question to clarify

danieldoglas (dismissed):
LGTM

lib/OnyxCache.js
Szymon20000: We don't want to remove key from storageKeys as that's used for getAllKeys. We don't remove the key from the db only from cache.

Szymon20000: https://github.com/Expensify/react-native-onyx/pull/362

mountiny (approved):
Makes sense to me, thanks!

lib/OnyxCache.js
```diff
@@ -179,12 +179,23 @@ class OnyxCache {
      * Remove keys that don't fall into the range of recently used keys
      */
     removeLeastRecentlyUsedKeys() {
-        while (this.recentKeys.size > this.maxRecentKeysSize) {
-            const iterator = this.recentKeys.values();
+        let toRemove = this.recentKeys.size - this.maxRecentKeysSize;
```
roryabraham, line 182:
```suggestion
        let numKeysToRemove = this.recentKeys.size - this.maxRecentKeysSize;
```

lib/OnyxCache.js
```diff
@@ -179,12 +179,23 @@ class OnyxCache {
...
+        if (toRemove <= 0) {
+            return;
+        }
+        const iterator = this.recentKeys.values();
+        const temp = [];
+        while (toRemove > 0) {
             const value = iterator.next().value;
             if (value !== undefined) {
```
roryabraham, line 190: If `value` is undefined, should we `break`?

lib/OnyxCache.js
```diff
@@ -179,12 +179,23 @@ class OnyxCache {
...
-                this.drop(value);
+                temp.push(value);
             }
+            toRemove--;
+        }
+
+        for (let i = 0; i < temp.length; ++i) {
```
roryabraham, line 196: Why aggregate `temp` in a separate array? Could we just do these two lines inline instead of in a separate loop?

```diff
- temp.push(value);
+ delete this.storageMap[value];
+ this.recentKeys.delete(value);
```

roryabraham (changes requested):

lib/OnyxCache.js
Szymon20000: because it modifies the bst and most likely breaks iterator

lib/OnyxCache.js
Szymon20000: it shouldn't every be undefined. I will remove it.

danieldoglas (approved):

mountiny (approved):
@roryabraham all yours

roryabraham (approved):

PR #364, opened 2023-09-23
Use fast path when we ask for just one key in Connect()

Szymon20000:
We found out that getAllKeys can be extremely slow as it's used very frequently and it just copies keys from cache to just created array. Because on heavy account number of keys can easily reach 50k that takes quite some time. 
In most of the cases we only ask for one keys so to speed it up we just check if it's in the set of keys if it is then we return array with just this item. 
### Related Issues

GH_LINK

### DIFF
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -755,8 +755,12 @@ function connect(mapping) {
     deferredInitTask.promise
         .then(() => addKeyToRecentlyAccessedIfNeeded(mapping))
         .then(() => {
-            if (typeof mapping?.key === 'string' && !(mapping?.key.endsWith('_')) && cache.storageKeys.has(mapping?.key)) {
-                return [mapping.key]
+            if (Boolean(mapping.key)
+                && typeof mapping.key === 'string'
+                && !(mapping.key.endsWith('_'))
+                && cache.storageKeys.has(mapping.key)
+            ) {
+                return [mapping.key];
             }
             return getAllKeys();
         })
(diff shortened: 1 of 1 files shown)


lib/Onyx.js
```diff
@@ -754,7 +754,16 @@ function connect(mapping) {
...
+        .then(() => {
+            if (Boolean(mapping.key)
+                && typeof mapping.key === 'string'
+                && !(mapping.key.endsWith('_'))
+                && cache.storageKeys.has(mapping.key)
+            ) {
+                return [mapping.key];
+            }
+            return getAllKeys();
+        })
```
mountiny, line 766: I think it would be handy to leave a short comment about this.

mountiny (changes requested):

mountiny (approved):
LGTM, thanks!

lib/Onyx.js
```diff
@@ -754,7 +754,19 @@ function connect(mapping) {
...
+        .then(() => {
+            // Performance improvement
+            // If the mapping is connected to an onyx key that is not a collection
+            // we can skip the call to getAllKeys() and return an array with a single item
+            if (Boolean(mapping.key)
+                && typeof mapping.key === 'string'
+                && !(mapping.key.endsWith('_'))
+                && cache.storageKeys.has(mapping.key)
+            ) {
+                return [mapping.key];
```
marcaaron, line 766: NAB, but now that we do this - some of the comments below are a little misleading. Since this is not a key that we found (we just assumed it exists).

PR #365, opened 2023-09-23
Use map for keeping pending promises

Szymon20000:
Map is faster than object when we have a lot of items. Using map in this place seems to improve chat switching performance.
### Related Issues

GH_LINK

### DIFF
--- a/lib/OnyxCache.js
+++ b/lib/OnyxCache.js
@@ -34,7 +34,9 @@ class OnyxCache {
         /**
          * @private
          * Captured pending tasks for already running storage methods
-         * @type {Record<string, Promise>}
+         * Using a map yields better performance on operations such a delete
+         * https://www.zhenghao.io/posts/object-vs-map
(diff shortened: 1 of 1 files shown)


lib/OnyxCache.js
```diff
@@ -36,7 +36,7 @@ class OnyxCache {
          * Captured pending tasks for already running storage methods
          * @type {Record<string, Promise>}
          */
-        this.pendingPromises = {};
+        this.pendingPromises = new Map();
```
mountiny, line 39: Could you update the docs above and leave a comment why we chose to use Map?

ospfranco: What do you mean with the docs though? Docs are only generated from onyx.js, and not from other internal implementation files such as this one

mountiny: `@type {Record<string, Promise>}` I meant this

mountiny (approved):
Thanks!

PR #368, opened 2023-09-26
Prevent assignment of values to undefined 'tempState' property

getusha:
@roryabraham @parasharrajat @aimane-chnaif 

### Details

We faced a problem after updating to `react-native-web 0.19.7`, where an attempt to assign a value to a property results in a console error because `tempState` is undefined.

`this.tempState` is deleted here which makes it undefined.

https://github.com/Expensify/react-native-onyx/blob/82b5372aa6fef416408bbe71986029e18793e43a/lib/withOnyx.js#L168

when `this.state.loading` is true and `this.tempState` is undefined  we get the following console error while trying to set property of undefined.

https://github.com/Expensify/react-native-onyx/blob/82b5372aa6fef416408bbe71986029e18793e43a/lib/withOnyx.js#L159

```bash
Uncaught (in promise) TypeError: Cannot set properties of undefined (setting statePropertyName)
```

This simple solution will stop this error from happening.

### Related Issues

GH_LINK

https://github.com/Expensify/App/issues/16660

### Linked PRs

https://github.com/Expensify/App/pull/24482

### DIFF
--- a/lib/withOnyx.js
+++ b/lib/withOnyx.js
@@ -146,7 +146,7 @@ export default function (mapOnyxToState, shouldDelayUpdates = false) {
                 const prevValue = this.state[statePropertyName];
 
                 // If the component is not loading (read "mounting"), then we can just update the state
-                if (!this.state.loading) {
+                if (!this.state.loading || !this.tempState) {
                     // Performance optimization, do not trigger update with same values
                     if (prevValue === val || utils.areObjectsEmpty(prevValue, val)) {
                         return;


getusha:

getusha:
recheck

roryabraham (approved):

roryabraham:
I also don't have full context about all the recent changes in Onyx, going to need to spend more time with this PR

getusha:
@parasharrajat added more details to help you understand the issue.

> I also don't have full context about all the recent changes in Onyx, going to need to spend more time with this PR

@roryabraham I don't think this change will make a big difference/have other side-effects because if this.tempState is undefined, we shouldn't be trying to set a value or access a key from it.

I am still investigating why `this.state.loading` is true as it should always be false when this.tempState is undefined beforehand. should we tag someone who is involved in recent changes?

getusha:
This change is already made https://github.com/Expensify/react-native-onyx/pull/376

PR #369, opened 2023-09-26
Propagate merge events across tabs

ospfranco:
### Details

There is currently [a bug](https://github.com/Expensify/App/issues/26190) with data propagation across tabs. After some debugging I found out that `merge` events are not propagated across tabs. This fixes the issue, but not sure if it is the correct solution, since it will propagate all merge events across tabs, which might increase the loads on the pages. A review is required.

### Related Issues

https://github.com/Expensify/App/issues/26190


luacmartins (approved):
LGTM and I think this is what we want

luacmartins:
@ospfranco wanna work on the App version bump?

ospfranco:
Will do

PR #376, opened 2023-09-29
Fix race condition when tempState is deleted before loading flag is set to false

ospfranco:
### Details

There is a small race condition. When calling setWithOnyxState we delete the `tempState` object that is used to hold temporary state updates while the HOC is gathering data. However the loading flag is only set on the `setState` callback down below. setState however is an async operation that is also batched, therefore there is a small window of time where the loading flag is not false but the tempState is already gone (while the update is queued and waiting to be applied). This simply bypasses the loading check if the tempState is gone and the update can be safely queued with a normal `setStateProxy`.

### Related Issues

GH_LINK https://github.com/Expensify/App/issues/28419

### DIFF
--- a/lib/withOnyx.js
+++ b/lib/withOnyx.js
@@ -146,7 +146,11 @@ export default function (mapOnyxToState, shouldDelayUpdates = false) {
                 const prevValue = this.state[statePropertyName];
 
                 // If the component is not loading (read "mounting"), then we can just update the state
-                if (!this.state.loading) {
+                // There is a small race condition here. On line 172 the setState is used with a callback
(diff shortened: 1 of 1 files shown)


mountiny (dismissed):
Thanks! This looks good but I ma not fan of the line numbers in the comment, that will inevitably change and become out of date. Could you please rewrite it without the numbers

mountiny (approved):
Thanks!

getusha:
@mountiny we were also attempting to get this change merged here https://github.com/Expensify/react-native-onyx/pull/368 😄, anyway thanks for handling it.

mountiny:
Ah yeah, saw that, this was faster for our purposes :D

PR #382, opened 2023-10-03
#15321: Only one tab writes to the DB

BeeMargarida:
### Details

This is the 1st part solution to fix the problem of the linked issue. The links to the broader discussion are below. This PR is big because it contains a RN app that contains an example for the e2e tests.

The goal of this PR is to change Onyx execution so that when there are several tabs open, only one of them can write to the DB. This is done by defining a leader tab, and all other tabs defer their writes to this one leader.

Changes included in this PR:
- All writes to the DB are made by the leader tab - this involves using the BroadcastAPI to communicate each set, multiSet, merge and clear to the leader tab.
- Ported `ActiveClientManager` to react-native-onyx and used `BroadcastAPI` instead of the indexedDB to reach a consensus between tabs about who is the leader (the last tab to be open). It uses timestamps in the messages to check that the last tab to send the message is the leader. If for a change the timestamps are the same, the client ID of the tab is used as a tiebreaker.
- When the leader tab is closed, a new consensus is reached about the new leader tab. The old leader tab communicates that it is closing so that the others can try to reach a consensus.
-  The leader tab is listening to all write requests from the other tabs and execute them once they are received.
- The leader tab does not execute writes to the DB while clear is running. These writes are ignored while the clear is running
- The callback for the Onyx.clear is communicated. This is necessary to allow actions to execute after clear, so a non-leader needs a way of executing those actions without having access to the promise. So, a new Onyx.onClear should be set with the callback to execute after the Onyx.clear completes.

The GH comment with the explanation: https://github.com/Expensify/App/issues/15321#issuecomment-1677179697

The doc with the discussion: https://docs.google.com/document/d/1Psr4Q4A2uH2b7l9CKBuB_dBT-15V4lJ7SOaH4uQOve4/edit

### Related Issues

GH_LINK: https://github.com/Expensify/App/issues/15321

### Automated Tests

Added e2e tests using playwright. To run them:

- `npm install` (should install onyx dependencies and the e2e app as well)
- `npm run e2e` or `npm run e2e-ui` to run the e2e tests

Current result:

Screenshot 2023-10-03 at 11-05-00 Playwright Test Report

### Linked PRs

This will require future changes in newDot, since `ActiveClientManager` is now available in react-native-onyx.


BeeMargarida:

lib/ActiveClientManager/index.web.js
```diff
@@ -0,0 +1,95 @@
...
+ * Subscribe to the broadcast channel to listen for messages from other tabs, so that
+ * all tabs agree on who the leader is, which should always be the last tab to open.
+ */
+function init() {
+    Broadcast.subscribe((message) => {
+        switch (message.data.type) {
+            case NEW_LEADER_MESSAGE:
+                // Only update the active leader if the message received was from another
+                // tab that initialized after the current one
+                if (clientID === activeClient && ((timestamp > message.data.timestamp) || (timestamp === message.data.timestamp && clientID > message.data.clientID))) {
```
rezkiy37, line 58: Can we create variables to simplify a condition. It looks complex. 
```javascript
const isClientTheLeader = isClientTheLeader()
const isCurrentTimestampOlder = timestamp > message.data.timestamp
const isCurrentTimestampSame = timestamp === message.data.timestamp
``` 
Something like that, names can be different.

lib/ActiveClientManager/index.web.js
```diff
@@ -0,0 +1,95 @@
...
+                    return;
+                }
+                activeClient = message.data.clientID;
+
+                subscribers.forEach(callback => callback());
```
rezkiy37, line 63:
```suggestion
                subscribers.forEach(callback);
```

lib/ActiveClientManager/index.web.js
```diff
@@ -0,0 +1,95 @@
...
+                break;
+            case REMOVED_LEADER_MESSAGE:
+                activeClient = clientID;
+                timestamp = Date.now();
+                Broadcast.sendMessage({type: NEW_LEADER_MESSAGE, clientID, timestamp});
+                subscribers.forEach(callback => callback());
```
rezkiy37, line 69:
```suggestion
                subscribers.forEach(callback);
```

lib/ActiveClientManager/index.web.js
rezkiy37, line 63:
```suggestion
                subscribers.forEach(callback);
```

lib/Onyx.js
```diff
@@ -1049,7 +1068,7 @@ function set(key, value) {
     const updatePromise = broadcastUpdate(key, valueWithNullRemoved, hasChanged, 'set');
 
     // If the value has not changed, calling Storage.setItem() would be redundant and a waste of performance, so return early instead.
-    if (!hasChanged) {
+    if (!hasChanged || ongoingClear) {
```
rezkiy37, line 1071: Looks like the second operand never works here, because of a condition above, where you already checked `ongoingClear` equals `true`.

lib/Onyx.js
```diff
@@ -1189,7 +1226,7 @@ function merge(key, changes) {
                 const updatePromise = broadcastUpdate(key, modifiedData, hasChanged, 'merge');
 
                 // If the value has not changed, calling Storage.setItem() would be redundant and a waste of performance, so return early instead.
-                if (!hasChanged) {
+                if (!hasChanged || ongoingClear) {
```
rezkiy37, line 1229: Looks like the second operand never works here, because of a condition above, where you already checked `ongoingClear` equals `true`.

lib/ActiveClientManager/index.d.ts
```diff
@@ -0,0 +1,22 @@
+/**
+ * Determines when the client is ready. We need to wait till the init method is called and the leader message is sent.
+ */
+declare function isReady(): Promise<void>;
+
+/**
+ * Subscribe to the broadcast channel to listen for messages from other tabs, so that
```
rezkiy37, line 7:
```suggestion
 * Subscribes to the broadcast channel to listen for messages from other tabs, so that
```

lib/ActiveClientManager/index.web.js
```diff
@@ -0,0 +1,95 @@
...
+
+const NEW_LEADER_MESSAGE = 'NEW_LEADER';
+const REMOVED_LEADER_MESSAGE = 'REMOVE_LEADER';
+
+const clientID = Str.guid();
+const subscribers = [];
+let timestamp = null;
+
+let activeClient = null;
+let resolveSavedSelfPromise;
```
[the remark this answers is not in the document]
BeeMargarida: Hum, I'll perhaps set the default value as an empty function

lib/broadcast/index.web.js
```diff
@@ -0,0 +1,42 @@
...
+    channel.onmessage = (message) => {
+        subscriptions.forEach(c => c(message));
+    };
+}
+
+/**
+ * Unsubscribes from the broadcast channel.
+ */
+function unsubscribe() {
+    subscriptions = [];
```
rezkiy37, line 30: Just to clarify: should we unsubscribe from all channels, or a specific one?

playwright.config.js
```diff
@@ -0,0 +1,61 @@
...
+            name: 'firefox',
+            use: {...devices['Desktop Firefox']},
+        },
+
+        /*
+        {
+            name: 'webkit',
+            use: {...devices['Desktop Safari']},
+        },
+        */
```
rezkiy37, line 51: If we need it, let's add a comment how to use it.
```suggestion
```

tests/e2e/app/package.json
```diff
@@ -0,0 +1,40 @@
...
+  "scripts": {
+    "android": "react-native run-android",
+    "ios": "react-native run-ios",
+    "web": "craco start",
+    "start": "react-native start"
+  },
+  "dependencies": {
+    "react": "18.2.0",
+    "react-dom": "^18.2.0",
+    "react-native": "0.71.2",
```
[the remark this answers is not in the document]
BeeMargarida: I used this one because it still allows for node 16 and I wanted to match the one in package.json, so that no weird behaviour happens.

tests/e2e/simple.spec.ts
```diff
@@ -0,0 +1,42 @@
+/* eslint-disable @lwc/lwc/no-async-await */
+import {test, expect} from '@playwright/test';
+
+test.describe('simple', () => {
```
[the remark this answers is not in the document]
BeeMargarida: I'll rename to tab, since it's only tests with only one tab

package.json
```diff
@@ -37,7 +37,10 @@
     "lint-tests": "eslint tests/**",
     "test": "jest",
     "build": "webpack --config webpack.config.js",
```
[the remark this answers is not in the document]
BeeMargarida: Nope, only the code

tests/e2e/app/src/App.js
```diff
@@ -0,0 +1,40 @@
...
+        <View>
+            {isAuthenticated && (
+                <Text aria-label="logged-in">You are logged in</Text>
+            )}
+            <Main />
+        </View>
+    );
+}
+
+App.propTypes = propTypes;
```
rezkiy37, line 33: Let's add `displayName`.

tests/e2e/app/src/App.js
```diff
@@ -0,0 +1,40 @@
...
+
+const defaultProps = {
+    session: {},
+};
+
+function App(props) {
+    const isAuthenticated = useMemo(
+        () => Boolean(lodashGet(props.session, 'login', null)),
+        [props.session],
+    );
```
rezkiy37, line 21: `useMemo` looks like unnecessary here. 
```suggestion
    const isAuthenticated = Boolean(lodashGet(props.session, 'login', null));
```

tests/e2e/app/src/Main.js
```diff
@@ -0,0 +1,175 @@
...
+        flexDirection: 'row',
+        justifyContent: 'center',
+        alignItems: 'center',
+        columnGap: 20,
+    },
+});
+
+function Main(props) {
+    const [isLeader, setIsLeader] = useState(
+        Onyx.isClientTheLeader ? Onyx.isClientTheLeader() : false,
```
rezkiy37, line 48: Is it possible that `Onyx.isClientTheLeader` does not exist? AFAIK, it is a noop, but exists.

tests/e2e/app/src/Main.js
```diff
@@ -0,0 +1,175 @@
...
+    );
+
+    const isAuthenticated = useMemo(
+        () => Boolean(lodashGet(props.session, 'login', null)),
+        [props.session],
+    );
```
rezkiy37, line 54:
```suggestion
    const isAuthenticated = Boolean(lodashGet(props.session, 'login', null));
```

tests/e2e/app/src/Main.js
```diff
@@ -0,0 +1,175 @@
...
+            </Text>
+            <Text aria-label="data-meteorites" numberOfLines={10}>
+                {_.filter(_.map(Object.entries(props.allMeteorites || {}),
+                    ([key, value]) => (value ? `${key}-${value && value.length}` : undefined)), v => !!v)
+                    .join(',')}
+            </Text>
+            <Text aria-label="data-asteroids" numberOfLines={10}>
+                {_.filter(_.map(Object.entries(props.allAsteroids || {}),
+                    ([key, value]) => (value ? `${key}-${value && value.length}` : undefined)), v => !!v)
+                    .join(',')}
```
rezkiy37, line 142: Looks like it is identical. Can we create a function to re-use?

tests/e2e/app/src/Main.js
```diff
@@ -0,0 +1,175 @@
...
+                numberOfLines={10}
+            >
+                {JSON.stringify(Updates.updates)}
+            </Text>
+            <Button title="Clear updates" onPress={Updates.clear} />
+        </View>
+    );
+}
+
+Main.propTypes = propTypes;
```
rezkiy37, line 156: Let's add `displayName`.

lib/ActiveClientManager/index.web.js
BeeMargarida: This would require defining this variables outside the switch, since it throws an eslint error `Unexpected lexical declaration in case block. eslint(no-case-declarations)`.

rezkiy37 (changes requested):
Left comments. However, in general, great job!

lib/ActiveClientManager/index.web.js
rezkiy37: We can create safely variable in a **case** block via `{}`:
```javascript
case "a": {
  const name = ""
  ....
}
```

rezkiy37: https://eslint.org/docs/latest/rules/no-case-declarations

BeeMargarida: TIL 🚀

lib/ActiveClientManager/index.web.js
BeeMargarida: That gives out 'callback is not defined', since it comes from the foreach loop

lib/Onyx.js
BeeMargarida: This one can be removed

lib/Onyx.js
BeeMargarida: This one might, because the `ongoingClear` might be false when the merge was called, but true by the time the `get` finishes executing

lib/broadcast/index.web.js
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * Subscribes to the broadcast channel. Every time a new message
+ * is received, the callback is called.
+ * @param {Function} callback
+ */
+function subscribe(callback) {
+    subscriptions.push(callback);
+    channel.onmessage = (message) => {
+        subscriptions.forEach(c => c(message));
```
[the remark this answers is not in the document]
BeeMargarida: Used this one so that it doesn't clash with the callback name above

lib/broadcast/index.web.js
BeeMargarida: This is actually not being used, but it should might be required in a possible cleanup of Onyx. I added this here by completeness, but I can actually remove it.

playwright.config.js
BeeMargarida: What do you mean? To use it is just to comment out this. Right now is commented because there's an issue with running with multiple browser contexts in WebKit.

tests/e2e/app/src/Main.js
BeeMargarida: Ah, this is dead code, thank you for pointing it out

tests/e2e/app/src/Main.js
BeeMargarida: I don't think it's worth the effort, it's just two entries, in a test app. Re-using would create more code and be more confusing

playwright.config.js
rezkiy37: Ah, okay, I meant if you are going to merge PR with the commented out code, let's left a comment like: "there's an issue with running with multiple browser contexts in WebKit."

lib/broadcast/index.web.js
rezkiy37: Let's keep it I think

rezkiy37: yeah, it is up to you

playwright.config.js
BeeMargarida: Done 👍

rezkiy37 (approved):
LGTM

fabioh8010 (approved):

marcochavezf:
Assigning @roryabraham as reviewer since he was assigned to the linked GH issue and has more context about the solution. cc @tienifr

koko57:

koko57:
conflicts fixed

koko57:
recheck

roryabraham (commented):
Can we please add some information to the README about the e2e tests and how to run them?

roryabraham:
Also @koko57 it looks like you need to sign the CLA

lib/Onyx.js
```diff
@@ -1525,6 +1619,21 @@ function init({
...
+        if (!ActiveClientManager.isClientTheLeader()) {
+            return;
+        }
+        subscribeToEvents();
+    });
+
+    // If the active client changes an the current client is the leader,
+    // subscribes to the events
+    ActiveClientManager.subscribeToClientChange(() => {
+        subscribeToEvents();
```
roryabraham, line 1634: I'm confused why we would need this, and concerned that we're adding subscriptions without every cleaning them up

lib/broadcast/index.web.js
```diff
@@ -0,0 +1,35 @@
+const BROADCAST_ONYX = 'BROADCAST_ONYX';
+
+const subscriptions = [];
```
roryabraham, line 3: There are many cases when we're adding subscriptions, but we never actually remove them. This means as you use the app this array will grow without bounds.

To address this, we should have `subscribe` return an `unsubscribe` cleanup function that removes the callback from the array:

```js
/**
 * Subscribes to the broadcast channel. Every time a new message
 * is received, the callback is called.
 *
 * @param {Function} callback
 * @returns {Function} unsubscribe
 */
function subscribe(callback) {
    subscriptions.push(callback);
    channel.onmessage = (message) => {
        subscriptions.forEach(c => c(message));
    };
    return () => {
        const index = subscriptions.indexOf(callback);
        subscriptions.splice(index, 1);
    };
}
```

Then in all places where this was used, keep a reference to all the cleanup functions and call them in a `beforeunload` handler.

tests/e2e/app/android/app/build.gradle
```diff
@@ -0,0 +1,123 @@
+apply plugin: "com.android.application"
+apply plugin: "com.facebook.react"
+
+/**
+ * This is the configuration block to customize your React Native Android app.
+ * By default you don't need to apply any configuration, just uncomment the lines you need.
+ */
+react {
+    /* Folders */
```
roryabraham, line 9: NAB Let's delete any unused boilerplate comments

tests/e2e/app/android/app/proguard-rules.pro
```diff
@@ -0,0 +1,10 @@
+# Add project specific ProGuard rules here.
```
roryabraham, line 1: Can we delete this file, along with any other boilerplate ones that aren't used?

tests/e2e/app/android/app/src/main/java/com/app/MainActivity.java
```diff
@@ -0,0 +1,32 @@
+package com.app;
+
+import com.facebook.react.ReactActivity;
+import com.facebook.react.ReactActivityDelegate;
+import com.facebook.react.defaults.DefaultNewArchitectureEntryPoint;
+import com.facebook.react.defaults.DefaultReactActivityDelegate;
+
+public class MainActivity extends ReactActivity {
```
roryabraham, line 8: Is Playwright current running on all platforms, or just on web?

roryabraham (changes requested):
Let's also hook up the e2e tests with our CI by adding `.github/workflows/e2e.yml` with contents something like this:

```yml
name: Test JavaScript

on:
  pull_request:
    types: [opened, synchronize]
    paths: ['tests/e2e/**', 'lib/**', 'package.json', 'package-lock.json']

jobs:
  test:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v3

      - uses: actions/cache@v3
        with:
          path: ~/.npm
          key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
          restore-keys: ${{ runner.os }}-node-

      - name: Setup Node
        uses: actions/setup-node@v3
        with:
          node-version-file: '.nvmrc'

      - run: npm ci

      - run: npx playwright install

      - run: npm run e2e
        env:
          CI: true

```

tests/e2e/app/android/app/src/main/java/com/app/MainActivity.java
koko57: no, it runs just for web

roryabraham: Right, so I'm wondering if we should get rid of all the boilerplate from the native app?

koko57:
@roryabraham I need to consult a few things to address the other comments, but I'll do my best to do it asap.

roryabraham:
@koko57 if you have questions or need clarification on any of my suggestions just let me know 🙂

koko57:
@roryabraham Sorry, I had to take care of my other issues, but I'll address the rest of the comments soon

lib/Onyx.js
```diff
@@ -1060,6 +1070,15 @@ function removeNullValues(key, value) {
  * @returns {Promise}
  */
 function set(key, value) {
+    if (!ActiveClientManager.isClientTheLeader()) {
+        Broadcast.sendMessage({type: METHOD.SET, key, value});
```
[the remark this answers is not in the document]
koko57: It would be ok in terms of separation of concerns, but still all the tabs would have to be connected to both channels (sending the data and leader election; we cannot subscribe only the leader to the data channel and unsubscribe the others).

roryabraham:
@koko57 thanks for pushing this forward. Please ping me when it's ready for review

lib/Onyx.js
koko57: I think we don't need it at all. Every time we opened or closed a tab (yes, it pushed a callback when a tab was closed instead of removing it)  it added the same [callback](https://github.com/Expensify/react-native-onyx/blob/8f4b3b0554b8955db7783001c1a72f3275e9b584/lib/Onyx.js#L1559-L1585) to the subscriptions array. And every time a message was received it run all the callbacks in the array were called. For non-leader tabs it returned early so it wasn't a problem but if there was i.e. 8 callbacks in the subscriptions array it would be called 8 times. For a new tab aka the leader tab we have always 2 different subscriptions and it's ok.

I removed this part of code and it works fine. For non leader tabs we also have only two callbacks in the subscriptions array. So I think now we don't need to add the unsubscribe logic here anymore.

koko57:
@roryabraham I think it's ready for re-review. On separate channels I commented here https://github.com/Expensify/react-native-onyx/pull/382#discussion_r1392757436 - do we really need it? For the unsubscribe logic - as we no longer need to subscribe so many times, I also think it's unnecessary: https://github.com/Expensify/react-native-onyx/pull/382#discussion_r1394284694. Let me know what do you think about it 🙂

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,31 @@
...
+on:
+    pull_request:
+        types: [opened, synchronize]
+        paths: ['tests/e2e/**', 'lib/**', 'package.json', 'package-lock.json']
+
+jobs:
+    test:
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@v3
```
roryabraham, line 12: Let's use v4 now since Onyx is on Node 20:

```suggestion
            - uses: actions/checkout@v4
```

lib/ActiveClientManager/index.web.js
```diff
@@ -0,0 +1,99 @@
...
+
+const NEW_LEADER_MESSAGE = 'NEW_LEADER';
+const REMOVED_LEADER_MESSAGE = 'REMOVE_LEADER';
+
+const clientID = Str.guid();
+const subscribers = [];
+let timestamp = null;
+
+let activeClient = null;
+let resolveSavedSelfPromise = () => {};
```
roryabraham, line 18:
```suggestion
let setIsReady = () => {};
```

lib/ActiveClientManager/index.web.js
```diff
@@ -0,0 +1,99 @@
...
+const savedSelfPromise = new Promise((resolve) => {
```
roryabraham, line 19:
```suggestion
const isReadyPromise = new Promise((resolve) => {
```

lib/ActiveClientManager/index.web.js
```diff
@@ -0,0 +1,99 @@
...
+import * as Broadcast from '../broadcast';
+
+const NEW_LEADER_MESSAGE = 'NEW_LEADER';
+const REMOVED_LEADER_MESSAGE = 'REMOVE_LEADER';
+
+const clientID = Str.guid();
+const subscribers = [];
+let timestamp = null;
+
+let activeClient = null;
```
roryabraham, line 17:
```suggestion
let activeClientID = null;
```

koko57:
> I still think we need to add some unsubscribe logic. Even if there are fewer subscriptions, we're never cleaning them up which means we're propagating data to tabs that no longer exist unnecessarily.

@roryabraham I checked it once again and if I understand this logic correctly, after closing tabs we no longer propagate the data to them. `subscriptions` array is separate for each tab and contains two callbacks - the one that we push while subscribing to in Onyx.init() and the second one in ActiveClientManager.init(). Both callbacks are then fired when a message from the channel is received, message object is passed to them both but they are looking for and handling different data types. The one from Onyx.init() (`subscribeToEvents()`) for data types to write to Onyx, the one from ActivveClientManager.init() for new tab selection. And we want these both callbacks for each tab. The part of code I removed `ActiveClientManager.subscribeToClientChange(() => {  subscribeToEvents();` ([comment](https://github.com/Expensify/react-native-onyx/pull/382#discussion_r1394284694)) added the callback from `subscribeToEvents()` so if we had i.e 8 tabs open the first one opened had one callback listening for leader change and 8 for listening for writing to Onyx. The second one had one callback less and so on and the leader tab had only these two different callbacks I mentioned earlier. After removing this part every tab has only these two callbacks from being open to being closed, no matter if we open a new tab or not. Each tab has its own `subscriptions` array, they don't share this array or store the info about subscriptions for other tabs, they are only communicating via sending/receiving messages through the broadcast. 
I only wonder if we should clear the whole array and/or disconnect from the channel in 'beforeunload' but it seems that when a BroadcastChannel instance is closed (when closing the tab) event listeners (onmessage) are typically removed as part of the cleanup process.

Let me know what do you think about it and correct me if I'm wrong with the logic 🙂

roryabraham:
Let's follow up ASAP with a PR to upgrade Onyx in E/App

koko57:
@roryabraham wdyt?

roryabraham:
Approved https://github.com/Expensify/react-native-onyx/pull/434

chrispader:
This PR caused [these regressions](https://github.com/Expensify/App/pull/33554#issuecomment-1894753320) after the Onyx bump.

Fixing PR created here: https://github.com/Expensify/react-native-onyx/pull/455

PR #384, opened 2023-10-04
Fix concurrent writes in the update operation

ospfranco:
### Details

The update operation was written in such manner that it fired all database operations at once. Some of the operations internally rely on other promises, meaning they will get re-scheduled on another tick. This causes unpredictable behavior as you cannot guarantee the order of the operations. The solution is clear and is the only one, all DB writes must respect the order of the operations.

The incorrect ordering of the operations being committed to the DB causes other random behavior around the app as well. Linked the two issues.

### Related Issues

https://github.com/Expensify/App/issues/28737
https://github.com/Expensify/App/pull/28584

### Automated Tests

Added one more test that makes sure the correct data is persisted. It failed with the previous implementation, now it passes.

### DIFF
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -1392,6 +1392,28 @@ function mergeCollection(collectionKey, collection) {
         });
 }
 
+function innerUpdate(data) {
+    if (data.length === 0) {
+        return Promise.resolve();
+    }
+
+    const {onyxMethod, key, value} = data.shift();
+    switch (onyxMethod) {
+        case METHOD.SET:
+            return set(key, value).then(() => innerUpdate(data));
+        case METHOD.MERGE:
+            return merge(key, value).then(() => innerUpdate(data));
+        case METHOD.MERGE_COLLECTION:
+            return mergeCollection(key, value).then(() => innerUpdate(data));
+        case METHOD.MULTI_SET:
+            return multiSet(value).then(() => innerUpdate(data));
+        case METHOD.CLEAR:
+            return clear().then(() => innerUpdate(data));
+        default:
+            return Promise.resolve();
+    }
+}
+
 /**
  * Insert API responses and lifecycle data into Onyx
  *
@@ -1414,32 +1436,7 @@ function update(data) {
         }
     });
 
-    const promises = [];
-    let clearPromise = Promise.resolve();
-
-    _.each(data, ({onyxMethod, key, value}) => {
-        switch (onyxMethod) {
-            case METHOD.SET:
-                promises.push(() => set(key, value));
-                break;
-            case METHOD.MERGE:
-                promises.push(() => merge(key, value));
-                break;
-            case METHOD.MERGE_COLLECTION:
-                promises.push(() => mergeCollection(key, value));
-                break;
-            case METHOD.MULTI_SET:
-                promises.push(() => multiSet(value));
-                break;
-            case METHOD.CLEAR:
-                clearPromise = clear();
-                break;
-            default:
-                break;
-        }
-    });
-
-    return clearPromise.then(() => Promise.all(_.map(promises, p => p())));
+    return innerUpdate(data);
 }
 
 /**
@@ -1534,6 +1531,7 @@ const Onyx = {
     connect,
     disconnect,
     set,
+    get,
     multiSet,
     merge,
     mergeCollection,
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -1392,6 +1392,12 @@ function mergeCollection(collectionKey, collection) {
         });
 }
 
+/**
+ * Internal recursive function to execute the functions in the correct order
+ *
+ * @param {Array} data An array of objects with shape {onyxMethod: oneOf('set', 'merge', 'mergeCollection', 'multiSet', 'clear'), key: string, value: *}
+ * @returns {Promise<Void>}
+ */
 function innerUpdate(data) {
     if (data.length === 0) {
         return Promise.resolve();
@@ -1430,8 +1436,9 @@ function innerUpdate(data) {
  */
 function update(data) {
     // First, validate the Onyx object is in the format we expect
+    const validMethods = [METHOD.CLEAR, METHOD.SET, METHOD.MERGE, METHOD.MERGE_COLLECTION, METHOD.MULTI_SET];
     _.each(data, ({onyxMethod, key, value}) => {
-        if (!_.contains([METHOD.CLEAR, METHOD.SET, METHOD.MERGE, METHOD.MERGE_COLLECTION, METHOD.MULTI_SET], onyxMethod)) {
+        if (!_.contains(validMethods, onyxMethod)) {
             throw new Error(`Invalid onyxMethod ${onyxMethod} in Onyx update.`);
         }
         if (onyxMethod === METHOD.MULTI_SET) {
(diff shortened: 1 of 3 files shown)


tests/unit/onyxTest.js
```diff
@@ -899,9 +899,9 @@ describe('Onyx', () => {
             {onyxMethod: Onyx.METHOD.MERGE_COLLECTION, key: ONYX_KEYS.COLLECTION.TEST_UPDATE, value: {[itemKey]: {a: 'a'}}},
         ])
             .then(() => {
-                expect(collectionCallback).toHaveBeenNthCalledWith(1, {[itemKey]: {a: 'a'}});
-                expect(testCallback).toHaveBeenNthCalledWith(1, 'taco', ONYX_KEYS.TEST_KEY);
-                expect(otherTestCallback).toHaveBeenNthCalledWith(1, 'pizza', ONYX_KEYS.OTHER_TEST);
+                expect(collectionCallback).toHaveBeenNthCalledWith(2, {[itemKey]: {a: 'a'}});
+                expect(testCallback).toHaveBeenNthCalledWith(2, 'taco', ONYX_KEYS.TEST_KEY);
+                expect(otherTestCallback).toHaveBeenNthCalledWith(2, 'pizza', ONYX_KEYS.OTHER_TEST);
```
ospfranco, line 904: This previously was receiving the write updates before the hydration callback, but now that the write operations take a little longer the order is correct.

lib/Onyx.js
```diff
@@ -1414,32 +1436,7 @@ function update(data) {
...
-            case METHOD.CLEAR:
-                clearPromise = clear();
-                break;
-            default:
-                break;
-        }
-    });
-
-    return clearPromise.then(() => Promise.all(_.map(promises, p => p())));
+    return innerUpdate(data);
```
ospfranco, line 1439: As a data point, [I tried the proposed solution by Marc](https://github.com/Expensify/react-native-onyx/compare/marcaaron-resolveOnBroadcast) which would try to resolve earlier while keeping the writes async, this however is not correct and broke a bunch of other tests.

tgolen (changes requested):
Has there been any testing done in App to see what kind of performance impact this will have? If so, can you please attach it to this PR?

lib/Onyx.js
```diff
@@ -1534,6 +1539,7 @@ const Onyx = {
     connect,
     disconnect,
     set,
+    get,
```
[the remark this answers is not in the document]
ospfranco: From what I saw Onyx.connect has caching implemented, which might not be what is actually persisted on the DB. I was about to directly query the DB which is at the end of the day what really matters. I think this `get` might also use caches right? if so, checking the DB should be the correct path.

ospfranco:
No, I haven't tested the app. I can try to get some metrics but I already asked in slack if there is any particular call to update that is heavy.

lib/Onyx.js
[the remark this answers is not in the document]
tgolen: I like the idea of querying the DB directly if that's possible.

ospfranco:
I added some react-native-performance measures on app start, I cannot tell which one exactly is OpenApp but they are all basically negible. Writing to DB should not be the bottleneck, and even then, correctness is not negotiable.

Oscar Franco Screen 000065

lib/Onyx.js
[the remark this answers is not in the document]
ospfranco: Done

tgolen:
OK, I just created a large account for testing things like this, so let me give it a try with this today.

lib/Onyx.js
```diff
@@ -1392,6 +1392,36 @@ function mergeCollection(collectionKey, collection) {
         });
 }
 
+function innerUpdate(data) {
```
tgolen, line 1395: Please add JSDoc

tgolen (changes requested):
I tested it in App with a large account and I couldn't notice any performance impact, so I think these are good changes. I just noticed this one thing.

marcaaron:
Doing some testing now...

marcaaron:
Alright well - it is defying my expectations a bit, but when timing the calls to `Onyx.update()` they are running way faster synchronously 😅 

### Sign in trials

**With sync updates:**

1

Timing:development.new.expensify.flushQueue 458
Timing.js:49 Timing:development.new.expensify.flushQueue 0
Timing.js:49 Timing:development.new.expensify.flushQueue 407

2

Timing:development.new.expensify.flushQueue 468
Timing.js:49 Timing:development.new.expensify.flushQueue 1
Timing.js:49 Timing:development.new.expensify.flushQueue 191

3 

Timing:development.new.expensify.flushQueue 311
Timing.js:49 Timing:development.new.expensify.flushQueue 0
Timing.js:49 Timing:development.new.expensify.flushQueue 226

**Without sync updates (but on latest `react-native-onyx`):**

1

Timing:development.new.expensify.flushQueue 488
Timing.js:49 Timing:development.new.expensify.flushQueue 4147
Timing.js:49 Timing:development.new.expensify.flushQueue 962

2

Timing:development.new.expensify.flushQueue 420
Timing.js:49 Timing:development.new.expensify.flushQueue 1126
Timing.js:49 Timing:development.new.expensify.flushQueue 170

3

Timing:development.new.expensify.flushQueue 464
Timing.js:49 Timing:development.new.expensify.flushQueue 4729
Timing.js:49 Timing:development.new.expensify.flushQueue 976

This was a web-only test. Haven't tested on iOS or Android - but looks promising.

lib/Onyx.js
```diff
@@ -1414,32 +1444,7 @@ function update(data) {
...
-                break;
-            case METHOD.CLEAR:
-                clearPromise = clear();
-                break;
-            default:
-                break;
-        }
-    });
-
-    return clearPromise.then(() => Promise.all(_.map(promises, p => p())));
```
marcaaron, line 1442: I see we are undoing the previous logic to move the `clear()` to the front of the update queue. Does running the updates sync solve the issue that caused us to make that change?

ospfranco: ah, probably not. I'll take care of it.

ospfranco: I'm not sure making sure clear runs first is a desired behavior but I guess now it is too late to change it, other parts of the system might rely on this behavior. I added sorting to the list to make sure clear operations get executed first.

deetergp (approved):

lib/Onyx.js
marcaaron: Yeah, I agree it is a bit of a weird hack. But changing it could have unexpected consequences. Thanks for updating.

marcaaron (approved):
🚀

lib/Onyx.js
```diff
@@ -1392,6 +1392,42 @@ function mergeCollection(collectionKey, collection) {
...
+            break;
+        case METHOD.MULTI_SET:
+            promise = multiSet(value);
+            break;
+        case METHOD.CLEAR:
+            promise = clear();
+            break;
+        default:
+            break;
+    }
```
tgolen, line 1426: It looks like this change possibly led to some regressions spotted in [this PR](https://github.com/Expensify/App/pull/29169#issuecomment-1755269426) and discussed in [this Slack thread](https://expensify.slack.com/archives/C035J5C9FAP/p1696941880028159?thread_ts=1696934662.824599&cid=C035J5C9FAP) that updated the Onyx version

PR #385, opened 2023-10-04
Allow infinite dependent keys when using withOnyx

tgolen:
This is a redo of https://github.com/Expensify/react-native-onyx/pull/355 which had to be reverted due to performance issues.

### Details
There were virtually no differences between this change and the previous change. This is what I did to try and ensure it doesn't leed to another performance issue:

- Test it in App with report switching (this is what we noticed was bricking the application before). With this change, I noticed no difference in the time it takes to switch chats:
 
image

### Related Issues
Related to https://github.com/Expensify/App/issues/28824

### Automated Tests
Covered by unit tests

### Linked PRs
https://github.com/Expensify/App/pull/28866

### DIFF
--- a/lib/withOnyx.js
+++ b/lib/withOnyx.js
@@ -20,6 +20,15 @@ function getDisplayName(component) {
     return component.displayName || component.name || 'Component';
 }
 
+/**
+ * Removes all the keys from state that are unrelated to the onyx data being mapped to the component.
+ *
+ * @param {Object} state of the component
+ * @param {Object} onyxToStateMapping the object holding all of the mapping configuration for the component
+ * @returns {Object}
+ */
+const getOnyxDataFromState = (state, onyxToStateMapping) => _.pick(state, _.keys(onyxToStateMapping));
+
 export default function (mapOnyxToState, shouldDelayUpdates = false) {
     // A list of keys that must be present in tempState before we can render the WrappedComponent
     const requiredKeysForInit = _.chain(mapOnyxToState)
@@ -92,16 +101,20 @@ export default function (mapOnyxToState, shouldDelayUpdates = false) {
                 this.checkEvictableKeys();
             }
 
-            componentDidUpdate(prevProps) {
+            componentDidUpdate() {
+                // When the state is passed to the key functions with Str.result(), omit anything
+                // from state that was not part of the mapped keys.
+                const onyxDataFromState = getOnyxDataFromState(this.state, mapOnyxToState);
+
                 // If any of the mappings use data from the props, then when the props change, all the
                 // connections need to be reconnected with the new props
-                _.each(mapOnyxToState, (mapping, propertyName) => {
-                    const previousKey = Str.result(mapping.key, prevProps);
-                    const newKey = Str.result(mapping.key, this.props);
+                _.each(mapOnyxToState, (mapping, propName) => {
+                    const previousKey = mapping.previousKey;
+                    const newKey = Str.result(mapping.key, {...this.props, ...onyxDataFromState});
                     if (previousKey !== newKey) {
                         Onyx.disconnect(this.activeConnectionIDs[previousKey], previousKey);
                         delete this.activeConnectionIDs[previousKey];
-                        this.connectMappingToOnyx(mapping, propertyName);
+                        this.connectMappingToOnyx(mapping, propName);
                     }
                 });
                 this.checkEvictableKeys();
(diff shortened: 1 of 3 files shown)


lib/withOnyx.js
```diff
@@ -110,9 +123,8 @@ export default function (mapOnyxToState, shouldDelayUpdates = false) {
             componentWillUnmount() {
                 // Disconnect everything from Onyx
                 _.each(mapOnyxToState, (mapping) => {
-                    const key = Str.result(mapping.key, this.props);
-                    const connectionID = this.activeConnectionIDs[key];
-                    Onyx.disconnect(connectionID, key);
+                    const key = Str.result(mapping.key, {...this.props, ...getOnyxDataFromState(this.state, mapOnyxToState)});
```
tgolen, line 126: This is one change that is different than the original PR. I realized that the `unmount()` was still trying to use the old method for getting the key name. This would probably lead to somewhat of a memory leak because it could cause Onyx to keep a bunch of connections that are never disconnected from when components unmount.

lib/withOnyx.js
```diff
@@ -244,7 +256,14 @@ export default function (mapOnyxToState, shouldDelayUpdates = false) {
...
             connectMappingToOnyx(mapping, statePropertyName) {
-                const key = Str.result(mapping.key, this.props);
+                const key = Str.result(mapping.key, {...this.props, ...getOnyxDataFromState(this.state, mapOnyxToState)});
+
+                // Remember the previous key so that if it ever changes, the component will reconnect to Onyx
+                // in componentDidUpdate
+                if (statePropertyName !== 'initialValue' && mapOnyxToState[statePropertyName]) {
+                    // eslint-disable-next-line no-param-reassign
+                    mapOnyxToState[statePropertyName].previousKey = key;
+                }
```
tgolen, line 266: This is another small change from the previous PR. When `initialValue` was integrated, it lead to a situation where `mapOnyxToState.initialValue: false`, which caused `mapOnyxToState[statePropertyName].previousKey` to throw an error (because you cannot set `.previousKey` on the value `false`).

tgolen:
I tested this on a large account (1000 reports each with 1000 comments) and the performance was still very good on switching chats:

image

lib/withOnyx.js
tgolen: I can give it a try. That's a good hunch

tgolen: No, it doesn't appear to. I think the changes for batched updates was probably the biggest help for preventing the performance problem we saw last time.

tgolen: >  I just feel like this might cause more issues down the road.

Are you referring to the change in `onUnmount()` specifically, or this PR in general?

tgolen: To dig into this further, I tried the following things:

1. Reverted Onyx to the commit where the original PR was reverted. Then I copied these `withOnyx` changes to the code and tried it out. **Result:** I was not able to duplicate the original performance problem.
2. Reverted Onyx to the commit of the original PR being merged, and tried it out in the app. **Result:** I experienced the following fatal error (caused by the `initialValue` change). After fixing _only that fatal error_ I was still not able to reproduce the original performance problem, though I did see that the report switching was taking about 2-3x longer.

image
image

I'm not sure if this fatal error was what we experienced when the original PR was deployed. We definitely witnessed the app hanging and no one could report any JS console errors or record performance profiles, but that was about it.

**Conclusion:** I am feeling pretty good about the changes in this PR in general. The worst that can happen is we have to do another quick revert, which would stink, but it's also not the end of the world.

tgolen:
@allroundexperts I have done some memory testing and tried to be very methodical about it so that it is properly comparing apples to apples. My testing procedure is:

1. Have 10 chats each with exactly 1000 comments
2. Visit chat 1
3. Refresh the page and wait for the chat and all comments to load
4. Select each of the other 9 chats, and wait for the comments to load on each one before moving to the next
5. After the final chat has fully loaded, record a memory heap snapshot

<details>
<summary>You can see a video of it here</summary>

https://github.com/Expensify/react-native-onyx/assets/1228807/6effb8f9-76cf-45ce-a92d-3845872e9ad5

</details>

I repeated this test two times, in three different code configurations (for a total of 6 results), to get a nice sampling. The code configurations I used were:

1. App on `main`, Onyx on `main` 
2. App on `main`, Onyx on `tgolen-dependent-keys2`
3. App on experimental branch, Onyx on `tgolen-dependent-keys2` - For the experimental branch on App, I removed all of the duplicate `withOnyx` code including the one that would impact this test this most which is in `OptionRowLHNData.js`. This is used for each row in the LHN and combines three separate `withOnyx` into a single one.

### Results
image

### Conclusion
There is no significant memory usage increase with this change.

tests/unit/withOnyxTest.js
```diff
@@ -244,31 +246,73 @@ describe('withOnyxTest', () => {
     });
 
     it('should pass a prop from one connected component to another', () => {
-        const collectionItemID = 1;
         const onRender = jest.fn();
         const markReadyForHydration = jest.fn();
-        Onyx.mergeCollection(ONYX_KEYS.COLLECTION.TEST_KEY, {test_1: {id: 1}});
-        Onyx.mergeCollection(ONYX_KEYS.COLLECTION.RELATED_KEY, {related_1: 'Test'});
+
+        // Given three collections with multiple items in each
```
MonilBhavsar, line 252: three collections?

MonilBhavsar (dismissed):
Sorry, was waiting for C+ review. Looks good to me overall!

tests/unit/withOnyxTest.js
tgolen: I'll change it to say "several"

MonilBhavsar: Thanks!

MonilBhavsar (approved):
Thanks!

PR #387, opened 2023-10-05
Add The Checklist to react-native-onyx

marcaaron:
### Details

Coming from https://expensify.slack.com/archives/C01GTK53T8Q/p1696292810843059

Adding an amended version of the App checklist to `react-native-onyx`. I tried to pull out some irrelevant things.

We won't have the same enforcements that we have in App since we are not using the same GitHub actions, but this should be a decent reminder for now. The bar is currently very low for testing. And we had a couple of reverts.

### Related Issues

❌ 

### Automated Tests

❌ 

### Linked PRs

❌

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -12,7 +12,95 @@ GH_LINK
 Most changes to Onyx should have accompanying tests. Describe the tests you added or if no tests were added an explanation about why one was not needed.
 --->
 
-### Linked PRs
+### Manual Tests
 <!---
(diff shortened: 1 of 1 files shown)


arosiclair (approved):
Nice this is long overdue 👍. Do we have a follow up issue for adding github actions to enforce this?

marcaaron:
Not at the moment, but I can create one. I think @roryabraham has some plans to do "dynamic" checklists. So maybe whenever we figure that out we can try to mirror the changes to other repos. Might also not be totally "worth" doing if it's complex. I haven't had time to learn about the planned changes so would defer to @roryabraham on that one.

PR #411, opened 2023-10-31
Don't remove nulls in cache merge

janicduplessis:
### Details

When merging cache we should not remove null values, they are useful to know an item was fetched but doesn't exist.

### Related Issues
https://github.com/Expensify/react-native-onyx/pull/408#issuecomment-1785038665

### Automated Tests
Added a unit test to assert behaviour of null values in OnyxCache.

### Manual Tests

Tested in Expensify app and made sure the money request button renders at the same time as the rest of the header.

### Author Checklist

    - [x] Android / native
    - [x] Android / Chrome
    - [x] iOS / native
    - [x] iOS / Safari
    - [x] MacOS / Chrome / Safari
    - [x] MacOS / Desktop

### Screenshots/Videos
<details>
<summary>Web</summary>

</details>

<details>
<summary>Mobile Web - Chrome</summary>

</details>

<details>
<summary>Mobile Web - Safari</summary>

</details>

<details>
<summary>Desktop</summary>

</details>

<details>
<summary>iOS</summary>

</details>

<details>
<summary>Android</summary>

</details>

### DIFF
--- a/lib/OnyxCache.js
+++ b/lib/OnyxCache.js
@@ -121,7 +121,7 @@ class OnyxCache {
 
         // lodash adds a small overhead so we don't use it here
         // eslint-disable-next-line prefer-object-spread, rulesdir/prefer-underscore-method
-        this.storageMap = Object.assign({}, utils.fastMerge(this.storageMap, data));
+        this.storageMap = Object.assign({}, utils.fastMerge(this.storageMap, data, false));
 
         const storageKeys = this.getAllKeys();
         const mergedKeys = _.keys(data);
--- a/tests/unit/onyxCacheTest.js
+++ b/tests/unit/onyxCacheTest.js
@@ -347,6 +347,18 @@ describe('Onyx', () => {
                 expect(() => cache.merge(0)).toThrow();
                 expect(() => cache.merge({})).not.toThrow();
             });
+
+            it('Should merge `null` values', () => {
+                // `null` values can override existing values and should also
+                // be preserved during merges.
+                cache.set('mockKey', {ID: 5});
+                cache.set('mockNullKey', null);
+
+                cache.merge({mockKey: null});
+
+                expect(cache.getValue('mockKey')).toEqual(null);
+                expect(cache.getValue('mockNullKey')).toEqual(null);
+            });
         });
 
         describe('hasPendingTask', () => {


chiragsalian (approved):
I think the code is fine but i thought we were using null to clear unused onyx data to free up memory so I'm unsure if this code would cause a negative impact. 
Either way I'll let @marcaaron also review since he's more familiar here.

marcaaron:
>When merging cache we should not remove null values, they are useful to know an item was fetched but doesn't exist.

Sorry, I am not sure what this means. Is there more context? Can you show where we are removing the `null` values? And how exactly "they are useful to know an item was fetched but doesn't exist".

lib/OnyxCache.js
```diff
@@ -121,7 +121,7 @@ class OnyxCache {
 
         // lodash adds a small overhead so we don't use it here
         // eslint-disable-next-line prefer-object-spread, rulesdir/prefer-underscore-method
-        this.storageMap = Object.assign({}, utils.fastMerge(this.storageMap, data));
+        this.storageMap = Object.assign({}, utils.fastMerge(this.storageMap, data, false));
```
marcaaron, line 124: Does this mean we would just leave in the storage map a bunch of keys with `null` and never delete them? I think it does. 

It is a bit hard to wrap one's head around what the side effects of this change would be 😄 

One possible problem is here:

https://github.com/Expensify/react-native-onyx/blob/cd54318794c1682ffb8aee650e9a7983fbceb6f8/lib/Onyx.js#L125-L129

Since a value that we are setting again after deleting previously would be `undefined` but now would be regarded as `null` so we would not check to see if it should be read from storage.

marcaaron (commented):
@chiragsalian maybe we can get more eyes on this one. I don't have the bandwidth to help validate this right now, but left some comments.

marcaaron:
at the very least I think we should do some testing for this PR against the App

janicduplessis:
Sorry I think this PR lacked a little bit more context around this change. The idea was to improve Onyx caching by caching `null` for missing values. Previously if a key did not exist we never cached it which means `withOnyx` components will start in a loading state then have to re-render, even if the data was loaded before. The PR to implement this is https://github.com/Expensify/react-native-onyx/pull/401. More specifically in the caching layer it means that `undefined` represent a value that we do not know if it exists and `null` represents a value that we know does not exist.

This PR is a bug fix since calling merge will wipe all `null` values from cache and cause all those components to start in loading state again. I don't think there is much downside to keeping these null values in cache. It seems like wiping them is kind of an accidental side effect of https://github.com/Expensify/react-native-onyx/commit/69c11fa7cd48ec9b09be65fe3e304d071230e36d. Previously they were not removed. Maybe @chrispader can confirm this is fine as I'm not familiar with the original intent of that PR.

lib/OnyxCache.js
janicduplessis: We will have more null values in cache storage yes, but this is only the in memory cache and not the persisted storage so I think it is fine to speedup lookup of values that we know do not exist in storage.

The behavior for deleted values should not change since the cache has a drop method which should be used to remove values completely. The only case I know of that we remove values from cache so they will be fetched from storage again is https://github.com/Expensify/react-native-onyx/blob/cd54318794c1682ffb8aee650e9a7983fbceb6f8/lib/OnyxCache.js#L185 which doesn't use merge. In other cases if we use merge with nulls to signal values were deleted it is fine to have them in cache still since we know their latest value is that they do not exist.

I hope that clears up any confusion.

janicduplessis:
@chiragsalian We have a method that clears least recently used keys from cache, but it doesn't use the merge method that is affected here, the cache has a drop method which should be used to fully remove items from cache and cause them to be fetched again from storage. Also note this is only the caching layer and doesn't affect persisted storage.

janicduplessis:
@chiragsalian @marcaaron Thanks for having a look, lmk if there's anything I forgot to address.

janicduplessis:
Also going to ping @mountiny and @tgolen which reviewed the previous PRs related to that.

tgolen:
> i thought we were using null to clear unused onyx data to free up memory

I think there is a subtle difference here. null is used to clear onyx data to free up disk space, not memory. To my knowledge, we haven't done much with optimizing memory.

> at the very least I think we should do some testing for this PR against the App

I definitely always agree with this one (I have learned my own hard lessons from it). To alleviate this concern a little bit, this PR is coming immediately on the heels of doing a lot of extensive in-app testing and it fixes one of the last remaining bugs that was found which prevents Onyx from being updated in App.

tgolen:
More context around the discussion and testing was in [this Slack thread](https://expensify.slack.com/archives/C05LX9D6E07/p1698193417564459).

lib/OnyxCache.js
marcaaron: >it is fine to speedup lookup of values that we know do not exist in storage.

That's a good point - I was thinking more along the lines of whether this could lead to anything unexpected e.g. could it somehow be possible for the value to exist in storage, but not the cache? But doesn't seem likely.

marcaaron:
Thanks for the thorough explanation @janicduplessis. With the additional context there are no blocking concerns from me.

mountiny (approved):
Thank you for the explanation indeed

chrispader:
> Sorry I think this PR lacked a little bit more context around this change. The idea was to improve Onyx caching by caching `null` for missing values. Previously if a key did not exist we never cached it which means `withOnyx` components will start in a loading state then have to re-render, even if the data was loaded before. The PR to implement this is #401. More specifically in the caching layer it means that `undefined` represent a value that we do not know if it exists and `null` represents a value that we know does not exist.
> 
> This PR is a bug fix since calling merge will wipe all `null` values from cache and cause all those components to start in loading state again. I don't think there is much downside to keeping these null values in cache. It seems like wiping them is kind of an accidental side effect of [69c11fa](https://github.com/Expensify/react-native-onyx/commit/69c11fa7cd48ec9b09be65fe3e304d071230e36d). Previously they were not removed. Maybe @chrispader can confirm this is fine as I'm not familiar with the original intent of that PR.

Sorry for the late reply! I also don't think there's anything wrong with this PR 👍 @janicduplessis

chrispader (approved):

Issue #266, opened 2023-06-27
Onyx.update doesn't apply updates in order with set and merge


neil-marcellini:
## Problem
Onyx.update doesn't apply updates in order when mixing set and merge

If `Onyx.update` is called with an update to merge an object on a key, and then the next update removes it by calling set with the value as `null`, then the updates are not applied in order. The Onyx.set update is applied first and the end state of the system is that the key is not removed.

Also, subscribers are notified with each update. In some cases we want to only notify subscribers with the correct end state of the system.

Related to [[Tracking] "Replay effect" when sequential actions are taken offline, and user subsequently comes online](https://github.com/Expensify/App/issues/12775#top)

### Example
I found this to be the root problem of this app issue https://github.com/Expensify/App/issues/15550. Here is the user action and results.

> ## Action Performed:
> 
> 1. Turn off network
> 2. Create a workspace
> 3. Delete the workspace
> 4. Turn on network
> 5. Notice that the deleted workspace blinks
> 
> ## Expected Result:
> The deleted workspace should not be visible
> 
> ## Actual Result:
> The deleted workspace blinks

After we go back online and the sequential queue finishes processing, we [call Onyx.update](https://github.com/Expensify/App/blob/b8d45763b6d9dcf011ab66936df7eba85b80bc24/src/libs/Network/SequentialQueue.js#L86) with a list of queued updates. We have already optimistically created and deleted the policy, and the list of updates includes creating and deleting the policy from the network responses. We create the policy with a call to `merge` and delete it with a call to `set`. To test I added some log lines and ran the following code in the JS console. You can see that the `set` call is applied first when it should be applied second.

<details>
<summary>Logs</summary>

```
Onyx.update([
    {
        onyxMethod: 'merge',
        key: 'policy_BEB826F4F75AD818',
        value: {
            id: 'BEB826F4F75AD818',
            name: "Expensifail's Workspace 2",
            role: 'admin',
            type: 'free',
            owner: 'neil+chat@expensifail.com',
            outputCurrency: 'USD',
            employeeList: [
                {
                    email: 'neil+chat@expensifail.com',
                    forwardsTo: '',
                    role: 'admin',
                    submitsTo: 'neil+chat@expensifail.com',
                },
            ],
        },
    },
    {
        onyxMethod: 'set',
        key: 'policy_BEB826F4F75AD818',
        value: null,
    },
]);
[...]

PR #437, opened 2023-12-19
Fix: Onyx.update order

chrispader:
@neil-marcellini 

### Details

Fixes a problem where `Onyx.update` seems to not apply updates in order. This was due to the merge queue not being cleared when we call `Onyx.set`.

Additionally, we returned early from `Onyx.set` and `Onyx.merge`, when we removed keys from storage and therefore didn't "broadcast" the update and update subscribers.

This PR fixed the problem with the order of `Onyx.update` by clearing the merge queue on every `Onyx.set` and additionally update subscribers with the cleared (null) value after removing a key intentionally with `Onyx.set(SOME_KEY, null)` or `Onyx.merge(SOME_KEY, null)`.

This PR also fixes some inconsistencies with key removal in `Onyx.multiSet` and `Onyx.multiMerge`.

### Related Issues

https://github.com/Expensify/react-native-onyx/issues/266

### Automated Tests

Added a Test "should apply updates in order with Onyx.update" in `onyxTest`, which tests `Onyx.update` for applying the changes in the correct order.

### Manual Tests

1. Open NewDot
2. Turn off network
3. Create a workspace
4. Delete the workspace
5. Turn on network
6. Check that the workspace is removed and disappears and it doesn't flash

### Author Checklist

    - [x] Android / native
    - [x] Android / Chrome
    - [x] iOS / native
    - [x] iOS / Safari
    - [x] MacOS / Chrome / Safari
    - [x] MacOS / Desktop

### Screenshots/Videos
<details>

</details>

<details>

</details>

<details>

</details>

<details>

https://github.com/Expensify/react-native-onyx/assets/20173411/9bdf48ad-86a9-4d39-bdb8-aea495c2128c

</details>

<details>

https://github.com/Expensify/react-native-onyx/assets/20173411/5c174f7d-84ac-4317-ba49-b3c49c07966b

</details>

<details>

</details>

Added screen recordings for web and iOS web

### DIFF
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -1117,7 +1117,17 @@ function set(key, value) {
  * @return {Array} an array of key - value pairs and a boolean which indicates if the key got removed <[key, value]>
  */
 function prepareKeyValuePairsForStorage(data) {
-    return _.map(data, (value, key) => [key, value]);
+    const keyValuePairs = [];
+
+    _.forEach(data, (value, key) =>  {
+        const {value: valueAfterRemoving, wasRemoved} = removeNullValues(key, value);
+
+        if (wasRemoved) return
+
+        keyValuePairs.push([key, valueAfterRemoving])
+    });
+
+    return keyValuePairs
 }
 
 /**
@@ -1140,24 +1150,13 @@ function multiSet(data) {
 
     const keyValuePairs = prepareKeyValuePairsForStorage(data);
 
-    const keyValuePairsWithoutNull = []
-    _.forEach(keyValuePairs, ([key, value]) => {
-        const {value: valueAfterRemoving, wasRemoved} = removeNullValues(key, value);
-
-        if (wasRemoved) {
-            return;
-        }
-
-        keyValuePairsWithoutNull.push([key, valueAfterRemoving]);
-    });
-
-    const updatePromises = _.map(keyValuePairsWithoutNull, ([key, value]) => {
+    const updatePromises = _.map(keyValuePairs, ([key, value]) => {
         // Update cache and optimistically inform subscribers on the next tick
         cache.set(key, value);
         return scheduleSubscriberUpdate(key, value);
     });
 
-    return Storage.multiSet(keyValuePairsWithoutNull)
+    return Storage.multiSet(keyValuePairs)
         .catch((error) => evictStorageAndRetry(error, multiSet, data))
         .then(() => Promise.all(updatePromises));
 }
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -1098,13 +1098,8 @@ function set(key, value) {
     // This approach prioritizes fast UI changes without waiting for data to be stored in device storage.
     const updatePromise = broadcastUpdate(key, valueAfterRemoving, 'set', hasChanged, wasRemoved);
 
-    // If the key got removed earlier, we don't have to set the value in storage, so return early instead
-    if (wasRemoved) {
-        return updatePromise;
-    }
-
-    // If the value has not changed, calling Storage.setItem() would be redundant and a waste of performance, so return early instead.
-    if (!hasChanged) {
+    // If the value has not changed or the key got removed, calling Storage.setItem() would be redundant and a waste of performance, so return early instead.
+    if (!hasChanged || wasRemoved) {
         return updatePromise;
     }
 
@@ -1255,10 +1250,7 @@ function merge(key, changes) {
             delete mergeQueuePromise[key];
 
             // If the batched changes equal null, we want to remove the key from storage, to reduce storage size
-            if (_.isNull(batchedChanges)) {
-                remove(key);
-                return;
-            }
+            const {wasRemoved} = removeNullValues(key, batchedChanges)
 
             // After that we merge the batched changes with the existing value
             // We can remove null values from the "modifiedData", because "null" implicates that the user wants to remove a value from storage.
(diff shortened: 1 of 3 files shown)


chrispader:
@neil-marcellini this PR should fix the problems with `Onyx.update` order and the "replay" effect. Feel free to review this, i'm gonna add screenshots tmrw. Testing fine though 👍

lib/Onyx.js
```diff
@@ -1053,18 +1054,18 @@ function hasPendingMergeForKey(key) {
...
  * @param {String} key
  * @param {Mixed} value
- * @returns {Mixed} `null` if the key got removed completely, otherwise the value without null values
+ * @returns {Mixed} The value without null values and a boolean "wasRemoved", which indicates if the key got removed completely
  */
 function removeNullValues(key, value) {
     if (_.isNull(value)) {
         remove(key);
-        return null;
+        return {value, wasRemoved: true};
```
neil-marcellini, line 1062: NAB: It seems pointless to have this `wasRemoved` flag when the caller could just check if the value is null to determine if the key was removed.

lib/Onyx.js
```diff
@@ -1085,41 +1086,48 @@ function set(key, value) {
...
+    const hasChanged = cache.hasValueChanged(key, valueAfterRemoving);
 
     // This approach prioritizes fast UI changes without waiting for data to be stored in device storage.
-    const updatePromise = broadcastUpdate(key, valueWithoutNull, hasChanged, 'set');
+    const updatePromise = broadcastUpdate(key, valueAfterRemoving, 'set', hasChanged, wasRemoved);
 
-    // If the value has not changed, calling Storage.setItem() would be redundant and a waste of performance, so return early instead.
-    if (!hasChanged) {
+    // If the value has not changed or the key got removed, calling Storage.setItem() would be redundant and a waste of performance, so return early instead.
+    if (!hasChanged || wasRemoved) {
```
neil-marcellini, line 1102: If we removed keys by setting their values to null, don't we need to update storage with the new object without those keys?

lib/Onyx.js
```diff
@@ -1021,17 +1021,18 @@ function evictStorageAndRetry(error, onyxMethod, ...args) {
...
  */
-function broadcastUpdate(key, value, hasChanged, method) {
+function broadcastUpdate(key, value, method, hasChanged, wasRemoved = false) {
     // Logging properties only since values could be sensitive things we don't want to log
     Logger.logInfo(`${method}() called for key: ${key}${_.isObject(value) ? ` properties: ${_.keys(value).join(',')}` : ''}`);
 
     // Update subscribers if the cached value has changed, or when the subscriber specifically requires
     // all updates regardless of value changes (indicated by initWithStoredValues set to false).
-    if (hasChanged) {
+    if (hasChanged && !wasRemoved) {
```
neil-marcellini, line 1035: Why would we not update the cache when the key is removed?

lib/Onyx.js
```diff
@@ -1085,41 +1086,48 @@ function set(key, value) {
...
 
 /**
  * Storage expects array like: [["@MyApp_user", value_1], ["@MyApp_key", value_2]]
  * This method transforms an object like {'@MyApp_user': myUserValue, '@MyApp_key': myKeyValue}
- * to an array of key-value pairs in the above format
+ * to an array of key-value pairs in the above format and removes key-value pairs that are being set to null
  * @private
  * @param {Record} data
- * @return {Array} an array of key - value pairs <[key, value]>
+ * @return {Array} an array of key - value pairs and a boolean which indicates if the key got removed <[key, value]>
```
neil-marcellini, line 1117: I don't see where we're returning the `wasRemoved` boolean.

neil-marcellini (changes requested):
Looking really good, thank you! Would you please write manual test and link to the video showing it working for creating and deleting a workspace offline? That was great, so nice to see it finally working!

I do have a handful of comments and questions, and I would like to see some small changes to clean up a bit please.

lib/Onyx.js
jasperhuangg: +1, I don't understand this

lib/Onyx.js
chrispader: `wasRemoved` here only applies when the whole key was set to `null`, not when nested `null` values got removed.

The storage is already being updated, when the `remove` function is called here: https://github.com/margelo/react-native-onyx/blob/ac8faed9890bca04f889f93e0f249191ee6301b5/lib/Onyx.js#L1061

That's why we have to omit calling `Storage.setItem` here again

lib/Onyx.js
chrispader: Same as in the the other thread, the call to `remove` already drops the key from cache: https://github.com/margelo/react-native-onyx/blob/ac8faed9890bca04f889f93e0f249191ee6301b5/lib/Onyx.js#L965

That's why we have to omit updating the storage here, and instead just add the key to the list of accessed keys.

lib/Onyx.js
chrispader: The reason i introduced this extra flag was to have a consistent variable/condition of whether the key got removed as part of an operation. We check for this in multiple spots (both of your other comments), and i felt that having a dedicated `wasRemoved` flag was better than always checking for `if (valueAfterRemoving === null)`.

lib/Onyx.js
chrispader: updated the comment! this was a relict from previous changes

chrispader:
@neil-marcellini the video showcasing the fix of the replay effect is under "Screenshots"

jasperhuangg:
Heading OOO for the holidays, so it might be best to assign another engineer if you want another review

lib/Onyx.js
neil-marcellini: I generally prefer limiting the number of parameters. Also, wasRemoved is always simply `_.isNull(value)`, except in one case where we pass the batched changes. In fact you have a comment there explaining that we're checking if the batched changes are null. That's kind of silly.

lib/Onyx.js
neil-marcellini: Ooooooh I see. Thank you for explaining.

lib/Onyx.js
neil-marcellini: Makes sense thanks

neil-marcellini (approved):
Thanks for addressing my concerns, it makes sense now and looks good. This is awesome, and I think it's going to fix a lot of rare bugs we had lurking in the App!

lib/Onyx.js
neil-marcellini: I'll leave it up to you if you want to change this is a follow up PR.

PR #439, opened 2023-12-21
Use `MemoryOnlyProvider` as fallback to allow the app to work without storage

chrispader:
@marcaaron @tgolen 

### Details

This PR adds a `MemoryOnlyProvider` fallback solution and changes the interface of `Storage` (`lib/storage/index.js`) to allow multiple simultaneous storage providers.

This PR also improves the structure of everything in `lib/storage` and establishes a consistent structure for providers and order for all the functions within them.

### Related Issues

### Automated Tests

### Manual Tests

### Author Checklist

    - [x] Android / native
    - [x] Android / Chrome
    - [x] iOS / native
    - [x] iOS / Safari
    - [x] MacOS / Chrome / Safari
    - [x] MacOS / Desktop

### Screenshots/Videos
<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

https://github.com/Expensify/react-native-onyx/assets/20173411/fc014b32-4877-443e-88de-b743269c1319

</details>

<details>

</details>


chrispader:
As described in https://github.com/Expensify/App/issues/29403#issuecomment-1869513946, i designed this new `Storage` interface to allow more than 1 storage provider in the future. Currently, we just fallback to the `MemoryOnlyProvider` if the platform storage is not working.

In a follow-up PR, we could potentially remove `OnyxCache` completely, since it's essentially just the same as the extra `MemoryOnlyProvider`.

In the future we can also think about adding other persistable storage solutions as fallback, instead of memory-only. This way, the user wouldn't have to log-in again after a refresh in case of storage problems

chrispader:
Tested thoroughly on web, adding more screen recordings for other platforms soon.

lib/storage/providers/MemoryOnlyProvider.js
```diff
@@ -0,0 +1,155 @@
...
+     */
+    init() {},
+
+    /**
+     * Get the value of a given key or return `null` if it's not available in memory
+     * @param {String} key
+     * @return {Promise<*>}
+     */
+    getItem(key) {
+        return Promise.resolve(store[key]);
```
arosiclair, line 33: `store[key]` will return `undefined` for a missing key. We should coerce that to `null` like we do in the IDB provider, right?

lib/storage/providers/MemoryOnlyProvider.js
```diff
@@ -0,0 +1,155 @@
...
+    },
+
+    /**
+     * Stores multiple key-value pairs in a batch
+     * @param {Array<[key, value]>} pairs
+     * @return {Promise<void>}
+     */
+    multiSet(pairs) {
+        const setPromises = _.map(pairs, ([key, value]) => this.setItem(key, value));
+        return new Promise((resolve) => Promise.all(setPromises).then(() => resolve(store)));
```
arosiclair, line 63: Does this actually need to resolve with the `store` as the result? The jsDoc labels this as a `void` return

lib/storage/providers/MemoryOnlyProvider.js
```diff
@@ -0,0 +1,155 @@
...
+     */
+    multiMerge(pairs) {
+        _.forEach(pairs, ([key, value]) => {
+            const existingValue = store[key];
+            const newValue = utils.fastMerge(existingValue, value);
+
+            set(key, newValue);
+        });
+
+        return Promise.resolve(store);
```
arosiclair, line 92: Same thing here. This result doesn't match the jsDoc

arosiclair (changes requested):
Looks mostly good to me. Would appreciate a review from @tgolen or @marcaaron as well

lib/storage/InstanceSync/index.web.js
```diff
@@ -0,0 +1,63 @@
...
+        // The keys must be retrieved before storage is cleared or else the list of keys would be empty
+        return this.getAllKeys()
+            .then((keys) => {
+                allKeys = keys;
+            })
+            .then(() => clearImplementation())
+            .then(() => {
+                // Now that storage is cleared, the storage sync event can happen which is a more atomic action
+                // for other browser tabs
+                _.each(allKeys, raiseStorageSyncEvent);
```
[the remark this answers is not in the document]
chrispader: yes, definitely,thanks! This was copied over from the original `keepInstancesSync` function

lib/storage/providers/MemoryOnlyProvider.js
chrispader: yes indeed

lib/storage/providers/MemoryOnlyProvider.js
chrispader: yes, indeed!

lib/Onyx.js
```diff
@@ -1665,12 +1669,20 @@ function init({
...
         Storage.keepInstancesSync((key, value) => {
             cache.set(key, value);
             keyChanged(key, value);
         });
     }
+
+    return isUsingFallback;
+}
+
+function isUsingFallbackStorageProvider() {
```
marcaaron, line 1682: doc the return value

lib/Onyx.js
```diff
@@ -1604,6 +1605,7 @@ function subscribeToEvents() {
  * @param {Boolean} [options.shouldSyncMultipleInstances] Auto synchronize storage events between multiple instances
  * of Onyx running in different tabs/windows. Defaults to true for platforms that support local storage (web/desktop)
  * @param {Boolean} [options.debugSetState] Enables debugging setState() calls to connected components.
+ * @returns {Boolean} Whether there was a problem initializing the platform storage provider and Onyx uses the fallback provider instead.
```
marcaaron, line 1608: What will we use this for?

lib/Onyx.js
marcaaron, line 1682: I'm a little confused about why we have `isUsingFallbackStorageProvider()` and also `Storage.init()` that _seems like_ it would answer the same question.

lib/Onyx.js
```diff
@@ -1684,6 +1696,7 @@ const Onyx = {
     clear,
     getAllKeys,
     init,
+    isUsingFallbackStorageProvider,
```
marcaaron, line 1699: Similarly, to having the return value for `Onyx.init()` answer this question - not sure who needs to know this.

marcaaron: Maybe it would be useful to return the provider name vs. a boolean and then the client code could decide if they need this or not.

lib/storage/index.js
```diff
@@ -0,0 +1,266 @@
...
+
+let storageProviders = [PlatformStorage, MemoryOnlyProvider];
+let finishInitalization;
+const initPromise = new Promise((resolve) => {
+    finishInitalization = resolve;
+});
+let shouldKeepInstancesSync = false;
+
+/**
+ * Runs an operation consecutively on all storage providers
```
marcaaron, line 15: **To improve this comment answer these questions:**

What is an  example of an "operation"? 
Why it must run on "all storage providers"? 
Why do we want to run it "consecutively"?

lib/storage/index.js
```diff
@@ -0,0 +1,266 @@
...
+                .then(resolve)
+                .catch((error) => {
+                    if (providerIndex + 1 >= storageProviders.length) {
+                        reject(error);
+                        return;
+                    }
+
+                    runOperation(providerIndex + 1)
+                        .then(resolve)
+                        .catch(reject);
```
marcaaron, line 35: **gut feeling:** The level of abstraction here obfuscates things. Curious to learn the answers to my last questions though before assuming it is unnecessary.

lib/storage/index.js
```diff
@@ -0,0 +1,266 @@
...
+        return new Promise((resolve, reject) => {
+            provider[operationName](...params)
+                .then(resolve)
+                .catch((error) => {
+                    if (providerIndex + 1 >= storageProviders.length) {
+                        reject(error);
+                        return;
+                    }
+
+                    runOperation(providerIndex + 1)
```
marcaaron, line 33: Will this double the memory footprint of the app? It looks like we will have values stored in both the Onyx cache + "memory" provider.

marcaaron (changes requested):
Leaving some initial questions before doing a full review. I like the code organization stuff happening here. But concerned about memory usage.

lib/storage/index.js
tgolen: I am pretty sure it will double the memory footprint. Which makes sense because this is replacing disc storage with an in-memory storage.

I feel like it's probably OK until we identify that the doubled memory footprint is an actual problem.

Ideally, it would be nice that the Onyx cache and this in memory provider are one-in-the-same, but it is kind of out of scope for this and something we can look at doing in the future.

You can see I commented about the same thing in the GH and you can read Chris's response to it: https://github.com/Expensify/App/issues/29403#issuecomment-1870409516

quinthar:
Uhh, doubling footprint sounds bad.

tgolen:
Yes, while it could be bad, let's please keep in mind that this only
happens when the primary on-disk storage fails for some reason. In this
case, I feel having a doubled memory footprint is better than the app not
working at all. Would you agree?

quinthar:
Ohh true.

lib/Onyx.js
chrispader: To indicate to the user, that the app has degraded performance and is not using storage. Maybe, to also react to this or adapt the app...

> It might be nice to have a hybrid approach where Onyx is only using memory storage, and then the App displays a global banner like "There could be degraded performance, blah blah" to let the user know that something isn't quite right.

From https://github.com/Expensify/App/issues/29403#issuecomment-1821216687

lib/Onyx.js
chrispader: It's answering the same question, yes. For `Onyx.init` we would use the return value at app startup and react with showing a banner or disabling certain features, or whatever.

Th `isUsingFallbackStorageProvider` was just an extra function to check for this after initialization phase.

Also, since i designed this new storage interface in a way that you can potentially also use another fallback storage and/or add/remove providers manually, we can use this to check if the "main" storage provider is working after the replacement.

Does that make sense? 😅

lib/Onyx.js
chrispader: We can also do that, definitely. Just thought it might make sense to assume that the user/developer who uses Onyx isn't aware of all the internal providers and therefore only wants to know if it's working 100% or not.

We could also change this function to something like `getOnyxStats` or sth similar, which gives all infos relevant to performance/etc...

lib/storage/index.js
chrispader: Yes, but only in case the main storage provider (IDBKeyVal/SQLite) fails to initialize.

As described in https://github.com/Expensify/App/issues/29403#issuecomment-1870222315 i planned on completely removing `OnyxCache` in favor of `MemoryOnlyProvider` in a follow-up PR, since it's essentially doing the same thing.

For now, the app would only have a double memory footprint in case of the storage failing.

Is that an approach we can go with?

chrispader:
> Uhh, doubling footprint sounds bad.

Also to answer your concerns @quinthar 

https://github.com/Expensify/react-native-onyx/pull/439#discussion_r1438134597

arosiclair (commented):
Looks good on my end other than @marcaaron's concerns

lib/storage/index.js
marcaaron: Did we consider the suggestion [here](https://github.com/Expensify/App/issues/29403#issuecomment-1821937151)?

>Having a "provider" that just returns resolved promises for all the methods

marcaaron: >completely removing OnyxCache in favor of MemoryOnlyProvider in a follow-up PR, since it's essentially doing the same thing

Just my opinion, but sounds like more work for unclear gain.

marcaaron:
I'd love a solution where the "app does not break and the memory is not doubled". I _think_ we can have that if it's not too hard. Though it could be I am not understanding the memory provider idea or why it's valuable to have.

lib/Onyx.js
marcaaron: Ah, got it thanks! We should pick one way to do this. Either have it get returned by `Onyx.init()` or expose some method like you are doing with `isUsingFallbackStorageProvider()`

lib/Onyx.js
marcaaron: Yes! It makes sense. I am envisioning an alternate approach where if the storage fails then we just set a boolean like `Onyx.isMemoryOnly()` and the app developers can check to see if storage is working or not.

lib/Onyx.js
marcaaron: I think that gets me closer to having this just say what it means. Instead of alluding to a "fallback storage provider" maybe something that clearly indicates the "user can't persist data because storage doesn't work".

tgolen: I think Marc is getting close now to what the frontend developer wants to know. They don't want to know the storage provider being used, or if it's memory only, they only really care that the data is being persisted to disk or not (since the whole point of Onyx is to persist the state to disk). I also agree there should only be one way to check for this. 

However, when I look over my previous comment suggesting the application could show a banner that performance is degraded, I fear I was pre-optimizing.

Let's keep it simple for now and only build what we need. We have no plans to display any kind of banner like I suggested, so let's not try to build in support for it now. I think we can remove:

1. The return value from `Onyx.init()`
2. The exported method `isUsingFallbackStorageProvider()`

lib/storage/InstanceSync/index.native.js
```diff
@@ -0,0 +1,6 @@
+/**
+ * Noop on native
```
tgolen, line 2: Can you please add a little more to the comment do explain why it's a noop (because this feature is only used to keep multiple browser tabs in sync)?

lib/storage/index.js
tgolen: I agree, Marc. From looking at `init()`, it doesn't look like there would ever be two storage providers at once. Either the on-disk one initializes and the in-memory one is removed, or vice versa. It feels a bit over-engineered to assume there would be multiple data providers running at the same time.

Although, if this would allow the in-memory provider to be used for the in-memory cache, then I could understand why `runConsecutively()` would be necessary (but we aren't doing this, so it's still over-engineered).

tgolen (changes requested):
I like the general clean-up of the providers.

chrispader:
> I'd love a solution where the "app does not break and the memory is not doubled". I _think_ we can have that if it's not too hard. Though it could be I am not understanding the memory provider idea or why it's valuable to have.

@marcaaron just left a high-level comment here: https://github.com/Expensify/App/issues/29403#issuecomment-1880902424

lib/storage/index.js
chrispader: > Did we consider the suggestion [here](https://github.com/Expensify/App/issues/29403#issuecomment-1821937151)?
> 
> > Having a "provider" that just returns resolved promises for all the methods

As just suggested [here](https://github.com/Expensify/App/issues/29403#issuecomment-1880902424), i can just (temporarily) create somewhat of a `NoopProvider` or adapt the `MemoryOnlyProvider` to noop. Though in follow-up PR i'd like to replace `OnyxCache`

chrispader: > > completely removing OnyxCache in favor of MemoryOnlyProvider in a follow-up PR, since it's essentially doing the same thing
> 
> Just my opinion, but possibly sounds like more work for unclear gain.

I think the gain is definitely a less complex implementation of `Onyx.js` and the ability to add additional fallback storage providers, and/or just add more storage providers.

lib/storage/index.js
chrispader: Currently, the `Onyx.init` and `Storage.init` functions are not yet adapted to this, but this was the future plan

lib/storage/index.js
chrispader: i updated the comments of these helper functions

lib/storage/index.js
tgolen: I see the noop provider was added, but I don't see it referenced anywhere. Maybe I misunderstood the purpose of it?

tgolen: @chrispader you might have missed this, but I'm still waiting for an answer to my previous question here to conclude my review of this PR.

tgolen: Bump @chrispader I'm still waiting for an answer on this.

chrispader: @tgolen sorry for the delay.

I added the `NoopProvider` just to show what i meant for now. I'm going to simplify this PR and remove and unnecessary overhead for now and just use the `NoopProvider` instead of actually using a working fallback provider.

Based on what @marcaaron suggested in https://github.com/Expensify/App/issues/29403#issuecomment-1881775744

tgolen:
> I'm going to simplify this PR and remove and unnecessary overhead for now and just use the NoopProvider instead of actually using a working fallback provider.

@chrispader Is this ready to review again? I wasn't sure and I don't want to be the hold-up here.

chrispader:
@tgolen @marcaaron i just removed the additional "multi-storage" logic and just fallback to the `NoopProvider` in case the main platform storage provider couldn't be initialised.

This whole re-structuring and cleanup as well as this extra "storage" layer comes in handy though, since we can now handle the initialisation phase first, and only then start handling operations.

I'm going to test this in the app now, but make sure to review this PR, as it should be ready!

lib/storage/index.js
```diff
@@ -0,0 +1,215 @@
...
+        return isUsingFallback;
+    },
+
+    /**
+     * Get the value of a given key or return `null` if it's not available in memory
+     * @param {String} key
+     * @return {Promise<*>}
+     */
+    getItem(key) {
+        return runAfterInit(() => provider.getItem(key));
```
marcaaron, line 64: **confusion:** Why do we need `runAfterInit()`?

lib/storage/providers/IDBKeyValProvider.js
```diff
@@ -4,30 +4,69 @@ import utils from '../../utils';
...
+     */
+    name: 'IDBKeyValProvider',
+
+    /**
+     * Initializes the storage provider
+     */
+    init() {
+        const newIdbKeyValStore = createStore('OnyxDB', 'keyvaluepairs');
+
+        if (newIdbKeyValStore == null) throw Error('IDBKeyVal store could not be created');
```
marcaaron, line 21: What problem specifically does this solve? Do we have any documented cases where `createStore()` returned `null`? What was the problem we saw? What about when an irrecoverable error happens when trying to save after initializing the store successfully? 

(that second situation is why this issue was created in the first place - and I'm not convinced we have solved it)

marcaaron (changes requested):

lib/Onyx.js
```diff
@@ -1664,13 +1673,6 @@ function init({
 
     // Initialize all of our keys with data provided then give green light to any pending connections
     Promise.all([addAllSafeEvictionKeysToRecentlyAccessedList(), initializeWithDefaultKeyStates()]).then(deferredInitTask.resolve);
-
-    if (shouldSyncMultipleInstances && _.isFunction(Storage.keepInstancesSync)) {
-        Storage.keepInstancesSync((key, value) => {
-            cache.set(key, value);
-            keyChanged(key, value);
-        });
-    }
```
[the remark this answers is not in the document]
chrispader: Just to keep the code that basically initiates the Storage layer together. Technically, it makes no difference

lib/storage/index.js
chrispader: Since i moved the initialization of the provider to `provider.init()`, we need this, to prevent any operations from running before the storage provider actually got intialized/created

chrispader:
One more time, i'm asking for you're opinions on this new approach. @marcaaron @tgolen 

(no need to completely review yet, i'm just curious what you think :))

The current approach handles errors during initialization and runtime, while also utilizing the extra storage layer and the cleaned up provider structure.

Still can't really test this in a real-life scenario, since we don't know how to trigger this `Internal error opening backing store for indexedDB.open` error, but it basically works with any error thrown for any operation.

As In the `tryOrDegradePerformance` function we could additionally check for specific errors. As long as we don't know which errors exactly to expect, we can start adding all the problematic errors to a list of error types and check for them in the functions `catch` block...

tgolen:
> it basically works with any error thrown for any operation.

My only concern with this is the chance of false positives. It seems like we are struggling between the extremes of not knowing which errors get thrown or only knowing one error that gets thrown with no way to reproduce it.

Maybe this hints that we need better error reporting first so that we can make a better assumption about which errors are actually occurring.

chrispader:
> > it basically works with any error thrown for any operation.
> 
> My only concern with this is the chance of false positives. It seems like we are struggling between the extremes of not knowing which errors get thrown or only knowing one error that gets thrown with no way to reproduce it.

Yes you're right, it's either this or that extreme case. I don't see any other way right now to check for these errors, unless we're able to confidently reproduce it.

After doing some internet research, i'm like 80% sure the `Internal error opening backing store for indexedDB.open` error described in https://github.com/Expensify/App/issues/29403 comes from the storage being full.

> Maybe this hints that we need better error reporting first so that we can make a better assumption about which errors are actually occurring.

My suggestion here would be to basically have this error check for initialization and every operation (`tryOrDegradePerformance`) and only degrade performance if we detect a known error that will prevent the storage from keep working. This way, we can easily add more errors on the go.

There really is no official way from `idb-keyval` or `IndexedDB` to detect these kind of errors and handle them. I think the current approach is the most generic and will fix this issue at least for the known error.

marcaaron:
On the subject of error reporting... I worked on improving it a bit in the past and [an internal log search](https://www.expensify.com/_devportal/tools/logSearch/#sort=asc&size=10000&query=blob%3A%20%22Out%20of%20storage.%20But%20found%20no%20acceptable%20keys%20to%20remove.%22%20AND%20timestamp%3A%5B2024-01-11T00%3A00%20TO%202024-01-26T23%3A59%5D) will show us when storage is failing and why it failed (some explanations more cryptic than others).

This will give us the ability to catch something like a "critical" vs. "benign" loss of client data (if there is such a thing).

With reliable updates, it seems extremely important to ensure storage failures don't lead to data loss at all e.g. if you are able to store one recent update, but for some reason could not store some previous update - the previous update is now lost forever. Maybe I am wrong, but the current system we have depends on a near perfect reliability. Anytime we fail to store something the update chain will essentially be broken.

So, I might actually argue that we should **always** stop storing things **immediately** when any error happens and address these errors with urgency.

chrispader:
> On the subject of error reporting... I worked on improving it a bit in the past and [an internal log search](https://www.expensify.com/_devportal/tools/logSearch/#sort=asc&size=10000&query=blob%3A%20%22Out%20of%20storage.%20But%20found%20no%20acceptable%20keys%20to%20remove.%22%20AND%20timestamp%3A%5B2024-01-11T00%3A00%20TO%202024-01-26T23%3A59%5D) will show us when storage is failing and why it failed (some explanations more cryptic than others).
> 
> This will give us the ability to catch something like a "critical" vs. "benign" loss of client data (if there is such a thing).

unfortunately i don't have access to that link.. are you referring to some sort of analytics where we collect data about (the type of) errors occurring (in production)? 

> With reliable updates, it seems extremely important to ensure storage failures don't lead to data loss at all e.g. if you are able to store one recent update, but for some reason could not store some previous update - the previous update is now lost forever. Maybe I am wrong, but the current system we have depends on a near perfect reliability. Anytime we fail to store something the update chain will essentially be broken.
> 
> So, I might actually argue that we should **always** stop storing things **immediately** when any error happens and address these errors with urgency.

If i understand you correctly, this means we would need to implement some sort of "transaction" system, to basically rollback changes if there an error occurs later in a chain of operations?

We can definitely check for initialization errors - probably like the original one described in the issue - and degrade performance before any operations are even triggered.

But if a fatal error occurs at some other point during Onyx's lifetime, we can only degrade performance from there... meaning after the failing operation. The data will still be stored in cache (in-memory), though that data will obviously be lost after a refresh. Without some other persisted fallback storage solution, i'm not sure if we can prevent this sort of data loss.

Maybe i got something wrong, because i'm not totally sure what you are referring to here and if it isn't extending the scope of this PR/issue massively.

chrispader:
JFYI @marcaaron @tgolen 

I'm going to be OOO from 31/01/2024 - 17/03/2024, which means i can either complete a (limited) version of this issue fix/PR by tomorrow/tuesday, or i'll have to hand it over to someone else to takeover.

From my perspective, i don't think the current state of this PR affects the behaviour of Onyx negatively at all. The current implementation of `tryOrDegradePerformance` checks for only this specific error (`Internal error opening backing store for indexedDB.open`) and only then degrades performance.

This PR also includes lots of re-structuring and code quality improvements and paves the way for future TypeScript typings and clean-up.

Do you guys think we can merge this (as-is) and work on advanced error reporting and preventing in a follow-up PR?

tgolen:
I think with such large changes in this PR, and you being gone for so long, I'd actually prefer holding this PR until you are back. Unless someone else is willing and able to take over maintaining it. We've had a lot of issues where Onyx PRs need to be reverted and then it causes as cascade effect on other PRs. I think the risk of causing a regression here is pretty high. The best way to combat that is to split this PR up into multiple smaller pieces.

chrispader:
> I think with such large changes in this PR, and you being gone for so long, I'd actually prefer holding this PR until you are back. Unless someone else is willing and able to take over maintaining it. We've had a lot of issues where Onyx PRs need to be reverted and then it causes as cascade effect on other PRs. I think the risk of causing a regression here is pretty high. The best way to combat that is to split this PR up into multiple smaller pieces.

I understand! I can hand this PR over to another Margelo engineer and ask them to split it up into more easily reviewable PRs. Otherwise (if this not a priority) i can also continue working on it in March.

marcaaron:
>If i understand you correctly, this means we would need to implement some sort of "transaction" system, to basically rollback changes if there an error occurs later in a chain of operations

Hmm no I wasn't trying to imply any kind of transaction system (unsure what that would entail). I am just stating a fact - which is that if you are unable to store an update, but allow storing a later update then you will have irrecoverable "gaps" in the client side data.

hannojg:
Hey, as Chris is OOO pretty long we are internally assigning someone new to carry on this PR.
The idea is as Tim pointed out to split up the PR in smaller PRs to reduce the risk of hard to fix regressions.

tgolen:
Thanks @hannojg. I'm going to close this PR for now then.

PR #303, opened 2023-08-17
Only use cache value when merge is not pending for key

marcaaron:
### Details

See documentation in code comment for more info and [this Slack thread](https://expensify.slack.com/archives/C035J5C9FAP/p1692145133839509?thread_ts=1691655217.992579&cid=C035J5C9FAP).

### Automated Tests

We should add these later...

### Linked PRs

https://github.com/Expensify/App/compare/marcaaron-reapply-onyx-upgrade-use-cache-with-fixes?expand=1

PR #462, opened 2024-01-25
Implement useOnyx hook

fabioh8010:
### Details

This PR adds a new way to consume Onyx data in the application, the `useOnyx` hook.

Relevant links:

- [Slack proposal](https://expensify.slack.com/archives/C01GTK53T8Q/p1706718802348819)
- [Proposal doc](https://docs.google.com/document/d/1oUDZH73gaeSM_8A5wuG1hc6L1-8c3XmVHfxLXxjN7YY/edit)

### Related Issues

https://github.com/Expensify/App/issues/34339

### Automated Tests

N/A

### Manual Tests

N/A

### Author Checklist

    - [x] Android / native
    - [x] Android / Chrome
    - [x] iOS / native
    - [x] iOS / Safari
    - [x] MacOS / Chrome / Safari
    - [x] MacOS / Desktop

### Screenshots/Videos
<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

### DIFF
--- a/lib/Onyx.js
+++ b/lib/Onyx.js
@@ -215,7 +215,7 @@ function isCollectionMemberKey(collectionKey, key) {
 /**
  * Splits a collection member key into the collection key part and the ID part.
  * @param {String} key - The collection member key to split.
- * @returns {[String, String]} A tuple where the first element is the collection part and the second element is the ID part.
+ * @returns {Array<String>} A tuple where the first element is the collection part and the second element is the ID part.
  */
 function splitCollectionMemberKey(key) {
     const underscoreIndex = key.indexOf('_');
--- a/API.md
+++ b/API.md
@@ -22,8 +22,15 @@ cause react to schedule the updates at once instead of after each other. This is
 and runs it through a reducer function to return a subset of the data according to a selector.
 The resulting collection will only contain items that are returned by the selector.</p>
 </dd>
+<dt><a href="#isCollectionKey">isCollectionKey(key)</a> ⇒ <code>Boolean</code></dt>
+<dd><p>Checks to see if the a subscriber&#39;s supplied key
+is associated with a collection of keys.</p>
+</dd>
 <dt><a href="#isCollectionMemberKey">isCollectionMemberKey(collectionKey, key)</a> ⇒ <code>Boolean</code></dt>
 <dd></dd>
+<dt><a href="#splitCollectionMemberKey">splitCollectionMemberKey(key)</a> ⇒ <code>Array.&lt;String&gt;</code></dt>
+<dd><p>Splits a collection member key into the collection key part and the ID part.</p>
+</dd>
 <dt><a href="#tryGetCachedValue">tryGetCachedValue(key, mapping)</a> ⇒ <code>Mixed</code></dt>
 <dd><p>Tries to get a value from the cache. If the value is not present in cache it will return the default value or undefined.
 If the requested key is a collection, it will return an object with all the collection members.</p>
@@ -34,7 +41,7 @@ If the requested key is a collection, it will return an object with all the coll
 <dt><a href="#disconnect">disconnect(connectionID, [keyToRemoveFromEvictionBlocklist])</a></dt>
 <dd><p>Remove the listener for a react component</p>
 </dd>
-<dt><a href="#scheduleSubscriberUpdate">scheduleSubscriberUpdate(key, value, [canUpdateSubscriber])</a> ⇒ <code>Promise</code></dt>
+<dt><a href="#scheduleSubscriberUpdate">scheduleSubscriberUpdate(key, value, prevValue, [canUpdateSubscriber])</a> ⇒ <code>Promise</code></dt>
 <dd><p>Schedules an update that will be appended to the macro task queue (so it doesn&#39;t update the subscribers immediately).</p>
 </dd>
 <dt><a href="#scheduleNotifyCollectionSubscribers">scheduleNotifyCollectionSubscribers(key, value)</a> ⇒ <code>Promise</code></dt>
@@ -89,13 +96,6 @@ value will be saved to storage after the default value.</p>
 </dd>
 <dt><a href="#setMemoryOnlyKeys">setMemoryOnlyKeys(keyList)</a></dt>
 <dd><p>When set these keys will not be persisted to storage</p>
-</dd>
-<dt><a href="#onClear">onClear(callback)</a></dt>
-<dd><p>Sets the callback to be called when the clear finishes executing.</p>
-</dd>
-<dt><a href="#subscribeToEvents">subscribeToEvents()</a></dt>
-<dd><p>Subscribes to the Broadcast channel and executes actions based on the
-types of events.</p>
 </dd>
 <dt><a href="#init">init([options])</a></dt>
 <dd><p>Initialize the store with actions and listening for storage events</p>
@@ -153,6 +153,18 @@ The resulting collection will only contain items that are returned by the select
 | selector | <code>String</code> \| <code>function</code> | (see method docs for getSubsetOfData() for full details) |
 | [withOnyxInstanceState] | <code>Object</code> |  |
 
+<a name="isCollectionKey"></a>
+
+## isCollectionKey(key) ⇒ <code>Boolean</code>
+Checks to see if the a subscriber's supplied key
+is associated with a collection of keys.
+
+**Kind**: global function  
+
+| Param | Type |
+| --- | --- |
+| key | <code>String</code> | 
+
 <a name="isCollectionMemberKey"></a>
 
 ## isCollectionMemberKey(collectionKey, key) ⇒ <code>Boolean</code>
@@ -162,6 +174,18 @@ The resulting collection will only contain items that are returned by the select
 | --- | --- |
 | collectionKey | <code>String</code> | 
 | key | <code>String</code> | 
+
+<a name="splitCollectionMemberKey"></a>
+
+## splitCollectionMemberKey(key) ⇒ <code>Array.&lt;String&gt;</code>
+Splits a collection member key into the collection key part and the ID part.
+
+**Kind**: global function  
+**Returns**: <code>Array.&lt;String&gt;</code> - A tuple where the first element is the collection part and the second element is the ID part.  
+
+| Param | Type | Description |
+| --- | --- | --- |
+| key | <code>String</code> | The collection member key to split. |
 
 <a name="tryGetCachedValue"></a>
 
@@ -221,7 +245,7 @@ Onyx.disconnect(connectionID);
 ```
 <a name="scheduleSubscriberUpdate"></a>
 
-## scheduleSubscriberUpdate(key, value, [canUpdateSubscriber]) ⇒ <code>Promise</code>
+## scheduleSubscriberUpdate(key, value, prevValue, [canUpdateSubscriber]) ⇒ <code>Promise</code>
 Schedules an update that will be appended to the macro task queue (so it doesn't update the subscribers immediately).
 
 **Kind**: global function  
@@ -230,6 +254,7 @@ Schedules an update that will be appended to the macro task queue (so it doesn't
 | --- | --- | --- |
 | key | <code>String</code> |  |
 | value | <code>\*</code> |  |
+| prevValue | <code>\*</code> |  |
 | [canUpdateSubscriber] | <code>function</code> | only subscribers that pass this truth test will be updated |
 
 **Example**  
@@ -410,24 +435,6 @@ When set these keys will not be persisted to storage
 | --- | --- |
 | keyList | <code>Array.&lt;string&gt;</code> | 
 
-<a name="onClear"></a>
-
-## onClear(callback)
-Sets the callback to be called when the clear finishes executing.
-
-**Kind**: global function  
-
-| Param | Type |
-| --- | --- |
-| callback | <code>function</code> | 
-
-<a name="subscribeToEvents"></a>
-
-## subscribeToEvents()
-Subscribes to the Broadcast channel and executes actions based on the
-types of events.
-
-**Kind**: global function  
 <a name="init"></a>
 
 ## init([options])


lib/useOnyx.ts
```diff
@@ -0,0 +1,132 @@
...
+            Onyx.addToEvictionBlockList(key, connectionIDRef.current);
+        }
+    }, [key, options?.canEvict]);
+
+    return value;
+}
+
+const cache = new Map<string, unknown>();
+
+function useOnyxWithSyncExternalStore<TKey extends OnyxKey>(key: TKey, options?: UseOnyxOptions<TKey>): OnyxValue<TKey> {
```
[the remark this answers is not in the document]
fabioh8010: This is just for testing purposes @roryabraham , I will keep only the `useSyncExternalStore` version before opening the PR 😄

.eslintrc.js
```diff
@@ -1,5 +1,6 @@
 module.exports = {
-    extends: ['expensify', 'prettier'],
+    extends: ['expensify', 'prettier', 'plugin:react-hooks/recommended'],
```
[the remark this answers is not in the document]
blazejkustra: [Done](https://github.com/Expensify/eslint-config-expensify/pull/85) ✅ 

But this resulted in a cascade of updates in the config, not sure if we want to proceed.

lib/useOnyx.ts
```diff
@@ -0,0 +1,144 @@
...
+    const connectionIDRef = useRef<number | null>(null);
+    const previousKey = usePrevious(key);
+
+    /**
+     * According to React docs, `getSnapshot` is a function that returns a snapshot of the data in the store that’s needed by the component.
+     * **While the store has not changed, repeated calls to getSnapshot must return the same value.**
+     * If the store changes and the returned value is different (as compared by Object.is), React re-renders the component.
+     *
+     * When the `key` is changed (e.g. to get a different record from a collection) and it's not yet in the cache,
+     * we return the value from the previous key to avoid briefly returning a `null` value to the component, thus avoiding a useless re-render.
```
blazejkustra, line 87: I raised a concern in a DM that this may led to some unexpected bugs in NewDot. Fabio's argument was that this is to omit unnecessary re-renders, but with a status property when a key is changed it will be re-rendered either way. 

I'm not convinced that this logic is needed.

fabioh8010: I've removed after my last changes (wasn't being able to reproduce it anymore)

lib/useOnyx.ts
```diff
@@ -0,0 +1,200 @@
...
+
+    /**
+     * If set to false, then no data will be prefilled into the component.
+     */
+    initWithStoredValues?: boolean;
+
+    /**
+     * TODO: Check if we still need this flag and associated logic.
+     */
+    // allowStaleData?: boolean;
```
fabioh8010, line 23: @roryabraham @tgolen Could you explain me what this `allowStaleData` property do exactly in `withOnyx` HOC?

.eslintrc.js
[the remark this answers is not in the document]
blazejkustra: @fabioh8010 Let's update eslint-config-expensify in this PR and remove 'plugin:react-hooks/recommended' here

lib/useOnyx.ts
```diff
@@ -0,0 +1,200 @@
...
+}
+
+function getCachedValue<TKey extends OnyxKey>(key: TKey, selector?: Selector<TKey, unknown, unknown>): OnyxValue<TKey> {
+    return (Onyx.tryGetCachedValue(key, {selector}) ?? null) as OnyxValue<TKey>;
+}
+
+function useOnyx<TKey extends OnyxKey, TReturnData = OnyxValue<TKey>>(key: TKey, options?: UseOnyxOptions<TKey, TReturnData>): UseOnyxData<TReturnData> {
+    const connectionIDRef = useRef<number | null>(null);
+    const previousKey = usePrevious(key);
+    const previousDataRef = useRef<TReturnData | null>(null);
```
blazejkustra, line 57: Why not just `dataRef`? OR `currentDataRef`?
```suggestion
    const currentDataRef = useRef<TReturnData | null>(null);
```

lib/useOnyx.ts
```diff
@@ -0,0 +1,200 @@
...
+         * If they are not equal, we update the internal data and return it.
+         *
+         * If they are equal, we just return the previous internal data.
+         */
+        const newData = getCachedValue(key, options?.selector);
+        if (!deepEqual(previousDataRef.current, newData)) {
+            previousDataRef.current = newData as TReturnData;
+        }
+
+        return previousDataRef.current as TReturnData;
```
blazejkustra, line 122: This is the same code as above in Case 2, right?

lib/useOnyx.ts
```diff
@@ -0,0 +1,200 @@
...
+        (onStoreChange: () => void) => {
+            connectionIDRef.current = Onyx.connect({
+                key: key as CollectionKeyBase,
+                callback: () => {
+                    /**
+                     * We don't need to update the Onyx cache again here, when `callback` is called the cache is already
+                     * expected to be updated, so we just signal that the store changed and `getSnapshot()` can be called.
+                     */
+                    fetchStatusRef.current = 'loaded';
+                    onStoreChange();
```
blazejkustra, line 135: Does it matter if we swap these?
```suggestion
                    onStoreChange();
                    fetchStatusRef.current = 'loaded';
```

blazejkustra (changes requested):

lib/useOnyx.ts
fabioh8010: Is almost the same, in Case 2 we have a condition. I kept them separated to explain better the possible cases.

lib/useOnyx.ts
```diff
@@ -0,0 +1,206 @@
...
+
+function isCollectionMemberKey<TKey extends OnyxKey>(key: TKey): boolean {
+    return key.includes('_') && !key.endsWith('_');
+}
+
+function getCachedValue<TKey extends OnyxKey>(key: TKey, selector?: Selector<TKey, unknown, unknown>): OnyxValue<TKey> {
+    return (Onyx.tryGetCachedValue(key, {selector}) ?? null) as OnyxValue<TKey>;
+}
+
+function useOnyx<TKey extends OnyxKey, TReturnData = OnyxValue<TKey>>(key: TKey, options?: UseOnyxOptions<TKey, TReturnData>): UseOnyxData<TReturnData> {
```
fabioh8010, line 54: Currently there is a minor type problem here when using the selector, I'm working on a fix.

package.json
```diff
@@ -58,11 +58,12 @@
     "@typescript-eslint/eslint-plugin": "^6.19.0",
     "@typescript-eslint/parser": "^6.19.0",
     "eslint": "^8.56.0",
-    "eslint-config-expensify": "^2.0.43",
+    "eslint-config-expensify": "^2.0.44",
     "eslint-config-prettier": "^8.8.0",
     "eslint-plugin-import": "^2.29.1",
     "eslint-plugin-jsx-a11y": "^6.8.0",
     "eslint-plugin-react": "^7.31.10",
+    "eslint-plugin-react-hooks": "^4.6.0",
```
blazejkustra, line 66: You can remove this plugin now, please double check `eslint-plugin-react-hooks` from eeslint-config-expensify works 🫣

blazejkustra:
Lint is failing, you can either fix these errors (add {} to each new Promise), or turn of the rule

lib/useOnyx.ts
tgolen: I'll defer most of my knowledge about it to this code comment:

https://github.com/Expensify/react-native-onyx/blob/10b74b7fd8b68ba9280b14015d62d1c6cefe3026/lib/withOnyx.js#L66-L76

tgolen: My basic understanding is that it allows the component to be rendered immediately instead of the next tick due to cache optimization.

package.json
fabioh8010: Seems to be working 🙂

lib/useOnyx.ts
hannojg: Yeah I think this is still needed.

- A component can render immediately if the value for a certain onyx key has already been cached
- However, the component we want to render might misses to be rendered immediately if the value is currently being updated
- This flag enables to render the component with "stale" data (the data before the update will happen)

(Reading through that code makes me wonder why we can't update the cache directly once we merge new values, but thats probably a different topic)

.eslintrc.js
```diff
@@ -67,6 +67,7 @@ module.exports = {
                         tsx: 'never',
                     },
                 ],
+                'rulesdir/prefer-onyx-connect-in-libs': 'off',
```
hannojg, line 70: [DUMB QUESTION] Why exactly did we turn that off here?

lib/useOnyx.ts
hannojg: One side note: I still have to review the whole PR, maybe the usage of that prop become obsolete (let me double check 🔍 )

lib/useOnyx.ts
```diff
@@ -0,0 +1,208 @@
...
+type FetchStatus = 'loading' | 'loaded';
+
+type UseOnyxDataValue<TKey extends OnyxKey, TValue> = TValue extends OnyxValue<TKey> ? TValue : TKey extends CollectionKeyBase ? NonNullable<OnyxCollection<TValue>> : TValue;
+
+type UseOnyxData<TKey extends OnyxKey, TValue> = {
+    value: UseOnyxDataValue<TKey, TValue>;
+    status: FetchStatus;
+};
+
+function isCollectionMemberKey<TKey extends OnyxKey>(key: TKey): boolean {
```
hannojg, line 48: Can we reuse `isCollectionMemberKey` from Onyx.js here?

hannojg (commented):
Code looks 👌👌👌 to me, just wondering if we should add some unit tests for the hook?

[Need to check one more use case in detail, will update in a moment]

lib/useOnyx.ts
fabioh8010: @hannojg Please share if you have some basic example about how to test this flag with `withOnyx` (we have only a few usages in the app), so I can also compare here if the behavior is consistent with `useOnyx`.

hannojg: Push okay, so actually after reviewing the PR I feel like maybe this exact option is needed anymore actually!

However, I am wondering, what happens to the use case of `!Onyx.hasPendingMergeForKey(key)`:

https://github.com/Expensify/react-native-onyx/blob/10b74b7fd8b68ba9280b14015d62d1c6cefe3026/lib/withOnyx.js#L77-L79

It has been initially added here:

- https://github.com/Expensify/react-native-onyx/pull/303

I believe right now we are missing that behaviour and the `useOnyx` hook would show stale data, e.g. meaning this behaviour:

```js
Onyx.merge('report_123', newValue); 
Navigation.navigate('report_screen', 'report_123');

// In ReportScreen

const data = useOnyx('report_123')
```

data would first be `staleData` and then in a second render become `newValue`.

Right now we are explicitly not using the cache value if has `Onyx.hasPendingMergeForKey(key)`.
However we made it possible to allow for reading stale data by that flag. 

So I think both should be added back:

- The  `Onyx.hasPendingMergeForKey(key)` for reading from cache or nah
- The `allowStaleData` flag

hannojg (changes requested):
After more careful consideration I think we need to be double careful about this one use case to not introduce a regression

lib/useOnyx.ts
```diff
@@ -0,0 +1,189 @@
...
+         *
+         * If they are equal, we just return the previous internal data.
+         */
+        const newData = getCachedValue(key, options?.selector);
+        if (!deepEqual(currentDataRef.current, newData)) {
+            currentDataRef.current = newData as CachedValue<TKey, TReturnData>;
+        }
+
+        return currentDataRef.current as CachedValue<TKey, TReturnData>;
+    }, [key, options?.selector]);
```
roryabraham, line 112: It's worth noting that if the selector or options are unstable references, then `getSnapshot` will be redefined with each render.

We could consider using `useRef` to stabilize the reference internally so that caller's don't have to be wary of this pitfall:

```diff
diff --git a/lib/useOnyx.ts b/lib/useOnyx.ts
index 965cf4f..9b454f9 100644
--- a/lib/useOnyx.ts
+++ b/lib/useOnyx.ts
@@ -78,6 +78,9 @@ function useOnyx<TKey extends OnyxKey, TReturnData = OnyxValue<TKey>>(key: TKey,
         );
     }, [previousKey, key]);
 
+    const selector = useRef(options?.selector);
+    selector.current = options?.selector;
+
     /**
      * According to React docs, `getSnapshot` is a function that returns a snapshot of the data in the store that’s needed by the component.
      * **While the store has not changed, repeated calls to getSnapshot must return the same value.**
@@ -89,7 +92,7 @@ function useOnyx<TKey extends OnyxKey, TReturnData = OnyxValue<TKey>>(key: TKey,
          *
          * We just return the data from the Onyx cache.
          */
-        if (!Onyx.isCollectionKey(key) && !options?.selector) {
+        if (!Onyx.isCollectionKey(key) && !selector.current) {
             return getCachedValue(key);
         }
 
@@ -103,13 +106,13 @@ function useOnyx<TKey extends OnyxKey, TReturnData = OnyxValue<TKey>>(key: TKey,
          *
          * If they are equal, we just return the previous internal data.
          */
-        const newData = getCachedValue(key, options?.selector);
+        const newData = getCachedValue(key, selector.current);
         if (!deepEqual(currentDataRef.current, newData)) {
             currentDataRef.current = newData as CachedValue<TKey, TReturnData>;
         }
 
         return currentDataRef.current as CachedValue<TKey, TReturnData>;
-    }, [key, options?.selector]);
+    }, [key, selector]);
 
     const subscribe = useCallback(
         (onStoreChange: () => void) => {
```

I've also seen some people wrap this pattern with a simple `useLiveRef` utility hook:

```ts
function useLiveRef<T>(value: T) {
  const ref = useRef(value);
  ref.current = value;
  return ref;
}

...

const selector = useLiveRef(options?.selector);
```

That would allow us to safely use this hook like so:

```ts
const report = useOnyx(ONYXKEYS.COLLECTION.REPORT, {
    selector: (reports) => reports[reportID],
});
```

Instead of having to do some verbose memoization:

```ts
const selector = useCallback((reports) => reports[reportID], [reportID]);
const options = useMemo(() => ({selector}), [selector]);
const report = useOnyx(ONYXKEYS.COLLECTION.REPORT, options);
```

lib/useOnyx.ts
```diff
@@ -0,0 +1,189 @@
...
+
+    const subscribe = useCallback(
+        (onStoreChange: () => void) => {
+            connectionIDRef.current = Onyx.connect({
+                key: key as CollectionKeyBase,
```
roryabraham, line 117: Why would this be a `CollectionKeyBase`? Couldn't it be any `OnyxKey` _or_ `CollectionKeyBase`?

roryabraham (changes requested):
Also, let's add some automated UI tests to cover this hook? There are some examples in `withOnyxTest` we can follow.

Otherwise this is looking pretty good. It's so simple compared with `withOnyx` 🙂

lib/useOnyx.ts
fabioh8010: Thanks for the example @hannojg ❤️ I will try to reproduce here and do the adjustments if necessary

.eslintrc.js
fabioh8010: This rule is complaining in my TS files because I'm calling `Onyx` methods

lib/useOnyx.ts
fabioh8010: I'm not sure, I saw that we already have a `isCollectionMemberKey` in `Onyx.js` but it actually expects two params (`collectionKey` and `key`), while I have only `key` in the hook. I will check again if we can reuse it somehow

fabioh8010: Actually I think we can reuse it, will do it

lib/useOnyx.ts
```diff
@@ -0,0 +1,189 @@
...
+         * Case 1 - We have a non-collection key without selector
+         *
+         * We just return the data from the Onyx cache.
+         */
+        if (!Onyx.isCollectionKey(key) && !options?.selector) {
+            return getCachedValue(key);
+        }
+
+        /**
+         * Case 2 - We have a non-collection/collection key with/without selector
```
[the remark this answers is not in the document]
fabioh8010: I will adjust to `Case 2 - We have a non-collection/collection key and/or selector` since Case 2 also handles non-collection keys, wdyt?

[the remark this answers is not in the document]
roryabraham: collection key and/or selector expands to:

"collection key and selector or selector"

Which is logically equivalent to:

"collection key and selector or (not collection key and selector)"

whatever you think is clearest though is fine. It's just a comment

lib/useOnyx.ts
fabioh8010: That makes sense to me, thanks for the suggestion!

lib/useOnyx.ts
fabioh8010: It's because I'm supplying a boolean to `waitForCollectionCallback` and that is making TS require a `CollectionKeyBase` here, but as this connect usage is generic it's fine to assert to `CollectionKeyBase` imo.

fabioh8010:
Most of the comments addressed, the only pending is the `allowStaleData` flag and the tests which I will continue tomorrow.

lib/Onyx.d.ts
```diff
@@ -115,6 +120,21 @@ declare const METHOD: {
...
+declare function isCollectionKey(key: OnyxKey): key is CollectionKeyBase;
+
+declare function isCollectionMemberKey<TCollectionKey extends CollectionKeyBase>(collectionKey: TCollectionKey, key: string): key is `${TCollectionKey}${string}`;
+
+/**
+ * Splits a collection member key into the collection key part and the ID part.
+ * @param key - The collection member key to split.
+ * @returns A tuple where the first element is the collection part and the second element is the ID part.
+ */
+declare function splitCollectionMemberKey<TKey extends CollectionKey>(key: TKey): [TKey extends `${infer Prefix}_${string}` ? `${Prefix}_` : never, string];
```
fabioh8010, line 136: @blazejkustra Maybe you have some idea how to improve these typings? It should allow only collection member keys, but currently is allowing collection keys too.

```ts
Onyx.splitCollectionMemberKey(`${ONYXKEYS.COLLECTION.POLICY}${policyID}`); // ["policy_", string], correct
Onyx.splitCollectionMemberKey(ONYXKEYS.COLLECTION.POLICY); // ["policy_", string], should raise an error in the param
```

blazejkustra: How about this?

```ts
type CollectionKeyBase = 'policy_' | 'report_';
type RemoveUnderscore<T> = T extends `${infer Word}_` ? Word : never;

const ONYXKEYS = {COLLECTION: {POLICY: 'policy_'}} as const;

declare function splitCollectionMemberKey<TKey extends RemoveUnderscore<CollectionKeyBase>, TRest extends string>(key: TRest extends '' ? never : `${TKey}_${TRest}`): [TKey, TRest];

splitCollectionMemberKey(`${ONYXKEYS.COLLECTION.POLICY}${123}`); // ["policy_", string], correct
splitCollectionMemberKey(ONYXKEYS.COLLECTION.POLICY);
```

blazejkustra: I'm not sure if it's worth it, as this type is quite complicated, up to you 👍

lib/Onyx.js
```diff
@@ -213,6 +212,15 @@ function isCollectionMemberKey(collectionKey, key) {
     return Str.startsWith(key, collectionKey) && key.length > collectionKey.length;
 }
 
+/**
+ * Splits a collection member key into the collection key part and the ID part.
+ * @param {String} key - The collection member key to split.
+ * @returns {[String, String]} A tuple where the first element is the collection part and the second element is the ID part.
+ */
+function splitCollectionMemberKey(key) {
+    return key.split('_');
```
tgolen, line 221: Should this include `isCollectionMemberKey(key) && key.split('_')` so that it's a little bit more safe (say someone tries to use an `_` in the name of a key that's not a collection)?

lib/useOnyx.ts
```diff
@@ -0,0 +1,193 @@
...
+     */
+    initWithStoredValues?: boolean;
+
+    /**
+     * TODO: Check if we still need this flag and associated logic.
+     */
+    // allowStaleData?: boolean;
+
+    /**
+     * Sets an initial value to be returned by the hook during the first render.
```
tgolen, line 27:
```suggestion
     * This value will be returned by the hook on the first render while the data is being read from Onyx.
```

lib/useOnyx.ts
```diff
@@ -0,0 +1,193 @@
...
+     */
+    initialValue?: OnyxValue<TKey>;
+
+    /**
+     * If included, this will be used to subscribe to a subset of an Onyx key's data.
```
tgolen, line 32:
```suggestion
     * This will be used to subscribe to a subset of an Onyx key's data.
```

tgolen (changes requested):

lib/Onyx.d.ts
fabioh8010: Yeah, let's avoid overcomplicating and skip this for now.

.eslintrc.js
fabioh8010: @tgolen Actually I just saw that [we are suppressing the rule in withOnyx.js too](https://github.com/Expensify/react-native-onyx/blob/main/lib/withOnyx.js#L288). I don't think is a problem with the ESLint, just that it doesn't make too much sense to have this rule enabled in this repository since `rulesdir/prefer-onyx-connect-in-libs` looks like tailored for E/App. For example, this is the error that is shown:

> Only call Onyx.connect() from inside a /src/libs/** file. React components and non-library code should not use Onyx.connect()

It doesn't apply to here.

lib/useOnyx.ts
```diff
@@ -0,0 +1,193 @@
...
+         * These conditions will ensure we can only handle dynamic collection member keys from the same collection.
+         */
+        if (previousKey === key) {
+            return;
+        }
+
+        const previousCollectionKey = Onyx.splitCollectionMemberKey(previousKey)[0];
+        const collectionKey = Onyx.splitCollectionMemberKey(key)[0];
+
+        if (Onyx.isCollectionMemberKey(previousCollectionKey, previousKey) && Onyx.isCollectionMemberKey(collectionKey, key) && previousCollectionKey === collectionKey) {
```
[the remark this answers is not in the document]
fabioh8010: These conditions serve to ensure we can only handle dynamic collection member keys from the same collection, for example:

- From `account` to `session`: ❌ 
- From `policy_` to `report_`: ❌ 
- From `report_` to `report_1283`: ❌ 
- From `report_385` to `report_9385`: ✅ 

Basically we have only one case where we would supply a different key to `useOnyx` over the time, when we are dealing with collection member keys:

```ts
const {value: policy} = useOnyx(`${ONYXKEYS.COLLECTION.POLICY}${policyID}`);
```

This way we ensure that `useOnyx` subscription is dealing with only one type of data and guarantee stable typings and safeness to the consumer.

fabioh8010:
I've addressed review comments from @tgolen and implemented `allowStaleData` logic cc @hannojg . Currently I'm investigating an issue I'm facing with the loading states, which are not being updated correctly when mounting a component that uses `useOnyx`.

.eslintrc.js
tgolen: OK, thanks! I guess it might make more sense if the rule was turned off by default and then only enabled in the App repo, since it's very specific to that repo. I'll have to keep this in mind for future custom ESLint rules.

tgolen (dismissed):

blazejkustra (approved):

lib/useOnyx.ts
```diff
@@ -0,0 +1,171 @@
...
+     * cause the component to re-render (and that can be expensive from a performance standpoint).
+     */
+    selector?: Selector<TKey, unknown, TReturnData>;
+};
+
+type FetchStatus = 'loading' | 'loaded';
+
+type CachedValue<TKey extends OnyxKey, TValue> = TValue extends OnyxValue<TKey> ? TValue : TKey extends CollectionKeyBase ? NonNullable<OnyxCollection<TValue>> : TValue;
+
+type UseOnyxData<TKey extends OnyxKey, TValue> = {
```
roryabraham, line 44: Hmmm, I thought we had aligned on this as a better pattern for `useOnyx` return data that's the best of both worlds:

- simple to get the "main" return value
- easy to extend with more configs / extra fields

```suggestion
type UseOnyxData<TKey extends OnyxKey, TValue> = [CachedValue<TKey, TValue>, {status: FetchStatus}];
```

that way we can do something like this:

```ts
const [account] = useOnyx(ONYXKEYS.ACCOUNT)l
const [reports, {status: reportsLoadingStatus}] = useOnyx(ONYXKEYS.COLLECTION.REPORT);
```

roryabraham:
Also, would love to see some automated UI tests added, similar to what we have for `withOnyx`

fabioh8010:
Hi @tgolen I'm currently fixing the loading state problems, after that will proceed with the tests.

fabioh8010:
@roryabraham @tgolen @blazejkustra I've pushed the changes to re-write implementation fix the loading states (and other issues I've found during tests). Key changes:
- Change `getSnapshot()` to return the entire hook data (`[CachedValue<TKey, TValue>, {status: FetchStatus}]`) in a stable way so it prevents issues with wrong loading state.
- Moved `initialValue` and `allowStaleData` logic to inside `getSnapshot()` function in order to consolidate all logic and fix several issues with stale data.

fabioh8010:
@roryabraham @tgolen @blazejkustra I've added some unit tests for `useOnyx`, I decided to use `renderHook` approach to test the hook and focus more on its result behavior during the tests, please let me know what do you think.

I didn't implement unit tests for `allowStaleData`, not sure yet how to implement them properly and I also noticed that we don't have it for `withOnyx` too.

lib/useOnyx.ts
```diff
@@ -0,0 +1,188 @@
+import {deepEqual} from 'fast-equals';
+import {useCallback, useEffect, useRef, useSyncExternalStore} from 'react';
+import Onyx from './Onyx';
+import type {CollectionKeyBase, KeyValueMapping, OnyxCollection, OnyxEntry, OnyxKey, Selector} from './types';
+import useLiveRef from './useLiveRef';
+import usePrevious from './usePrevious';
+
+type OnyxValue<TKey extends OnyxKey> = TKey extends CollectionKeyBase ? OnyxCollection<KeyValueMapping[TKey]> : OnyxEntry<KeyValueMapping[TKey]>;
```
blazejkustra, line 8: Let's move this type next to OnyxKey

lib/useOnyx.ts
```diff
@@ -0,0 +1,188 @@
...
+
+type FetchStatus = 'loading' | 'loaded';
+
+type CachedValue<TKey extends OnyxKey, TValue> = TValue extends OnyxValue<TKey> ? TValue : TKey extends CollectionKeyBase ? NonNullable<OnyxCollection<TValue>> : TValue;
+
+type UseOnyxData<TKey extends OnyxKey, TValue> = [
+    CachedValue<TKey, TValue>,
+    {
+        status: FetchStatus;
+    },
```
blazejkustra, line 48: How about extracting it to a separate type?

lib/useOnyx.ts
```diff
@@ -0,0 +1,188 @@
...
+            newData = options?.initialValue as CachedValue<TKey, TReturnData>;
+            newFetchStatus = 'loaded';
+        }
+
+        // If the previously cached data is different from the new data, we update both cached data
+        // and the result data to be returned by the hook.
+        // We can't directly compare the value from `dataRef` with `newData` because we default `undefined`
+        // to `null` when setting `dataRef` value to ensure we have a consistent return, but we need to be able to differentiate
+        // between `undefined` and `null` during the comparison, so `cachedValueRef` is used to store the real value without changing
+        // to `null`.
```
blazejkustra, line 129: This part seems quite complicated, had to read through this couple of times before I understood what is going on 😅 

One question, why can't we use null value for the loading status as well? This way we could simplify the logic a bit here

fabioh8010: Let's just simplify then, wdyt?

```ts
        // If the previously cached data is different from the new data, we update both cached data
        // and the result data to be returned by the hook.
        if (!deepEqual(cachedValueRef.current, newData)) {
            cachedValueRef.current = newData as CachedValue<TKey, TReturnData>;

            // If the new data is `undefined` we default it to `null` to ensure the consumer get a consistent result from the hook.
            dataRef.current = [(cachedValueRef.current ?? null) as CachedValue<TKey, TReturnData>, {status: newFetchStatus ?? 'loaded'}];
        }
```

> One question, why can't we use null value for the loading status as well? This way we could simplify the logic a bit here

I didn't understand quite well the question 😅 The comments on that condition was more directed to the use and difference between `cachedValueRef` and `dataRef`, the `newFetchStatus ?? 'loaded'` part I'm just defaulting to `loaded` in case it was not set before.

tgolen (dismissed):

lib/useOnyx.ts
```diff
@@ -0,0 +1,189 @@
...
+function useOnyx<TKey extends OnyxKey, TReturnValue = OnyxValue<TKey>>(key: TKey, options?: UseOnyxOptions<TKey, TReturnValue>): UseOnyxResult<TKey, TReturnValue> {
+    const connectionIDRef = useRef<number | null>(null);
+    const previousKey = usePrevious(key);
+
+    // Used to stabilize the selector reference and avoid unnecessary calls to `getSnapshot()`.
+    const selectorRef = useLiveRef(options?.selector);
+
+    // Stores the previous cached value as it's necessary to compare with the new value in `getSnapshot()`.
+    // We initialize it to `undefined` to simulate that we don't have any value from cache yet.
+    const cachedValueRef = useRef<CachedValue<TKey, TReturnValue> | undefined>(undefined);
```
roryabraham, line 61: *question:* since this is used to store the previous cached value, why don't we use `usePrevious` for this?

lib/useOnyx.ts
```diff
@@ -0,0 +1,189 @@
...
+    /**
+     * This will be used to subscribe to a subset of an Onyx key's data.
+     * Using this setting on `useOnyx` can have very positive performance benefits because the component will only re-render
+     * when the subset of data changes. Otherwise, any change of data on any property would normally
+     * cause the component to re-render (and that can be expensive from a performance standpoint).
+     */
+    selector?: Selector<TKey, unknown, TReturnValue>;
+};
+
+type FetchStatus = 'loading' | 'loaded';
```
roryabraham, line 38: let's use an object / const here and export it so we can check loading status with a const in E/App?

lib/useOnyx.ts
roryabraham: seems like you could do this:

```ts
const prevCachedValue = usePrevious(resultRef.current[0]);
```

and that might be a bit more concise

tests/unit/useOnyxTest.ts
```diff
@@ -0,0 +1,384 @@
...
+                fail("Expected to don't throw any errors.");
+            }
+        });
+    });
+
+    describe('misc', () => {
+        it('should return value and loaded state when loading cached key', async () => {
+            Onyx.set(ONYXKEYS.TEST_KEY, 'test');
+
+            const {result} = renderHook(() => useOnyx(ONYXKEYS.TEST_KEY));
```
roryabraham, line 94: `renderHook` seems like a great utility for this! nice TIL

roryabraham (changes requested):
tests look great! everything looks great! I seriously considered merging this but I think the last blocker for me is that we should use a constant for loading state 🙂

lib/useOnyx.ts
blazejkustra: Sorry! I meant why do we need to use undefined here:
```ts
        // If we have pending merge operations for the key during the first connection, we set the new value to `undefined`
        // and fetch status to `loading` to simulate that it is still being loaded until we have the most updated data.
        // If `allowStaleData` is `true` this logic will be ignored and cached value will be used, even if it's stale data.
        if (isFirstConnectionRef.current && Onyx.hasPendingMergeForKey(key) && !options?.allowStaleData) {
            newValue = undefined;
            newFetchStatus = 'loading';
        }
```

Most likely I'm missing something but if we were to use null in the logic above we could compare `dataRef` with `newData`, no?

blazejkustra (approved):
LGTM! 🚀 

I tested it in the App today, also migrated some withOnyx usage to check if it works as expected, looks very promising! Great work @fabioh8010

tgolen (dismissed):

tgolen:
Does there need to be anything generated for the API docs for this?

fabioh8010:
@tgolen I've ran the script to update the docs

lib/useOnyx.ts
roryabraham: Not sure I got all that but I'll take your word for it 😅

lib/useOnyx.ts
roryabraham: sounds good

roryabraham (approved):

roryabraham:
LGTM 🚀

tgolen (approved):

mountiny:
As noted by others, the code changes are looking great! Not leaving a review so I don't take credits for weeks worth of work, but wanted to note that you have done great job with this project, not only Fabio and Blazej, but also Rory and Tim (and others leaving a review on the doc) for pushing this forwards, thank you! 🏅

PR #474, opened 2024-02-21
Migrate remaining files to TS

blazejkustra:
### Details
This PR aims to migrate remaining files to TS (other than Onyx.js and withOnyx.js). Changes are almost only on the type level, so it shouldn't create any regressions.

### Related Issues

https://github.com/Expensify/App/issues/34342


BartoszGrajdek (approved):

lib/createDeferredTask.ts
```diff
@@ -1,13 +1,16 @@
+type DeferredTask = {
+    promise: Promise<unknown>;
+    resolve: (value: unknown) => void;
+};
```
fabioh8010, line 4:
```suggestion
type DeferredTask = {
    promise: Promise<unknown>;
    resolve?: (value: unknown) => void;
};
```

lib/SyncQueue.js
fabioh8010, line 1: Is this file not being used anymore?

lib/createDeferredTask.ts
fabioh8010: `resolve` is initially undefined until the promise is resolved.

lib/createDeferredTask.ts
```diff
@@ -1,13 +1,16 @@
...
  * The returned promise will complete when we call `resolve`
  * Useful when we want to wait for a tasks that is resolved from an external action
- *
- * @template T
- * @returns {{ resolve: function(*), promise: Promise<T|void> }}
  */
-export default function createDeferredTask() {
-    const deferred = {};
+export default function createDeferredTask(): DeferredTask {
+    const deferred: DeferredTask = {} as DeferredTask;
```
fabioh8010, line 12:
```suggestion
    const deferred = {} as DeferredTask;
```

wdyt?

lib/createDeferredTask.ts
fabioh8010: I updated my code suggestion, it should be `void`.

lib/SyncQueue.js
blazejkustra: Yes 👍

fabioh8010 (approved):

arosiclair (approved):

PR #475, opened 2024-02-27
refactor: unify storage/providers (for further InMemory storage integration) [1/3]

kirillzyusko:
### Details

This is the first PR that prepares an "infrastructure" for next PR. In this PR I did:
- all calls to providers are managed via additional `Storage` layer (in future this layer will intercept errors and will substitute provider to InMemory);
- synchronization between tabs is not a responsibility of provider anymore - now it's handled in separate `InstanceSync` class/object - such approach allows us to avoid code duplication if we add `InMemory` provider on web;
- all providers are initialized lazily (all providers got a new `init` method);

### Related Issues
 

### Automated Tests

### Manual Tests

### Author Checklist

    - [x] Android / native
    - [x] Android / Chrome
    - [x] iOS / native
    - [x] iOS / Safari
    - [x] MacOS / Chrome / Safari
    - [x] MacOS / Desktop

### Screenshots/Videos
<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>


kirillzyusko:

kirillzyusko:
recheck

danieldoglas:
Adding original reviewers from https://github.com/Expensify/react-native-onyx/pull/439

lib/storage/index.ts
```diff
@@ -1,3 +1,162 @@
...
+    /**
+     * Stores multiple key-value pairs in a batch
+     * @param {Array<[key, value]>} pairs
+     * @return {Promise<void>}
+     */
+    multiSet: (pairs) => provider.multiSet(pairs),
+
+    /**
+     * Merging an existing value with a new one
+     * @param {String} key
```
tgolen, line 68: For a lot of these methods docs, I think it should be using defined types in the param definition. Is that right? Like `mergeItem:(string key)`

lib/storage/platforms/NativeStorage.ts
```diff
@@ -0,0 +1,3 @@
+import SQLiteStorage from '../providers/SQLiteProvider';
```
tgolen, line 1: This file has some odd names. The file is called `NativeStorage`, but it exports `SQLiteStorage` while importing `SQLiteProvider`. So... which one is it?

Our naming conventions state that a file should be named for the same thing that it exports so I think that would result in something more like this:

```
import {default as NativeStorage} from '../providers/SQLiteProvider';

export NativeStorage;
```
```

File: lib/storage/platforms/WebStorage.ts [hunk 3]
```diff
@@ -0,0 +1,3 @@
+import IDBKeyValProvider from '../providers/IDBKeyValProvider';
+
+export default IDBKeyValProvider;
```
tgolen, line 3: Same here, let's `export WebStorage;`

lib/storage/platforms/index.ts
```diff
@@ -0,0 +1,3 @@
+import WebStorage from './WebStorage';
```
tgolen, line 1: Can you remind me what the purpose is of this file? I'm wondering why it can't import `IDBKeyValProvider` directly.

tgolen (changes requested):

lib/storage/platforms/index.ts
kirillzyusko: @tgolen you are right, seems like there is no sense to re-export these files - I removed these index files 👍

kirillzyusko:
@tgolen I addressed your notes - would you mind to check this PR again? 👀

lib/storage/platforms/index.ts
tgolen: Ah, sorry, I actually like the index files (because that is a better system for determining which platform uses which provider). Could you remove the other storage files instead?

kirillzyusko: @tgolen sorry, I didn't describe correctly what I did 🤦 

I left `index` and `index.native` files and removed `WebStorage`/`NativeStorage` - this is a commit for a reference https://github.com/Expensify/react-native-onyx/pull/475/commits/53bd1e7173696e3f5a5fb6358923a089194468bf

lib/storage/InstanceSync/index.ts
```diff
@@ -0,0 +1,16 @@
+import NOOP from 'lodash/noop';
+
+/**
+ *  This is used to keep multiple browser tabs in sync, therefore only needed on web
+ *  On native platforms, we omit this syncing logic by setting this to mock implementation.
+ */
+const InstanceSync = {
+    init: NOOP,
```
tgolen, line 8: Does this need to use lodash, or can it just be an empty arrow function like `init: () => {}`?

tgolen (changes requested):
Gotcha, I see what you did now! I like it.

lib/Onyx.js
```diff
@@ -1563,6 +1563,16 @@ function init({
     shouldSyncMultipleInstances = Boolean(global.localStorage),
     debugSetState = false,
 } = {}) {
+    Storage.init();
+
+    if (shouldSyncMultipleInstances) {
```
tgolen, line 1568: Was there any specific reason to move this up here? I think it makes sense but I'm also surprised that it wasn't here to begin with.

marcaaron:
Got some competing priorities so will have to pull myself off the review to avoid becoming a blocker. If there is anything that needs my attention I am happy to take a look!

lib/storage/InstanceSync/index.ts
kirillzyusko: @tgolen if I add empty functions I'm getting eslint errors:

image

Do you think it would be better to specify our own function?

tgolen: Hm, that's interesting. OK, it's probably fine to leave it as `NOOP` for now. Thanks for trying!

tgolen (dismissed):

kirillzyusko:
@tgolen I fixed eslint errors in https://github.com/Expensify/react-native-onyx/pull/475/commits/1932e25263ab6222736018d604bb0d1aa12afb31 - can you please check if it's okay? 👀

tgolen:
Looks like the lint errors are fixed. Sorry, I didn't catch this one earlier either:

image

Looks like you need to setup commit signing. It can be a pain to go back and sign previous commits... so what you _could_ do is close this PR and create a brand new one after you've setup commit signing. Up to you!

kirillzyusko:
> Looks like you need to setup commit signing. 

My bad 🤦‍♂️ 

I signed them now and pushed to this PR (hopefully now everything is okay and it can be merged 😊)

tgolen (approved):

PR #483, opened 2024-02-28
feat: fallback to NoopProvider if we run into OOM [2/3]

kirillzyusko:
### Details

This is a second PR in chain of multiple PRs. In this PR I added:
- `name` property to all providers (used in logger if we ran into OOM);
- added `NoopProvider` that is used as a fallback one when OOM happens;
- added `tryOrDegradePerformance` function that attempt to perform operation and in case of specific error substitutes the storage to a fallback one;

### Related Issues

### Automated Tests

### Manual Tests

### Author Checklist

    - [x] Android / native
    - [x] Android / Chrome
    - [x] iOS / native
    - [x] iOS / Safari
    - [x] MacOS / Chrome / Safari
    - [x] MacOS / Desktop

### Screenshots/Videos
<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>


kirillzyusko:
@danieldoglas @tgolen tagging you again - this PR is ready for review 😊

lib/storage/index.ts
```diff
@@ -22,112 +61,121 @@ const Storage: Storage = {
      * and enables fallback providers if necessary
      */
     init() {
-        provider.init();
+        tryOrDegradePerformance(() => {
+            provider.init();
+        }).finally(() => {
```
tgolen, line 66:
```suggestion
        tryOrDegradePerformance(provider.init).finally(() => {
```
Is it OK to simplify this?

lib/storage/index.ts
```diff
@@ -1,13 +1,52 @@
...
+            } catch (error) {
+                // Test for known critical errors that the storage provider throws, e.g. when storage is full
+                if (error instanceof Error) {
+                    // IndexedDB error when storage is full (https://github.com/Expensify/App/issues/29403)
+                    if (error.message.includes('Internal error opening backing store for indexedDB.open')) {
+                        degradePerformance(error);
+                    }
+                }
+
+                reject(error);
```
tgolen, line 45: It doesn't every look like anything is using this rejection. If so, maybe it should be removed?

tgolen (changes requested):
Thanks for the ping! I was waiting for the rebase.

lib/storage/index.ts
kirillzyusko: @tgolen I think no, it's better to have it here

The function `tryOrDegradePerformance` just acts as an interceptor - it simply executes the function and in case of a specific exception just substitues a provider. And when we develop such generic functions it's important to propagate exceptions further.

And this function keeps full backward compatibility - if `getItem` or any other function would throw exception before, then it would be thrown further (and potentially we may have a code that handles these errors). So I'd like to keep this line and propagate exceptions and don't silently swallow them.

kirillzyusko:
@tgolen may I ask you to review it again? 👀

lib/storage/index.ts
```diff
@@ -1,13 +1,52 @@
...
 let shouldKeepInstancesSync = false;
+let finishInitalization: (value?: unknown) => void;
+const initPromise = new Promise((resolve) => {
+    finishInitalization = resolve;
+});
 
 type Storage = {
     getStorageProvider: () => StorageProvider;
-} & StorageProvider;
+} & Omit<StorageProvider, 'name'>;
```
danieldoglas, line 17: Why are we omititng the name here?

lib/storage/index.ts
```diff
@@ -1,13 +1,52 @@
...
+
+/**
+ * Degrade performance by removing the storage provider and only using cache
+ */
+function degradePerformance(error: Error) {
+    Logger.logAlert(`Error while using ${provider.name}. Falling back to only using cache and dropping storage.`);
```
danieldoglas, line 23: Just to confirm, starting from here, the app won't actually update anything based on what the user received, and the data will always be stale, right?

lib/storage/index.ts
kirillzyusko: @danieldoglas `StorageProvider` contains `name` property because this property is used in logger (to know which provider threw the error).

And `Storage` acts as a gate, i. e. underhood this class/object can substitute internal providers, but the gateway will be always the same. So I thought it senseless to have this property on a gateway level 🤷‍♂️

lib/storage/index.ts
kirillzyusko: > Just to confirm, starting from here, the app won't actually update anything based on what the user received, and the data will always be stale, right?

@danieldoglas Yes, in case of error (out-of-memory) we will not store anything and the data will be always stale

danieldoglas (dismissed):
LGTM

tgolen (approved):

blazejkustra:
Can't we just remove `setMemoryOnlyKeys` from react-native-onyx? I think it serves no purpose in the app:
image

☝️ All these are just empty functions.

Only this part is relevant:
```ts
/**
 * When set these keys will not be persisted to storage
 * @param {string[]} keyList
 */
function setMemoryOnlyKeys(keyList) {
    // When in memory only mode for certain keys we do not want to ever drop items from the cache as the user will have no way to recover them again via storage.
    cache.setRecentKeysLimit(Infinity);
}
```

cc @tgolen @kirillzyusko

kirillzyusko:
Good point @blazejkustra 

@tgolen will be happy to hear your thoughts 😊

tgolen:
I think it's OK to remove at this point. It has a little bit of complex history. We used to use it, but it was never fully implemented either. https://github.com/Expensify/App/issues/26304

Let's just kill it 🔪

blazejkustra:
Do I have a green light to remove `Onyx.setMemoryOnlyKeys` method as well?

tgolen:
Yep!

tgolen:
Actually, no... sorry. I should have checked App first. It's still being referenced in App. So we need to remove those first.

PR #485, opened 2024-02-29
feat: fallback to `NoopProvider` if OOM happens

kirillzyusko:
### Details

This PR accumulates al changes from https://github.com/Expensify/react-native-onyx/pull/483 and https://github.com/Expensify/react-native-onyx/pull/475 and adds a usage of `MemoryOnlyProvider` in unit-tests.

Originally https://github.com/Expensify/react-native-onyx/pull/483 and https://github.com/Expensify/react-native-onyx/pull/475 were reverted - in this PR I added fixes:
- https://github.com/Expensify/react-native-onyx/pull/485/commits/c461b993af46703fa03006a3bef1750a5db592e6 <- fixes a crash when `this` is `undefined` (which prevented a synchronization between tabs).
- https://github.com/Expensify/react-native-onyx/pull/485/commits/42c42eac988f17760ada9518534f3f390965ed92 <- fixes infinite initialization

In this PR I added [1/3]:
- all calls to providers are managed via additional `Storage` layer (in future this layer will intercept errors and will substitute provider to InMemory);
- synchronization between tabs is not a responsibility of provider anymore - now it's handled in separate `InstanceSync` class/object - such approach allows us to avoid code duplication if we add `InMemory` provider on web;
- all providers are initialized lazily (all providers got a new `init` method);

In this PR I added [2/3]:
- `name` property to all providers (used in logger if we ran into OOM);
- added `NoopProvider` that is used as a fallback one when OOM happens;
- added `tryOrDegradePerformance` function that attempt to perform operation and in case of specific error substitutes the storage to a fallback one;

In this PR I added [3/3]:
- `MemoryOnlyProvider` implementation that stores all info in memory;
- re-worked mocks and used `MemoryOnlyProvider` there;

### Related Issues

### Automated Tests

### Manual Tests

### Author Checklist

    - [x] Android / native
    - [x] Android / Chrome
    - [x] iOS / native
    - [x] iOS / Safari
    - [x] MacOS / Chrome / Safari
    - [x] MacOS / Desktop

### Screenshots/Videos
<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>


hayata-suenaga:
@marcaaron @tgolen I assigned you as you were reviewer of the closed PR and because you're familiar with the Onyx library

kirillzyusko:
It's not needed to review now, because it contains changes from part 2. Let's review & merge part 2 and after that I'll rebase this Pr and we can review it 👀

kirillzyusko:
@danieldoglas @tgolen this PR is ready for review 😊

lib/storage/providers/MemoryOnlyProvider.ts
```diff
@@ -0,0 +1,144 @@
...
+     */
+    getAllKeys() {
+        return Promise.resolve(_.keys(store));
+    },
+
+    /**
+     * Gets the total bytes of the store.
+     * `bytesRemaining` will always be `Number.POSITIVE_INFINITY` since we don't have a hard limit on memory.
+     */
+    getDatabaseSize() {
```
hayata-suenaga, line 132: why do we need to measure the database size?

lib/storage/providers/MemoryOnlyProvider.ts
```diff
@@ -0,0 +1,144 @@
...
+
+        return Promise.resolve([]);
+    },
+
+    /**
+     * Remove given key and it's value from memory
+     */
+    removeItem(key) {
+        delete store[key];
+        return Promise.resolve();
```
tgolen, line 95: Should this resolve with `null` like `getItem()`?

lib/storage/providers/MemoryOnlyProvider.ts
```diff
@@ -0,0 +1,144 @@
...
+
+    /**
+     * Clear everything from memory
+     */
+    clear() {
+        store = {};
+        return Promise.resolve();
+    },
+
+    // This is a noop for now in order to keep clients from crashing see https://github.com/Expensify/Expensify/issues/312438
```
tgolen, line 116:
```suggestion
    /*
     * Since this is an in-memory only provider, nothing special needs to happen here and it can just be a noop
     */
```

tgolen (changes requested):

lib/storage/providers/MemoryOnlyProvider.ts
marcaaron: We added this for logging purposes. It only gets used when we encounter an error so we probably don't need this. In theory, the `MemoryOnlyProvider` should never fail or throw any errors (at least doesn't look like it can).

marcaaron: See: https://github.com/Expensify/react-native-onyx/blob/69b2d5973b4aef19ce783440df55dedb30d4efd2/lib/Onyx.js#L1037-L1045

And: https://github.com/Expensify/react-native-onyx/blob/69b2d5973b4aef19ce783440df55dedb30d4efd2/lib/Onyx.js#L1069-L1078

marcaaron (approved):
Changes LGTM - I agree with @hayata-suenaga that we do not need to measure the size of the store (or at least can't think of any useful reason to do it).

lib/storage/providers/MemoryOnlyProvider.ts
kirillzyusko: Corrected 🙌

kirillzyusko:
> I agree with @hayata-suenaga that we do not need to measure the size of the store (or at least can't think of any useful reason to do it).

@marcaaron so do you want to return a mocked value from `getDatabaseSize` without `object-sizeof` usage?

lib/storage/providers/MemoryOnlyProvider.ts
tgolen: OK, great! That sounds fine as long as that's what's happening in the other providers. Also, I just checked and I can't find any uses of `.remove()` in Expensify/App and it looks like this method is only used internally.

tgolen (approved):

lib/storage/providers/MemoryOnlyProvider.ts
hayata-suenaga: ah I can see from the code that it's only used in a very rare case. Thank you so much for the explanation!

hayata-suenaga (dismissed):
the code looks good to me

there is [a pending question](https://github.com/Expensify/react-native-onyx/pull/485#issuecomment-1978669078)

danieldoglas (approved):

marcaaron:
>@marcaaron so do you want to return a mocked value from getDatabaseSize without object-sizeof usage?

yeah, it does not need to be accurate so I'd rather not add a dependency if we don't need one.

kirillzyusko:
> yeah, it does not need to be accurate so I'd rather not add a dependency if we don't need one.

@marcaaron okay, make sense - I'll remove it 🙌

marcaaron (approved):

kirillzyusko:
@tgolen @hayata-suenaga I think it's ready for second review. I wasn't able to reproduce https://github.com/Expensify/App/issues/37752 since I think it's only possible in staging env?

I checked a code that responsible for synchronization - found and fixed there an issue. Checked how synchronization code works and it seems like it works okay.

If you have any suggestions which other local tests I need to perform in order not to introduce regressions - please tell me, I'll be happy to perform these tests 😊

tgolen (approved):
Great, thanks for squashing this all together. To help protect against regressions, I suggest the following:

- Create a PR on E/App that points to this branch (ie. the last commit like I did [here](https://github.com/Expensify/App/pull/37804/files))
- Request from our @appdeployers team on Slack to create an Ad-Hoc build for the E/App PR
- Then we can request for QA to do a full regression test on that ad-hoc build

That will allow us to find any possible regressions before committing to merge this PR.

hayata-suenaga (dismissed):
the test plan sounds good 👍

lib/storage/index.ts
```diff
@@ -1,3 +1,185 @@
...
+        promise.then(() => {
+            try {
+                resolve(fn());
+            } catch (error) {
+                // Test for known critical errors that the storage provider throws, e.g. when storage is full
+                if (error instanceof Error) {
+                    // IndexedDB error when storage is full (https://github.com/Expensify/App/issues/29403)
+                    if (error.message.includes('Internal error opening backing store for indexedDB.open')) {
+                        degradePerformance(error);
+                    }
```
marcaaron, line 44: Should we also catch the error we are explicitly throwing here?

https://github.com/Expensify/react-native-onyx/blob/3df4fef6d78fbb8a986304e071403f7a9d8418c4/lib/storage/providers/IDBKeyValProvider.ts#L22

kirillzyusko: @marcaaron good catch! Yes, since we can not open a connection to DB, so all calls to set/get/remove will throw exceptions.

So for me it makes sense to substitute a storage in order not to crash the app 👍

lib/storage/index.ts
```diff
@@ -1,3 +1,190 @@
...
+    getStorageProvider: () => StorageProvider;
+} & Omit<StorageProvider, 'name'>;
+
+/**
+ * Degrade performance by removing the storage provider and only using cache
+ */
+function degradePerformance(error: Error) {
+    Logger.logAlert(`Error while using ${provider.name}. Falling back to only using cache and dropping storage.`);
+    console.error(error);
+    provider = NoopProvider;
```
kirillzyusko, line 25: @tgolen @marcaaron also a question about this - do we want to fallback to `NoopProvider` (does nothing) or to `MemoryProvider` stores all data in RAM?

tgolen: I think the memory provider would be better. That's what the log message assumes.

tgolen:
I got word from QA this morning that they have begun testing the E/App PR. They will report any bugs they find on that PR and we can triage them from there.

kirillzyusko:
@mountiny @tgolen just a friendly ping to check the status of this PR 👀 

I think QA have tested and didn't find anything new (only old issues that can be reproduced in any preview deploys - CORS issues).

Just curious whether it can this PR be merged or not?

tgolen (approved):

tgolen:
OK, I took one last look at the code and it looks good to me. I'm glad QA hasn't found any more issues with this and sounds like it's ready to be released. 

@marcaaron @danieldoglas @hayata-suenaga I wouldn't mind having at least one more approval before merging this. Would one of you like the honors?

hayata-suenaga:
thank you for the review, Tim! Unfortunately, I don't have time to go over the code again, so I'd wait for Daniel or Marc 😄

marcaaron:
Sure, taking a look!

lib/storage/index.ts
```diff
@@ -1,3 +1,187 @@
...
+
+type Storage = {
+    getStorageProvider: () => StorageProvider;
+} & Omit<StorageProvider, 'name'>;
+
+/**
+ * Degrade performance by removing the storage provider and only using cache
+ */
+function degradePerformance(error: Error) {
+    Logger.logAlert(`Error while using ${provider.name}. Falling back to only using cache and dropping storage.`);
```
marcaaron, line 23: NAB, it would be useful to log the error message as well. Without knowing anything about _why_ we fell back to using the cache this alert is not really actionable.

marcaaron (approved):
I have a small note. I am not sure if it's worth blocking on and this PR has been open for quite a while so suggesting a follow up to address it instead.

lib/storage/index.ts
tgolen: I always got tripped up on this too, but it is logged on the very next line 😁

tgolen: Oh, but maybe you mean to send the error as part of `logAlert()`?

marcaaron: Yep! That's what I meant. I do see the `console.error()` which is local (good) but might be desirable to have server logs. 

I backpedaled here because I'm not sure if we are going to be tracking the frequency of these errors or doing anything about them.

That said, I also wonder if logging alert will lead to issues that nobody can action... I am not sure if the JS logs create issues or not. Maybe they should and if so - this could probably be a `hmmm()`.

kirillzyusko: @tgolen @marcaaron should I do these changes in follow up PR? 👀

tgolen: Yeah, I think it would probably be good to change this to a `hmmm()` and also log the specific error. It would sure be nice to look at server logs to determine _IF_ the fallback is happening and also _WHY_.

kirillzyusko: @tgolen make sense! I've created a follow up PR: https://github.com/Expensify/react-native-onyx/pull/524

PR #490, opened 2024-03-03
Try to synchronize operations within one update

paultsimura:
- Held on https://github.com/Expensify/App/pull/42772 which updates the version of Onyx and includes this fix.

### Details

This PR changes how we process the `Onyx.update` operation:
1. Enqueue all the operations per individual key;
2. Group the collection-related keys into a single `mergeCollection` call;
3. Process the rest of the individual keys with `merge` calls;

### Related Issues

https://github.com/Expensify/App/issues/37560

### Automated Tests

I've added tests that cover the `Onyx.update()` with multiple merge operations of the same keys to verify that all the changes are batched into a single `mergeCollection` operation.

1. Perform a complex `Onyx.update()` operation with multiple updates of the same item in `merge` and `mergeCollection` operations.
2. Verify the collection items are updated with all the changes combined.

_Operation example ref:_ https://github.com/Expensify/react-native-onyx/blob/eb17a5dd0eba6b7b27f1e92e0fc81863091ba6c8/tests/unit/onyxTest.js#L1046-L1103

### Author Checklist

    - [ ] Android / native
    - [ ] Android / Chrome
    - [ ] iOS / native
    - [ ] iOS / Safari
    - [ ] MacOS / Chrome / Safari
    - [ ] MacOS / Desktop

### Screenshots/Videos
<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>


paultsimura:
@tgolen I think I found a way to fix the issue of the out-of-order updates.

The main idea is the following: when one `update` call comes with multiple merge/set/mergeCollection operations, – do not process anything asynchronously. Instead, build stack all the operations by keys, and then persist everything in one `multiGet` + `multiSet` take.

This PR is super draft, it lacks the validations as well as the broadcasting of the updates, but I would appreciate your opinion on the idea as a whole.

lib/Onyx.js
```diff
@@ -1571,7 +1587,25 @@ function update(data) {
...
 
-    return clearPromise.then(() => Promise.all(_.map(promises, (p) => p())));
+    return clearPromise.then(() => Storage.multiGet(_.keys(updateQueue)).then((pairs) => {
+        const existingSnapshot = _.reduce(pairs, (acc, [key, value]) => {
+            acc[key] = value;
+            return acc;
+        }, {});
+
+        const updatedSnapshot = {};
+        _.each(updateQueue, (operations, key) => {
```
tgolen, line 1597: Could this be a `_.reduce` instead?

lib/Onyx.js
```diff
@@ -1571,7 +1587,25 @@ function update(data) {
...
+            let existingValue = existingSnapshot[key];
+            if (operations[0].onyxMethod === METHOD.SET) {
+                existingValue = operations[0].value;
+            }
+
+            const mergeOperations = _.filter(operations, ({onyxMethod}) => onyxMethod === METHOD.MERGE);
```
tgolen, line 1603: Since a single operation can't be both a `SET` and a `MERGE` there might be a little cleaning up to this logic which could happen (maybe an early return in the `SET` case?)

tgolen (commented):
I love the direction this is going in and I think your approach is a good one. I hope it holds up as you keep working on it!

lib/Onyx.js
paultsimura: The logic behind this is the following: I'm putting the `merge` and `set` operations in an array of a certain object key in the order they appear in the update operation.
However, since the `set` operation fully overwrites the existing object, it makes sense to keep only the operations that come after the last `set` (if it is present).

Therefore, eventually, a single key's update queue will consist of an array of `merge` operations with (optionally) only the first one in the array being `set`: `[set?, merge, merge, ...]`.

Does it make sense?

lib/Onyx.js
paultsimura: Yes, will update 👍

paultsimura:
In general, it does work – I gave it a little test run.
However, there are a lot of different callbacks in the current implementation and I'd appreciate your guidance on a more general level @tgolen.

Calls like these:

https://github.com/Expensify/react-native-onyx/blob/7cbf836cc1c107e82da42f372b82c91e5ecd1640/lib/Onyx.js#L1511-L1516

https://github.com/Expensify/react-native-onyx/blob/7cbf836cc1c107e82da42f372b82c91e5ecd1640/lib/Onyx.js#L1153-L1154

And all the places where we call the `sendActionToDevTools`. With the new approach, if we treat `update` as a single transaction with the eventual `merge`, it will be much harder to differentiate `mergeCollection` from `merge`.

What broadcasting approach would you suggest using here? If I broadcast each key update separately – would the collection listener handle each update individually?
Should I instead broadcast all collection-related keys as a single collection update, and non-collection keys as individual updates?

lib/Onyx.js
tgolen: Ah, OK. That makes sense. Adding some code comments could help with that.

Does this also work in a case like this where a single key might have multiple operations that are separated by other keys?

1. key1 - merge
2. key2 - merge
3. key1 - merge

That should result in a single merge instruction for `key1`.

paultsimura: > Does this also work in a case like this where a single key might have multiple operations that are separated by other keys?
> 
> 1. key1 - merge
> 2. key2 - merge
> 3. key1 - merge
> 
> That should result in a single merge instruction for `key1`.

Yes, I'll perform some tests and attach different scenarios in a while

tgolen:
> If I broadcast each key update separately – would the collection listener handle each update individually?

I believe this would work yes, and the collection listener will handle each individual update. This would be the safest approach against causing regressions.

> Should I instead broadcast all collection-related keys as a single collection update, and non-collection keys as individual updates?

The collection updates get a little tricky because some collection mappings will be expecting the callback to be triggered with the entire collection, while others might be expecting the callback to be triggered with each object in the collection. This is dependent on the `waitForCollectionCallback` mapping option.

I think I prefer the first method because that will make the least amount of behavior changes for Onyx.

paultsimura:
This looks like a promising picture to me when coming online, 46 operations were batched into one `update`:

image

Some of the keys had up to 7 `merge` operations stacked:

image

But eventually, the broadcast was called for each key only once, containing the final updated version of the object:
image

paultsimura:
**_I have read the CLA Document and I hereby sign the CLA_**

paultsimura:
@tgolen please make a sanity check when you have a minute, along with [this comment](https://github.com/Expensify/react-native-onyx/pull/490#issuecomment-1977262222).

The only thing that bothers me is that one failing test. I don't fully understand the nature of those subscriptions, and based on the comment in code, that first call is expected to be with `null, undefined` params 🤔 
https://github.com/Expensify/react-native-onyx/blob/7cbf836cc1c107e82da42f372b82c91e5ecd1640/lib/Onyx.js#L915-L918

tgolen:
Wow, that is a great improvement!

I don't think the test is too much to worry about. You can update the tests to change the `nth` to be `0` instead of `1` now. It makes sense in the context of the change you're making. 

Before your PR:
- [This](https://github.com/Expensify/react-native-onyx/blob/7cbf836cc1c107e82da42f372b82c91e5ecd1640/tests/unit/onyxTest.js#L919-L923) `Onyx.update()` would have triggered each subscriber twice (first with `null`, second with the actual value)

After your PR:
- The updates are squished into a single callback with the final value 🎉 

I think the important part of `null` that needs to be kept in mind (and you might want to add this to the unit tests with these changes) is that if I subscribe to a key that doesn't have any data stored on it, then the subscriber should be triggered once with `null` for a value. This allows the component to render and then the `null` value is substituted for `undefined` in `withOnyx` [here](https://github.com/Expensify/react-native-onyx/blob/7cbf836cc1c107e82da42f372b82c91e5ecd1640/lib/withOnyx.js#L322). That way, when the component gets the `undefined` prop, the `defaultProp` will be picked up and used.

Does that all make sense?

paultsimura:
> Does that all make sense?

Yes, updated the test as you recommended – to verify that the first call for the non-existing keys is made with `undefined`.

paultsimura:
@tgolen how terrible is the idea to make the following change?

```js
case METHOD.MERGE_COLLECTION: 
    _.each(value, (_value, _key) => promises.push(() => merge(_key, _value)));
    break; 
```
https://github.com/Expensify/react-native-onyx/blob/69b2d5973b4aef19ce783440df55dedb30d4efd2/lib/Onyx.js#L1560-L1562

From what I understand, it will trigger the re-render of components listening to a collection N times?

tgolen:
We use `mergecollection` quite a bit so I would definitely be worried about performance issues with that.

paultsimura:
@tgolen I _think_ I've made a working solution: it updates the data in the correct order, **but** it's a bit slower than the current version.

Would it be possible to request a look from the side at the potential bottlenecks and improvements, especially the cache-related operations?
This PR is risky because it changes one of the mostly used Onyx functions, so an extra pair of eyes would really be helpful here.

tgolen:
OK, great. Let me ping some of our experts on Slack to see if they can have a look at this PR. You can also see my comment [here](https://github.com/Expensify/react-native-onyx/pull/485#pullrequestreview-1922833969) about how it can be reliably tested to help protect against major regressions.

lib/Onyx.js
```diff
@@ -1571,7 +1593,99 @@ function update(data) {
...
+            console.log('Onyx: updatedSnapshot', updatedSnapshot);
+
+            const updatesToBroadcast = [];
+            onyxCollectionKeySet.forEach((collectionKey) => {
+                const updatedCollection = _.reduce(
+                    _.keys(updatedSnapshot),
+                    (acc, key) => {
+                        if (isKeyMatch(collectionKey, key)) {
+                            acc[key] = updatedSnapshot[key];
+                            delete updatedSnapshot[key];
```
kidroca, line 1646: A comment explaining why we delete here would be handy. It seems it's to prevent duplicate notification of subscribers, because `keysChanged` probably triggers individual key subscribers

kidroca (commented):
This Pull Request represents a positive step forward in enhancing performance by batching updates into a single persist-to-storage operation.

**Considerations for follow-up work:**
- **Refactoring for Code Deduplication:** There's an opportunity to further refine our codebase by reducing duplicate logic. Specifically, methods like `set` could be simplified to merely forward calls to `Onyx.update` with the appropriate parameters. Similarly, `Onyx.merge` might be refactored to leverage `Onyx.update` for batched updates. These changes would streamline our code and potentially simplify future maintenance.

- **Optimizing Merge Logic:** The current implementation's use of `multiGet` to read current values seems to diverge from our optimized merge strategy that emphasizes delta merges. Previously, we focused on reading full values only as needed to facilitate updates via `keyChanged`, while ensuring that only the delta of the merge was persisted to storage. The observed logic in this PR, which appears to send back the full JSON rather than just the delta, might warrant a reevaluation to ensure it aligns with our efficiency goals.


The overall changes introduced by this PR are commendable and contribute to ongoing efforts to improve performance.

lib/Onyx.js
paultsimura: Yes, exactly. So the main idea is that we collect all the keys that belong to collections (even if they were updated individually), and process them as one collection while processing the non-collection keys as individuals.
The `keysChanged` notifies the subscribers subscribed to the whole collection as well as those subscribed to individual keys. So we remove the individual collection keys to avoid double notifications.

lib/Onyx.js
```diff
@@ -1571,7 +1593,99 @@ function update(data) {
         }
     });
 
-    return clearPromise.then(() => Promise.all(_.map(promises, (p) => p())));
+    return clearPromise.then(() =>
+        multiGet(_.keys(updateQueue)).then((pairs) => {
```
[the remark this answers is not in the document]
paultsimura: After a closer look, don't we eventually read all the values from cache/storage anyway before updating the cached values?

https://github.com/Expensify/react-native-onyx/blob/3f77c6bd2030979da3d581ea01fc9fd5e6d646e5/lib/Onyx.js#L1525-L1530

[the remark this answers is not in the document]
paultsimura: Since we need to wait for all the Storage write operations to complete, and then perform the `cache.get()` for each key and merge in-memory anyway to update the cache, is there any benefit from the `JSON_PATCH` in the current implementation?

I see there already was a similar discussion [here](https://github.com/Expensify/react-native-onyx/pull/238#discussion_r1123966066), looks like it was not completely resolved 🤔

paultsimura:
@kidroca could you please help me understand whether the `mergeQueue` was introduced to chain the merge operations within one `update` call, or if there's more to it? How broad is the "single tick" definition?

https://github.com/Expensify/react-native-onyx/blob/3f77c6bd2030979da3d581ea01fc9fd5e6d646e5/lib/Onyx.js#L23-L25

https://github.com/Expensify/react-native-onyx/blob/3f77c6bd2030979da3d581ea01fc9fd5e6d646e5/lib/Onyx.js#L1285-L1291

paultsimura:
I just discovered a behavior that looks like a blocker to me. Explained in detail here: https://github.com/Expensify/react-native-onyx/issues/301#issuecomment-1987950312

lib/Onyx.js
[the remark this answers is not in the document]
kidroca: Yes, we do `get` the latest value and perform a full merge with it to update cache, but we only try to send the delta for the write to storage.
The only benefit seems is we're sparing sending the full data for the write operation (and in fact this is only on native on web we send the full data for the merge)

kidroca:
> @kidroca could you please help me understand whether the `mergeQueue` was introduced to chain the merge operations within one `update` call, or if there's more to it? How broad is the "single tick" definition?

The merge queue was created way before the `update` interface existed, AFAIK it serves use cases where `Onyx.merge` is called multiple times from actions in the same tick, to be fair I don't know exactly when it succeeds in batching more than one item to merge, I guess it would be during init or updates

lib/Onyx.js
```diff
@@ -1606,6 +1625,58 @@ function update(data) {
...
+    const promises = [];
+
+    // Group all the collection-related keys and update each collection in a single `mergeCollection` call
+    onyxCollectionKeySet.forEach((collectionKey) => {
+        const collectionItemKeys = _.filter(_.keys(updateQueue), (key) => isKeyMatch(collectionKey, key));
+        if (_.size(collectionItemKeys) <= 1) {
+            // If there are no items of this collection in the updateQueue, we should skip it.
+            // If there is only one item, we should update it individually, therefore retain it in the updateQueue.
+            return;
+        }
```
kidroca, line 1637: I'm interested in the decision-making process for using `mergeCollection` versus individual merges based on the count of items in the update.

A few questions arise from this approach:

1. **Rationale Behind Individual Merges for Single Items**: Could you provide more details on the decision to favor individual updates over `mergeCollection` when there's only one item in the update? Understanding the rationale could clarify whether it's a performance consideration, a matter of update specificity, or another reason.

2. **Notification to Collection Subscribers**: How does this approach impact the notification of collection subscribers, especially in scenarios where a key is added or removed? The decision to handle single and multiple item updates differently could affect how subscribers are notified, potentially leading to inconsistencies.

3. **Consistency in Handling Updates**: If bypassing `mergeCollection` for single-item updates achieves the intended functionality and subscriber notification, could a similar logic not be applied to multiple item updates? Understanding the distinction between handling one versus multiple items could help in assessing potential improvements or optimizations in the update mechanism.

Your insights on these points would enhance my understanding and ensure that the logic aligns with the intended behavior for collection updates.

lib/Onyx.js
```diff
@@ -1606,6 +1625,58 @@ function update(data) {
...
+
+    _.each(updateQueue, (operations, key) => {
+        const batchedChanges = applyMerge(undefined, operations, false);
+
+        const shouldOverwriteExistingValue = _.includes(operations, null);
+        if (shouldOverwriteExistingValue) {
+            promises.push(() => set(key, batchedChanges));
+        } else {
+            promises.push(() => merge(key, batchedChanges));
+        }
```
kidroca, line 1677: It seems possible for `null` values to appear in positions other than index 0, which implies that `merge` operations are allowed to pass `null` as a merge value.

Instead of attempting to identify a `null` value here, wouldn't it be more effective to address it during the `enqueueMergeOperation`? If the value provided for a key is `null`, we could directly modify the update queue for this key to `updateQueue[key] = [null];`, effectively treating the update for the key as a `set` operation since the initial value would be overridden by `null`.

This approach would enable us to eliminate irrelevant merges that precede the `null` value, and also simplify the check here having to iterate all operations looking for `null`

tests/unit/onyxTest.js
```diff
@@ -1023,25 +1018,232 @@ describe('Onyx', () => {
...
                 expect(testKeyValue).toEqual(null);
+                return Onyx.merge(ONYX_KEYS.TEST_KEY, 2);
+            })
+            .then(() => {
+                expect(testKeyValue).toEqual(2);
+            });
+    });
+
+    describe('update', () => {
+        it('should squash all updates of collection-related keys into a single mergeCollection call', () => {
```
kidroca, line 1037: Great work on adding more, refining existing tests and relocating them here

lib/Onyx.js
kidroca: I speculate that the decision to forego `multiSet` and `multiMerge` methods was made to capitalize on the existing logic for notifying subscribers and updating cache. 

It's still possible to try and leverage the `multi` methods in a potential follow-up

kidroca (commented):
Overall, this PR is in great shape. However, I have raised a concern regarding the merge collection process, detailed in a separate note. Additionally, I've provided some comments on the new changes introduced in this PR since my last review.

lib/Onyx.js
kidroca: Thank you for your comprehensive explanations, @paultsimura.

I see the logic behind the current method and its potential for performance improvement. However, I'd like to offer a suggestion - not as a directive but as a point for consideration - that we might not need to differentiate between single-item and multiple-item updates right now. This thought is guided by a preference for simplicity, aiming to ease the review from others.

Perhaps the special handling for "1 collection key" scenarios could be tabled as an area for future exploration. This way, we can focus on the core functionality at present, leaving room for optimization down the line. Of course, this is just my perspective, and I'm open to further discussion or to hear other team members' thoughts on this matter.

tgolen:
> @tgolen this looks like a pretty unpredictable hole to me since the mergeCollection basically makes the object-dependencies in hooks useless, and we use such dependencies all across the App.

I agree, that is a bad hole. Did you consider changing `mergeCollection()` so that subscribers are only triggered if the object changes (so it works like `merge()`)? I think that should be a reasonable change to make. I think it would also be fine to do that in a follow-up issue so that PR you are working on here doesn't have the scope increased.

I'm just getting caught up from being on vacation last week. Is there anything that I can help with for a next step here? How is the QA coming on the original PR and what is the status/ETA of getting this merged?

tgolen:

blazejkustra:
@paultsimura I should be able to help with this! 😄 

[OnyxUtils.js migration](https://github.com/Expensify/react-native-onyx/pull/523) is close to a merge (@fabioh8010 is testing it in NewDot). 

Would it be possible to wait with this PR until the migration is merged? This could resolve some of your type problems + you won't have to care about declaration OnyxUtils.d.ts file.

blazejkustra:
Update: [OnyxUtils migration](https://github.com/Expensify/react-native-onyx/pull/523#pullrequestreview-1996416637) is in it's final state, waiting to be merged 😄

blazejkustra:
@paultsimura Update: [OnyxUtils](https://github.com/Expensify/react-native-onyx/pull/523#pullrequestreview-1996416637) was merged yesterday, we can remove HOLD from the title!

blazejkustra:
@paultsimura Please have a look at the fix [here](https://github.com/Expensify/react-native-onyx/pull/533). Thanks for reporting!

lib/OnyxUtils.ts
```diff
@@ -107,10 +109,10 @@ function initStoreValues(keys: DeepRecord<string, OnyxKey>, initialKeyStates: Pa
     // We need the value of the collection keys later for checking if a
     // key is a collection. We store it in a map for faster lookup.
     const collectionValues = Object.values(keys.COLLECTION ?? {});
-    onyxCollectionKeyMap = collectionValues.reduce((acc, val) => {
-        acc.set(val, true);
+    onyxCollectionKeySet = collectionValues.reduce((acc, val) => {
+        acc.add(val as OnyxKey); // todo
```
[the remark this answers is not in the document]
blazejkustra: Instead of assertion here, let's validate that `collectionValues` is of type `string[]`, we could also just do this:
```ts
const collectionValues = Object.values(keys.COLLECTION ?? {}) as string[];
```

lib/OnyxUtils.ts
```diff
@@ -256,11 +258,18 @@ function getAllKeys(): Promise<Set<OnyxKey>> {
 }
 
 /**
- * Checks to see if the a subscriber's supplied key
+ * @returns {Set} A set of collection keys.
```
blazejkustra, line 261: Let's remove "types" from JSDoc

lib/OnyxUtils.ts
```diff
@@ -256,11 +258,18 @@ function getAllKeys(): Promise<Set<OnyxKey>> {
...
+ */
+function getCollectionKeys() {
```
blazejkustra, line 263: Don't forget to add return types to every function, we could also enforce this with an eslint rule

lib/OnyxUtils.ts
```diff
@@ -1014,7 +1023,7 @@ function applyMerge(existingValue: OnyxValue<OnyxKey>, changes: Array<OnyxValue<
         return lastChange;
     }
 
-    if (changes.some((change) => typeof change === 'object')) {
+    if (changes.some((change) => change && typeof change === 'object')) {
```
blazejkustra, line 1026: +1 Thanks for noticing this regression!

lib/OnyxUtils.ts
```diff
@@ -1041,6 +1050,30 @@ function initializeWithDefaultKeyStates(): Promise<void> {
     });
 }
 
+function isValidMergeCollection<TKey extends CollectionKeyBase, TMap>(collectionKey: TKey, collection: Collection<TKey, TMap, NullishDeep<KeyValueMapping[TKey]>>) {
```
blazejkustra, line 1053: Missing return type here as well

lib/Onyx.ts
```diff
@@ -573,23 +554,43 @@ function update(data: OnyxUpdate[]): Promise<void> {
...
                 break;
             case OnyxUtils.METHOD.MERGE:
-                promises.push(() => merge(key, value));
+                enqueueMergeOperation(key, value);
                 break;
             case OnyxUtils.METHOD.MERGE_COLLECTION:
-                // eslint-disable-next-line @typescript-eslint/no-explicit-any -- We validated that the value is a collection
-                promises.push(() => mergeCollection(key, value as any));
+                // eslint-disable-next-line @typescript-eslint/no-explicit-any
+                if (OnyxUtils.isValidMergeCollection(key, value as any)) {
```
blazejkustra, line 588:
```suggestion
                if (OnyxUtils.isValidMergeCollection(key, value)) {
```

lib/OnyxUtils.ts
blazejkustra, line 1053: Also how about this declaration?
```suggestion
function isValidMergeCollection<TKey extends CollectionKeyBase>(collectionKey: TKey, collection: Record<string, unknown>): boolean {
```

We can use `Record<string, unknown>` here because this function verifies that `collection` is a valid argument for `enqueueMergeOperation` function

blazejkustra (changes requested):

blazejkustra:
Sorry for not posting these comments earlier, I reviewed the PR TS-wise. In case of any questions let me know! 😄

blazejkustra:
@paultsimura I'm not feeling too good today 🤒 @fabioh8010 could you help reviewing here?

fabioh8010 (approved):
LGTM from TS side!

tgolen:
Sorry, I have not been following the Onyx releases very closely. Where does this PR stand and does anyone know if we are close to removing the HOLD on this and moving it forward?

tgolen:
OK, thanks! I just merged that PR, so I'll take this one off hold.

lib/Onyx.ts
```diff
@@ -642,6 +644,54 @@ function update(data: OnyxUpdate[]): Promise<void> {
...
+    OnyxUtils.getCollectionKeys().forEach((collectionKey) => {
+        const collectionItemKeys = Object.keys(updateQueue).filter((key) => OnyxUtils.isKeyMatch(collectionKey, key));
+        if (collectionItemKeys.length <= 1) {
+            // If there are no items of this collection in the updateQueue, we should skip it.
+            // If there is only one item, we should update it individually, therefore retain it in the updateQueue.
+            return;
+        }
+
+        const batchedCollectionUpdates = collectionItemKeys.reduce(
+            (acc: MixedOperationsQueue, key: string) => {
```
tgolen, line 657: Instead of the name `acc`, please us a full variable name.

lib/Onyx.ts
```diff
@@ -642,6 +644,54 @@ function update(data: OnyxUpdate[]): Promise<void> {
...
+                const operations = updateQueue[key];
+
+                // Remove the collection-related key from the updateQueue so that it won't be processed individually later.
```
tgolen, line 660: What does "later" imply? Can you please update the comment to be more clear?

lib/Onyx.ts
```diff
@@ -642,6 +644,54 @@ function update(data: OnyxUpdate[]): Promise<void> {
         }
     });
 
+    // Group all the collection-related keys and update each collection in a single `mergeCollection` call
```
tgolen, line 647: Please add more context to this comment to answer _why_ this needs to happen.

lib/Onyx.ts
```diff
@@ -616,23 +599,42 @@ function update(data: OnyxUpdate[]): Promise<void> {
         }
     });
 
+    const updateQueue: Record<OnyxKey, Array<OnyxValue<OnyxKey>>> = {};
```
tgolen, line 602: Please add some comments about what this queue is for, what it's format is, and how it should be used.

tgolen (changes requested):
The code looks good. I have some requests for improved comments.

I believe that QA is currently doing a regression test against the latest version of Onyx. We should consider doing the same with this change before it is merged, since it is such a fundamental change. If you haven't done that before, let me know and I can help coordinate it.

srikarparsi (approved):
This lgtm but I'll leave it to @tgolen for the final review

mallenexpensify:
@tgolen it looks like there are [a couple](https://github.com/Expensify/App/issues/34686) of [issues](https://github.com/Expensify/App/issues/40589) held on this PR (and this PR is held on `App onyx bump`?)

I'm unsure of how big of a priority this is but I noticed you're focused on another project all week. If this is a priority, does it make sense to assign someone else for review? If not... we can hang tight.

tgolen (dismissed):

tgolen:
@paultsimura Can you please fix the conflicts? Also, can you link to whatever issue or PR this is on HOLD for?

tgolen:
@mallenexpensify I'm around this week, so I'll work on getting this merged. I don't think it needs reassigned at all.

tgolen:
@paultsimura Can you fix the conflicts here and link to whatever issue or PR this is on HOLD for?

tgolen:
Maybe @mountiny knows if there is a current PR for bumping Onyx? I don't think we track it anywhere, but you could maybe look through the titles of the open PRs to see if you can spot anything.

tgolen (approved):

mountiny:
Yep we got [this one](https://github.com/Expensify/App/pull/42772) and in the final stages of getting it done hopefully, its been a big effort 😅  

We are also discussing in this canvas thread if you want to follow https://expensify.slack.com/archives/C05R2V5GM1S/p1718007552204639?thread_ts=1717193187.147939&cid=C05R2V5GM1S

tgolen:
Great, thanks!

tgolen:
It looks like the version bump PR was merged, so I think I can take the hold off of this one. We might want to still let it sit for a couple of days to ensure the version bump isn't reverted until any regressions.

@paultsimura if you want to fix the conflicts again in the meantime, we can look at merging this next week. Maybe Tuesday or Wednesday.

mountiny:
Yes! Luckily, it seems like no regressions have been found really in connection to the bump, which is... sus, but I welcome it 😂 

@paultsimura I think you can now work on getting this PR ready for review and re-test again.

tgolen:
Bump @paultsimura

tgolen (dismissed):

tgolen:
OK, that's probably wise. In order to do that, you need to make an App PR that points to this specific branch. Then I can create an ad-hoc build that we can give to QA for full regression testing.

tgolen:
That's the correct way to do it, yeah. There should be a couple other examples in the `package.json` which do this. I think, what I forgot to mention is that you need to temporarily include the built source files on your branch. I believe you can do this with the `git add -f` command (`-f` to force the file to be added and bypass the `.ignore` file).

tgolen:
From the issue, it sounds like this can be closed for now.

tgolen (approved):
Looks great, thank you! Here we go!

aldo-expensify:
@paultsimura with these changes, does `mergecollection` onyx update have a performance advantage vs a bunch of `merge` updates if they belong to the same collection? or it should be roughly the same because the separate `merge` updates will be put together as a `mergecollection`?

aldo-expensify:
Nice, thanks!

aldo-expensify:
@paultsimura I found a bug in the version of react-native-onyx we are currently using and it seems like your PR here is fixing it. The bug: https://github.com/Expensify/App/issues/46135#issue-2428111404

When are we planning to release the changes made here?

PR #520, opened 2024-03-26
keyChange update key/connectionID for faster lookup

gedu:
### Details

Following proposal: https://github.com/Expensify/App/issues/35809

### Manual Tests

https://github.com/Expensify/App/pull/36226#issuecomment-1938661752

### Author Checklist

    - [X] Android / native
    - [X] Android / Chrome
    - [X] iOS / native
    - [X] iOS / Safari
    - [X] MacOS / Chrome / Safari
    - [X] MacOS / Desktop
- [X] I verified there are no console errors (if there's a console error not related to the PR, report it or open an issue for it to be fixed)
    - [X] The file has a description of what it does and/or why is needed at the top of the file if the code is not self explanatory
- [X] If the `main` branch was merged into this PR after a review, I tested again and verified the outcome was still expected according to the `Test` steps.
- [X] I have checked off every checkbox in the PR author checklist, including those that don't apply to this PR.

### Screenshots/Videos
<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>


gedu:

lib/OnyxUtils.d.ts
```diff
@@ -51,6 +51,14 @@ declare function getCallbackToStateMapping(): Record<string, Mapping<string>>;
...
 declare function getDefaultKeyStates(): Record<string, OnyxValue<string>>;
 
+/**
+ *
+ * @param connectionID connectionID of the subscriber
+ * @param key key that the subscriber is connected to
+ */
+declare function storeKeyByConnections(connectionID: string, key: OnyxKey): void;
+
+declare function deleteKeyByConnections(connectionID: number): void;
```
Julesssss, line 61: NAB: Just incase these get refactored or a new function is created which separates them, would you mind adding a function comment for `deleteKeyByConnections` too please?

lib/OnyxUtils.js
```diff
@@ -311,6 +314,21 @@ function isSafeEvictionKey(testKey) {
     return _.some(evictionAllowList, (key) => isKeyMatch(key, testKey));
 }
 
+function storeKeyByConnections(key, connectionID) {
+    if (!onyxKeyToConnectionIDs.has(key)) {
+        onyxKeyToConnectionIDs.set(key, []);
+    }
```
Julesssss, line 320: This is simply setting a default array right? I miss-read this initially as it looked like we were overwriting the value with `[]`

Julesssss:
> [gedu](https://github.com/gedu) requested a review from https://github.com/orgs/Expensify/teams/pullerbear as a [code owner](https://github.com/Expensify/react-native-onyx/blob/667e0d2daa67e06ee02cc1162ce3731c3cac18f2/.github/CODEOWNERS#L2) [6 hours ago](https://github.com/Expensify/react-native-onyx/pull/520#event-12319149999)

Is this off HOLD?

cc @mountiny who is assigned to the linked issue

mountiny:
It is not off hold until we merge the onyx bump and deploy it, but we can review in the meantime

lib/OnyxUtils.js
gedu: Yes, it is setting a default array

Julesssss:
> It is not off hold until we merge the onyx bump and deploy it, but we can review in the meantime

I think we've likely bumped Onyx by now?

gedu:
@mountiny can we keep reviewing this PR? 
Will be more improvements: https://callstack-hq.slack.com/archives/C05LX9D6E07/p1715178705054759

lib/Onyx.ts
```diff
@@ -99,6 +99,8 @@ function connect<TKey extends OnyxKey>(connectOptions: ConnectOptions<TKey>): nu
     callbackToStateMapping[connectionID] = mapping as Mapping<OnyxKey>;
     callbackToStateMapping[connectionID].connectionID = connectionID;
 
+    OnyxUtils.storeKeyByConnections(mapping.key, callbackToStateMapping[connectionID].connectionID);
```
mountiny, line 102: In onyx, its best practice to be more verbose with comments than less. Could you please add a comment explaining here why we are doing this? Thanks!

lib/OnyxUtils.ts
```diff
@@ -295,6 +298,32 @@ function isSafeEvictionKey(testKey: OnyxKey): boolean {
     return evictionAllowList.some((key) => isKeyMatch(key, testKey));
 }
 
+/**
+ * Stores a connection ID associated with a given key.
+ *
+ * @param connectionID connectionID of the subscriber
```
mountiny, line 304:
```suggestion
 * @param connectionID - a connection ID of the subscriber
```

lib/OnyxUtils.ts
```diff
@@ -295,6 +298,32 @@ function isSafeEvictionKey(testKey: OnyxKey): boolean {
...
+ * @param key key that the subscriber is connected to
```
mountiny, line 305:
```suggestion
 * @param key - a key that the subscriber is connected to
```

lib/OnyxUtils.ts
```diff
@@ -629,7 +666,16 @@ function keyChanged<TKey extends OnyxKey>(
     // We are iterating over all subscribers to see if they are interested in the key that has just changed. If the subscriber's  key is a collection key then we will
     // notify them if the key that changed is a collection member. Or if it is a regular key notify them when there is an exact match. Depending on whether the subscriber
     // was connected via withOnyx we will call setState() directly on the withOnyx instance. If it is a regular connection we will pass the data to the provided callback.
-    const stateMappingKeys = Object.keys(callbackToStateMapping);
+    let stateMappingKeys = onyxKeyToConnectionIDs.get(key);
+
+    if (!stateMappingKeys) {
+        // Getting the collection key from the specific key 'cos only collection keys were stored in the mapping.
+        stateMappingKeys = onyxKeyToConnectionIDs.get(getPureKey(key));
+        if (!stateMappingKeys) {
```
mountiny, line 674: Could you update the comment above to explain this thoroughly? Also can you explain here and in comment why do we not need the same change for the `keysChanged` method?

lib/OnyxUtils.ts
```diff
@@ -382,6 +411,13 @@ function addToEvictionBlockList(key: OnyxKey, connectionID: number): void {
     evictionBlocklist[key].push(connectionID);
 }
 
+function getPureKey(key: OnyxKey): string {
```
mountiny, line 414: Can you add a short doc explanation for this method please?

lib/Onyx.ts
```diff
@@ -99,6 +99,11 @@ function connect<TKey extends OnyxKey>(connectOptions: ConnectOptions<TKey>): nu
     callbackToStateMapping[connectionID] = mapping as Mapping<OnyxKey>;
     callbackToStateMapping[connectionID].connectionID = connectionID;
 
+    // When keyChanged is called, the key is passed and looks through all the callbackToStateMapping the right key
+    // to avoid having to loop through all the keys all the time (even when just one connection belongs to one key),
+    // we store the key by connectionID, to access the specific list of connectionIDs
```
mountiny, line 104: @gedu sorry can you please try to rewrite this a bit more, I am not sure if I am following, especially this part does not seem correct to me `the key is passed and looks through all the callbackToStateMapping the right key`.

What `looks through` callbackToStateMapping? What `the right key`?

Thanks!

lib/OnyxUtils.ts
```diff
@@ -626,10 +674,21 @@ function keyChanged<TKey extends OnyxKey>(
...
+    let stateMappingKeys = onyxKeyToConnectionIDs.get(key);
+
+    if (!stateMappingKeys) {
+        // Getting the collection key from the specific key because only collection keys were stored in the mapping.
+        stateMappingKeys = onyxKeyToConnectionIDs.get(getPureKey(key));
+        if (!stateMappingKeys) {
+            return;
+        }
+    }
+
```
hannojg, line 691: also NAB and just code style discussion / how explicit we want to be.

It might be more explicit:
```js
const connectionKey = isCollectionKey(key) ? getPureKey(key) : key;
const stateMappingKeys = onyxKeyToConnectionIDs.get(connectionKey)
if (!stateMappingKeys) {
  return;
}
```

Here we explicitly check for a collection key, instead of implicitly doing that by having a fallback check, which is more verbose

hannojg (commented):
LGTM (just some stylistic discussion, but the main improvement is 🔥!)

lib/OnyxUtils.ts
gedu: Most of the time the key exists and isn't a collection, also when it enters the main loop 
```
for (let i = 0; i < stateMappingKeys.length; i++) {
```
it also checks for `isCollectionKey`, wanna avoid extra computation

lib/OnyxUtils.ts
```diff
@@ -295,6 +298,32 @@ function isSafeEvictionKey(testKey: OnyxKey): boolean {
...
+/**
+ * Deletes a connection ID associated with its corresponding key.
+ *
+ * @param {number} connectionID - The connection ID to be deleted.
+ */
+function deleteKeyByConnections(connectionID: number) {
+    const subscriber = callbackToStateMapping[connectionID];
+
+    if (subscriber && onyxKeyToConnectionIDs.has(subscriber.key)) {
+        onyxKeyToConnectionIDs.set(subscriber.key, _.without(onyxKeyToConnectionIDs.get(subscriber.key), connectionID));
```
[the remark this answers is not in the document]
gedu: Thanks for point this out, I also improve it using `filter` instead of `without`

Screenshot 2024-05-10 at 13 16 37

lib/OnyxUtils.ts
hannojg: okay, makes sense. If its also that way for performance reasons, might be worth to add a comment to avoid future regressions

lib/OnyxUtils.ts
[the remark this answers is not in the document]
hannojg: Sorry, don't want to be the nit picker here - I had a typo in my first example, it should be `updatedConnectionIDs` with a d instead of an s 😅 @gedu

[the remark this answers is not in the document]
gedu: Don't worry, this is the moment to change it 💪

gedu:
@mountiny should I check something more?

mountiny (dismissed):
Looks good, @mollfpr please, how is the testing going? Ideally if there are no clear blockers found, I think we should merge now and ask applause to do regression testing on the onyx bump

lib/OnyxUtils.ts
```diff
@@ -381,6 +411,24 @@ function addToEvictionBlockList(key: OnyxKey, connectionID: number): void {
...
+ * For example:
+ * - `getPureKey("report_123")` would return "report_"
+ * - `getPureKey("report")` would return "report"
+ * - `getPureKey("report_")` would return "report_"
+ * - `getPureKey(null)` would return ""
+ *
+ * @param {OnyxKey} key - The key to process.
+ * @return {string} The pure key without any numeric
+ */
+function getPureKey(key: OnyxKey): string {
```
chrispader, line 426: I find the name of this function a bit misleading. What is a pure key? Why is the "collection part" pure?

Also, think we should not just remove the non-numeric part to get the key part that's defining the collection. 

Technically, it would be possible to have a collection called something like `101products_` or `report5_`

Instead we should scrap the "collection part" based on the valid/allowed collection keys configured here: https://github.com/Expensify/react-native-onyx/blob/d7ab22617fedc1faba251e9680909793cd01ea9c/lib/OnyxUtils.ts#L109-L113

chrispader (changes requested):
Not sure, if the minor requested changes are worth holding this PR, because i'm not sure if it's not a very minor edge-case.

Feel free to merge this anyway. If we expect collection keys to always follow a non-numeric pattern, this should work fine!

lib/OnyxUtils.ts
```diff
@@ -427,7 +427,7 @@ function getPureKey(key: OnyxKey): string {
     if (!key) {
         return '';
     }
-    return key.replace(/\d+/g, '');
+    return key.replace(/_\w+/g, '_');
 }
```
chrispader, line 431: I think this is definitely better and should cover most use-cases. But just hypothetically, isn't it possible to have more than one `_` in the collection key or have no `_` at all? Sth. like `reports.` or `reports$`?

mountiny: I think if we would do this, it would be larger change

lib/OnyxUtils.ts
gedu: @chrispader  I can make that change, that name I just keep it when I was doing some proof of concepts, I can change it. 
I also just change it, to add more support to cases like "report_add", "report_test1". That function is to clean, and get the main key, from "report_123" get "report_", using the key from keys.COLLECTION I would have to loop through the collection keys, and check the startsWith.

mollfpr:
I have tested the web and the desktop doesn't find any issues, but I am still unable to run the iOS and Android with Onyx installed from the local path.

Simulator Screenshot - iPhone 15 Pro 17 2 - 2024-05-11 at 01 26 52

mountiny:
@mollfpr thank you! that is helpful, I think we are good with web testing on this change for now and full testing would be done in the App bump.

@gedu what is your ETA for addressing the test failures?

mollfpr:
We have couples of failing test case https://github.com/Expensify/react-native-onyx/actions/runs/9034415824/job/24828280471?pr=520

lib/OnyxUtils.ts
gedu: Something that I'm just realize, is when sending a key that is `report_1234` it sends to any component listening to that specific key, but it also sends updates to all components listening to `report_`, I'm looking into a way to update my code + keep the performance

gedu:
@mountiny I just pushed a fix

mollfpr:
The test is pass now!

Screenshot 2024-05-14 at 02 40 15

mountiny:
Re running

gedu:
@mountiny is it all good?

chrispader (changes requested):
LGTM aside form one required JSDoc comment update

lib/OnyxUtils.ts
```diff
@@ -381,6 +411,24 @@ function addToEvictionBlockList(key: OnyxKey, connectionID: number): void {
...
+ *
+ * For example:
+ * - `getPureKey("report_123")` would return "report_"
+ * - `getPureKey("report")` would return "report"
+ * - `getPureKey("report_")` would return "report_"
+ * - `getPureKey(null)` would return ""
+ *
+ * @param {OnyxKey} key - The key to process.
+ * @return {string} The pure key without any numeric
+ */
```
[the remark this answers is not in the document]
gedu: @chrispader what needs to be updated? can you help me I don't recall

gedu:
All good from my side, this is still valid, just need to double check @chrispader comment about JSDoc

lib/OnyxUtils.ts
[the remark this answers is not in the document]
chrispader: Just the name and maybe we can change the description to

`It extracts the non-numeric collection identifier of a given key.` or sth.

`getCollectionKey` instead of `getPureKey`

gedu:
@chrispader updated JSDoc

chrispader (approved):
Thanks for the changes. LGTM!

Julesssss (approved):

lib/OnyxUtils.ts
```diff
@@ -397,6 +427,24 @@ function addToEvictionBlockList(key: OnyxKey, connectionID: number): void {
...
+function getCollectionKey(key: OnyxKey): string {
+    if (!key) {
+        return '';
+    }
+    return key.replace(/_\w+/g, '_');
+}
```
mountiny, line 447: @chrispader @gedu Soooo as part of the workspace feeds and billing docs (I think) we are actually going to add onyx collection keys that have 2 `_` for example `sharedNVP_private_expensifyCardSettings_12345`

Will these changes work with it? cc @fabioh8010

mountiny (changes requested):

lib/OnyxUtils.ts
chrispader: No this will not work in this case. We would need to search for the last `_` and only remove the rest after that. cc @gedu

mountiny: Raised [here](https://expensify.slack.com/archives/C05LX9D6E07/p1719310826344159)

chrispader:
```suggestion
function getCollectionKey(key: OnyxKey): string {
    if (!key) {
        return '';
    }
    const lastUnderscoreIndex = key.lastIndexOf('_');
    if (lastUnderscoreIndex === -1) {
        return key;
    }
    return key.substring(0, lastUnderscoreIndex);
}
```

gedu: won't work, will be always the last one?

fabioh8010: Maybe we could use fix and use `splitCollectionMemberKey` instead? Regardless of the decision, could we use this PR to fix `splitCollectionMemberKey` to use `lastIndexOf()`?

chrispader: > won't work, will be always the last one?

Good question. I think we should have some convention around this. Allowing underscores in both the collection key and the member key might cause problems where the correct key cannot be determined correctly.

Imagine having a collection `some_collection_` and the member key `member_key1`. Which would result in the full key `some_collection_member_key1`.

In this case it would not work...

mountiny: I think we can have underscore in non-collection keys like some private_nvpname. We cannot easily change that since that is how its stored in the db.

But we can have the convention that the collection has _<number> in the end I think

gedu: @mountiny what should I do? I got lost

gedu: I updated and added some tests

mountiny: > Instead we should scrap the "collection part" based on the valid/allowed collection keys configured here:

I think this might not be that easy either as we are going to add more complex collection keys

I think for now that a reasonable way to detect a collection key is that after the last `_`, there is only one number left. So that would be a way to detect a collection key

If you have a key that has _ in it and its not a collection, it would not have any number at the end

mountiny: Or we can add sharedNVP_...... to the collection keys

gedu: @mountiny  I think in this PR we should keep just this collection update, and in a follow up PR look into the new single key with `_`, I'm seeing that we will need to plan that part a bit more

mountiny (approved):
Thanks for adding the tests.

chrispader (approved):

Julesssss (approved):

Julesssss:
I think we're ready to go here, are we waiting on any final reviews?

gedu:
Any news? @mountiny @Julesssss

Julesssss (approved):

Julesssss:
> Any news? @mountiny @Julesssss

Sorry, I missed this one. Can you please create an onyx bump PR and assign me please?

Julesssss:
Ah shit. Your first two commits weren't signed. Could you please raise another PR against this [issue](https://github.com/Expensify/App/issues/35809)?

gedu:
Ohhh maan, ok on my way, sorry for that

gedu:
@Julesssss new PR created: https://github.com/Expensify/react-native-onyx/pull/565

PR #523, opened 2024-03-27
[TS migration] Migrate Onyx private methods to TS 

blazejkustra:
### Details
This PR migrates "private" Onyx.js methods that we use in NewDot, "private" methods were moved to `OnyxUtils.js` in this [PR](https://github.com/Expensify/react-native-onyx/pull/512).

### Related Issues
Part of https://github.com/Expensify/App/issues/34344

### Automated Tests
N/A

### Manual Tests

- Tested NewDot thoroughly on every platform, requested money, logged in and out, sent messages

### Author Checklist

    - [x] Android / native
    - [x] Android / Chrome
    - [x] iOS / native
    - [x] iOS / Safari
    - [x] MacOS / Chrome / Safari
    - [x] MacOS / Desktop

### Screenshots/Videos
<details>

[android.webm](https://github.com/Expensify/react-native-onyx/assets/46095609/0b28f030-c827-471c-9066-ab9ec5ae813c)

</details>

<details>

[android-web.webm](https://github.com/Expensify/react-native-onyx/assets/46095609/a86b67b7-22e9-40e3-adf6-cf35db7bdf58)

</details>

<details>

https://github.com/Expensify/react-native-onyx/assets/46095609/3838f03b-ad4f-4a86-8c24-d782ea7140a5

</details>

<details>

https://github.com/Expensify/react-native-onyx/assets/46095609/4a07de6f-086f-427a-9f75-8977c682b561

</details>

<details>

https://github.com/Expensify/react-native-onyx/assets/46095609/3db5759f-f1c2-41d2-adb3-1d01066e0d30

</details>

<details>

https://github.com/Expensify/react-native-onyx/assets/46095609/feea403b-2ae4-40a0-bb72-5abe73118e95

</details>


tgolen:

blazejkustra:
Absolutely! The PR is still WIP, so no need to worry that is going to be merged too soon

blazejkustra:
Sorry @aldo-expensify, I miss-clicked. @fabioh8010 is still reviewing this PR, once it's approved I'll make it as ready for review 😄

lib/OnyxUtils.ts
```diff
@@ -742,8 +670,8 @@ function keyChanged(key, data, prevData, canUpdateSubscriber = () => true, notif
             // returned by the selector and only if the selected data has changed.
             if (subscriber.selector) {
                 subscriber.withOnyxInstance.setStateProxy(() => {
-                    const previousValue = getSubsetOfData(prevData, subscriber.selector, subscriber.withOnyxInstance.state);
```
fabioh8010, line 745: I'm starting to wonder if removing `getSubsetOfData` was a good idea, it was encapsulating this logic of using the selector to get the data and for know it was just calling the selector, but in the future if we want to change or make it more complex we will have to replicate it everywhere now. wdyt?

lib/OnyxUtils.ts
```diff
@@ -783,71 +711,59 @@ function keyChanged(key, data, prevData, canUpdateSubscriber = () => true, notif
...
- * @param {Function} [mapping.callback]
- * @param {String} [mapping.selector]
- * @param {*|null} val
- * @param {String|undefined} matchedKey
- * @param {Boolean} isBatched
  */
-function sendDataToConnection(mapping, val, matchedKey, isBatched) {
+function sendDataToConnection<TKey extends OnyxKey>(
+    mapping: Mapping<TKey>,
+    val: OnyxCollection<OnyxValue<TKey>> | OnyxEntry<OnyxValue<TKey>>,
```
fabioh8010, line 717:
```suggestion
    val: OnyxValue<TKey>,
```

In this example makes sense to use `OnyxValue`

lib/OnyxUtils.ts
```diff
@@ -783,71 +711,59 @@ function keyChanged(key, data, prevData, canUpdateSubscriber = () => true, notif
...
-        let newData = val;
+    if ('withOnyxInstance' in mapping) {
+        let newData: OnyxValue<OnyxKey> = val;
 
         // If the mapping has a selector, then the component's state must only be updated with the data
         // returned by the selector.
         if (mapping.selector) {
             if (isCollectionKey(mapping.key)) {
-                newData = reduceCollectionWithSelector(val, mapping.selector, mapping.withOnyxInstance.state);
+                newData = reduceCollectionWithSelector(val as OnyxCollection<OnyxValue<TKey>>, mapping.selector, mapping.withOnyxInstance.state);
```
fabioh8010, line 734:
```suggestion
                newData = reduceCollectionWithSelector(val as OnyxCollection<KeyValueMapping[TKey]>, mapping.selector, mapping.withOnyxInstance.state);
```

but we need to fix here and `reduceCollectionWithSelector` to accept `OnyxCollection<KeyValueMapping[TKey]>`

fabioh8010 (commented):
Also let's make sure all functions have return types!

lib/Onyx.ts
```diff
@@ -206,33 +207,35 @@ function disconnect(connectionID: number, keyToRemoveFromEvictionBlocklist?: Ony
...
     Logger.logInfo(`set called for key: ${key}${_.isObject(value) ? ` properties: ${_.keys(value).join(',')}` : ''} hasChanged: ${hasChanged}`);
 
     // This approach prioritizes fast UI changes without waiting for data to be stored in device storage.
-    const updatePromise = OnyxUtils.broadcastUpdate(key, valueAfterRemoving, hasChanged, wasRemoved);
+    const updatePromise = OnyxUtils.broadcastUpdate(key, valueWithoutNullValues, hasChanged, wasRemoved);
 
     // If the value has not changed or the key got removed, calling Storage.setItem() would be redundant and a waste of performance, so return early instead.
     if (!hasChanged || wasRemoved) {
-        return updatePromise;
+        return updatePromise.then(() => undefined);
```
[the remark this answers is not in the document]
blazejkustra: True, it's not necessary here!

lib/Onyx.ts
```diff
@@ -258,7 +261,8 @@ function multiSet(data: Partial<NullableKeyValueMapping>): Promise<void[]> {
         .then(() => {
             OnyxUtils.sendActionToDevTools(OnyxUtils.METHOD.MULTI_SET, undefined, data);
             return Promise.all(updatePromises);
-        });
+        })
+        .then(() => undefined);
```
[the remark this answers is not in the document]
blazejkustra: Yes, I tested this in the App. I've done it to clean the return types of Onyx public methods, as it was confusing to return Promise<void[]> or Promise<[void, void]>

lib/OnyxUtils.ts
blazejkustra: I have type errors after this change, any idea how to fix it?

blazejkustra: Other than this: https://github.com/Expensify/react-native-onyx/pull/523/commits/6807e085aa4f566208a91616b30a670c0f0bda86

lib/OnyxUtils.ts
```diff
@@ -939,53 +854,40 @@ function getCollectionDataAndSendAsObject(matchingKeys, mapping) {
...
- * @param {String} key
- * @param {*} value
- * @param {*} prevValue
- * @param {Function} [canUpdateSubscriber] only subscribers that pass this truth test will be updated
- * @returns {Promise}
  */
-function scheduleSubscriberUpdate(key, value, prevValue, canUpdateSubscriber = () => true) {
+function scheduleSubscriberUpdate<TKey extends OnyxKey>(
+    key: TKey,
+    value: OnyxEntry<OnyxValue<TKey>>,
```
fabioh8010, line 860:
```suggestion
    value: OnyxValue<TKey>,
```

lib/OnyxUtils.ts
```diff
@@ -939,53 +854,40 @@ function getCollectionDataAndSendAsObject(matchingKeys, mapping) {
...
- * @private
- * @param {String} key
- * @return {Promise}
  */
-function remove(key) {
-    const prevValue = cache.getValue(key, false);
+function remove<TKey extends OnyxKey>(key: TKey): Promise<void> {
+    const prevValue = cache.getValue(key, false) as OnyxValue<TKey>;
     cache.drop(key);
     scheduleSubscriberUpdate(key, null, prevValue);
```
fabioh8010, line 886:
```suggestion
    scheduleSubscriberUpdate(key, null as OnyxValue<TKey>, prevValue);
```

lib/OnyxUtils.ts
```diff
@@ -1027,20 +927,16 @@ function evictStorageAndRetry(error, onyxMethod, ...args) {
...
- * @param {String} key
- * @param {*} value
- * @param {Boolean} hasChanged
- * @param {Boolean} wasRemoved
- * @returns {Promise}
+ * Notifies subscribers and writes current value to cache
  */
-function broadcastUpdate(key, value, hasChanged, wasRemoved = false) {
-    const prevValue = cache.getValue(key, false);
+function broadcastUpdate<TKey extends OnyxKey>(key: TKey, value: OnyxEntry<OnyxValue<TKey>>, hasChanged?: boolean, wasRemoved = false): Promise<void> {
```
fabioh8010, line 938:
```suggestion
function broadcastUpdate<TKey extends OnyxKey>(key: TKey, value: OnyxValue<TKey>, hasChanged?: boolean, wasRemoved = false): Promise<void> {
```

lib/OnyxUtils.ts
```diff
@@ -650,18 +586,17 @@ function keysChanged(collectionKey, partialCollection, notifyRegularSubscibers =
...
- * @param {*} data
- * @param {*} prevData
- * @param {Function} [canUpdateSubscriber] only subscribers that pass this truth test will be updated
- * @param {boolean} [notifyRegularSubscibers=true]
- * @param {boolean} [notifyWithOnyxSubscibers=true]
  */
-function keyChanged(key, data, prevData, canUpdateSubscriber = () => true, notifyRegularSubscibers = true, notifyWithOnyxSubscibers = true) {
+function keyChanged<TKey extends OnyxKey>(
+    key: TKey,
+    data: OnyxEntry<OnyxValue<TKey>>,
```
fabioh8010, line 592:
```suggestion
    data: OnyxValue<TKey>,
```

Also you need to apply some fixes I commented below, and also these files:

```diff
diff --git a/lib/storage/providers/MemoryOnlyProvider.ts b/lib/storage/providers/MemoryOnlyProvider.ts
index 00583a5..0dd0397 100644
--- a/lib/storage/providers/MemoryOnlyProvider.ts
+++ b/lib/storage/providers/MemoryOnlyProvider.ts
@@ -36,7 +36,7 @@ const provider: StorageProvider = {
     getItem(key) {
         const value = store[key] as OnyxValue<typeof key>;
 
-        return Promise.resolve(value === undefined ? null : value);
+        return Promise.resolve(value === undefined ? (null as OnyxValue<typeof key>) : value);
     },
 
     /**
diff --git a/lib/storage/providers/NoopProvider.ts b/lib/storage/providers/NoopProvider.ts
index 07af0a4..f99af06 100644
--- a/lib/storage/providers/NoopProvider.ts
+++ b/lib/storage/providers/NoopProvider.ts
@@ -1,3 +1,4 @@
+import type {OnyxValue} from '../../types';
 import type StorageProvider from './types';
 
 const provider: StorageProvider = {
@@ -15,11 +16,9 @@ const provider: StorageProvider = {
 
     /**
      * Get the value of a given key or return `null` if it's not available in memory
-     * @param {String} key
-     * @return {Promise<*>}
      */
-    getItem() {
-        return Promise.resolve(null);
+    getItem(key) {
+        return Promise.resolve(null as OnyxValue<typeof key>);
     },
 
     /**
diff --git a/lib/storage/providers/types.ts b/lib/storage/providers/types.ts
index 0161ecd..6552602 100644
--- a/lib/storage/providers/types.ts
+++ b/lib/storage/providers/types.ts
@@ -5,7 +5,7 @@ type KeyValuePair = [OnyxKey, OnyxValue<OnyxKey>];
 type KeyList = OnyxKey[];
 type KeyValuePairList = KeyValuePair[];
 
-type OnStorageKeyChanged = <TKey extends OnyxKey>(key: TKey, value: OnyxValue<TKey> | null) => void;
+type OnStorageKeyChanged = <TKey extends OnyxKey>(key: TKey, value: OnyxValue<TKey>) => void;
 
 type StorageProvider = {
     /**
@@ -19,7 +19,7 @@ type StorageProvider = {
     /**
      * Gets the value of a given key or return `null` if it's not available in storage
      */
-    getItem: <TKey extends OnyxKey>(key: TKey) => Promise<OnyxValue<TKey> | null>;
+    getItem: <TKey extends OnyxKey>(key: TKey) => Promise<OnyxValue<TKey>>;
 
     /**
      * Get multiple key-value pairs for the given array of keys in a batch

```

lib/OnyxUtils.ts
```diff
@@ -689,25 +624,26 @@ function keyChanged(key, data, prevData, canUpdateSubscriber = () => true, notif
                 continue;
             }
 
-            subscriber.callback(data, key);
+            subscriber.callback(data as OnyxCollection<OnyxValue<TKey>>, key);
```
fabioh8010, line 627:
```suggestion
            const subscriberCallback = subscriber.callback as DefaultConnectCallback<TKey>;
            subscriberCallback(data as OnyxEntry<KeyValueMapping[TKey]>, key);
```

lib/OnyxUtils.ts
fabioh8010: Well I don't agree but it's up to you to keep this way, it's also ok for me 🙂

lib/OnyxUtils.ts
```diff
@@ -857,21 +776,17 @@ function addKeyToRecentlyAccessedIfNeeded(mapping) {
...
     // Tasks that are pending
-    const pendingTasks = [];
+    const pendingTasks: Array<Promise<OnyxValue<TKey>>> = [];
     // Keys for the tasks that are pending
-    const pendingKeys = [];
+    const pendingKeys: OnyxKey[] = [];
 
     // We are going to combine all the data from the matching keys into a single object
-    const data = {};
+    const data: OnyxCollection<OnyxValue<TKey>> | OnyxEntry<OnyxValue<TKey>> = {};
```
fabioh8010, line 789:
```suggestion
    const data: OnyxCollection<KeyValueMapping[TKey]> = {};
```

lib/OnyxUtils.ts
```diff
@@ -920,17 +835,17 @@ function getCollectionDataAndSendAsObject(matchingKeys, mapping) {
             }
 
             // temp object is used to merge the missing data into the cache
-            const temp = {};
-            values.forEach((value) => {
-                data[value[0]] = value[1];
-                temp[value[0]] = value[1];
+            const temp: OnyxCollection<OnyxValue<TKey>> | OnyxEntry<OnyxValue<TKey>> = {};
```
fabioh8010, line 838:
```suggestion
            const temp: OnyxCollection<KeyValueMapping[TKey]> = {};
```

lib/OnyxUtils.ts
```diff
@@ -939,53 +854,40 @@ function getCollectionDataAndSendAsObject(matchingKeys, mapping) {
...
  * This method is similar to notifySubscribersOnNextTick but it is built for working specifically with collections
  * so that keysChanged() is triggered for the collection and not keyChanged(). If this was not done, then the
  * subscriber callbacks receive the data in a different format than they normally expect and it breaks code.
- *
- * @param {String} key
- * @param {*} value
- * @returns {Promise}
  */
-function scheduleNotifyCollectionSubscribers(key, value) {
+function scheduleNotifyCollectionSubscribers(key: OnyxKey, value: OnyxCollection<OnyxValue<OnyxKey>>): Promise<void> {
```
fabioh8010, line 874:
```suggestion
function scheduleNotifyCollectionSubscribers<TKey extends OnyxKey>(key: TKey, value: OnyxCollection<KeyValueMapping[TKey]>): Promise<void> {
```

lib/OnyxUtils.ts
```diff
@@ -1050,51 +946,47 @@ function broadcastUpdate(key, value, hasChanged, wasRemoved = false) {
...
  * Otherwise removes all nested null values in objects and returns the object
- * @param {String} key
- * @param {Mixed} value
- * @returns {Mixed} The value without null values and a boolean "wasRemoved", which indicates if the key got removed completely
+ *
+ * @returns The value without null values and a boolean "wasRemoved", which indicates if the key got removed completely
  */
-function removeNullValues(key, value) {
-    if (_.isNull(value)) {
+function removeNullValues(key: OnyxKey, value: OnyxValue<OnyxKey>): {value: Record<string, unknown> | unknown[] | null; wasRemoved: boolean} {
```
fabioh8010, line 962: Could you extract `{value: Record<string, unknown> | unknown[] | null; wasRemoved: boolean}` to a separate type?

lib/types.ts
```diff
@@ -265,44 +265,57 @@ type Collection<TKey extends CollectionKeyBase, TMap, TValue> = {
 };
 
 type WithOnyxInstance = Component<unknown, WithOnyxInstanceState<NullableKeyValueMapping>> & {
-    setStateProxy: (cb: (state: NullableKeyValueMapping) => OnyxValue<OnyxKey>) => void;
+    setStateProxy: (cb: (state: Record<string, OnyxCollection<OnyxValue<OnyxKey>>>) => OnyxValue<OnyxKey>) => void;
```
fabioh8010, line 268:
```suggestion
    setStateProxy: (cb: (state: Record<string, OnyxCollection<KeyValueMapping[OnyxKey]>>) => OnyxValue<OnyxKey>) => void;
```

lib/OnyxUtils.ts
fabioh8010: It's okay in my opinion (i do the same on `useOnyx`)

lib/OnyxUtils.ts
blazejkustra: This causes errors when calling this function

blazejkustra:
@aldo-expensify All yours! It was already tested in NewDot by me and @fabioh8010 😄

lib/useOnyx.ts
```diff
@@ -140,16 +140,16 @@ function useOnyx<TKey extends OnyxKey, TReturnValue = OnyxValue<TKey>>(key: TKey
...
+                key,
                 callback: () => {
                     // We don't need to update the Onyx cache again here, when `callback` is called the cache is already
                     // expected to be updated, so we just signal that the store changed and `getSnapshot()` can be called again.
                     isFirstConnectionRef.current = false;
                     onStoreChange();
                 },
                 initWithStoredValues: options?.initWithStoredValues,
-                waitForCollectionCallback: OnyxUtils.isCollectionKey(key),
+                waitForCollectionCallback: OnyxUtils.isCollectionKey(key) as true,
```
aldo-expensify, line 152: I don't think these types are well defined. Why are we using the case `as true`? cast should be avoided, why do we need one here? If we are using `CollectionKeyBase` above, shouldn't we just put `waitForCollectionCallback: true` instead of calling `OnyxUtils.isCollectionKey(key)`? If this may not be a collection, then why did we hardcode `CollectionKeyBase` up there in the first place?

lib/OnyxUtils.ts
```diff
@@ -208,7 +216,7 @@ function get(key) {
 
     // When a value retrieving task for this key is still running hook to it
     if (cache.hasPendingTask(taskName)) {
-        return cache.getTaskPromise(taskName);
+        return cache.getTaskPromise(taskName) as Promise<OnyxValue<OnyxKey>>;
     }
```
aldo-expensify, line 220: While I see that this won't cause an error because I checked the implementation of these functions, it again uses cast which I think signals not a good design in the code. What about just doing:

```
const taskPromise = cache.getTaskPromise(taskName);
if (taskPromise !== undefined) {
    return taskPromise;
}
```

Another approach is that `cache.getTaskPromise` should throw when it doesn't find the task, then `undefined` can be removed from the return type of `cache.getTaskPromise` and you can use `cache.hasPendingTask` like you do now.

lib/DevTools.ts
```diff
@@ -58,7 +58,7 @@ class DevTools {
      * @param payload - data written to the storage
      * @param stateChanges - partial state that got updated after the changes
      */
-    registerAction(type: string, payload: unknown, stateChanges: Record<string, unknown> = {}) {
+    registerAction(type: string, payload: unknown, stateChanges: Record<string, unknown> | null = {}) {
         try {
             if (!this.remoteDev) {
                 return;
```
aldo-expensify, line 64: NAB: if `payload` can be `null` (or even an empty object `{}`), the state will remain the same right? shouldn't we just return instead of creating a copy and setting it as a new state?

```
{...{a: 1}, ...null}
{a: 1}
```

NAB because I have no idea of what DevTools.ts is and maybe this matters very little

lib/OnyxUtils.ts
```diff
@@ -238,7 +242,7 @@ function getAllKeys() {
 
     // When a value retrieving task for all keys is still running hook to it
     if (cache.hasPendingTask(taskName)) {
-        return cache.getTaskPromise(taskName);
+        return cache.getTaskPromise(taskName) as Promise<Set<OnyxKey>>;
```
aldo-expensify, line 245: Same as above

lib/OnyxUtils.ts
```diff
@@ -248,78 +252,54 @@ function getAllKeys() {
         return cache.getAllKeys();
     });
 
-    return cache.captureTask(taskName, promise);
+    return cache.captureTask(taskName, promise) as Promise<Set<OnyxKey>>;
```
aldo-expensify, line 255: Why did we cast this to `Promise<Set<OnyxKey>>`, isn't the right type `Promise<OnyxValue<OnyxKey>>`?

lib/OnyxUtils.ts
```diff
@@ -248,78 +252,54 @@ function getAllKeys() {
...
     return onyxCollectionKeyMap.has(key);
 }
 
-/**
- * @param {String} collectionKey
- * @param {String} key
- * @returns {Boolean}
- */
-function isCollectionMemberKey(collectionKey, key) {
+function isCollectionMemberKey<TCollectionKey extends CollectionKeyBase>(collectionKey: TCollectionKey, key: string): key is `${TCollectionKey}${string}` {
```
aldo-expensify, line 266: NAB: I see that `CollectionKeyBase` is just `string`: 

image


Is it supposed to be narrowed down later?

aldo-expensify (changes requested):
Overall looks good to me, I just left some comments regarding casts that I think we should avoid if possible

lib/OnyxUtils.ts
aldo-expensify: Ideally we should fix the types in my opinion, and if for some reason that is not possible, I think we should leave a comment on why it is ok to cast here

lib/OnyxUtils.ts
blazejkustra: It's a string in react-native-onyx but later it's narrowed down in the end repo (e.g. NewDot)

lib/useOnyx.ts
blazejkustra: It is a standard practice to use assertion in library code, TS isn't ideal and sometimes we have to use either assertions or ts-expect-error comment.

Here it is necessary to use an assertion due to how connect method is defined:
```ts
function connect<TKey extends OnyxKey>(connectOptions: ConnectOptions<TKey>): number {

👇 

type ConnectOptions<TKey extends OnyxKey> = (CollectionConnectOptions<TKey> | DefaultConnectOptions<TKey>) & (BaseConnectOptions | WithOnyxConnectOptions<TKey>);

👇 

type DefaultConnectOptions<TKey extends OnyxKey> = {
    key: TKey;
    callback?: DefaultConnectCallback<TKey>;
    waitForCollectionCallback?: false;
};

/** Represents the callback function used in `Onyx.connect()` method with a collection key. */
type CollectionConnectOptions<TKey extends OnyxKey> = {
    key: TKey extends CollectionKeyBase ? TKey : never;
    callback?: CollectionConnectCallback<TKey>;
    waitForCollectionCallback: true;
};
```

Because we use `waitForCollectionCallback`, Typescript assumes that the key is `CollectionKeyBase` 👉 `waitForCollectionCallback` has to be true in this case. I tried different approaches to this but wasn't able to do it without assertion.

The only idea that comes to my mind is to change the types in `DefaultConnectOptions` and `CollectionConnectOptions` so that waitForCollectionCallback is boolean in both cases:
```ts
/** Represents the callback function used in `Onyx.connect()` method with a regular key. */
type DefaultConnectOptions<TKey extends OnyxKey> = {
    key: TKey;
    callback?: DefaultConnectCallback<TKey>;
    waitForCollectionCallback?: boolean;
};

/** Represents the callback function used in `Onyx.connect()` method with a collection key. */
type CollectionConnectOptions<TKey extends OnyxKey> = {
    key: TKey extends CollectionKeyBase ? TKey : never;
    callback?: CollectionConnectCallback<TKey>;
    waitForCollectionCallback?: boolean;
};
```

lib/OnyxUtils.ts
blazejkustra: That's actually a really cool idea! What do you think about doing it as a follow-up PR to avoid bloating this PR any further?

lib/useOnyx.ts
aldo-expensify: I appreciate the explanation, and I'm not sure really if that would be better. I tried playing around adding a super type for `DefaultConnectOptions` and `CollectionConnectOptions`, but couldn't make it work as I wanted.

I think the types that you defined will work well when we use `Onyx.connect` from the App, so the assertion needed here is a minor detail.

aldo-expensify: Also, I learned here that in typescript this is called `assertion` as you say and not a `cast` as a I said before :P

lib/OnyxUtils.ts
aldo-expensify: Also playing with `ConnectOptions<TKey extends OnyxKey>` I see that it does matter if we use `CollectionKeyBase` or `Key` even if visual code just says `string`

lib/OnyxUtils.ts
aldo-expensify: Sounds good to me!

lib/types.ts
```diff
@@ -265,44 +265,61 @@ type Collection<TKey extends CollectionKeyBase, TMap, TValue> = {
...
+    callback?: DefaultConnectCallback<TKey>;
+    waitForCollectionCallback?: false;
+};
+
+/** Represents the callback function used in `Onyx.connect()` method with a collection key. */
+type CollectionConnectOptions<TKey extends OnyxKey> = {
+    key: TKey extends CollectionKeyBase ? TKey : never;
+    callback?: CollectionConnectCallback<TKey>;
+    waitForCollectionCallback: true;
+};
```
aldo-expensify, line 303: Last comment/question: Wondering why these types were not defined like this:
```suggestion
type DefaultConnectOptions<TKey extends Key> = {
    key: TKey;
    callback?: DefaultConnectCallback<TKey>;
    waitForCollectionCallback?: false;
};

/** Represents the callback function used in `Onyx.connect()` method with a collection key. */
type CollectionConnectOptions<TKey extends CollectionKeyBase> = {
    key: TKey;
    callback?: CollectionConnectCallback<TKey>;
    waitForCollectionCallback: true;
};
```

`DefaultConnectOptions<TKey extends OnyxKey>` feels a bit wrong since `OnyxKey` can be a collection key, then the `waitForCollectionCallback: false` would be wrong for such key. Also `CollectionConnectOptions<TKey extends CollectionKeyBase> ` allows you to do `key: TKey` instead of the more complex `TKey extends CollectionKeyBase ? TKey : never;`.

Do you see any downside of specifying these types like this?

aldo-expensify (commented):
Left a comment

lib/useOnyx.ts
blazejkustra: Interesting, isn't cast and assertion the same thing? 😮

aldo-expensify: > Interesting, isn't cast and assertion the same thing? 😮

Honestly, not sure, maybe it depends on the language? You made me doubt because in languages like C++ / java, casting happens at runtime and may fail, here it is different since these types is only used when building and since you called it `assertion`, I thought I was wrong about calling it a `cast`. 🤷

lib/types.ts
blazejkustra: I was like, the type looks weird indeed but it got me thinking that it was defined like this for a reason 😅 

I checked in NewDot, this type works until you use it with a key (not a collection key) and add `waitForCollectionCallback: true`:
```ts
let allCards: OnyxValues[typeof ONYXKEYS.CARD_LIST] = {};
Onyx.connect({
    key: ONYXKEYS.CARD_LIST,
    callback: (val) => {
        if (!val || Object.keys(val).length === 0) {
            return;
        }

        allCards = val;
    },
});
```

It makes `val: OnyxCollection<CardList>` while it should be `val: OnyxEntry<CardList>`. With the old type errors properly: `Type 'true' is not assignable to type 'false'`

aldo-expensify: Oh, interesting, wouldn't that mean that `ONYXKEYS.CARD_LIST` is a valid `CollectionKeyBase` when it shouldn't be?

aldo-expensify: In you example, if things work properly, I would expect two problems:

- The callback takes as param `val: OnyxCollection<CardList>`, and
- `key: ONYXKEYS.CARD_LIST` is wrong because `key` must be a `CollectionKeyBase` and is not

blazejkustra: > Oh, interesting, wouldn't that mean that ONYXKEYS.CARD_LIST is a valid CollectionKeyBase when it shouldn't be?

I checked it in ONYXKEYS.ts in NewDot and `ONYXKEYS.CARD_LIST` is not a valid `CollectionKeyBase`:
```ts
const cardList1: OnyxCollectionKey = 'cardList'; // errors
const cardList2: OnyxValueKey = 'cardList'; // works fine
```

For context, `OnyxCollectionKey` and `OnyxValueKey` are used in `src/types/modules/react-native-onyx.d.ts` to augment onyx keys:
```ts
declare module 'react-native-onyx' {
    // eslint-disable-next-line @typescript-eslint/consistent-type-definitions
    interface CustomTypeOptions {
        keys: OnyxValueKey | OnyxFormKey | OnyxFormDraftKey;
        collectionKeys: OnyxCollectionKey;
        values: OnyxValues;
    }
}
```

This way keys, and collectionKeys are no longer `string` but instead are union of [literal strings](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#literal-types)

blazejkustra: > In your example, if things worked properly, I would have expected one of these two errors:
>
> key: ONYXKEYS.CARD_LIST is wrong because key must be a CollectionKeyBase and is not, or
waitForCollectionCallback: true has an error because ONYXKEYS.CARD_LIST is a Key and therefore, waitForCollectionCallback must be false or undefined

As you already noted, all `Key`, `OnyxKey` and `CollectionKeyBase` are of type `string` in react-native-onyx. It's later augmented in the end-user repo like NewDot, that's why it is so tricky to understand. 

While [this](https://github.com/Expensify/react-native-onyx/pull/523#discussion_r1566273457) seems like a great idea, it comes with a problem. It does work until all `Key`, `OnyxKey` and `CollectionKeyBase` are of type `string`, but once we augment these values through `src/types/modules/react-native-onyx.d.ts` file, `ConnectOptions` type breaks:
```ts
type Key = 'account' | 'isSidebarLoaded' ;
type CollectionKeyBase = 'report_';
type OnyxKey = Key | CollectionKey;

type DefaultConnectOptions<TKey extends Key> = {
    key: TKey;
    callback?: DefaultConnectCallback<TKey>;
    waitForCollectionCallback?: false;
};

type CollectionConnectOptions<TKey extends CollectionKeyBase> = {
    key: TKey;
    callback?: CollectionConnectCallback<TKey>;
    waitForCollectionCallback?: true;
};

type ConnectOptions<TKey extends OnyxKey> = CollectionConnectOptions<TKey> | DefaultConnectOptions<TKey>; 
// Type 'TKey' does not satisfy the constraint 'Key'.
  // Type 'OnyxKey' is not assignable to type 'Key'.
``` 

That is why TS struggles to show an error [here](https://github.com/Expensify/react-native-onyx/pull/523#discussion_r1566360988).

blazejkustra: I think we should leave `ConnectOptions` as is, @fabioh8010 maybe you have some idea as you created this type?

aldo-expensify: I see, well, thanks a lot for trying and following up with this. Meanwhile, I'll merge since the PR leaves us in a better place anyway and I don't want to keep blocking here. If you come up with some improvements later, just lets do them in a follow up PR.

aldo-expensify: This is the first time I see type "augmentation" in typescript, pretty cool feature!

aldo-expensify (approved):

lib/types.ts
fabioh8010: I think is better to keep the way it is for now, this was the best solution we found at that time to add this type safety around `waitForCollectionCallback` and collection keys.

lib/OnyxUtils.ts
```diff
@@ -1105,22 +1005,21 @@ function prepareKeyValuePairsForStorage(data) {
...
+function applyMerge(existingValue: OnyxValue<OnyxKey>, changes: Array<OnyxValue<OnyxKey>>, shouldRemoveNullObjectValues: boolean): OnyxValue<OnyxKey> {
+    const lastChange = changes?.at(-1);
 
-    if (_.isArray(lastChange)) {
+    if (Array.isArray(lastChange)) {
         return lastChange;
     }
 
-    if (_.some(changes, _.isObject)) {
+    if (changes.some((change) => typeof change === 'object')) {
```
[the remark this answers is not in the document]
CortneyOfstad: Any update? Thanks!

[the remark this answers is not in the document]
blazejkustra: Hey! It was addressed here: https://github.com/Expensify/react-native-onyx/pull/533

PR #526, opened 2024-03-28
Fix: (Nested) null values will never be applied to SQLite when batched in Onyx

chrispader:
@tgolen 

### Details

When we batch queued changes in `Onyx.merge` and they contain (nested) null values, SQLite will never be able to apply the individual deltas. Therefore, when we pass `null` or a nested `null` in an object, SQLite will not delete the (nested) keys from storage.

Imagine having something like this:

```javascript
const initialData = {
    a: 'a',
    b: 'b',
    c: {
        d: 'd',
        e: 'e',
    },
};
const change1 = {
    b: null
}
const change2 = null;
const change3 = {
    f: 'f',
    c: {
        g: 'g',
        h: 'h',
    },
};
const change4 = {
    c: {
        g: null,
    },
};
const changes = [change1, change2, change3, change4];

const batchedChanges = OnyxUtils.applyMerge(undefined, changes, false);
```

This will result in the following `batchedChanges`:

```javascript
{
    f: 'f',
    c: {
        g: 'g',
    },
}
```

When we then apply these batched changes in the native storage layer (SQLite) with `JSON_PATCH`, the initial data will not be erased and keys "a", "b", "d" and "e" will not be removed either, because SQLite never applied the top-level and nested `null`'s.

The expected value in storage after applying the batched changes should be:

```js
{
    f: 'f',
    c: {
        h: 'h',
    },
}
```

Whereas the actual value will be:

```js
{
    a: 'a',
    b: 'b',
    c: {
        d: 'd',
        e: 'e',
        h: 'h'
    },
    f: 'f'
};
```

Generally, when SQLite sees a top-level or nested null value, it will remove this (nested) key from the database. But since we already batched the changes beforehand and don't pass every delta from the queue individually.

Cases that will prevent SQLite from removing the key from storage:

- any top-level `null` will not affect SQLite, if there's a change after, because the `batchedChanges` will get merged with the last values
- any nested `null` before a `top-level` null will be cancelled out, since the top-level `null` just clears the batched changes

# Notes multiple follow-up PRs

## Creating a SQLite mock in `react-native-quick-sqlite` 

I'm going to fix this issue in this PR. Additionally, i'm thinking about creating a in-memory or local SQLite mock and ship it with https://github.com/margelo/react-native-quick-sqlite. The mock could then either run in-memory directly - if possible - or we would need to spin up a local SQLite database.

## Creating storage provider specific tests (with actual mocks)

With this mock we could add more native-specific tests in `react-native-onyx` and potentially improve other tests so that we test for all platforms.

### Related Issues

GH_LINK

### Author Checklist

- [ ] I linked the correct issue in the `### Related Issues` section above
    - [ ] Android / native
    - [ ] Android / Chrome
    - [ ] iOS / native
    - [ ] iOS / Safari
    - [ ] MacOS / Chrome / Safari
    - [ ] MacOS / Desktop

### Screenshots/Videos
<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>


tgolen:
Interesting issue! I'm not sure I fully understand it, even after reading it a couple of times, but as long as you have some tests that prove the bug and the fix, then I'm good with it.

As for the SQLite mock, I think it would be ideal to be able to spin up an actual SQLite database and run tests directly against it. That would give us extremely accurate and reliable tests. What would need to happen to have that?

chrispader:
> Interesting issue! I'm not sure I fully understand it, even after reading it a couple of times, but as long as you have some tests that prove the bug and the fix, then I'm good with it.



> As for the SQLite mock, I think it would be ideal to be able to spin up an actual SQLite database and run tests directly against it. That would give us extremely accurate and reliable tests. What would need to happen to have that?

The idea about an in-memory SQLite database is as in https://www.sqlite.org/inmemorydb.html. It's possible to temporarily store the database only in-memory and this seems more than sufficient for testing/mocking purposes. If this is possible, i think we should prefer this, instead of relying on a local SQLite database and installation.

I'm gonna fix this actual issue before i potentially start work on a mock, as the mock isn't really high-priority right now. Basically we would just ship the mock in `react-native-quick-sqlite` and then we would need to setup the mock in `jestSetup.js`, same as the other mocks.

tgolen:
Ah, I see. So the in memory database would still use the exact same SQLite
interface? If so, then I agree that would be fine.

chrispader:
> Ah, I see. So the in memory database would still use the exact same SQLite
> 
> interface? If so, then I agree that would be fine.

yes 👍

mountiny:
@chrispader Lint is failing

lib/storage/providers/IDBKeyValProvider.ts
```diff
@@ -40,9 +40,9 @@ const provider: StorageProvider = {
                 return Promise.all(upsertMany);
             });
         }),
-    mergeItem(key, _changes, modifiedData) {
+    mergeItem(key, _deltaChanges, preMergedValue, _shouldSetValue) {
```
mountiny, line 43: So we need the `_shouldSetValue` here since its not used?

lib/storage/providers/types.ts
```diff
@@ -43,10 +43,10 @@ type StorageProvider = {
 
     /**
      * Merges an existing value with a new one by leveraging JSON_PATCH
-     * @param changes - the delta for a specific key
-     * @param modifiedData - the pre-merged data from `Onyx.applyMerge`
+     * @param deltaChanges - the delta for a specific key
+     * @param preMergedValue - the pre-merged data from `Onyx.applyMerge`
```
mountiny, line 47: San you also add the `shouldSetValue ` param here?

mountiny (commented):
@chrispader Changes look good to me, one comment, is there a way to test these in App now?

chrispader:
@mountiny addressed your comments.

> @chrispader Changes look good to me, one comment, is there a way to test these in App now?

There isn't really a use-case in which you can test exactly this fixed behaviour, since at the time this only occurs in `SQLite` on native platforms. Therefore, in [this test PR](https://github.com/Expensify/App/pull/39802) i do some Onyx operations and then fetch the value from storage directly instead of using something like `withOnyx`, `Onyx.connect` or `Onyx.get`, since for these functions the cache would be used first...

lib/Onyx.ts
```diff
@@ -303,48 +303,43 @@ function merge<TKey extends OnyxKey>(key: TKey, changes: OnyxEntry<NullishDeep<K
...
         try {
             // We first only merge the changes, so we can provide these to the native implementation (SQLite uses only delta changes in "JSON_PATCH" to merge)
-            // We don't want to remove null values from the "batchedChanges", because SQLite uses them to remove keys from storage natively.
-            let batchedChanges = OnyxUtils.applyMerge(undefined, mergeQueue[key], false);
+            // We don't want to remove null values from the "batchedDeltaChanges", because SQLite uses them to remove keys from storage natively.
+            const batchedDeltaChanges = OnyxUtils.applyMerge(undefined, mergeQueue[key], false);
 
-            // The presence of a `null` in the merge queue instructs us to drop the existing value.
+            // Case (1): When there is no existing value in storage, we want to set the value instead of merge it.
+            // Case (2): The presence of a top-level`null` in the merge queue instructs us to drop the whole existing value.
```
mountiny, line 310:
```suggestion
            // Case (2): The presence of a top-level `null` in the merge queue instructs us to drop the whole existing value.
```

mountiny (dismissed):
@chrispader thanks! I have spotted one little typo before merging then

mountiny (approved):
Thanks

PR #534, opened 2024-04-17
useOnyx type improvements

fabioh8010:
### Details

This PR does the following changes:

- Changes `useOnyx` to return `undefined` instead of `null`, [thus allowing us to use default values more easily](https://github.com/Expensify/App/issues/34339#issuecomment-2057486233).
- Changes `OnyxEntry` and `OnyxCollection` types to allow `undefined` values (necessary because of the change above). 
- Changes `Onyx.set` and `Onyx.merge` types to disallow passing `undefined` directly (like it was before).
- Changes `Onyx.connect` callback's value type to exclude `undefined` (`undefined` will never be passed).
- Deprecates `withOnyx`.
- Makes `useOnyx` return read-only value to avoid future situations [like this one](https://github.com/Expensify/App/pull/36420#issuecomment-2058973502) (it's not committed yet but the problem was because an Onyx value was being mutated during runtime, which causes `getSnapshot()` to always fail the deep-equality check and re-render infinitely).
- Fix `initialValue` type to avoid infer `useOnyx`'s return value
- Bump TS to 5.4.5 in order to be able to use [NoInfer type](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-4.html#the-noinfer-utility-type) to fix `initialValue`.

**⚠️ Follow-up PR on E/App when these changes are available there:** https://github.com/Expensify/App/pull/40367

**⚠️ Playground branch:** https://github.com/fabioh8010/expensify-app/tree/feature/playground-use-onyx

### Related Issues

$ https://github.com/Expensify/App/issues/34339

### Manual Tests

Some common tasks on E/App:

1. Start a new chat and search for someone (uses `useOnyx`).
2. Send a message to someone.
3. Open someone's profile and change settings (uses `useOnyx`).
4. Open Preferences and change settings (uses `useOnyx`).

### Author Checklist

    - [x] Android / native
    - [x] Android / Chrome
    - [x] iOS / native
    - [x] iOS / Safari
    - [x] MacOS / Chrome / Safari
    - [x] MacOS / Desktop

### Screenshots/Videos
<details>

I'm having issue to build Android app.

</details>

<details>

https://github.com/Expensify/react-native-onyx/assets/20051562/b6c57db7-c4f2-42fb-ac4c-88d14ed10b29

</details>

<details>

It's failing runtime due to changes on this repo's `main`.

Simulator Screenshot - iPhone 15 Pro Max - 2024-04-26 at 15 52 16

</details>

<details>

https://github.com/Expensify/react-native-onyx/assets/20051562/f8f779c4-7c7e-4589-8925-cd342327e6d9

</details>

<details>

https://github.com/Expensify/react-native-onyx/assets/20051562/97460201-260d-42aa-aebc-1a78f5f931bf

https://github.com/Expensify/react-native-onyx/assets/20051562/7b8aadf1-11fc-47f4-8b43-798541542887

</details>

<details>

https://github.com/Expensify/react-native-onyx/assets/20051562/d67d7f58-a492-42cc-8258-4f7144757aa9

</details>


tests/unit/useOnyxTest.ts
```diff
@@ -102,7 +102,7 @@ describe('useOnyx', () => {
 
             const {result} = renderHook(() => useOnyx(ONYXKEYS.TEST_KEY));
 
-            expect(result.current[0]).toEqual(null);
+            expect(result.current[0]).toEqual(undefined);
```
rezkiy37, line 105: You can use this method instead of those `toEqual`.
```suggestion
            expect(result.current[0]). toBeUndefined;
```

fabioh8010:
@roryabraham @blazejkustra I've also upgraded both Onyx and E/App TS to 5.4.5 to be able to use [NoInfer type](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-4.html#the-noinfer-utility-type), necessary to fix `initialValue` issues with TS.

lib/withOnyx.d.ts
```diff
@@ -125,6 +125,8 @@ type OnyxPropCollectionMapping<TComponentProps, TOnyxProps, TOnyxProp extends ke
 }[CollectionKeyBase];
 
 /**
+ * @deprecated Use `useOnyx` instead of `withOnyx` whenever possible.
```
blazejkustra, line 128: Let's go! 🚀

lib/types.ts
```diff
@@ -168,10 +178,10 @@ type Selector<TKey extends OnyxKey, TOnyxProps, TReturnType> = (value: OnyxEntry
  * })(Component);
  * ```
  */
-type OnyxEntry<TOnyxValue> = TOnyxValue | null;
+type OnyxEntry<TOnyxValue> = TOnyxValue | null | undefined;
```
blazejkustra, line 181: That is a big change for `withOnyx` and other Onyx methods in terms of typings. Does this type (and `OnyxCollection`) work well in NewDot?

lib/types.ts
```diff
@@ -3,6 +3,16 @@ import type {Merge} from 'type-fest';
...
+/**
+ * Utility type that excludes `null` from the type `TValue`.
+ */
+type NonNull<TValue> = TValue extends null ? never : TValue;
+
+/**
+ * Utility type that excludes `undefined` from the type `TValue`.
+ */
+type NonUndefined<TValue> = TValue extends undefined ? never : TValue;
+
```
blazejkustra, line 15: Would [NonNullable](https://www.typescriptlang.org/docs/handbook/utility-types.html#nonnullabletype) work here?

blazejkustra (changes requested):

lib/types.ts
fabioh8010: `NonNullable` excludes both `undefined` and `null`, which is not what I want here 😕

lib/types.ts
fabioh8010: @blazejkustra Yeah I know, I tried to do different approaches but at the end they looked worse than this one, it will require some changes in E/App but it would be the one that makes more sense imo. Please have a look at this PR where I apply the changes -> https://github.com/Expensify/App/pull/40367

blazejkustra (approved):

roryabraham (commented):
This code seems good to me

fabioh8010:
Putting this on HOLD because I'm going to be OOO next week, and will return on May 6th.

fabioh8010:
@roryabraham @techievivek I'm taking this PR off the hold.

roryabraham (approved):

PR #561, opened 2024-06-10
More minor fixes for the Onyx bump in E/App

chrispader:
@mountiny @s77rt @ikevin127 

### Details

This PR is adding more fixes to Onyx which are needed to resolve issues in the Onyx bump PR in E/App.

Fixes:

- `useOnyx` will never resolve loading state if the value from cache is `undefined` (because of `deepEqual` in `getSnapshot`)
- Update tests to reflect the new behavior of Onyx setting the cache. When we call Onyx (setter) methods this will set the value in cache synchronously (before storage promises are resolved). Therefore, we must first check for `initialValue` being set, before we can call these methods.

### Related Issues

https://github.com/Expensify/App/pull/42772

### Author Checklist

    - [x] Android / native
    - [x] Android / Chrome
    - [x] iOS / native
    - [x] iOS / Safari
    - [x] MacOS / Chrome / Safari
    - [x] MacOS / Desktop

### Screenshots/Videos
<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

### DIFF
--- a/lib/useOnyx.ts
+++ b/lib/useOnyx.ts
@@ -118,7 +118,7 @@ function useOnyx<TKey extends OnyxKey, TReturnValue = OnyxValue<TKey>>(key: TKey
     const getSnapshot = useCallback(() => {
         // We get the value from the cache, supplying a selector too in case it's defined.
         // If `newValue` is `undefined` it means that the cache doesn't have a value for that key yet.
-        // If `newValue` is `null` or any other value if means that the cache does have a value for that key.
+        // If `newValue` is `null` or any other value it means that the cache does have a value for that key.
         // This difference between `undefined` and other values is crucial and it's used to address the following
         // conditions and use cases.
         let newValue = getCachedValue<TKey, TReturnValue>(key, selectorRef.current);
--- a/lib/useOnyx.ts
+++ b/lib/useOnyx.ts
@@ -131,11 +131,6 @@ function useOnyx<TKey extends OnyxKey, TReturnValue = OnyxValue<TKey>>(key: TKey
             newFetchStatus = 'loading';
         }
 
-        // If `options.initWithStoredValues` is explicitly set to `false`, the fetch status should always be `loaded`.
-        if (options?.initWithStoredValues === false) {
-            newFetchStatus = 'loaded';
-        }
-
         // If data is not present in cache and `initialValue` is set during the first connection,
         // we set the new value to `initialValue` and fetch status to `loaded` since we already have some data to return to the consumer.
         if (isFirstConnectionRef.current && !hasCacheForKey && options?.initialValue !== null && options?.initialValue !== undefined) {
--- a/lib/useOnyx.ts
+++ b/lib/useOnyx.ts
@@ -6,6 +6,7 @@ import type {CollectionKeyBase, OnyxCollection, OnyxKey, OnyxValue, Selector} fr
 import useLiveRef from './useLiveRef';
 import usePrevious from './usePrevious';
 import Onyx from './Onyx';
+import cache from './OnyxCache';
 
 type BaseUseOnyxOptions = {
     /**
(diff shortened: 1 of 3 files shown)


lib/useOnyx.ts
```diff
@@ -133,24 +131,29 @@ function useOnyx<TKey extends OnyxKey, TReturnValue = OnyxValue<TKey>>(key: TKey
             newFetchStatus = 'loading';
         }
 
-        // If data is not present in cache (if it's `undefined`) and `initialValue` is set during the first connection,
+        // If `options.initWithStoredValues` is explicitly set to `false`, the fetch status should always be `loaded`.
+        if (options?.initWithStoredValues === false) {
```
fabioh8010, line 135: Is this only applicable during the first connection or all the time? If only the first time we should use `isFirstConnectionRef.current`.

```ts
        if (isFirstConnectionRef.current && options?.initWithStoredValues === false) {
```

lib/useOnyx.ts
```diff
@@ -133,24 +131,29 @@ function useOnyx<TKey extends OnyxKey, TReturnValue = OnyxValue<TKey>>(key: TKey
...
+            newFetchStatus = 'loaded';
+        }
+
+        // If data is not present in cache and `initialValue` is set during the first connection,
         // we set the new value to `initialValue` and fetch status to `loaded` since we already have some data to return to the consumer.
-        if (isFirstConnectionRef.current && newValue === undefined && options?.initialValue !== undefined) {
-            newValue = options?.initialValue as CachedValue<TKey, TReturnValue>;
+        if (isFirstConnectionRef.current && !hasCacheForKey && options?.initialValue !== null && options?.initialValue !== undefined) {
```
fabioh8010, line 141:
```suggestion
        if (isFirstConnectionRef.current && !hasCacheForKey && options?.initialValue !== undefined) {
```

According to the types we shouldn't be able to pass `null` to `initialValue`

tests/unit/useOnyxTest.ts
```diff
@@ -257,6 +255,8 @@ describe('useOnyx', () => {
             expect(result.current[0]).toEqual('initial value');
             expect(result.current[1].status).toEqual('loaded');
 
+            Onyx.set(ONYXKEYS.TEST_KEY, 'test_id_1');
```
fabioh8010, line 258: Just curious: Why did you move the operation to here? Shouldn't we have tests for both cases?

tests/unit/useOnyxTest.ts
```diff
@@ -335,6 +329,12 @@ describe('useOnyx', () => {
             expect(result.current[0]).toEqual('initial value');
             expect(result.current[1].status).toEqual('loaded');
 
+            Onyx.set(ONYXKEYS.TEST_KEY, 'test1');
```
fabioh8010, line 332: Same: Why did you move the operation to here? Shouldn't we have tests for both cases?

lib/useOnyx.ts
chrispader: we can actually remove this.

tests/unit/useOnyxTest.ts
chrispader: The moment we use `Onyx.set` or `Onyx.merge` (or any setter function) it will store the value in cache synchronously and update the `storageKeys` Set in `OnyxCache`.\

If we set the key before asserting for the `initialValue`, we will always get `undefined`, because there IS a value in cache already, even though the storage promise hasn't resolved yet.

tests/unit/useOnyxTest.ts
chrispader: Same as above

tests/unit/useOnyxTest.ts
fabioh8010: Thanks for the explanation!

fabioh8010 (approved):

mountiny (dismissed):
Just a NAB though seems like we are good to go

mountiny (approved):
Merging this to get it incorporated to the onyx bump PR per discussion here https://expensify.slack.com/archives/C05R2V5GM1S/p1718119019498549?thread_ts=1717193187.147939&cid=C05R2V5GM1S

PR #563, opened 2024-06-20
perf: improve native apps startup

hurali97:
### Details

The changes in the PR are part of improving the app startup which was profiled for the native Apps. The main change is explained below, the other changes are self explanatory.

<hr/> 
Prior to the following change, we had 2.4 seconds being consumed in the loop. We can improve this by doing a `multiGet` and then doing a loop over returned keys. This reduces the execution time to 250 ms, saving us 2 seconds.

Screenshot 2024-06-20 at 6 12 54 PM

Before:
Screenshot 2024-06-20 at 6 17 19 PM

After:
Screenshot 2024-06-20 at 6 18 03 PM

<hr/> 

### Related Issues

https://github.com/Expensify/App/issues/43746

### Author Checklist

    - [X] Android / native
    - [X] Android / Chrome
    - [X] iOS / native
    - [X] iOS / Safari
    - [X] MacOS / Chrome / Safari
    - [X] MacOS / Desktop

### Screenshots/Videos
<details>

https://github.com/Expensify/react-native-onyx/assets/47336142/a0e2ff33-eb62-45ad-8aed-fb2bddcb18ea

</details>

<details>

https://github.com/Expensify/react-native-onyx/assets/47336142/16ab1920-f82f-4b1e-8160-7594bddaa272

</details>

<details>

https://github.com/Expensify/react-native-onyx/assets/47336142/a215f721-a281-4937-8ba9-931404a353bd

</details>

<details>

https://github.com/Expensify/react-native-onyx/assets/47336142/7f24c6fa-ac8c-482e-842a-1c5f7d778ff9

</details>

<details>

https://github.com/Expensify/react-native-onyx/assets/47336142/728ef8b3-86c1-404f-8945-276258902834

</details>

<details>

https://github.com/Expensify/react-native-onyx/assets/47336142/983b0783-5836-460b-8c26-e5716542f9b9

</details>

### DIFF
--- a/lib/Onyx.ts
+++ b/lib/Onyx.ts
@@ -156,7 +156,7 @@ function connect<TKey extends OnyxKey>(connectOptions: ConnectOptions<TKey>): nu
 
                     // We did not opt into using waitForCollectionCallback mode so the callback is called for every matching key.
                     OnyxUtils.multiGet(matchingKeys).then((values) => {
-                        values.forEach(([key, val]) => {
+                        values.forEach((val, key) => {
                             OnyxUtils.sendDataToConnection(mapping, val as OnyxValue<TKey>, key as TKey, true);
                         });
                     });
--- a/lib/Onyx.ts
+++ b/lib/Onyx.ts
@@ -438,7 +438,8 @@ function mergeCollection<TKey extends CollectionKeyBase, TMap>(collectionKey: TK
 
     // Confirm all the collection keys belong to the same parent
     let hasCollectionKeyCheckFailed = false;
-    Object.keys(mergedCollection).forEach((dataKey) => {
+    const mergedCollectionKeys = Object.keys(mergedCollection);
+    mergedCollectionKeys.forEach((dataKey) => {
         if (OnyxUtils.isKeyMatch(collectionKey, dataKey)) {
             return;
         }
@@ -459,7 +460,7 @@ function mergeCollection<TKey extends CollectionKeyBase, TMap>(collectionKey: TK
     return OnyxUtils.getAllKeys()
         .then((persistedKeys) => {
             // Split to keys that exist in storage and keys that don't
-            const keys = Object.keys(mergedCollection).filter((key) => {
+            const keys = mergedCollectionKeys.filter((key) => {
                 if (mergedCollection[key] === null) {
                     OnyxUtils.remove(key);
                     return false;
@@ -470,8 +471,6 @@ function mergeCollection<TKey extends CollectionKeyBase, TMap>(collectionKey: TK
             const existingKeys = keys.filter((key) => persistedKeys.has(key));
 
             const cachedCollectionForExistingKeys = OnyxUtils.getCachedCollection(collectionKey, existingKeys);
-
-            const newKeys = keys.filter((key) => !persistedKeys.has(key));
 
             const existingKeyCollection = existingKeys.reduce((obj: OnyxInputKeyValueMapping, key) => {
                 const {isCompatible, existingValueType, newValueType} = utils.checkCompatibilityWithExistingValue(mergedCollection[key], cachedCollectionForExistingKeys[key]);
@@ -484,11 +483,13 @@ function mergeCollection<TKey extends CollectionKeyBase, TMap>(collectionKey: TK
                 return obj;
             }, {}) as Record<OnyxKey, OnyxInput<TKey>>;
 
-            const newCollection = newKeys.reduce((obj: OnyxInputKeyValueMapping, key) => {
-                // eslint-disable-next-line no-param-reassign
-                obj[key] = mergedCollection[key];
-                return obj;
-            }, {}) as Record<OnyxKey, OnyxInput<TKey>>;
+            const newCollection: Record<OnyxKey, OnyxInput<TKey>> = {};
+            keys.forEach((key) => {
+                if (persistedKeys.has(key)) {
+                    return;
+                }
+                newCollection[key] = mergedCollection[key];
+            });
 
             // When (multi-)merging the values with the existing values in storage,
             // we don't want to remove nested null values from the data that we pass to the storage layer,
--- a/lib/Onyx.ts
+++ b/lib/Onyx.ts
@@ -155,10 +155,11 @@ function connect<TKey extends OnyxKey>(connectOptions: ConnectOptions<TKey>): nu
                     }
 
                     // We did not opt into using waitForCollectionCallback mode so the callback is called for every matching key.
-                    // eslint-disable-next-line @typescript-eslint/prefer-for-of
-                    for (let i = 0; i < matchingKeys.length; i++) {
-                        OnyxUtils.get(matchingKeys[i]).then((val) => OnyxUtils.sendDataToConnection(mapping, val as OnyxValue<TKey>, matchingKeys[i] as TKey, true));
-                    }
+                    Storage.multiGet(matchingKeys).then((values) => {
+                        values.forEach(([key, val]) => {
+                            OnyxUtils.sendDataToConnection(mapping, val as OnyxValue<TKey>, key as TKey, true);
+                        });
+                    });
                     return;
                 }
 
--- a/lib/Onyx.ts
+++ b/lib/Onyx.ts
@@ -155,7 +155,7 @@ function connect<TKey extends OnyxKey>(connectOptions: ConnectOptions<TKey>): nu
                     }
 
                     // We did not opt into using waitForCollectionCallback mode so the callback is called for every matching key.
-                    Storage.multiGet(matchingKeys).then((values) => {
+                    OnyxUtils.multiGet(matchingKeys).then((values) => {
                         values.forEach(([key, val]) => {
                             OnyxUtils.sendDataToConnection(mapping, val as OnyxValue<TKey>, key as TKey, true);
                         });
--- a/lib/OnyxUtils.ts
+++ b/lib/OnyxUtils.ts
@@ -249,10 +249,13 @@ function get<TKey extends OnyxKey, TValue extends OnyxValue<TKey>>(key: TKey): P
 function multiGet<TKey extends OnyxKey>(keys: CollectionKeyBase[]): Promise<Map<OnyxKey, OnyxValue<TKey>>> {
     // Keys that are not in the cache
     const missingKeys: OnyxKey[] = [];
+
     // Tasks that are pending
     const pendingTasks: Array<Promise<OnyxValue<TKey>>> = [];
+
     // Keys for the tasks that are pending
     const pendingKeys: OnyxKey[] = [];
+
     // Data to be sent back to the invoker
     const dataMap = new Map<OnyxKey, OnyxValue<TKey>>();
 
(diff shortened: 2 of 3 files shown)


lib/Onyx.ts
```diff
@@ -155,10 +155,11 @@ function connect<TKey extends OnyxKey>(connectOptions: ConnectOptions<TKey>): nu
...
                     // We did not opt into using waitForCollectionCallback mode so the callback is called for every matching key.
-                    // eslint-disable-next-line @typescript-eslint/prefer-for-of
-                    for (let i = 0; i < matchingKeys.length; i++) {
-                        OnyxUtils.get(matchingKeys[i]).then((val) => OnyxUtils.sendDataToConnection(mapping, val as OnyxValue<TKey>, matchingKeys[i] as TKey, true));
-                    }
+                    Storage.multiGet(matchingKeys).then((values) => {
+                        values.forEach(([key, val]) => {
+                            OnyxUtils.sendDataToConnection(mapping, val as OnyxValue<TKey>, key as TKey, true);
+                        });
+                    });
```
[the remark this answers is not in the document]
chrispader: @hannojg yes i right, we might already have a value in cache for a key that we connect to with `Onyx.connect`.

I agree, we should either create the logic @hannojg described to `OnyxUtils.multiGet` or directly in `Onyx.connect`. Right now we won't make any use of this anywhere else

lib/Onyx.ts
```diff
@@ -484,11 +484,13 @@ function mergeCollection<TKey extends CollectionKeyBase, TMap>(collectionKey: TK
...
-                obj[key] = mergedCollection[key];
-                return obj;
-            }, {}) as Record<OnyxKey, OnyxInput<TKey>>;
+            const newCollection: Record<OnyxKey, OnyxInput<TKey>> = {};
+            keys.forEach((key) => {
+                if (persistedKeys.has(key)) {
+                    return;
+                }
+                newCollection[key] = mergedCollection[key];
+            });
```
chrispader, line 493: Looks good to me, but just curios... is the reason we use `forEach` instead of `reduce` because it's faster or more readable?

lib/Onyx.ts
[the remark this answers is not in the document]
hurali97: @hannojg I have updated the PR description, which explains why this change was important.

>So, maybe we want to see if we can add OnyxUtils.multiGet, which will try to do a multiKey lookup from cache, and only get the keys from storage that weren't found (which would be the fastest I guess)?

I didn't think of it, thanks for flagging this. However, even without going with this approach, we get only ~250ms execution time. Nevertheless, I agree that we should create this functionality.

lib/Onyx.ts
hurali97: @chrispader Only reason here is to avoid 2 loops. First we do a filter to get the `newKeys` and then we do a reduce over them. However, we could do this in one loop as `keys` is a `Set` which gives us a `forEach` helper function.

lib/Onyx.ts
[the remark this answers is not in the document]
hurali97: @hannojg @chrispader 

I have added a function in `OnyxUtils` for `multiGet`, which first looks for the value in the cache and then from the storage if it's not present. In my observation, from the execution time perspective it's more or less the same result we were getting by using `multiGet` directly from the `Storage` 👍

lib/OnyxUtils.ts
```diff
@@ -245,6 +246,69 @@ function get<TKey extends OnyxKey, TValue extends OnyxValue<TKey>>(key: TKey): P
...
+                if (missingKeys.length === 0) {
+                    return Promise.resolve(undefined);
+                }
+
+                return Storage.multiGet(missingKeys);
+            })
+            // We are going to add the data from the missing keys to the data object and also merge it to the cache.
+            .then((values): Promise<KeyValuePairList> => {
+                if (!values || values.length === 0) {
+                    return Promise.resolve(data);
```
hannojg, line 297: hm, confused why we are returning a promise here, it could be just:

```suggestion
            .then((values): KeyValuePairList => {
                if (!values || values.length === 0) {
                    return data;
```

lib/OnyxUtils.ts
```diff
@@ -245,6 +246,69 @@ function get<TKey extends OnyxKey, TValue extends OnyxValue<TKey>>(key: TKey): P
...
+                }
+
+                // temp object is used to merge the missing data into the cache
+                const temp: OnyxCollection<KeyValueMapping[TKey]> = {};
+                values.forEach(([key, value]) => {
+                    data.push([key, value]);
+                    temp[key] = value as OnyxValue<TKey>;
+                });
+                cache.merge(temp);
+                return Promise.resolve(data);
```
hannojg, line 307: same here

neil-marcellini (changes requested):
Looks great overall. I think we can re-use some existing code.

I'm also curious if the for in loops are really helping performance / memory use or if that was just a preference. Disabling eslint is probably fine because I don't think we'll have prototype properties getting in the way, but let's only disable the rule if there's a good reason.

lib/OnyxUtils.ts
```diff
@@ -245,6 +246,69 @@ function get<TKey extends OnyxKey, TValue extends OnyxValue<TKey>>(key: TKey): P
     return cache.captureTask(taskName, promise) as Promise<TValue>;
 }
 
+// multiGet the data first from the cache and then from the storage for the missing keys.
```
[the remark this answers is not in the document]
hurali97: @neil-marcellini They are only different in one case and that is `getCollectionDataAndSendAsObject` generates a map and we need an array in `multiGet`. I will see what parts we can split into smaller functions.

lib/OnyxUtils.ts
chrispader: +1

chrispader (changes requested):
LGTM except for the changes that were already mentioned

hurali97:
> Looks great overall. I think we can re-use some existing code.
> 
> I'm also curious if the for in loops are really helping performance / memory use or if that was just a preference. Disabling eslint is probably fine because I don't think we'll have prototype properties getting in the way, but let's only disable the rule if there's a good reason.

Comparing the before and after profiles, we get 400ms of reduction if we use `for...in` mainly because we are not doing `Object.keys` first for `target` and second for `source`. These numbers are really noticeable on a large data like ~15k reports.

hurali97:
@chrispader @hannojg  suggested changes are addressed, would be great to if you guys can re-review 👍

lib/OnyxUtils.ts
```diff
@@ -245,6 +245,77 @@ function get<TKey extends OnyxKey, TValue extends OnyxValue<TKey>>(key: TKey): P
...
+function multiGet<TKey extends OnyxKey>(keys: CollectionKeyBase[]): Promise<Map<OnyxKey, OnyxValue<TKey>>> {
+    // Keys that are not in the cache
+    const missingKeys: OnyxKey[] = [];
+    // Tasks that are pending
+    const pendingTasks: Array<Promise<OnyxValue<TKey>>> = [];
+    // Keys for the tasks that are pending
+    const pendingKeys: OnyxKey[] = [];
+    // Data to be sent back to the invoker
+    const dataMap = new Map<OnyxKey, OnyxValue<TKey>>();
```
mountiny, line 257: NAB: readability
```suggestion
    const missingKeys: OnyxKey[] = [];

    // Tasks that are pending
    const pendingTasks: Array<Promise<OnyxValue<TKey>>> = [];

    // Keys for the tasks that are pending
    const pendingKeys: OnyxKey[] = [];

    // Data to be sent back to the invoker
    const dataMap = new Map<OnyxKey, OnyxValue<TKey>>();
```

lib/OnyxUtils.ts
```diff
@@ -245,6 +245,77 @@ function get<TKey extends OnyxKey, TValue extends OnyxValue<TKey>>(key: TKey): P
...
+
+    /**
+     * We are going to iterate over all the matching keys and check if we have the data in the cache.
+     * If we do then we add it to the data object. If we do not then we check if there is a pending task
```
mountiny, line 261:
```suggestion
     * If we do then we add it to the data object. If we do not have them, then we check if there is a pending task
```

lib/OnyxUtils.ts
```diff
@@ -245,6 +245,77 @@ function get<TKey extends OnyxKey, TValue extends OnyxValue<TKey>>(key: TKey): P
...
+     * for the key. If there is then we add the promise to the pendingTasks array and the key to the pendingKeys
```
mountiny, line 262:
```suggestion
     * for the key. If there is such task, then we add the promise to the pendingTasks array and the key to the pendingKeys
```

lib/OnyxUtils.ts
```diff
@@ -245,6 +245,77 @@ function get<TKey extends OnyxKey, TValue extends OnyxValue<TKey>>(key: TKey): P
...
+     * array. If there is no pending task then we add the key to the missingKeys array.
+     *
+     * These missingKeys will be later to use to multiGet the data from the storage.
```
mountiny, line 265:
```suggestion
     * These missingKeys will be later used to multiGet the data from the storage.
```

lib/OnyxUtils.ts
```diff
@@ -245,6 +245,77 @@ function get<TKey extends OnyxKey, TValue extends OnyxValue<TKey>>(key: TKey): P
...
+            pendingTasks.push(cache.getTaskPromise(pendingKey) as Promise<OnyxValue<TKey>>);
+            pendingKeys.push(key);
+        } else {
+            missingKeys.push(key);
+        }
+    });
+
+    return (
+        Promise.all(pendingTasks)
+            // We are going to wait for all the pending tasks to resolve and then add the data to the data object.
```
mountiny, line 285:
```suggestion
            // Wait for all the pending tasks to resolve and then add the data to the data map.
```

lib/OnyxUtils.ts
```diff
@@ -245,6 +245,77 @@ function get<TKey extends OnyxKey, TValue extends OnyxValue<TKey>>(key: TKey): P
...
+            .then((values) => {
+                values.forEach((value, index) => {
+                    dataMap.set(pendingKeys[index], value);
+                });
+
+                return Promise.resolve();
+            })
+            // We are going to get the missing keys using multiGet from the storage.
```
mountiny, line 293:
```suggestion
            // Get the missing keys using multiGet from the storage.
```

lib/OnyxUtils.ts
```diff
@@ -245,6 +245,77 @@ function get<TKey extends OnyxKey, TValue extends OnyxValue<TKey>>(key: TKey): P
...
+            .then(() => {
+                if (missingKeys.length === 0) {
+                    return Promise.resolve(undefined);
+                }
+
+                return Storage.multiGet(missingKeys);
+            })
+            // We are going to add the data from the missing keys to the data object and also merge it to the cache.
```
mountiny, line 301:
```suggestion
            // Add the data from the missing keys to the data map and also merge it to the cache.
```

mountiny (commented):
@hurali97 mainly comment changes requested. 

@chrispader @hannojg can you give this another look please?

chrispader (approved):
LGTM!

neil-marcellini:
@hannojg do you want to review once more before we merge? @mountiny feel free to go ahead if you want.

mountiny (approved):
Lets get this one going thank you very much!

Issue #569, opened 2024-07-18
Onyx connection can be called twice for the same data


hannojg:
There exists a "race condition" where an Onyx connection callback will be called twice for the same data. Potentially this can lead to unnecessary rerenders in the react applications.

The race condition happens when creating a connection and shortly after setting the data:

![Screenshot 2024-07-18 at 15 12 29](https://github.com/user-attachments/assets/2181d9a4-9f05-4025-b2e8-5dfe0a79a2ac)

It once gets called here in `.set`:

https://github.com/Expensify/react-native-onyx/blob/c53826bbfe31e423fad0c33b0cb9ab097b88f128/lib/Onyx.ts#L274-L275

and then here once the connection promise resolved:

https://github.com/Expensify/react-native-onyx/blob/c53826bbfe31e423fad0c33b0cb9ab097b88f128/lib/Onyx.ts#L175

hannojg:
I think I could build a failing unit test and fix the issue (cc @mountiny)

mountiny:
@hannojg yes please, feel free to investigate a fix for this

hannojg:
PR is up here:

- https://github.com/Expensify/react-native-onyx/pull/570

mountiny:
@hannojg are you able to create the onyx bump now with the checklist filled in? thanks!

hannojg:
Hey, yes I could, but I think you mentioned me [in some other issue](https://github.com/Expensify/App/issues/37560#issuecomment-2251989745) and someone else is doing the bump, no?

mountiny:
Yeah that was after i commented here, i think they can handle that for you

hannojg:
My first PR [caused a few test to fail in the newdot code](https://github.com/Expensify/App/pull/46297#issuecomment-2252785350). I addressed the problems in these PRs (would appreciate a review):

- https://github.com/Expensify/react-native-onyx/pull/573
- https://github.com/Expensify/react-native-onyx/pull/574

mountiny:
@hannojg I believe these changes are now live in the App, thank you! Do you reckon we can close this issue out?

hannojg:
Yes, thats fixed now!

PR #570, opened 2024-07-22
fix: connection could be called twice on set

hannojg:
### Details

This PR fixes cases where a call to `Onyx.set` could invoke a `Onyx.connect` listener twice. 
For `withOnyx` connections this isn't happening, as they are already guarded by previous equal comparisons.

The fix here is to add a map that will store all previous values that were send to the `Onyx.connect` callback. We can't reuse the cache here unfortunately, as it can contain values while no data has been send to the connection yet (and thus a comparison would lead to us not sending to the connection.

For useOnyx I had to add a special parameter to `Onyx.connect` called `alwaysNotify` which is false by default.
`useOnyx` might uses a selector to transform the data. The output of that selector can change, even though the input data stay the same.
Thus for `useOnyx` we always have to call the connect callback.

### Related Issues

https://github.com/Expensify/react-native-onyx/issues/569

### Automated Tests

- Added a test that explicitly tests for the edge case where a connection would be called twice
- Fixed previous test that now has a reduced connection callback count which is expected 

### Manual Tests

n/a

### Author Checklist

    - [x] Android / native
    - [x] Android / Chrome
    - [x] iOS / native
    - [x] iOS / Safari
    - [x] MacOS / Chrome / Safari
    - [x] MacOS / Desktop

### Screenshots/Videos
Note: pressing the send button might be buggy in the recordings, its a know issue [first reported here](https://expensify.slack.com/archives/C01GTK53T8Q/p1721039243892689) and is not related to this PR.

<details>

https://github.com/user-attachments/assets/a3ca2efc-78a4-4c7f-8b92-4f00b6fe7fa7

</details>

<details>

https://github.com/user-attachments/assets/1d83dd10-3e12-41a2-b885-ad61120f2a92

</details>

<details>

https://github.com/user-attachments/assets/07a76549-87ac-42b7-a65c-eb4f6148c9ce

</details>

<details>

https://github.com/user-attachments/assets/3e962c75-ab38-44cb-b78b-a78adebcfca4

</details>

<details>

https://github.com/user-attachments/assets/b031aa42-2241-4628-aa9f-d77032f10bf9

</details>

<details>

https://github.com/user-attachments/assets/50ad6d61-170b-4b5a-af9f-414aa32de967

</details>

### DIFF
--- a/lib/OnyxUtils.ts
+++ b/lib/OnyxUtils.ts
@@ -978,7 +978,7 @@ function sendDataToConnection<TKey extends OnyxKey>(mapping: Mapping<TKey>, valu
         return;
     }
 
-    (mapping as DefaultConnectOptions<TKey>).callback?.(value === null ? undefined : value, matchedKey as TKey);
+    (mapping as DefaultConnectOptions<TKey>).callback?.(valueToPass, matchedKey as TKey);
 }
 
 /**
--- a/lib/OnyxUtils.ts
+++ b/lib/OnyxUtils.ts
@@ -72,6 +72,9 @@ let defaultKeyStates: Record<OnyxKey, OnyxValue<OnyxKey>> = {};
 let batchUpdatesPromise: Promise<void> | null = null;
 let batchUpdatesQueue: Array<() => void> = [];
 
+// Holds a map representing the latest data that was used to invoke a connection callback
+const lastConnectionCallbackData = new Map<number, OnyxValue<OnyxKey>>();
+
 let snapshotKey: OnyxKey | null = null;
 
 function getSnapshotKey(): OnyxKey | null {
@@ -774,8 +777,8 @@ function keyChanged<TKey extends OnyxKey>(
     value: OnyxValue<TKey>,
     previousValue: OnyxValue<TKey>,
     canUpdateSubscriber: (subscriber?: Mapping<OnyxKey>) => boolean = () => true,
-    notifyRegularSubscibers = true,
-    notifyWithOnyxSubscibers = true,
+    notifyConnectSubscribers = true,
+    notifyWithOnyxSubscribers = true,
 ): void {
     // Add or remove this key from the recentlyAccessedKeys lists
     if (value !== null) {
@@ -809,9 +812,13 @@ function keyChanged<TKey extends OnyxKey>(
 
         // Subscriber is a regular call to connect() and provided a callback
         if (typeof subscriber.callback === 'function') {
-            if (!notifyRegularSubscibers) {
+            if (!notifyConnectSubscribers) {
                 continue;
             }
+            if (lastConnectionCallbackData.has(subscriber.connectionID) && lastConnectionCallbackData.get(subscriber.connectionID) === value) {
+                continue;
+            }
+
             if (isCollectionKey(subscriber.key) && subscriber.waitForCollectionCallback) {
                 const cachedCollection = getCachedCollection(subscriber.key);
 
@@ -822,12 +829,14 @@ function keyChanged<TKey extends OnyxKey>(
 
             const subscriberCallback = subscriber.callback as DefaultConnectCallback<TKey>;
             subscriberCallback(value, key);
+
+            lastConnectionCallbackData.set(subscriber.connectionID, value);
             continue;
         }
 
         // Subscriber connected via withOnyx() HOC
         if ('withOnyxInstance' in subscriber && subscriber.withOnyxInstance) {
-            if (!notifyWithOnyxSubscibers) {
+            if (!notifyWithOnyxSubscribers) {
                 continue;
             }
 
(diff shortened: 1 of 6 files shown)


fabioh8010 (approved):

mountiny (dismissed):
Changes look good to me too, waiting for a c+ testing

lib/types.ts
```diff
@@ -263,6 +263,11 @@ type Collection<TKey extends CollectionKeyBase, TValue, TMap = never> = {
 /** Represents the base options used in `Onyx.connect()` method. */
 type BaseConnectOptions = {
     initWithStoredValues?: boolean;
+    /**
+     * If true, it will call the callback function even if the value might hasn't changed.
```
blazejkustra, line 267: I think you left an error here, right? "Value might hasn't changed"

blazejkustra (approved):

lib/types.ts
hannojg: Hm maybe my wording wasn't precise here.
When this value is `true` it can happen that the Onyx.conenct callback gets called with the same value multiple times (basically bypass the improvement)

Do you have an idea how to rephrase it to make it more obvious / better understandable?

hannojg:
hm, maybe one second before we continue, want to ask Fabio one thing who wrote the useOnxy hook. Maybe we don't need the `alwaysNotify` prop

lib/types.ts
hannojg: Currently discussing if this param is really needed

https://expensify.slack.com/archives/C05LX9D6E07/p1721746491534399?thread_ts=1721742421.298419&cid=C05LX9D6E07

hannojg:
Yes, thanks for being patient with me!
I figured that the `alwaysNotify` isn't needed.

There was this test that failed with the changes I made:

Screenshot 2024-07-23 at 17 46 23

The thing is that the test case was a bit broken. In the test we change the selector function reference. In a react application that will only happen when the hook gets called again aka. the component re-renders (if we follow all strict rules, which we do). 
The test was working earlier early because the Onyx.connect callback was called another time, which is what this PR fixes.

lib/OnyxUtils.ts
```diff
@@ -72,6 +72,9 @@ let defaultKeyStates: Record<OnyxKey, OnyxValue<OnyxKey>> = {};
 let batchUpdatesPromise: Promise<void> | null = null;
 let batchUpdatesQueue: Array<() => void> = [];
 
+// Holds a map representing the latest data that was used to invoke a connection callback
```
marcaaron, line 75: **non blocking thought:** maybe better to explain why we need this and what it is used for.

lib/OnyxUtils.ts
```diff
@@ -960,6 +969,15 @@ function sendDataToConnection<TKey extends OnyxKey>(mapping: Mapping<TKey>, valu
     // If we would pass undefined to setWithOnyxInstance instead, withOnyx would not set the value in the state.
     // withOnyx will internally replace null values with undefined and never pass null values to wrapped components.
     // For regular callbacks, we never want to pass null values, but always just undefined if a value is not set in cache or storage.
+    const valueToPass = value === null ? undefined : value;
```
marcaaron, line 972: **nitpick:** we can also update line 981 to use `valueToPass`

marcaaron (dismissed):
Changes look good. No blockers from me.

tests/unit/useOnyxTest.ts
```diff
@@ -276,6 +276,8 @@ describe('useOnyx', () => {
             expect(result.current[1].status).toEqual('loaded');
 
             selector = (entry: OnyxEntry<{id: string; name: string}>) => `id - ${entry?.id}, name - ${entry?.name} - selector changed`;
+            // In a react app we expect the selector ref to change during a rerender (see selectorRef/useLiveRef)
+            rerender(undefined);
 
             await act(async () => Onyx.merge(ONYXKEYS.TEST_KEY, {id: 'changed_id', name: 'changed_name'}));
```
fabioh8010, line 282:
```suggestion
            await act(async () => waitForPromisesToResolve());

            expect(result.current[0]).toEqual('id - test_id, name - test_name - selector changed');
```

I think we can remove the Onyx.merge since the rerender() is enough to prove the selector changed

hannojg: ah good idea, let me check!

hannojg: We don't even need waitForPromisesToResolve() 😊 looks good?

fabioh8010 (approved):

hannojg:
@mountiny I think we are good to merge!

marcaaron (approved):

hannojg:
They are failing due to another bug in Onyx for which I have a PR to fix here:

- https://github.com/Expensify/react-native-onyx/pull/574

PR #577, opened 2024-08-04
[PERF] OnyxUtils keyChanged cached collection retrieval optimisation

kacper-mikolajczak:
### Details

PR introduces an improvement to how `getCachedCollection` is called in `keyChanged`. 

After investigating [this trace](https://expensify.slack.com/archives/C05LX9D6E07/p1721107369742019?thread_ts=1721060786.438019&cid=C05LX9D6E07), we've noticed that when looping through subscribers list in `keyChanged`, the `getCachedCollection` is called for every iteration with the same key. 

To do:
- [X] cache `getCachedCollection` results to reuse them between iterations ([similar to what we have in `keysChanged`](https://github.com/Expensify/react-native-onyx/blob/cb15a800da404ce9e484bcd24564d1fb264fc4e3/lib/OnyxUtils.ts#L590))
- [x] check if cached collection should not be invalidated

I could not reproduce a case from the trace where `getCachedCollection` is called on each iteration to compare the performance gains. By comparing PR perf to main there was no regression seen and from static analysis we should help when the aforementioned edge case happens. 

### Related Issues

https://github.com/Expensify/App/issues/46863

### Author Checklist

    - [X] Android / native
    - [X] Android / Chrome
    - [X] iOS / native
    - [X] iOS / Safari
    - [X] MacOS / Chrome / Safari
    - [X] MacOS / Desktop
- [X] I verified there are no console errors (if there's a console error not related to the PR, report it or open an issue for it to be fixed)
    - [X] The file has a description of what it does and/or why is needed at the top of the file if the code is not self explanatory
- [X] If the `main` branch was merged into this PR after a review, I tested again and verified the outcome was still expected according to the `Test` steps.
- [X] I have checked off every checkbox in the PR author checklist, including those that don't apply to this PR.

### Screenshots/Videos
<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

### DIFF
--- a/lib/OnyxUtils.ts
+++ b/lib/OnyxUtils.ts
@@ -823,7 +823,12 @@ function keyChanged<TKey extends OnyxKey>(
             }
 
             if (isCollectionKey(subscriber.key) && subscriber.waitForCollectionCallback) {
-                const cachedCollection = cachedCollections[subscriber.key] ?? getCachedCollection(subscriber.key);
+                let cachedCollection = cachedCollections[subscriber.key];
+
+                if (!cachedCollection) {
+                    cachedCollection = getCachedCollection(subscriber.key);
+                    cachedCollections[subscriber.key] = cachedCollection;
+                }
 
                 cachedCollection[key] = value;
                 subscriber.callback(cachedCollection);
--- a/lib/OnyxUtils.ts
+++ b/lib/OnyxUtils.ts
@@ -804,6 +804,9 @@ function keyChanged<TKey extends OnyxKey>(
             return;
         }
     }
+
+    const cachedCollections: Record<string, ReturnType<typeof getCachedCollection>> = {};
+
     for (let i = 0; i < stateMappingKeys.length; i++) {
         const subscriber = callbackToStateMapping[stateMappingKeys[i]];
         if (!subscriber || !isKeyMatch(subscriber.key, key) || !canUpdateSubscriber(subscriber)) {
@@ -820,7 +823,7 @@ function keyChanged<TKey extends OnyxKey>(
             }
 
             if (isCollectionKey(subscriber.key) && subscriber.waitForCollectionCallback) {
-                const cachedCollection = getCachedCollection(subscriber.key);
+                const cachedCollection = cachedCollections[subscriber.key] ?? getCachedCollection(subscriber.key);
 
                 cachedCollection[key] = value;
                 subscriber.callback(cachedCollection);


mountiny (approved):
Comjng from Slack, this seems quite straightforward. Can you create App issue for this with details and link it? Also update the cla and checklist, thank you!

@hannojg @blazejkustra do you want to give it a quick review too

lib/OnyxUtils.ts
```diff
@@ -804,6 +804,9 @@ function keyChanged<TKey extends OnyxKey>(
             return;
         }
     }
+
+    const cachedCollections: Record<string, ReturnType<typeof getCachedCollection>> = {};
```
blazejkustra, line 808: `cachedCollections` stores all collections all in one object? Let's verify that we have enough memory for that 😅

blazejkustra (changes requested):
TS-wise looks good to me, I'm not an Onyx expert but it seems like a good improvement 👍

lib/OnyxUtils.ts
kacper-mikolajczak: Haha :D

On serious note, it won't store all the collections, only those from `subscriber.key`. Also the reference is hold for the `keyChange` lifespan and the object size is the same as we've got from `getCachedCollection`. 

Furthermore, it will hold a reference to only one of such collection object, whereas multiple calls to `getCachedCollection` spawn many of those. 

Let me know if my understanding is correct here, thanks!

hannojg:
Changes look good to me as well. Just wanted to check with y'all if you think we can solve this more holistically - [discussion here](https://expensify.slack.com/archives/C05LX9D6E07/p1722939983736009?thread_ts=1721060786.438019&cid=C05LX9D6E07)

kacper-mikolajczak:
> Can you create App issue for this with details and link it? Also update the cla and checklist, thank you!

Done ✅ !

blazejkustra (approved):

kacper-mikolajczak:
> Just wanted to check with y'all if you think we can solve this more holistically - [discussion here](https://expensify.slack.com/archives/C05LX9D6E07/p1722939983736009?thread_ts=1721060786.438019&cid=C05LX9D6E07)

Thanks for your input @hannojg ❤️ 

I think it's definitely worth looking into how the cache retrieval process works!

For now, let's go ahead with the current improvement. It should still have a positive impact, even when we consider the changes you've suggested. What do you think?

kacper-mikolajczak:
_Continuation of Slack thread where Hanno wrote:_

> I think one solution could be to have an additional map directly in the OnyxCache that stores for each collection key the member keys that are in cache. This way we could easily retrieve a collection from cache.

> Option A where we would just store the keys of the collection that we have cached
```ts
// OnyxCache.ts (pseudo code)
const collectionKeyMap = new Map<string, string>();

set(key: OnyxKey, value: OnyxValue<OnyxKey>): OnyxValue<OnyxKey> {

  if (isCollectionKey(key) {
    collectionKeyMap[key] = splitCollectionMemberKey(key)[1]
  }
}

getCachedCollection(key: CollectionKey) {
  const result = {}
  Object.values(collectionKeyMap[key]).forEach(...)
  return result
}
```

> And maybe there could be an option B, where we have a collectionCacheMap that stores the collections directly :thinking_face:

> in regards to the PR the Problem with solution A would be that we might still for the same key (lets say reports_) create the collection twice, so Option B might be better :thinking_face:

Options B seems more practical but I am not sure of what would be the memory impact of storing such collections once again just to serve as a middleman between in-memory cache and end-user. Let's think of how we can improve the idea even further. 

Having that said @mountiny, what are the next steps if there is a positive response from reviewers?

hannojg:
> For now, let's go ahead with the current improvement. It should still have a positive impact, even when we consider the changes you've suggested. What do you think?

Absolutely, lets do that 😊

hannojg (approved):

hannojg:
> Options B seems more practical but I am not sure of what would be the memory impact of storing such collections once again just to serve as a middleman between in-memory cache and end-user. Let's think of how we can improve the idea even further.

Maybe, instead of having an extra copy just for collections, we could store collections and regular keys separately in the cache (so the data is only there once)

hurali97:
> > Options B seems more practical but I am not sure of what would be the memory impact of storing such collections once again just to serve as a middleman between in-memory cache and end-user. Let's think of how we can improve the idea even further.
> 
> Maybe, instead of having an extra copy just for collections, we could store collections and regular keys separately in the cache (so the data is only there once)

@hannojg Though I agree with this but if we do this it would require us adding some if condition in `set` function which would check whether a key is collection or otherwise and then add to the right data structure.

I was thinking if we can do something like this:

```diff
diff --git a/lib/OnyxCache.ts b/lib/OnyxCache.ts
index 530f8e6..69da4f6 100644
--- a/lib/OnyxCache.ts
+++ b/lib/OnyxCache.ts
@@ -9,7 +9,7 @@ import type {OnyxKey, OnyxValue} from './types';
  */
 class OnyxCache {
     /** Cache of all the storage keys available in persistent storage */
-    private storageKeys: Set<OnyxKey>;
+    private storageKeys: Map<string, Set<OnyxKey>>;
 
     /** A list of keys where a nullish value has been fetched from storage before, but the key still exists in cache */
     private nullishStorageKeys: Set<OnyxKey>;
@@ -30,7 +30,7 @@ class OnyxCache {
     private maxRecentKeysSize = 0;
 
     constructor() {
-        this.storageKeys = new Set();
+        this.storageKeys = new Map();
         this.nullishStorageKeys = new Set();
         this.recentKeys = new Set();
         this.storageMap = {};
@@ -59,8 +59,8 @@ class OnyxCache {
     }
 
     /** Get all the storage keys */
-    getAllKeys(): Set<OnyxKey> {
-        return this.storageKeys;
+    getAllKeys(key: string): Set<OnyxKey> {
+        return this.storageKeys.get(key) || new Set();
     }
 
     /**
@@ -74,15 +74,15 @@ class OnyxCache {
      *
      * @param keys - an array of keys
      */
-    setAllKeys(keys: OnyxKey[]) {
-        this.storageKeys = new Set(keys);
+    setAllKeys(key: string, keys: OnyxKey[]) {
+        this.storageKeys.set(key, new Set(keys));
     }
 
     /** Saves a key in the storage keys list
      * Serves to keep the result of `getAllKeys` up to date
      */
     addKey(key: OnyxKey): void {
-        this.storageKeys.add(key);
+        this.storageKeys.set(key, this.storageKeys.get(key)?.add(key) || new Set([key]));
     }
 
     /** Used to set keys that are null/undefined in storage without adding null to the storage map */

```

kacper-mikolajczak:

mountiny (approved):
Thanks for the discussion, I agree that this should be beneficial and we can move ahead and brainstorm a holistic solution in parallel @hannojg @hurali97

mountiny:
All yours @MariaHCD

MariaHCD (approved):

PR #585, opened 2024-09-30
Feat: Call useOnyx selector for entire collection

kacper-mikolajczak:
### Details

This PR makes `useOnyx` call its `selector` once for entire collection object instead of calling selector per item.

❗ As `withOnyx` is deprecated the change to its selector functionality was not implemented. In order to achieve this, separation between `withOnyx` and `useOnyx` selectors were introduced.

Implementation steps:
1. Modify tests to work with new functionality
2. Refactor typings to match new functionality
3. Fix implementation to make both test runner and type checker happy :)

### Related Issues

https://github.com/Expensify/App/issues/49644

### Author Checklist

- [ ] I linked the correct issue in the `### Related Issues` section above
    - [ ] Android / native
    - [ ] Android / Chrome
    - [ ] iOS / native
    - [ ] iOS / Safari
    - [ ] MacOS / Chrome / Safari
    - [ ] MacOS / Desktop

### Screenshots/Videos
<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

### DIFF
--- a/lib/useOnyx.ts
+++ b/lib/useOnyx.ts
@@ -220,7 +220,7 @@ function useOnyx<TKey extends OnyxKey, TReturnValue = OnyxValue<TKey>>(key: TKey
             newFetchStatus = 'loaded';
         }
 
-        // We do a deep equality check if `selector` is defined, since each `OnyxUtils.tryGetCachedValue()` call will
+        // We do a deep equality check if `selector` is defined, since each `tryGetCachedValue()` call will
         // generate a plain new primitive/object/array that was created using the `selector` function.
         // For the other cases we will only deal with object reference checks, so just a shallow equality check is enough.
         let areValuesEqual: boolean;
--- a/lib/useOnyx.ts
+++ b/lib/useOnyx.ts
@@ -109,7 +109,7 @@ function useOnyx<TKey extends OnyxKey, TReturnValue = OnyxValue<TKey>>(key: TKey
 
     // Stores the previous cached value as it's necessary to compare with the new value in `getSnapshot()`.
     // We initialize it to `null` to simulate that we don't have any value from cache yet.
-    const previousValueRef = useRef<TReturnValue | OnyxValue<TKey> | undefined | null>(null);
+    const previousValueRef = useRef<CachedValue<TKey, TReturnValue> | undefined | null>(null);
 
     // Stores the newest cached value in order to compare with the previous one and optimize `getSnapshot()` execution.
     const newValueRef = useRef<CachedValue<TKey, TReturnValue> | undefined | null>(null);
--- a/lib/useOnyx.ts
+++ b/lib/useOnyx.ts
@@ -1,11 +1,10 @@
 import {deepEqual, shallowEqual} from 'fast-equals';
 import {useCallback, useEffect, useRef, useSyncExternalStore} from 'react';
-import type {IsEqual} from 'type-fest';
 import OnyxCache from './OnyxCache';
 import type {Connection} from './OnyxConnectionManager';
 import connectionManager from './OnyxConnectionManager';
 import OnyxUtils from './OnyxUtils';
-import type {CollectionKeyBase, OnyxCollection, OnyxEntry, OnyxKey, OnyxValue} from './types';
+import type {CollectionKeyBase, OnyxKey, OnyxValue} from './types';
 import useLiveRef from './useLiveRef';
 import usePrevious from './usePrevious';
 
@@ -33,6 +32,8 @@ type UseOnyxInitialValueOption<TInitialValue> = {
     initialValue?: TInitialValue;
 };
 
+type UseOnyxSelector<TKey extends OnyxKey, TReturnValue = OnyxValue<TKey>> = (data: OnyxValue<TKey> | undefined) => TReturnValue;
+
 type UseOnyxSelectorOption<TKey extends OnyxKey, TReturnValue> = {
     /**
      * This will be used to subscribe to a subset of an Onyx key's data.
@@ -47,17 +48,11 @@ type UseOnyxOptions<TKey extends OnyxKey, TReturnValue> = BaseUseOnyxOptions & U
 
 type FetchStatus = 'loading' | 'loaded';
 
-type SelectedValue<TKey, TValue> = TKey extends CollectionKeyBase ? OnyxCollection<TValue> : OnyxEntry<TValue>;
-
-type CachedValue<TKey extends OnyxKey, TValue> = IsEqual<TValue, OnyxValue<TKey>> extends true ? TValue : SelectedValue<TKey, TValue>;
-
 type ResultMetadata = {
     status: FetchStatus;
 };
 
-type UseOnyxResult<TKey extends OnyxKey, TValue> = [CachedValue<TKey, TValue>, ResultMetadata];
-
-type UseOnyxSelector<TKey extends OnyxKey, SelectorValue> = (data?: OnyxValue<TKey>) => SelectorValue;
+type UseOnyxResult<TValue> = [TValue, ResultMetadata];
 
 /**
  * Gets the cached value from the Onyx cache. If the key is a collection key, it will return all the values in the collection.
@@ -91,23 +86,15 @@ function getCachedValue<TKey extends OnyxKey>(key: TKey): OnyxValue<TKey> | unde
 /**
  * Gets the value from cache and maps it with selector.
  */
-function getSelectedValue<TKey extends OnyxKey, TValue = OnyxValue<TKey> | undefined>(key: TKey, selector?: UseOnyxSelector<TKey, TValue>) {
+function getSelectedValue<TKey extends OnyxKey, TValue>(key: TKey, selector?: UseOnyxSelector<TKey, TValue>) {
     const value = getCachedValue(key);
 
-    const selectedValue = selector ? selector(value) : value;
+    const selectedValue = (selector ? selector(value) : value) as TValue;
 
     return selectedValue;
 }
 
-function useOnyx<TKey extends OnyxKey, TReturnValue = OnyxValue<TKey>>(
-    key: TKey,
-    options?: BaseUseOnyxOptions & UseOnyxInitialValueOption<TReturnValue> & Required<UseOnyxSelectorOption<TKey, TReturnValue>>,
-): UseOnyxResult<TKey, TReturnValue>;
-function useOnyx<TKey extends OnyxKey, TReturnValue = OnyxValue<TKey>>(
-    key: TKey,
-    options?: BaseUseOnyxOptions & UseOnyxInitialValueOption<NoInfer<TReturnValue>>,
-): UseOnyxResult<TKey, TReturnValue>;
-function useOnyx<TKey extends OnyxKey, TReturnValue = OnyxValue<TKey>>(key: TKey, options?: UseOnyxOptions<TKey, TReturnValue>): UseOnyxResult<TKey, TReturnValue> {
+function useOnyx<TKey extends OnyxKey, TReturnValue = OnyxValue<TKey>>(key: TKey, options?: UseOnyxOptions<TKey, TReturnValue>) {
     const connectionRef = useRef<Connection | null>(null);
     const previousKey = usePrevious(key);
 
@@ -116,16 +103,16 @@ function useOnyx<TKey extends OnyxKey, TReturnValue = OnyxValue<TKey>>(key: TKey
 
     // Stores the previous cached value as it's necessary to compare with the new value in `getSnapshot()`.
     // We initialize it to `null` to simulate that we don't have any value from cache yet.
-    const previousValueRef = useRef<CachedValue<TKey, TReturnValue> | undefined | null>(null);
+    const previousValueRef = useRef<TReturnValue | undefined | null>(null);
 
     // Stores the newest cached value in order to compare with the previous one and optimize `getSnapshot()` execution.
-    const newValueRef = useRef<CachedValue<TKey, TReturnValue> | undefined | null>(null);
+    const newValueRef = useRef<TReturnValue | undefined | null>(null);
 
     // Stores the previously result returned by the hook, containing the data from cache and the fetch status.
     // We initialize it to `undefined` and `loading` fetch status to simulate the initial result when the hook is loading from the cache.
     // However, if `initWithStoredValues` is `false` we set the fetch status to `loaded` since we want to signal that data is ready.
-    const resultRef = useRef<UseOnyxResult<TKey, TReturnValue>>([
-        undefined as CachedValue<TKey, TReturnValue>,
+    const resultRef = useRef<UseOnyxResult<TReturnValue | undefined>>([
+        undefined,
         {
             status: options?.initWithStoredValues === false ? 'loaded' : 'loading',
         },
@@ -197,7 +184,7 @@ function useOnyx<TKey extends OnyxKey, TReturnValue = OnyxValue<TKey>>(key: TKey
             // If `newValueRef.current` is `null` or any other value it means that the cache does have a value for that key.
             // This difference between `undefined` and other values is crucial and it's used to address the following
             // conditions and use cases.
-            newValueRef.current = getSelectedValue(key, selectorRef.current) as CachedValue<TKey, TReturnValue>;
+            newValueRef.current = getSelectedValue(key, selectorRef.current);
 
             // We set this flag to `false` again since we don't want to get the newest cached value every time `getSnapshot()` is executed,
             // and only when `Onyx.connect()` callback is fired.
@@ -220,7 +207,7 @@ function useOnyx<TKey extends OnyxKey, TReturnValue = OnyxValue<TKey>>(key: TKey
         // If data is not present in cache and `initialValue` is set during the first connection,
         // we set the new value to `initialValue` and fetch status to `loaded` since we already have some data to return to the consumer.
         if (isFirstConnectionRef.current && !hasCacheForKey && options?.initialValue !== undefined) {
-            newValueRef.current = (options?.initialValue ?? undefined) as CachedValue<TKey, TReturnValue>;
+            newValueRef.current = options?.initialValue ?? undefined;
             newFetchStatus = 'loaded';
         }
 
@@ -242,7 +229,7 @@ function useOnyx<TKey extends OnyxKey, TReturnValue = OnyxValue<TKey>>(key: TKey
             previousValueRef.current = newValueRef.current;
 
             // If the new value is `null` we default it to `undefined` to ensure the consumer gets a consistent result from the hook.
-            resultRef.current = [previousValueRef.current as CachedValue<TKey, TReturnValue>, {status: newFetchStatus ?? 'loaded'}];
+            resultRef.current = [previousValueRef.current, {status: newFetchStatus ?? 'loaded'}];
         }
 
         return resultRef.current;
@@ -285,7 +272,7 @@ function useOnyx<TKey extends OnyxKey, TReturnValue = OnyxValue<TKey>>(key: TKey
         checkEvictableKey();
     }, [checkEvictableKey]);
 
-    const result = useSyncExternalStore<UseOnyxResult<TKey, TReturnValue>>(subscribe, getSnapshot);
+    const result = useSyncExternalStore(subscribe, getSnapshot);
 
     return result;
 }
--- a/lib/useOnyx.ts
+++ b/lib/useOnyx.ts
@@ -74,7 +74,7 @@ function getCachedValue<TKey extends OnyxKey>(key: TKey): OnyxValue<TKey> | unde
 
     const values: Record<string, unknown> = {};
     allCacheKeys.forEach((cacheKey) => {
-        if (cacheKey.startsWith(key)) {
+        if (!cacheKey.startsWith(key)) {
             return;
         }
 
--- a/lib/useOnyx.ts
+++ b/lib/useOnyx.ts
@@ -59,6 +59,10 @@ type UseOnyxResult<TKey extends OnyxKey, TValue> = [CachedValue<TKey, TValue>, R
 
 type UseOnyxSelector<TKey extends OnyxKey, SelectorValue> = (data?: OnyxValue<TKey>) => SelectorValue;
 
+/**
+ * Gets the cached value from the Onyx cache. If the key is a collection key, it will return all the values in the collection.
+ * It is a fork of `tryGetCachedValue` from `OnyxUtils` caused by different selector logic in `useOnyx`. It should be unified in the future, when `withOnyx` is removed.
+ */
 function getCachedValue<TKey extends OnyxKey>(key: TKey): OnyxValue<TKey> | undefined {
     if (!OnyxUtils.isCollectionKey(key)) {
         return OnyxCache.get(key) as OnyxValue<TKey> | undefined;
@@ -84,6 +88,9 @@ function getCachedValue<TKey extends OnyxKey>(key: TKey): OnyxValue<TKey> | unde
     return values as OnyxValue<TKey>;
 }
 
+/**
+ * Gets the value from cache and maps it with selector.
+ */
 function getSelectedValue<TKey extends OnyxKey, TValue = OnyxValue<TKey> | undefined>(key: TKey, selector?: UseOnyxSelector<TKey, TValue>) {
     const value = getCachedValue(key);
 
--- a/lib/useOnyx.ts
+++ b/lib/useOnyx.ts
@@ -61,7 +61,7 @@ type UseOnyxSelector<TKey extends OnyxKey, SelectorValue> = (data?: OnyxValue<TK
 
 function getCachedValue<TKey extends OnyxKey>(key: TKey): OnyxValue<TKey> | undefined {
     if (!OnyxUtils.isCollectionKey(key)) {
-        return OnyxCache.get(key);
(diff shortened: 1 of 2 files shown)


mjasikowski:
@kacper-mikolajczak are these breaking changes for App?

kacper-mikolajczak:
Hi @mjasikowski!

Yes, this PR will lead to breaking changes to `useOnyx` selector. 

Here is a PR that is going to contain all the necessary adjustments: https://github.com/Expensify/App/pull/49918

I am going to add a few clean-up commits to this PR and will let you know when it is ready for review if you don't mind.

mjasikowski:
@kacper-mikolajczak thanks and no worries - I'd appreciate if you'd convert it to a Draft PR before it's ready though.

kacper-mikolajczak:
@mjasikowski PR is ready for review! We can also improve typings of `useOnyx` but it is outside the scope of this PR (I will try to work on it in the meantime)

lib/useOnyx.ts
```diff
@@ -40,7 +40,7 @@ type UseOnyxSelectorOption<TKey extends OnyxKey, TReturnValue> = {
      * when the subset of data changes. Otherwise, any change of data on any property would normally
      * cause the component to re-render (and that can be expensive from a performance standpoint).
      */
-    selector?: Selector<TKey, unknown, TReturnValue>;
+    selector?: UseOnyxSelector<TKey, TReturnValue>;
```
fabioh8010, line 43: Could you update the `selector` JSDoc to account for this change?

lib/useOnyx.ts
```diff
@@ -57,8 +57,39 @@ type ResultMetadata = {
...
+    const values: Record<string, unknown> = {};
+    allCacheKeys.forEach((cacheKey) => {
+        if (!cacheKey.startsWith(key)) {
+            return;
+        }
+
+        values[cacheKey] = OnyxCache.get(cacheKey);
+    });
+
+    return values as OnyxValue<TKey>;
```
fabioh8010, line 84:
```suggestion
    return values as OnyxValue<TKey> | undefined;
```

lib/useOnyx.ts
```diff
@@ -57,8 +57,39 @@ type ResultMetadata = {
...
+}
+
+function getSelectedValue<TKey extends OnyxKey, TValue = OnyxValue<TKey> | undefined>(key: TKey, selector?: UseOnyxSelector<TKey, TValue>) {
```
fabioh8010, line 87: Any reason why you separated into two functions?

lib/useOnyx.ts
```diff
@@ -57,8 +57,39 @@ type ResultMetadata = {
 
 type UseOnyxResult<TKey extends OnyxKey, TValue> = [CachedValue<TKey, TValue>, ResultMetadata];
 
-function getCachedValue<TKey extends OnyxKey, TValue>(key: TKey, selector?: Selector<TKey, unknown, unknown>): CachedValue<TKey, TValue> | undefined {
-    return (OnyxUtils.tryGetCachedValue(key, {selector}) ?? undefined) as CachedValue<TKey, TValue> | undefined;
+type UseOnyxSelector<TKey extends OnyxKey, SelectorValue> = (data?: OnyxValue<TKey>) => SelectorValue;
+
+function getCachedValue<TKey extends OnyxKey>(key: TKey): OnyxValue<TKey> | undefined {
```
fabioh8010, line 62: Could you add a description for this function? Also mentioning why is different from `tryGetCachedValue` implementation.

lib/useOnyx.ts
fabioh8010, line 87: Could you add a description for this function?

lib/useOnyx.ts
kacper-mikolajczak: Sure thing! Could you tell a little bit more which fragment should be updated? Thanks!

lib/useOnyx.ts
kacper-mikolajczak: If I may use your help here - do you know in what circumstances Onyx collection could result to `undefined`? 

For this scope itself:

Value in this branch of the code (collection part) won't be set to `undefined` - it starts with empty object and is populated with collection items. 

`undefined` covers first branch for non-collection values where we could get undefined from cache.

Thanks for looking into it with such detail! <3

lib/useOnyx.ts
kacper-mikolajczak: Mainly, the separation of concerns:
- one function encapsulates value retrieval (consolidates the collection vs non-collection logic)
- second function maps the values without being interested in the retrieval itself

I think it is easier to work with that way, wdyt?

lib/useOnyx.ts
kacper-mikolajczak: Done ✅

lib/useOnyx.ts
kacper-mikolajczak: Done ✅

lib/useOnyx.ts
fabioh8010: Actually nvm, the description still makes sense with the change 😅

lib/useOnyx.ts
```diff
@@ -33,6 +32,8 @@ type UseOnyxInitialValueOption<TInitialValue> = {
     initialValue?: TInitialValue;
 };
 
+type UseOnyxSelector<TKey extends OnyxKey, TReturnValue = OnyxValue<TKey>> = (data: OnyxValue<TKey> | undefined) => TReturnValue;
```
fabioh8010, line 35:
```suggestion
type UseOnyxSelector<TKey extends OnyxKey, TReturnValue> = (data: OnyxValue<TKey>) => TReturnValue;
```

We don't need to specify `undefined` as `OnyxValue` already includes it. Also, no need to specify default for `TReturnValue`

lib/useOnyx.ts
```diff
@@ -84,23 +83,18 @@ function getCachedValue<TKey extends OnyxKey>(key: TKey): OnyxValue<TKey> | unde
...
-function useOnyx<TKey extends OnyxKey, TReturnValue = OnyxValue<TKey>>(
-    key: TKey,
-    options?: BaseUseOnyxOptions & UseOnyxInitialValueOption<TReturnValue> & Required<UseOnyxSelectorOption<TKey, TReturnValue>>,
-): UseOnyxResult<TKey, TReturnValue>;
-function useOnyx<TKey extends OnyxKey, TReturnValue = OnyxValue<TKey>>(
-    key: TKey,
-    options?: BaseUseOnyxOptions & UseOnyxInitialValueOption<NoInfer<TReturnValue>>,
-): UseOnyxResult<TKey, TReturnValue>;
-function useOnyx<TKey extends OnyxKey, TReturnValue = OnyxValue<TKey>>(key: TKey, options?: UseOnyxOptions<TKey, TReturnValue>): UseOnyxResult<TKey, TReturnValue> {
+function useOnyx<TKey extends OnyxKey, TReturnValue = OnyxValue<TKey>>(key: TKey, options?: UseOnyxOptions<TKey, TReturnValue>) {
```
fabioh8010, line 97: Please don't remove the overrides, it will mess with some edge cases

lib/useOnyx.ts
```diff
@@ -278,7 +272,7 @@ function useOnyx<TKey extends OnyxKey, TReturnValue = OnyxValue<TKey>>(key: TKey
         checkEvictableKey();
     }, [checkEvictableKey]);
 
-    const result = useSyncExternalStore<UseOnyxResult<TKey, TReturnValue>>(subscribe, getSnapshot);
+    const result = useSyncExternalStore(subscribe, getSnapshot);
```
fabioh8010, line 275: Why was this removed?

lib/useOnyx.ts
kacper-mikolajczak: Got this error when trying to remove `undefined` from union:

```
Argument of type 'OnyxValue<TKey> | undefined' is not assignable to parameter of type 'OnyxValue<TKey>'.
Type 'undefined' is not assignable to type 'OnyxValue<TKey>'.ts(2345)
  ```

File: lib/useOnyx.ts (hunk 6 above)
Comment (kacper-mikolajczak) [reply lib/useOnyx.ts:105]: Revived them!

File: lib/useOnyx.ts (hunk 7 above)
Comment (kacper-mikolajczak) [reply lib/useOnyx.ts:283]: Do we need to define it explicitly?

**Review by mjasikowski — APPROVED** [review]
Looks good for my very limited knowledge, but I'd recommend to defer to @fabioh8010 and @mountiny for proper reviews.

**Comment by mountiny** [comment]
@hungvu193 added for review and testing

File: lib/useOnyx.ts (hunk 5 above)
Comment (fabioh8010) [reply lib/useOnyx.ts:35]: I left some new comments that will solve this!

File: lib/useOnyx.ts [hunk 8]
```diff
@@ -33,43 +32,77 @@ type UseOnyxInitialValueOption<TInitialValue> = {
...
 };
 
-type UseOnyxResult<TKey extends OnyxKey, TValue> = [CachedValue<TKey, TValue>, ResultMetadata];
+type UseOnyxResult<TValue> = [TValue, ResultMetadata];
+
+/**
+ * Gets the cached value from the Onyx cache. If the key is a collection key, it will return all the values in the collection.
+ * It is a fork of `tryGetCachedValue` from `OnyxUtils` caused by different selector logic in `useOnyx`. It should be unified in the future, when `withOnyx` is removed.
+ */
+function getCachedValue<TKey extends OnyxKey>(key: TKey): OnyxValue<TKey> | undefined {
```
fabioh8010, line 61:
```suggestion
function getCachedValue<TKey extends OnyxKey>(key: TKey): OnyxValue<OnyxKey> {
```

This will allows us to remove some assertions

lib/useOnyx.ts
```diff
@@ -33,43 +32,77 @@ type UseOnyxInitialValueOption<TInitialValue> = {
...
+    if (!OnyxUtils.isCollectionKey(key)) {
+        return OnyxCache.get(key) as OnyxValue<TKey> | undefined;
```
fabioh8010, line 63:
```suggestion
        return OnyxCache.get(key);
```

lib/useOnyx.ts
```diff
@@ -33,43 +32,77 @@ type UseOnyxInitialValueOption<TInitialValue> = {
...
+    const values: Record<string, unknown> = {};
+    allCacheKeys.forEach((cacheKey) => {
+        if (!cacheKey.startsWith(key)) {
+            return;
+        }
+
+        values[cacheKey] = OnyxCache.get(cacheKey);
+    });
+
+    return values as OnyxValue<TKey>;
```
fabioh8010, line 83:
```suggestion
    return values;
```

lib/useOnyx.ts
```diff
@@ -33,43 +32,77 @@ type UseOnyxInitialValueOption<TInitialValue> = {
...
+
+    const allCacheKeys = OnyxCache.getAllKeys();
+
+    // It is possible we haven't loaded all keys yet so we do not know if the
+    // collection actually exists.
+    if (allCacheKeys.size === 0) {
+        return;
+    }
+
+    const values: Record<string, unknown> = {};
```
fabioh8010, line 74:
```suggestion
    const values: OnyxCollection<KeyValueMapping[TKey]> = {};
```

lib/useOnyx.ts
```diff
@@ -33,43 +32,77 @@ type UseOnyxInitialValueOption<TInitialValue> = {
...
+        values[cacheKey] = OnyxCache.get(cacheKey);
+    });
+
+    return values as OnyxValue<TKey>;
+}
+
+/**
+ * Gets the value from cache and maps it with selector.
+ */
+function getSelectedValue<TKey extends OnyxKey, TValue>(key: TKey, selector?: UseOnyxSelector<TKey, TValue>) {
```
fabioh8010, line 89:
```suggestion
function getSelectedValue<TKey extends OnyxKey, TValue>(key: TKey, selector?: UseOnyxSelector<TKey, TValue>): TValue {
```

Let's add return type as well

lib/useOnyx.ts
```diff
@@ -33,43 +32,77 @@ type UseOnyxInitialValueOption<TInitialValue> = {
...
+    const value = getCachedValue(key);
+
+    const selectedValue = (selector ? selector(value) : value) as TValue;
```
fabioh8010, line 92:
```suggestion
    const selectedValue = (selector ? selector(value as OnyxValue<TKey>) : value) as TValue;
```

So now we only have to do this assertion here, compared to previous code where we had more asssertions

lib/useOnyx.ts
```diff
@@ -78,16 +111,16 @@ function useOnyx<TKey extends OnyxKey, TReturnValue = OnyxValue<TKey>>(key: TKey
...
     // Stores the newest cached value in order to compare with the previous one and optimize `getSnapshot()` execution.
-    const newValueRef = useRef<CachedValue<TKey, TReturnValue> | undefined | null>(null);
+    const newValueRef = useRef<TReturnValue | undefined | null>(null);
 
     // Stores the previously result returned by the hook, containing the data from cache and the fetch status.
     // We initialize it to `undefined` and `loading` fetch status to simulate the initial result when the hook is loading from the cache.
     // However, if `initWithStoredValues` is `false` we set the fetch status to `loaded` since we want to signal that data is ready.
-    const resultRef = useRef<UseOnyxResult<TKey, TReturnValue>>([
-        undefined as CachedValue<TKey, TReturnValue>,
+    const resultRef = useRef<UseOnyxResult<TReturnValue | undefined | null>>([
```
fabioh8010, line 122:
```suggestion
    const resultRef = useRef<UseOnyxResult<TReturnValue | undefined>>([
```

Could we keep this way? We will have to do one assertion in the code but it protects the code from someone accidentally assigning `null` to `resultRef` value, which could never happen.

lib/useOnyx.ts
```diff
@@ -204,7 +237,7 @@ function useOnyx<TKey extends OnyxKey, TReturnValue = OnyxValue<TKey>>(key: TKey
             previousValueRef.current = newValueRef.current;
 
             // If the new value is `null` we default it to `undefined` to ensure the consumer gets a consistent result from the hook.
-            resultRef.current = [previousValueRef.current as CachedValue<TKey, TReturnValue>, {status: newFetchStatus ?? 'loaded'}];
+            resultRef.current = [previousValueRef.current, {status: newFetchStatus ?? 'loaded'}];
```
fabioh8010, line 240:
```suggestion
            resultRef.current = [previousValueRef.current as TReturnValue | undefined, {status: newFetchStatus ?? 'loaded'}];
```

The necessary assertion I mentioned above

lib/useOnyx.ts
fabioh8010: It's to ensure `getSnapshot` **has** to return this value `UseOnyxResult<TReturnValue | undefined>` as a code protection measure.

fabioh8010 (commented):
some minor comments, but it's looking great!

kacper-mikolajczak:
As of what @fabioh8010 shared internally, the `useOnyx` cannot return `null` so the selector will also override `null` values with `undefined` to remain consistent:

```tsx
return selectedValue ?? undefined;
```

kacper-mikolajczak:
Thanks for all the suggestions @fabioh8010! 

All the changes should be applied - let me know what you think 🙏

lib/useOnyx.ts
```diff
@@ -204,11 +238,11 @@ function useOnyx<TKey extends OnyxKey, TReturnValue = OnyxValue<TKey>>(key: TKey
             previousValueRef.current = newValueRef.current;
 
             // If the new value is `null` we default it to `undefined` to ensure the consumer gets a consistent result from the hook.
-            resultRef.current = [previousValueRef.current as CachedValue<TKey, TReturnValue>, {status: newFetchStatus ?? 'loaded'}];
+            resultRef.current = [previousValueRef.current ?? undefined, {status: newFetchStatus ?? 'loaded'}];
```
fabioh8010, line 241: Do you think we need to add `?? undefined`? I think at this point of the code it would never be `null`.

fabioh8010 (approved):
Left one question, besides LGTM 🚀  Thanks for the work and patience here!!

lib/useOnyx.ts
kacper-mikolajczak: Not sure from execution perspective, but the TS was complaining there about `previousValue` being possibly `null`, so just to be safe I added the `undefined` fallback. 

If you feel like we should replace it with assertion

```tsx
resultRef.current = [previousValueRef.current as NonNullable<TReturnValue> | undefined, {status: newFetchStatus ?? 'loaded'}];
```

We can swap it out

kacper-mikolajczak:
Wild idea as another potential improvement: What if we apply selector at the end of `useOnyx` pipeline and rely internally on actual Onyx cache data as source of truth?

It should simplify the typings and also enable possibility to return `null` values from `useOnyx` selectors as well. Also it should yield positive performance result as we won't do the `deepEquality` check whenever selector is applied - only rely on shallow checks for state (the requirement is cache state internally has stable reference, otherwise this idea is not feasible)

Here's a diff:

```diff
diff --git a/lib/useOnyx.ts b/lib/useOnyx.ts
index e294fe9..fa8e4d4 100644
--- a/lib/useOnyx.ts
+++ b/lib/useOnyx.ts
@@ -1 +1 @@
-import {deepEqual, shallowEqual} from 'fast-equals';
+import {shallowEqual} from 'fast-equals';
@@ -123,2 +123,2 @@ function useOnyx<TKey extends OnyxKey, TReturnValue = OnyxValue<TKey>>(key: TKey
-    const resultRef = useRef<UseOnyxResult<TReturnValue>>([
-        undefined,
+    const resultRef = useRef<[TReturnValue | undefined, ResultMetadata]>([
+        selectorRef.current?.(undefined),
@@ -196 +196 @@ function useOnyx<TKey extends OnyxKey, TReturnValue = OnyxValue<TKey>>(key: TKey
-            newValueRef.current = getCachedValue(key, selectorRef.current);
+            newValueRef.current = getCachedValue(key);
@@ -226,6 +226 @@ function useOnyx<TKey extends OnyxKey, TReturnValue = OnyxValue<TKey>>(key: TKey
-        let areValuesEqual: boolean;
-        if (selectorRef.current) {
-            areValuesEqual = deepEqual(previousValueRef.current ?? undefined, newValueRef.current);
-        } else {
-            areValuesEqual = shallowEqual(previousValueRef.current ?? undefined, newValueRef.current);
-        }
+        const areValuesEqual = shallowEqual(previousValueRef.current ?? undefined, newValueRef.current);
@@ -241 +236 @@ function useOnyx<TKey extends OnyxKey, TReturnValue = OnyxValue<TKey>>(key: TKey
-            resultRef.current = [previousValueRef.current ?? undefined, {status: newFetchStatus ?? 'loaded'}];
+            resultRef.current = [selectorRef.current(previousValueRef.current ?? undefined), {status: newFetchStatus ?? 'loaded'}];
```

fabioh8010:
> Wild idea as another potential improvement: What if we apply selector at the end of `useOnyx` pipeline and rely internally on actual Onyx cache data as source of truth?
> 
> It should simplify the typings and also enable possibility to return `null` values from `useOnyx` selectors as well. Also it should yield positive performance result as we won't do the `deepEquality` check whenever selector is applied - only rely on shallow checks for state (the requirement is cache state internally has stable reference, otherwise this idea is not feasible)
> 
> Here's a diff:
> 
> ```diff
> diff --git a/lib/useOnyx.ts b/lib/useOnyx.ts
> index e294fe9..fa8e4d4 100644
> --- a/lib/useOnyx.ts
> +++ b/lib/useOnyx.ts
> @@ -1 +1 @@
> -import {deepEqual, shallowEqual} from 'fast-equals';
> +import {shallowEqual} from 'fast-equals';
> @@ -123,2 +123,2 @@ function useOnyx<TKey extends OnyxKey, TReturnValue = OnyxValue<TKey>>(key: TKey
> -    const resultRef = useRef<UseOnyxResult<TReturnValue>>([
> -        undefined,
> +    const resultRef = useRef<[TReturnValue | undefined, ResultMetadata]>([
> +        selectorRef.current?.(undefined),
> @@ -196 +196 @@ function useOnyx<TKey extends OnyxKey, TReturnValue = OnyxValue<TKey>>(key: TKey
> -            newValueRef.current = getCachedValue(key, selectorRef.current);
> +            newValueRef.current = getCachedValue(key);
> @@ -226,6 +226 @@ function useOnyx<TKey extends OnyxKey, TReturnValue = OnyxValue<TKey>>(key: TKey
> -        let areValuesEqual: boolean;
> -        if (selectorRef.current) {
> -            areValuesEqual = deepEqual(previousValueRef.current ?? undefined, newValueRef.current);
> -        } else {
> -            areValuesEqual = shallowEqual(previousValueRef.current ?? undefined, newValueRef.current);
> -        }
> +        const areValuesEqual = shallowEqual(previousValueRef.current ?? undefined, newValueRef.current);
> @@ -241 +236 @@ function useOnyx<TKey extends OnyxKey, TReturnValue = OnyxValue<TKey>>(key: TKey
> -            resultRef.current = [previousValueRef.current ?? undefined, {status: newFetchStatus ?? 'loaded'}];
> +            resultRef.current = [selectorRef.current(previousValueRef.current ?? undefined), {status: newFetchStatus ?? 'loaded'}];
> ```

@kacper-mikolajczak I think this idea makes very much sense! I’m just worried about how we currently use `selector` inside `useOnyx` — we use a “live ref” and I think with this approach we may have stale renders, so we need to test this properly.

But do you think we could work on this separately in another Proposal / Issue / PR? So we keep the scope of the current PR small and allow a quicker merge, less chance of regressions, and more time to think about the solution.

WDYT?

hungvu193:
The last line here confuses me:
```diff
- resultRef.current = [previousValueRef.current ?? undefined, {status: newFetchStatus ?? 'loaded'}]
+ resultRef.current = [selectorRef.current(previousValueRef.current ?? undefined), {status: newFetchStatus ?? 'loaded'}];
``` 
According to the current code, I assume `selectorRef.current` can be undefined right?
```javascript
        if (selectorRef.current) {
            areValuesEqual = deepEqual(previousValueRef.current ?? undefined, newValueRef.current);
        } else {
            areValuesEqual = shallowEqual(previousValueRef.current ?? undefined, newValueRef.current);
        }
```

So I'd agree with @fabioh8010, we should work on it in a separate issue.

lib/useOnyx.ts
```diff
@@ -33,43 +32,78 @@ type UseOnyxInitialValueOption<TInitialValue> = {
...
     status: FetchStatus;
 };
 
-type UseOnyxResult<TKey extends OnyxKey, TValue> = [CachedValue<TKey, TValue>, ResultMetadata];
+type UseOnyxResult<TValue> = [NonNullable<TValue> | undefined, ResultMetadata];
+
+/**
+ * Gets the cached value from the Onyx cache. If the key is a collection key, it will return all the values in the collection.
+ * It is a fork of `tryGetCachedValue` from `OnyxUtils` caused by different selector logic in `useOnyx`. It should be unified in the future, when `withOnyx` is removed.
+ */
```
hungvu193, line 61: NAB, can you also update the comment at line 223, since we didn't use `OnyxUtils.tryGetCachedValue()` anymore?

mountiny:
@hungvu193 did you test the changes applied to App? Can you please do so now before we merge such change? Thanks!

hungvu193:
> @hungvu193 did you test the changes applied to App? Can you please do so now before we merge such change? Thanks!

Yes. So far so good. I can verify that the selector is called once for the entire collection. I haven't found any issues so far, but let me test it carefully to make sure it won't affect any existing flows.

https://github.com/user-attachments/assets/fc1ee2d0-af27-4017-8143-80dafaffc281

hungvu193:
All good @mountiny

lib/useOnyx.ts
kacper-mikolajczak: Done ✅

mountiny (approved):
Dont want to delay this too much but also going to ask Margelo to review

mountiny:
https://expensify.slack.com/archives/C035J5C9FAP/p1727998618219179

lib/useOnyx.ts
```diff
@@ -33,43 +32,78 @@ type UseOnyxInitialValueOption<TInitialValue> = {
...
+    return values;
+}
+
+/**
+ * Gets the value from cache and maps it with selector. It changes `null` to `undefined` for `useOnyx` compatibility.
+ */
+function getCachedValue<TKey extends OnyxKey, TValue>(key: TKey, selector?: UseOnyxSelector<TKey, TValue>) {
+    const value = tryGetCachedValue(key) as OnyxValue<TKey>;
+
+    const selectedValue = selector ? selector(value) : (value as TValue);
```
hannojg, line 94: just wondering for the future if the onyx cache could also have a cache of key + selector function, so that when the selector function remains the same, the data remains the same we don't have to rerun the selector (as selectors could  be expensive to run)

hannojg (approved):
LGTM!

lib/useOnyx.ts
kacper-mikolajczak: Thanks for review Hanno!

What came up as a next improvement while working on this issue was the idea to separate selector from `useOnyx` internals entirely so it won't interfere with Onyx cache and only transform the cached data at the end of the pipeline.

It should have many benefits, but I am still working on the [PoC](https://github.com/Expensify/react-native-onyx/pull/587) whether it is feasible for us to implement 🤞

mountiny:
Going to merge this one now

PR #588, opened 2024-10-08
Fix problems related to Onyx.clear and useOnyx

fabioh8010:
### Details

This PR fixes two problems we've discovered on E/App during migration to `useOnyx`, both related to `Onyx.clear`. More details [here](https://github.com/Expensify/App/issues/48725#issuecomment-2400683475).

#### Problem 1

Racing conditions between an `Onyx.clear()` execution and `useOnyx` subscription causes `useOnyx` to be always in `loading` state and never fully resolve its first connection.

This happens because we clear the cache during `Onyx.clear()` execution and if we are connecting to some key with `useOnyx` at the same time, there is no cache where the hook can look into (`OnyxCache.hasCacheForKey(key)` will be `false`) and it get stuck in `loading` state.

This problem doesn't happen for `withOnyx` basically because the racing condition doesn't occur. This can happen by the following reasons:

- As `withOnyx` is a HOC and consequently a wrapper component, it uses React lifecycles / batching mechanisms that will contribute to a slightly rendering/execution delay that is enough to allow `Onyx.clear()` finish its operation.
- The `useOnyx` hook uses `useSyncExternalStore` under the hood, which is designed to not batch its updates and execute and return as fast as possible, thus contributing to the race condition to happen.

#### Problem 2

The cache clearing performed by `Onyx.clear()` combined with the connection reusing logic we currently have causes an `useOnyx` connection to a key that was connected before to be always `loading` state and never fully resolve its first connection.

Specifically, if we are reusing the connection of a key that is still exists in the connection manager after an `Onyx.clear()` execution, the next `useOnyx` to connect to that same key gets stuck in the `loading` state as the cache doesn't really exist anymore, and we didn't populate it again because we are reusing the connection.

When logging out a lot of Onyx connections will be removed from the connection manager as expected, but some of them persists because they don't belong to the React's lifecycle e.g. any `Onyx.connect` in our lib files, helping create the issue described above.

This problem doesn't happen for `withOnyx` because the HOC doesn't reuse connections and create new ones upon every subscription, thus populating the cache again for that key and allowing the HOC to work as expected.

---

This PR also exposes the `reuseConnection` flag in `useOnyx` API for convenience purposes (it's already exposed in `Onyx.connect`).

### Related Issues

https://github.com/Expensify/App/issues/48725

### Automated Tests

Unit tests were added to cover the fixes mentioned.

### Manual Tests

1. Checkout [this branch](https://github.com/fabioh8010/expensify-app/tree/bugfix/48725) on E/App.
2. Link the changes of this Onyx PR branch to E/App.
3. Run `npm test` on E/App and assert all tests are passing.
4. Login with a user, logout and login with the same user again, assert the whole process works as expected.

❗️You will notice that after second login on Web/mWeb, user is redirected to settings page instead of home. Please have in mind that this issue is not related to these fixes, but instead one more fix that would be necessary to be made on E/App when migrating `AuthScreens.tsx` to `useOnyx`.❗️

### Author Checklist

    - [x] Android / native
    - [x] Android / Chrome
    - [x] iOS / native
    - [x] iOS / Safari
    - [x] MacOS / Chrome / Safari
    - [x] MacOS / Desktop

### Screenshots/Videos
<details>

https://github.com/user-attachments/assets/2ec75fd9-2a82-48ec-b900-15d61f0546f4

</details>

<details>

https://github.com/user-attachments/assets/20abd527-ad1f-4c5c-8533-b0e0f2d839ab

</details>

<details>

https://github.com/user-attachments/assets/c9389ba4-f4c6-4ea9-a9be-01adf1e80914

</details>

<details>

https://github.com/user-attachments/assets/858dff39-8952-4cf6-9282-05b8b60fd26e

</details>

<details>

https://github.com/user-attachments/assets/c05274ed-b940-44fa-b5a4-6056564ca7f0

https://github.com/user-attachments/assets/81490232-19a6-4415-b445-61fc2f2b19bf

</details>

<details>

https://github.com/user-attachments/assets/84204411-5c48-45b4-b7d7-3da22f686a49

</details>


hannojg:
LGTM !

Beamanator (dismissed):
This looks great to me, and looks like it was peer reviewed a few times in Slack 👍 👍 

@tgolen or @marcaaron do either of y'all want to take a look before we merge?

blazejkustra (approved):

fabioh8010:
Bump @tgolen @marcaaron

marcaaron:
Happy to review this, but I don't yet understand what problem we are solving. Can we update the description with the explanation? I took a look at the comment we linked out to and it says:

>From what I've seen this problem doesn't happen for withOnyx basically because the racing condition doesn't occur (maybe because it's a component and it's slightly slower during initialisation?).

Am I reading that "maybe" correctly? Are we not sure why it happens? Do we have a better explanation somewhere?

lib/OnyxConnectionManager.ts
```diff
@@ -73,12 +73,18 @@ class OnyxConnectionManager {
      */
     private lastCallbackID: number;
 
+    /**
+     * Stores the last generated session ID for the connection manager.
```
marcaaron, line 77: Maybe we can add some explanation about what this `sessionID` is used for and why we have it at a ll?

lib/OnyxConnectionManager.ts
```diff
@@ -89,7 +95,10 @@ class OnyxConnectionManager {
      */
     private generateConnectionID<TKey extends OnyxKey>(connectOptions: ConnectOptions<TKey>): string {
         const {key, initWithStoredValues, reuseConnection, waitForCollectionCallback} = connectOptions;
-        let suffix = '';
+
+        // The current session ID is appended to the connection ID so we can have different connections
+        // after an Onyx.clear() operation.
```
marcaaron, line 100: >so we can have different connections after an Onyx.clear() operation.

Let's also say why that is important or needed?

lib/OnyxConnectionManager.ts
```diff
@@ -229,6 +238,15 @@ class OnyxConnectionManager {
         this.connectionsMap.clear();
     }
 
+    /**
+     * Refreshes the connection manager's session ID. Used in Onyx.clear() in order to
+     * create new connections after Onyx is cleared, instead of reusing existing connections
+     * that can produce unexpected behavior in Onyx subscribers.
```
marcaaron, line 244: Can we get more specific? What is the "unexpected behavior"? I think including an example will help later on if someone needs to make changes here or questions whether it is still required.

lib/useOnyx.ts
```diff
@@ -23,6 +23,12 @@ type BaseUseOnyxOptions = {
      * If set to `true`, data will be retrieved from cache during the first render even if there is a pending merge for the key.
      */
     allowStaleData?: boolean;
+
+    /**
+     * If set to `false`, the connection won't be reused between other subscribers that are listening to the same Onyx key
+     * with the same connect configurations.
+     */
```
marcaaron, line 30: Can we include an example of _when_ might we want to use this?

lib/useOnyx.ts
```diff
@@ -232,8 +238,8 @@ function useOnyx<TKey extends OnyxKey, TReturnValue = OnyxValue<TKey>>(key: TKey
 
         // If the previously cached value is different from the new value, we update both cached value
         // and the result to be returned by the hook.
-        // If the cache was set for the first time, we also update the cached value and the result.
-        const isCacheSetFirstTime = previousValueRef.current === null && hasCacheForKey;
+        // If the cache was set for the first time or we have a pending Onyx.clear() task, we also update the cached value and the result.
+        const isCacheSetFirstTime = previousValueRef.current === null && (hasCacheForKey || OnyxCache.hasPendingTask(TASK.CLEAR));
```
marcaaron, line 242: What does it mean for the cache to be "set first time"? Like, as in the first time the hook runs? It makes sense to consider it "set" when it has cache value. But, can we improve the documentation to explain why we should consider the cache "set" if there is a pending clear task?

marcaaron (changes requested):
Most of my comments are about improving the documentation for now.

lib/OnyxConnectionManager.ts
fabioh8010: I moved all explanations to `private sessionID: string;`, please have a look!

lib/OnyxConnectionManager.ts
fabioh8010: I moved all explanations to `private sessionID: string;`, please have a look!

lib/useOnyx.ts
fabioh8010: I refactored the comments, please let me know if it's clearer now.

fabioh8010:
@marcaaron Addressed the comments and added explanations to the PR description, please let me know if it's clearer now.

lib/useOnyx.ts
marcaaron: Ok, I see now. Not really related to these changes.

marcaaron (approved):
These changes seem reasonable to me.

PR #618, opened 2025-02-19
Add Reassure tests (Phase 1)

fabioh8010:
### Details

This PR adds Reassure tests for the following files:

- `Onyx.ts`
- `OnyxUtils.ts`
- `OnyxCache.ts`
- `OnyxConnectionManager.ts`

The intention here is to have the most coverage as possible, even for very super simple functions (like getters). The reasoning is that if they stay the same, performance measures will be the same and we won't have to worry about. If someone adds crazy or super complex logic to simple functions (either by mistake or intentional), they will be flagged.

After it's merged I'm going to create a quick follow-up PR to test if the Reassure performance tests are stable enough when we don't have changes, in order to avoid flakiness.

### Related Issues

https://github.com/Expensify/App/issues/56483

### Automated Tests

The very ones created in the PR.

### Manual Tests

N/A

### Author Checklist

    - [x] Android / native
    - [x] Android / Chrome
    - [x] iOS / native
    - [x] iOS / Safari
    - [x] MacOS / Chrome / Safari
    - [x] MacOS / Desktop

### Screenshots/Videos
<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>


fabioh8010:
@hannojg Do you mind reviewing this PR too? Since you was interested in testing tools for Onyx a while ago.

.github/scripts/buildActions.sh
roryabraham, line 1: I know this script was copied from E/App, but I recently learned more about our bash style guide and best practices, and I think this is how I would write this script today:

```bash
#!/bin/bash
#
# Used to precompile all Github Action node.js scripts using ncc.
# This bundles them with their dependencies into a single executable node.js script.

# In order for this script to be safely run from anywhere, we cannot use the raw relative path '../actions'.
ACTIONS_DIR="$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")/../actions/javascript" &>/dev/null && pwd)"
readonly ACTIONS_DIR

# This stores all the process IDs of the ncc commands so they can run in parallel
ASYNC_BUILDS=()

# Build all the actions in the background
for ACTION in "$ACTIONS_DIR"/*/*.ts; do
    ACTION_DIR=$(dirname "$ACTION")
    npx ncc build --transpile-only --external encoding "$ACTION" -o "$ACTION_DIR" &
    ASYNC_BUILDS+=($!)
done

# Wait for the background build to finish
EXIT_CODE=0
for PID in "${ASYNC_BUILDS[@]}"; do
    if ! wait "$PID"; then
        EXIT_CODE=1
    fi
done

# Prepend this note at the top of all compiled files as a warning to devs.
readonly NOTE_DONT_EDIT='/**
 * NOTE: This is a compiled file. DO NOT directly edit this file.
 */
'
for OUTPUT_FILE in "$ACTIONS_DIR"/*/index.js; do
    echo "$NOTE_DONT_EDIT$(cat "$OUTPUT_FILE")" > "$OUTPUT_FILE"
done

if [[ EXIT_CODE -ne 0 ]]; then
    echo "❌ One or more builds failed"
    exit 1
fi

echo "✅ All builds succeeded"
```

I think that's a bit simpler. At minimum we should use 4 spaces instead of 2 per the style guide. (I say this knowing that I wrote the original version of this script that you copied here).

lib/OnyxUtils.ts
```diff
@@ -1436,6 +1438,65 @@ function unsubscribeFromKey(subscriptionID: number): void {
...
+function updateSnapshots(data: OnyxUpdate[], mergeFn: typeof Onyx.merge): Array<() => Promise<void>> {
+    const snapshotCollectionKey = OnyxUtils.getSnapshotKey();
+    if (!snapshotCollectionKey) return [];
+
+    const promises: Array<() => Promise<void>> = [];
+
+    const snapshotCollection = OnyxUtils.getCachedCollection(snapshotCollectionKey);
+    const snapshotCollectionKeyLength = snapshotCollectionKey.length;
+
+    Object.entries(snapshotCollection).forEach(([snapshotEntryKey, snapshotEntryValue]) => {
```
roryabraham, line 1450: Out-of-scope for this PR but I think we should globally disable the lint rule that prevents us using more efficient and readable `for ... of` loops.

.github/workflows/reassurePerfTests.yml
roryabraham, line 1: Few comments for this file:

- Should `tests/**` really be in `paths-ignore` in this file?
- Are the dummy git credentials doing anything? We're not doing any commits or pushes, so I'm not sure if we need that step
- All of our GitHub Workflows use 2 spaces, not 4. would be great to update this
- We should not need `fetch-depth: 0` here. It doesn't take long in react-native-onyx but is not a good practice.
- Why do we do `npm install --force`, then immediately after `npm install reassure`? Reassure is in the `package.json` so `npm install --force` should be enough on its own
- `shell: bash` is not needed in workflow job run steps, nor is `set -e`
- One blank line between run steps is generally the format we see in these files
- I think it would be made clearer by splitting up the baseline and delta checks into separate steps:

```yml
name: Reassure Performance Tests

on:
  pull_request:
    types: [opened, synchronize]
    branches-ignore: [staging, production]
    paths-ignore: ['**.md', '**.sh']

jobs:
  # Note: We run baseline and delta performance checks in the same runner to reduce hardware variance across machines
  perf-tests:
    if: ${{ github.actor != 'OSBotify' }}
    runs-on: ubuntu-latest
    steps:
      # v4
      - name: Checkout
        uses: actions/checkout@8ade135a41bc03ea155e62e844d188df1ea18608
        with:
          ref: ${{ github.event.pull_request.base.ref }}

      # v4
      - name: Setup Node
        uses: actions/setup-node@cdca7365b2dadb8aad0a33bc7601856ffabcc48e
        with:
          node-version-file: '.nvmrc'

      - name: Run reassure baseline tests
        run: npx reassure --baseline

      - name: Checkout PR head ref
        run: |
          git fetch origin ${{ github.event.pull_request.head.ref }} --no-tags --depth=1
          git switch --force --detach ${{ github.event.pull_request.head.ref }}
          
      - name: Reinstall node_modules
        run: npm install --force

      - name: Run reassure delta tests
        run: npx reassure --branch

      - name: Validate output.json
        id: validateReassureOutput
        uses: ./.github/actions/javascript/validateReassureOutput
        with:
          DURATION_DEVIATION_PERCENTAGE: 20
```

tests/perf-test/Onyx.perf-test.ts
```diff
@@ -0,0 +1,168 @@
...
+const collectionKey = ONYXKEYS.COLLECTION.TEST_KEY;
+const mockedReportActionsMap = getRandomReportActions(collectionKey);
+
+const clearOnyxAfterEachMeasure = async () => {
+    await Onyx.clear();
+};
+
+describe('Onyx', () => {
+    beforeAll(async () => {
+        Onyx.init({
```
roryabraham, line 32: What storage provider, if any, are we using for these tests? Just the in-memory provider?

roryabraham (changes requested):

tests/perf-test/Onyx.perf-test.ts
fabioh8010: We currently use [MemoryOnlyProvider.ts](https://github.com/Expensify/react-native-onyx/blob/main/lib/storage/providers/MemoryOnlyProvider.ts) for both unit and performance tests. I plan to add support to test in all providers (SQLiteProvider, IDBKeyValProvider) in the future, since we end up lacking coverage with current structure.

.github/workflows/reassurePerfTests.yml
fabioh8010: I tried the script but I'm running in some problems tried using the the Reassure script from E/App (with some adaptations) but still facing issues with the workflow, I will give it a try a couple more times.

lib/OnyxUtils.ts
fabioh8010: As discussed in NewDot, I will handle this in a separate PR 👍

.github/scripts/buildActions.sh
fabioh8010: Thanks @roryabraham ❤️ I tested this script with current file and adding more test ones, and it's working great.

.github/workflows/reassurePerfTests.yml
fabioh8010: Ok, I figured it out:

1. Using `github.event.pull_request.head.ref` was not working because, according to GPT, GitHub Actions checks out a shallow clone of the repository and doesn’t include all branches by default. It recommended using `github.event.pull_request.head.sha` as it will checkout the specific commit instead of relying on branch names.
2. Additionally, I had to add NPM installation step before the baseline Reassure test.

roryabraham (approved):

Issue #615, opened 2025-02-13
Incorrect result for Onyx update with delete and object property delete


neil-marcellini:
## Problem
When there is an item set in a collection in Onyx, and Onyx.update is called with two merge updates where the first sets the collection item to null in order to remove it, and the second one sets a property of the object to null, then the item remains unchanged in the collection when it should be set to an empty object.

Please see the [related App issue](https://github.com/Expensify/App/issues/55199) for more context.

<details>
<summary>Issue example</summary>

```js
// Given we have a report action to set into Onyx
window.Onyx.set('reportActions_8865516015258724', {
    '5738984309614092595': {
        reportActionID: '5738984309614092595',
        someKey: 'someValue',
    },
});

const queuedUpdates = [
    {
        key: 'reportActions_8865516015258724',
        onyxMethod: 'merge',
        value: {
            '5738984309614092595': null,
        },
    },
    {
        onyxMethod: 'merge',
        key: 'reportActions_8865516015258724',
        value: {
            '5738984309614092595': {
                pendingAction: null,
            },
        },
    },
];
window.Onyx.update(queuedUpdates);

// Then the object under the key 5738984309614092595 in the collection reportActions_8865516015258724, should be set to an empty object, but it's actually set to the original value.
window.Onyx.get('reportActions_8865516015258724').then((val) => console.log('Result after updates', val['5738984309614092595']));

```
</details>

<details>
<summary>Proof that setting an action to an empty object will remove it from the UI in Expensify/App</summary>

```js
// Get actions from a report. Replace the id (8865516015258724) with the reportID you want to test.
window.Onyx.get('reportActions_8865516015258724').then((val) => {
    const actionsArray = Object.values(val);
    actionsArray.sort((a, b) => new Date(a.created) - new Date(b.created));

    // Replace the most recent action with an empty object
    const replacedActionID = actionsArray[actionsArray.length - 1].reportActionID;
    actionsArray[actionsArray.length - 1] = {};

    const sortedActions = actionsArray.reduce((acc, action) => {
        acc[action.reportActionID ?? replacedActionID] = action;
        return acc;
    }, {});

    console.log('console test sortedActions', sortedActions);

    // Set the new collection into Onyx
    window.Onyx.set('reportActions_8865516015258724', sortedActions);

    // You should see that the most recent action is removed from the UI
});
```
</details>

## Solution
Find the root cause and fix it.

fabioh8010:
[...]

PR #620, opened 2025-03-03
Replace old value after `null` merges in nested properties when batching merges/updates

fabioh8010:
### Details

https://github.com/Expensify/react-native-onyx/issues/615#issuecomment-2681745461

This PR:

- Changes `Onyx.merge()`/`Onyx.update()` batching logic to replace the current value of a nested property after a `null` change in it, so the subsequent updates of that batch in this nested property can fully reset its data.
- Introduces `OnyxMerge` file that will handle the merging of data in different ways depending on the platform
  - For native, we use `mergeItem` that will take advantage of `JSON_PATCH` and `JSON_REPLACE` SQL operations to merge the object in a performant way. `JSON_PATCH` operations were already being used before, and now we also use `JSON_REPLACE` operations to fully replace the marked nested objects when merging data.
  - For web, we use `setItem` since the object was already merged with its changes before.
- Several changes to `fastMerge`/`mergeObject` in order to mark the nested objects we need to fully replace, more details are commented in the functions.
- Removes `OnyxUtils.removeNullValues` since it won't be necessary anymore after the changes in this PR.
- Creates additional `OnyxUtils.mergeChanges`/`OnyxUtils.mergeAndMarkChanges` to handle the batching of merge changes or just merge data into a value.
- Several unit tests, comments added and types improved.

### Related Issues

https://github.com/Expensify/react-native-onyx/issues/615

### Automated Tests

Unit tests were added to cover these changes.

### Manual Tests

Same as https://github.com/Expensify/App/issues/55199.

### Author Checklist

    - [x] Android / native
    - [x] Android / Chrome
    - [x] iOS / native
    - [x] iOS / Safari
    - [x] MacOS / Chrome / Safari
    - [x] MacOS / Desktop

### Screenshots/Videos
<details>

https://github.com/user-attachments/assets/74347569-094e-4519-a13d-51fb11315882

</details>

<details>

I'm having problems with my emulators when opening the Chrome app (they crash instantly), so I couldn't record videos for this platform.

</details>

<details>

https://github.com/user-attachments/assets/b79093c2-78a3-4333-a8e8-08cfa2be3a4c

</details>

<details>

https://github.com/user-attachments/assets/10a313ed-ddfc-47b9-94fa-dcea950e4ea0

</details>

<details>

https://github.com/user-attachments/assets/e3b389be-01ee-45f8-bc61-fa81b7274204

https://github.com/user-attachments/assets/872d0958-ab04-433f-b59f-458cb49d2309

</details>

<details>

https://github.com/user-attachments/assets/392971f8-0754-4cf7-83b6-40493ea7be81

</details>

### DIFF
--- a/API-INTERNAL.md
+++ b/API-INTERNAL.md
@@ -149,6 +149,11 @@ It will also mark deep nested objects that need to be entirely replaced during t
 <dt><a href="#unsubscribeFromKey">unsubscribeFromKey(subscriptionID)</a></dt>
 <dd><p>Disconnects and removes the listener from the Onyx key.</p>
 </dd>
+<dt><a href="#mergeCollectionWithPatches">mergeCollectionWithPatches(collectionKey, collection, mergeReplaceNullPatches)</a></dt>
+<dd><p>Merges a collection based on their keys.
+Serves as core implementation for <code>Onyx.mergeCollection()</code> public function, the difference being
+that this internal function allows passing an additional <code>mergeReplaceNullPatches</code> parameter.</p>
+</dd>
 <dt><a href="#clearOnyxUtilsInternals">clearOnyxUtilsInternals()</a></dt>
 <dd><p>Clear internal variables used in this file, useful in test environments.</p>
 </dd>
@@ -502,6 +507,21 @@ Disconnects and removes the listener from the Onyx key.
 | Param | Description |
 | --- | --- |
 | subscriptionID | Subscription ID returned by calling `OnyxUtils.subscribeToKey()`. |
+
+<a name="mergeCollectionWithPatches"></a>
+
+## mergeCollectionWithPatches(collectionKey, collection, mergeReplaceNullPatches)
+Merges a collection based on their keys.
+Serves as core implementation for `Onyx.mergeCollection()` public function, the difference being
+that this internal function allows passing an additional `mergeReplaceNullPatches` parameter.
+
+**Kind**: global function  
+
+| Param | Description |
+| --- | --- |
+| collectionKey | e.g. `ONYXKEYS.COLLECTION.REPORT` |
+| collection | Object collection keyed by individual collection member keys and values |
+| mergeReplaceNullPatches | Record where the key is a collection member key and the value is a list of tuples that we'll use to replace the nested objects of that collection member record with something else. |
 
 <a name="clearOnyxUtilsInternals"></a>
 
--- a/API.md
+++ b/API.md
@@ -28,8 +28,8 @@ Values of type <code>Object</code> get merged with the old value, whilst for <co
 applied in the order they were called. Note: <code>Onyx.set()</code> calls do not work this way so use caution when mixing
 <code>Onyx.merge()</code> and <code>Onyx.set()</code>.</p>
 </dd>
-<dt><a href="#mergeCollection">mergeCollection(collectionKey, collection, mergeReplaceNullPatches)</a></dt>
-<dd><p>Merges a collection based on their keys</p>
+<dt><a href="#mergeCollection">mergeCollection(collectionKey, collection)</a></dt>
+<dd><p>Merges a collection based on their keys.</p>
 </dd>
 <dt><a href="#clear">clear(keysToPreserve)</a></dt>
 <dd><p>Clear out all the data in the store</p>
@@ -157,8 +157,8 @@ Onyx.merge(ONYXKEYS.POLICY, {name: 'My Workspace'}); // -> {id: 1, name: 'My Wor
 ```
 <a name="mergeCollection"></a>
 
-## mergeCollection(collectionKey, collection, mergeReplaceNullPatches)
-Merges a collection based on their keys
+## mergeCollection(collectionKey, collection)
+Merges a collection based on their keys.
 
 **Kind**: global function  
 
@@ -166,7 +166,6 @@ Merges a collection based on their keys
 | --- | --- |
 | collectionKey | e.g. `ONYXKEYS.COLLECTION.REPORT` |
 | collection | Object collection keyed by individual collection member keys and values |
-| mergeReplaceNullPatches | Record where the key is a collection member key and the value is a list of tuples that we'll use to replace the nested objects of that collection member record with something else. |
 
 **Example**  
 ```js
--- a/lib/Onyx.ts
+++ b/lib/Onyx.ts
@@ -24,7 +24,6 @@ import type {
     OnyxValue,
     OnyxInput,
     OnyxMethodMap,
-    MultiMergeReplaceNullPatches,
 } from './types';
 import OnyxUtils from './OnyxUtils';
 import logMessages from './logMessages';
@@ -345,7 +344,7 @@ function merge<TKey extends OnyxKey>(key: TKey, changes: OnyxMergeInput<TKey>): 
 }
 
 /**
- * Merges a collection based on their keys
+ * Merges a collection based on their keys.
  *
  * @example
  *
@@ -356,125 +355,9 @@ function merge<TKey extends OnyxKey>(key: TKey, changes: OnyxMergeInput<TKey>): 
  *
  * @param collectionKey e.g. `ONYXKEYS.COLLECTION.REPORT`
  * @param collection Object collection keyed by individual collection member keys and values
- * @param mergeReplaceNullPatches Record where the key is a collection member key and the value is a list of
- * tuples that we'll use to replace the nested objects of that collection member record with something else.
  */
-function mergeCollection<TKey extends CollectionKeyBase, TMap>(
-    collectionKey: TKey,
-    collection: OnyxMergeCollectionInput<TKey, TMap>,
-    mergeReplaceNullPatches?: MultiMergeReplaceNullPatches,
-): Promise<void> {
-    if (!OnyxUtils.isValidNonEmptyCollectionForMerge(collection)) {
-        Logger.logInfo('mergeCollection() called with invalid or empty value. Skipping this update.');
-        return Promise.resolve();
-    }
-
-    let resultCollection: OnyxInputKeyValueMapping = collection;
-    let resultCollectionKeys = Object.keys(resultCollection);
-
-    // Confirm all the collection keys belong to the same parent
-    if (!OnyxUtils.doAllCollectionItemsBelongToSameParent(collectionKey, resultCollectionKeys)) {
-        return Promise.resolve();
-    }
-
-    const skippableCollectionMemberIDs = OnyxUtils.getSkippableCollectionMemberIDs();
-    if (skippableCollectionMemberIDs.size) {
-        resultCollection = resultCollectionKeys.reduce((result: OnyxInputKeyValueMapping, key) => {
-            try {
-                const [, collectionMemberID] = OnyxUtils.splitCollectionMemberKey(key, collectionKey);
-                // If the collection member key is a skippable one we set its value to null.
-                // eslint-disable-next-line no-param-reassign
-                result[key] = !skippableCollectionMemberIDs.has(collectionMemberID) ? resultCollection[key] : null;
-            } catch {
-                // Something went wrong during split, so we assign the data to result anyway.
-                // eslint-disable-next-line no-param-reassign
-                result[key] = resultCollection[key];
-            }
-
-            return result;
-        }, {});
-    }
-    resultCollectionKeys = Object.keys(resultCollection);
-
-    return OnyxUtils.getAllKeys()
-        .then((persistedKeys) => {
-            // Split to keys that exist in storage and keys that don't
-            const keys = resultCollectionKeys.filter((key) => {
-                if (resultCollection[key] === null) {
-                    OnyxUtils.remove(key);
-                    return false;
-                }
-                return true;
-            });
-
-            const existingKeys = keys.filter((key) => persistedKeys.has(key));
-
-            const cachedCollectionForExistingKeys = OnyxUtils.getCachedCollection(collectionKey, existingKeys);
-
-            const existingKeyCollection = existingKeys.reduce((obj: OnyxInputKeyValueMapping, key) => {
-                const {isCompatible, existingValueType, newValueType} = utils.checkCompatibilityWithExistingValue(resultCollection[key], cachedCollectionForExistingKeys[key]);
-
-                if (!isCompatible) {
-                    Logger.logAlert(logMessages.incompatibleUpdateAlert(key, 'mergeCollection', existingValueType, newValueType));
-                    return obj;
-                }
-
-                // eslint-disable-next-line no-param-reassign
-                obj[key] = resultCollection[key];
-                return obj;
-            }, {}) as Record<OnyxKey, OnyxInput<TKey>>;
-
-            const newCollection: Record<OnyxKey, OnyxInput<TKey>> = {};
-            keys.forEach((key) => {
-                if (persistedKeys.has(key)) {
-                    return;
-                }
-                newCollection[key] = resultCollection[key];
-            });
-
-            // When (multi-)merging the values with the existing values in storage,
-            // we don't want to remove nested null values from the data that we pass to the storage layer,
-            // because the storage layer uses them to remove nested keys from storage natively.
-            const keyValuePairsForExistingCollection = OnyxUtils.prepareKeyValuePairsForStorage(existingKeyCollection, false, mergeReplaceNullPatches);
-
-            // We can safely remove nested null values when using (multi-)set,
-            // because we will simply overwrite the existing values in storage.
-            const keyValuePairsForNewCollection = OnyxUtils.prepareKeyValuePairsForStorage(newCollection, true);
-
-            const promises = [];
-
-            // We need to get the previously existing values so we can compare the new ones
-            // against them, to avoid unnecessary subscriber updates.
-            const previousCollectionPromise = Promise.all(existingKeys.map((key) => OnyxUtils.get(key).then((value) => [key, value]))).then(Object.fromEntries);
-
-            // New keys will be added via multiSet while existing keys will be updated using multiMerge
-            // This is because setting a key that doesn't exist yet with multiMerge will throw errors
-            if (keyValuePairsForExistingCollection.length > 0) {
-                promises.push(Storage.multiMerge(keyValuePairsForExistingCollection));
-            }
-
-            if (keyValuePairsForNewCollection.length > 0) {
-                promises.push(Storage.multiSet(keyValuePairsForNewCollection));
-            }
-
-            // finalMergedCollection contains all the keys that were merged, without the keys of incompatible updates
-            const finalMergedCollection = {...existingKeyCollection, ...newCollection};
-
-            // Prefill cache if necessary by calling get() on any existing keys and then merge original data to cache
-            // and update all subscribers
-            const promiseUpdate = previousCollectionPromise.then((previousCollection) => {
-                cache.merge(finalMergedCollection);
-                return OnyxUtils.scheduleNotifyCollectionSubscribers(collectionKey, finalMergedCollection, previousCollection);
-            });
-
-            return Promise.all(promises)
-                .catch((error) => OnyxUtils.evictStorageAndRetry(error, mergeCollection, collectionKey, resultCollection))
-                .then(() => {
-                    OnyxUtils.sendActionToDevTools(OnyxUtils.METHOD.MERGE_COLLECTION, undefined, resultCollection);
-                    return promiseUpdate;
-                });
-        })
-        .then(() => undefined);
+function mergeCollection<TKey extends CollectionKeyBase, TMap>(collectionKey: TKey, collection: OnyxMergeCollectionInput<TKey, TMap>): Promise<void> {
+    return OnyxUtils.mergeCollectionWithPatches(collectionKey, collection);
 }
 
 /**
@@ -708,7 +591,11 @@ function update(data: OnyxUpdate[]): Promise<void> {
 
         if (!utils.isEmptyObject(batchedCollectionUpdates.merge)) {
             promises.push(() =>
-                mergeCollection(collectionKey, batchedCollectionUpdates.merge as Collection<CollectionKey, unknown, unknown>, batchedCollectionUpdates.mergeReplaceNullPatches),
+                OnyxUtils.mergeCollectionWithPatches(
+                    collectionKey,
+                    batchedCollectionUpdates.merge as Collection<CollectionKey, unknown, unknown>,
+                    batchedCollectionUpdates.mergeReplaceNullPatches,
+                ),
             );
         }
         if (!utils.isEmptyObject(batchedCollectionUpdates.set)) {
--- a/lib/OnyxUtils.ts
+++ b/lib/OnyxUtils.ts
@@ -25,6 +25,7 @@ import type {
     OnyxCollection,
     OnyxEntry,
     OnyxInput,
+    OnyxInputKeyValueMapping,
     OnyxKey,
     OnyxMergeCollectionInput,
     OnyxUpdate,
@@ -39,6 +40,7 @@ import createDeferredTask from './createDeferredTask';
 import * as GlobalSettings from './GlobalSettings';
 import decorateWithMetrics from './metrics';
 import type {StorageKeyValuePair} from './storage/providers/types';
+import logMessages from './logMessages';
 
 // Method constants
 const METHOD = {
@@ -361,7 +363,7 @@ function multiGet<TKey extends OnyxKey>(keys: CollectionKeyBase[]): Promise<Map<
                 values.forEach(([key, value]) => {
                     if (skippableCollectionMemberIDs.size) {
                         try {
-                            const [, collectionMemberID] = OnyxUtils.splitCollectionMemberKey(key);
+                            const [, collectionMemberID] = splitCollectionMemberKey(key);
                             if (skippableCollectionMemberIDs.has(collectionMemberID)) {
                                 // The key is a skippable one, so we skip this iteration.
                                 return;
@@ -387,7 +389,7 @@ function multiGet<TKey extends OnyxKey>(keys: CollectionKeyBase[]): Promise<Map<
  * Note: just using `.map`, you'd end up with `Array<OnyxCollection<Report>|OnyxEntry<string>>`, which is not what we want. This preserves the order of the keys provided.
  */
 function tupleGet<Keys extends readonly OnyxKey[]>(keys: Keys): Promise<{[Index in keyof Keys]: OnyxValue<Keys[Index]>}> {
-    return Promise.all(keys.map((key) => OnyxUtils.get(key))) as Promise<{[Index in keyof Keys]: OnyxValue<Keys[Index]>}>;
+    return Promise.all(keys.map((key) => get(key))) as Promise<{[Index in keyof Keys]: OnyxValue<Keys[Index]>}>;
 }
 
 /**
@@ -1444,12 +1446,12 @@ function unsubscribeFromKey(subscriptionID: number): void {
 }
 
 function updateSnapshots(data: OnyxUpdate[], mergeFn: typeof Onyx.merge): Array<() => Promise<void>> {
-    const snapshotCollectionKey = OnyxUtils.getSnapshotKey();
+    const snapshotCollectionKey = getSnapshotKey();
     if (!snapshotCollectionKey) return [];
 
     const promises: Array<() => Promise<void>> = [];
 
-    const snapshotCollection = OnyxUtils.getCachedCollection(snapshotCollectionKey);
+    const snapshotCollection = getCachedCollection(snapshotCollectionKey);
 
     Object.entries(snapshotCollection).forEach(([snapshotEntryKey, snapshotEntryValue]) => {
         // Snapshots may not be present in cache. We don't know how to update them so we skip.
@@ -1461,7 +1463,7 @@ function updateSnapshots(data: OnyxUpdate[], mergeFn: typeof Onyx.merge): Array<
 
         data.forEach(({key, value}) => {
             // snapshots are normal keys so we want to skip update if they are written to Onyx
-            if (OnyxUtils.isCollectionMemberKey(snapshotCollectionKey, key)) {
+            if (isCollectionMemberKey(snapshotCollectionKey, key)) {
                 return;
             }
 
@@ -1507,6 +1509,133 @@ function updateSnapshots(data: OnyxUpdate[], mergeFn: typeof Onyx.merge): Array<
     });
 
     return promises;
+}
+
+/**
+ * Merges a collection based on their keys.
+ * Serves as core implementation for `Onyx.mergeCollection()` public function, the difference being
+ * that this internal function allows passing an additional `mergeReplaceNullPatches` parameter.
+ *
+ * @param collectionKey e.g. `ONYXKEYS.COLLECTION.REPORT`
+ * @param collection Object collection keyed by individual collection member keys and values
+ * @param mergeReplaceNullPatches Record where the key is a collection member key and the value is a list of
+ * tuples that we'll use to replace the nested objects of that collection member record with something else.
+ */
+function mergeCollectionWithPatches<TKey extends CollectionKeyBase, TMap>(
+    collectionKey: TKey,
+    collection: OnyxMergeCollectionInput<TKey, TMap>,
+    mergeReplaceNullPatches?: MultiMergeReplaceNullPatches,
+): Promise<void> {
+    if (!isValidNonEmptyCollectionForMerge(collection)) {
+        Logger.logInfo('mergeCollection() called with invalid or empty value. Skipping this update.');
+        return Promise.resolve();
+    }
+
+    let resultCollection: OnyxInputKeyValueMapping = collection;
+    let resultCollectionKeys = Object.keys(resultCollection);
+
+    // Confirm all the collection keys belong to the same parent
+    if (!doAllCollectionItemsBelongToSameParent(collectionKey, resultCollectionKeys)) {
+        return Promise.resolve();
+    }
+
+    if (skippableCollectionMemberIDs.size) {
+        resultCollection = resultCollectionKeys.reduce((result: OnyxInputKeyValueMapping, key) => {
+            try {
+                const [, collectionMemberID] = splitCollectionMemberKey(key, collectionKey);
+                // If the collection member key is a skippable one we set its value to null.
+                // eslint-disable-next-line no-param-reassign
+                result[key] = !skippableCollectionMemberIDs.has(collectionMemberID) ? resultCollection[key] : null;
+            } catch {
+                // Something went wrong during split, so we assign the data to result anyway.
+                // eslint-disable-next-line no-param-reassign
+                result[key] = resultCollection[key];
+            }
+
+            return result;
+        }, {});
+    }
+    resultCollectionKeys = Object.keys(resultCollection);
+
+    return getAllKeys()
+        .then((persistedKeys) => {
+            // Split to keys that exist in storage and keys that don't
+            const keys = resultCollectionKeys.filter((key) => {
+                if (resultCollection[key] === null) {
+                    remove(key);
+                    return false;
+                }
+                return true;
+            });
+
+            const existingKeys = keys.filter((key) => persistedKeys.has(key));
+
+            const cachedCollectionForExistingKeys = getCachedCollection(collectionKey, existingKeys);
+
+            const existingKeyCollection = existingKeys.reduce((obj: OnyxInputKeyValueMapping, key) => {
+                const {isCompatible, existingValueType, newValueType} = utils.checkCompatibilityWithExistingValue(resultCollection[key], cachedCollectionForExistingKeys[key]);
+
+                if (!isCompatible) {
+                    Logger.logAlert(logMessages.incompatibleUpdateAlert(key, 'mergeCollection', existingValueType, newValueType));
+                    return obj;
+                }
+
+                // eslint-disable-next-line no-param-reassign
+                obj[key] = resultCollection[key];
+                return obj;
+            }, {}) as Record<OnyxKey, OnyxInput<TKey>>;
+
+            const newCollection: Record<OnyxKey, OnyxInput<TKey>> = {};
+            keys.forEach((key) => {
+                if (persistedKeys.has(key)) {
+                    return;
+                }
+                newCollection[key] = resultCollection[key];
+            });
+
+            // When (multi-)merging the values with the existing values in storage,
+            // we don't want to remove nested null values from the data that we pass to the storage layer,
+            // because the storage layer uses them to remove nested keys from storage natively.
+            const keyValuePairsForExistingCollection = prepareKeyValuePairsForStorage(existingKeyCollection, false, mergeReplaceNullPatches);
+
+            // We can safely remove nested null values when using (multi-)set,
+            // because we will simply overwrite the existing values in storage.
+            const keyValuePairsForNewCollection = prepareKeyValuePairsForStorage(newCollection, true);
+
+            const promises = [];
+
+            // We need to get the previously existing values so we can compare the new ones
+            // against them, to avoid unnecessary subscriber updates.
+            const previousCollectionPromise = Promise.all(existingKeys.map((key) => get(key).then((value) => [key, value]))).then(Object.fromEntries);
+
+            // New keys will be added via multiSet while existing keys will be updated using multiMerge
+            // This is because setting a key that doesn't exist yet with multiMerge will throw errors
+            if (keyValuePairsForExistingCollection.length > 0) {
+                promises.push(Storage.multiMerge(keyValuePairsForExistingCollection));
+            }
+
+            if (keyValuePairsForNewCollection.length > 0) {
+                promises.push(Storage.multiSet(keyValuePairsForNewCollection));
+            }
+
+            // finalMergedCollection contains all the keys that were merged, without the keys of incompatible updates
+            const finalMergedCollection = {...existingKeyCollection, ...newCollection};
+
+            // Prefill cache if necessary by calling get() on any existing keys and then merge original data to cache
+            // and update all subscribers
+            const promiseUpdate = previousCollectionPromise.then((previousCollection) => {
+                cache.merge(finalMergedCollection);
+                return scheduleNotifyCollectionSubscribers(collectionKey, finalMergedCollection, previousCollection);
+            });
+
+            return Promise.all(promises)
+                .catch((error) => evictStorageAndRetry(error, mergeCollectionWithPatches, collectionKey, resultCollection))
+                .then(() => {
+                    sendActionToDevTools(METHOD.MERGE_COLLECTION, undefined, resultCollection);
+                    return promiseUpdate;
+                });
+        })
+        .then(() => undefined);
 }
 
 function logKeyChanged(onyxMethod: Extract<OnyxMethod, 'set' | 'merge'>, key: OnyxKey, value: unknown, hasChanged: boolean) {
@@ -1578,6 +1707,7 @@ const OnyxUtils = {
     addKeyToRecentlyAccessedIfNeeded,
     reduceCollectionWithSelector,
     updateSnapshots,
+    mergeCollectionWithPatches,
     logKeyChanged,
     logKeyRemoved,
 };


fabioh8010:
Strangely the `should replace the old value after a null merge in a nested property when batching updates` and `should replace the old value after a null merge in a nested property when batching merges` tests are partially passing, which I was not expecting to happen.

Maybe the `fastMerge`/`mergeObject` original logic are partially working somehow? I need to investigate.

fabioh8010:
My tests were "wrong", now it will output the errors in the two scenarios of both Onyx.update() and Onyx.merge()

fabioh8010:
cc @chrispader @neil-marcellini

lib/Onyx.ts
```diff
@@ -307,8 +307,10 @@ function merge<TKey extends OnyxKey>(key: TKey, changes: OnyxMergeInput<TKey>):
         }
 
         try {
-            // We first only merge the changes, so we can provide these to the native implementation (SQLite uses only delta changes in "JSON_PATCH" to merge)
-            // We don't want to remove null values from the "batchedDeltaChanges", because SQLite uses them to remove keys from storage natively.
+            // We first only merge the changes, so we can use our custom merging strategy by signalling OnyxUtils.applyMerge()
```
lakchote, line 310:
```suggestion
            // We first only merge the changes, so we can use our custom merging strategy by signaling OnyxUtils.applyMerge()
```

lib/storage/providers/SQLiteProvider.ts
```diff
@@ -60,28 +61,51 @@ const provider: StorageProvider = {
...
-        `;
+        const nonNullishPairs: KeyValuePairList = [];
+        const nonNullishPairsKeys: OnyxKey[] = [];
 
-        const nonNullishPairs = pairs.filter((pair) => pair[1] !== undefined);
-        const queryArguments = nonNullishPairs.map((pair) => {
-            const value = JSON.stringify(pair[1]);
-            return [pair[0], value];
-        });
+        // eslint-disable-next-line @typescript-eslint/prefer-for-of
```
lakchote, line 67: Why did you add this rule disabling (that is repeated through the PR)?

Couldn't you do something like `for (const pair of pairs) {...}`?

lib/storage/providers/SQLiteProvider.ts
```diff
@@ -60,28 +61,51 @@ const provider: StorageProvider = {
...
+        if (nonNullishPairs.length === 0) {
+            return Promise.resolve();
         }
 
-        return this.multiMerge([[key, deltaChanges]]) as Promise<BatchQueryResult>;
+        return this.multiGet(nonNullishPairsKeys).then((storagePairs) => {
+            // multiGet() is not guaranteed to return the data in the same order we asked with "nonNullishPairsKeys",
+            // so we use a map to associate keys to their existing values correctly.
+            const existingMap = new Map<OnyxKey, OnyxValue<OnyxKey>>();
+            // eslint-disable-next-line @typescript-eslint/prefer-for-of
```
lakchote, line 84: Same comment as above related to disabling this rule.

lib/storage/providers/SQLiteProvider.ts
```diff
@@ -60,28 +61,51 @@ const provider: StorageProvider = {
...
+            for (let i = 0; i < storagePairs.length; i++) {
+                existingMap.set(storagePairs[i][0], storagePairs[i][1]);
+            }
+
+            const newPairs: KeyValuePairList = [];
+
+            // eslint-disable-next-line @typescript-eslint/prefer-for-of
```
lakchote, line 91: Same comment as above related to disabling this rule.

lib/utils.ts
```diff
@@ -78,7 +90,28 @@ function mergeObject<TObject extends Record<string, unknown>>(target: TObject |
...
                 // to ensure that nested null values are removed from the merged object.
                 const targetValueWithFallback = (targetValue ?? {}) as TObject;
-                destination[key] = fastMerge(targetValueWithFallback, sourceValue, shouldRemoveNestedNulls);
+
+                // If we are batching merge changes and the previous merge change (targetValue) is null,
+                // it means we want to fully replace this object when merging the batched changes with the Onyx value.
+                // To achieve this, we first mark these nested objects with an internal flag. With the desired objects
+                // marked, when calling this method again with "shouldReplaceMarkedObjects" set to true we can proceed
+                // to effectively replace them in the next condition.
+                if (isBatchingMergeChanges && targetValue === null) {
```
lakchote, line 99: Could we do something like this?

```suggestion
                if (isBatchingMergeChanges && !targetValue) {
```

lib/utils.ts
```diff
@@ -78,7 +90,28 @@ function mergeObject<TObject extends Record<string, unknown>>(target: TObject |
...
+                    (targetValueWithFallback as Record<string, unknown>)[ONYX_INTERNALS__REPLACE_OBJECT_MARK] = true;
+                }
+
+                // Then, when merging the batched changes with the Onyx value, if a nested object of the batched changes
+                // has the internal flag set, we replace the entire destination object with the source one and remove
+                // the flag.
+                if (shouldReplaceMarkedObjects && sourceValue[ONYX_INTERNALS__REPLACE_OBJECT_MARK] === true) {
```
lakchote, line 106: Could we do something like this?

```suggestion
                if (shouldReplaceMarkedObjects && sourceValue[ONYX_INTERNALS__REPLACE_OBJECT_MARK]) {
```

lakchote (commented):
Left some comments

tests/unit/fastMergeTest.ts
```diff
@@ -86,14 +86,14 @@ describe('fastMerge', () => {
...
         expect(result).toEqual([1, 2, 3]);
     });
 
     it('should replace an array with an object', () => {
         // eslint-disable-next-line @typescript-eslint/no-explicit-any
-        const result = utils.fastMerge([1, 2, 3] as any, testObject);
+        const result = utils.fastMerge([1, 2, 3] as any, testObject, true, false, false);
 
         expect(result).toEqual(testObject);
     });
```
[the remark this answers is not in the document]
chrispader: Basically test for the extra two parameters that were added

tests/unit/onyxTest.ts
```diff
@@ -1698,7 +1698,386 @@ describe('Onyx', () => {
...
+                        id: 'sub_entry2',
+                        someKey: 'someValue',
+                        someNestedObject: {
+                            someNestedKey: 'someNestedValue',
+                            anotherNestedObject: {
+                                anotherNestedKey: 'anotherNestedValue',
+                            },
+                            anotherNestedObject2: {
+                                anotherNestedKey2: 'anotherNestedValue2',
+                            },
```
neil-marcellini, line 1773: NAB: What case or cases exactly are we testing here? It's kind of chaotic. I would prefer multiple tests of specific cases so it's a bit easier to understand.

tests/unit/onyxTest.ts
```diff
@@ -1698,7 +1698,386 @@ describe('Onyx', () => {
...
+                    onyxMethod: 'merge',
+                    value: {
+                        sub_entry2: {
+                            someNestedObject: {
+                                anotherNestedObject: {
+                                    newNestedKey: 'newNestedValue',
+                                },
+                                // Removing "anotherNestedObject3" property in this update.
+                                // This property was only introduced in a previous update, so we don't need to care
+                                // about an old value because there isn't one.
```
neil-marcellini, line 1856: I find it really hard to follow these tests. What would make it easier for me is if we appended updates to the updates array one at a time while setting up the test, and as we do so update an `expectedValue` variable. I think that would make it much more clear.

Otherwise I need to break out a pen and paper to track how all the changes should happen in the test 😅 📝

lib/storage/providers/SQLiteProvider.ts
```diff
@@ -60,28 +61,51 @@ const provider: StorageProvider = {
...
+
+                const existingValue = existingMap.get(key) ?? {};
+
+                const mergedValue = utils.fastMerge(existingValue, newValue, true, false, true);
+
+                newPairs.push([key, mergedValue]);
+            }
+
+            return this.multiSet(newPairs);
+        });
```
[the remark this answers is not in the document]
neil-marcellini: Yes I think it would be a good idea to measure performance in the app before and after this change to make sure it doesn't cause a large performance hit.

marcaaron:
>now we have some customizations (this very PR) that is impossible to replicate with JSON_PATCH

Can you please update the description to mention which customizations are not compatible with JSON_PATCH? I will review today, thanks!

marcaaron:
Is it possible to reword some of this for clarity?

>Changes our merging strategy (defined in applyMerge/fastMerge/mergeObject) to have two additional flags: isBatchingMergeChanges and shouldReplaceMarkedObjects. The isBatchingMergeChanges flag will be used when batching any queued merge changes, as it has a special logic to handle the replacement of old values after null merges. 

Specifically, "handle the replacement of old values after null merges" didn't quite make sense to me.

>The shouldReplaceMarkedObjects flag will be used when applying this batched merge changes to the existing value, using this special logic to effectively replace the old values we desire.

This part of the description is also a bit confusing. Maybe you can also explain what a "marked object" is in the description? Thanks!

lib/Onyx.ts
```diff
@@ -307,8 +307,10 @@ function merge<TKey extends OnyxKey>(key: TKey, changes: OnyxMergeInput<TKey>):
         }
 
         try {
-            // We first only merge the changes, so we can provide these to the native implementation (SQLite uses only delta changes in "JSON_PATCH" to merge)
```
marcaaron, line 310: A part of me is confused about why we are getting rid of JSON_PATCH. Another part believes that this was a _benefit_ since when we do this we end up passing a smaller delta change instead of a potentially large object and therefore it should be more efficient (hand wave there as I do not entirely recall anymore).

lib/Onyx.ts
```diff
@@ -320,23 +322,23 @@ function merge<TKey extends OnyxKey>(key: TKey, changes: OnyxMergeInput<TKey>):
             if (!validChanges.length) {
                 return Promise.resolve();
             }
-            const batchedDeltaChanges = OnyxUtils.applyMerge(undefined, validChanges, false);
+            const batchedDeltaChanges = OnyxUtils.applyMerge(undefined, validChanges, false, true, false);
```
marcaaron, line 325: NAB, `applyMerge()` is getting a bit complex with all the boolean flags. Curious if there is a way to maybe create a new function instead that does what we want and improve the clarity?

lib/Onyx.ts
```diff
@@ -320,23 +322,23 @@ function merge<TKey extends OnyxKey>(key: TKey, changes: OnyxMergeInput<TKey>):
...
 
             // Case (1): When there is no existing value in storage, we want to set the value instead of merge it.
             // Case (2): The presence of a top-level `null` in the merge queue instructs us to drop the whole existing value.
-            // In this case, we can't simply merge the batched changes with the existing value, because then the null in the merge queue would have no effect
+            // In this case, we can't simply merge the batched changes with the existing value, because then the null in the merge queue would have no effect.
```
marcaaron, line 329: I wonder if this comment is making sense still or perhaps just requires clarification? Has "simply merge the batched changes with the existing value" lost it's meaning now that our "merge" has a different behavior?

lib/Onyx.ts
```diff
@@ -346,11 +348,12 @@ function merge<TKey extends OnyxKey>(key: TKey, changes: OnyxMergeInput<TKey>):
                 return Promise.resolve();
             }
 
-            // For providers that can't handle delta changes, we need to merge the batched changes with the existing value beforehand.
```
marcaaron, line 349: flagging this since I still have questions about `JSON_PATCH` before we remove it...

lib/Onyx.ts
```diff
@@ -346,11 +348,12 @@ function merge<TKey extends OnyxKey>(key: TKey, changes: OnyxMergeInput<TKey>):
...
-            // The "preMergedValue" will be directly "set" in storage instead of being merged
-            // Therefore we merge the batched changes with the existing value to get the final merged value that will be stored.
+            // The "preMergedValue" will be directly "set" in storage instead of being merged, so we merge
```
marcaaron, line 351: **thinking out loud:** we should update this to mention why we have a "preMergedValue" at all and why it will be "set". Without that explanation that follows is very confusing. Consider:

- variable is named "preMergedValue" implying that it's a step that will happen before a merge.
- we say it will be "set" instead of "merged"
- we then say we will "merge" the batched changes to get the final "merged" value

Maybe we can use more specific language for these things?

lib/Onyx.ts
```diff
@@ -346,11 +348,12 @@ function merge<TKey extends OnyxKey>(key: TKey, changes: OnyxMergeInput<TKey>):
...
+            // the batched changes with the existing value to get the final merged value that will be stored.
             // We can remove null values from the "preMergedValue", because "null" implicates that the user wants to remove a value from storage.
-            const preMergedValue = OnyxUtils.applyMerge(shouldSetValue ? undefined : existingValue, [batchedDeltaChanges], true);
+            // Additionally, we will signal OnyxUtils.applyMerge() to replace any nested properties previously marked in "batchedDeltaChanges"
+            // by our custom merging strategy.
```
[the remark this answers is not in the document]
marcaaron: I agree. I would love if the idea of "marking" could be broken down or documented somewhere if it's not already. Both "marked" and "pre merge" are a bit difficult to grasp without rolling your sleeves up 😄

lib/Onyx.ts
```diff
@@ -365,7 +368,7 @@ function merge<TKey extends OnyxKey>(key: TKey, changes: OnyxMergeInput<TKey>):
                 return updatePromise;
             }
 
-            return Storage.mergeItem(key, batchedDeltaChanges as OnyxValue<TKey>, preMergedValue as OnyxValue<TKey>, shouldSetValue).then(() => {
+            return Storage.mergeItem(key, preMergedValue as OnyxValue<TKey>).then(() => {
```
marcaaron, line 371: I think maybe `preMergedValue` should be `batchValue`. That seems more clear to me. Like yes, we still have another "merge" to perform. But since we do a "merge" to get to that value it is confusing to refer to it as "pre merge". Just my 2 cents.

lib/OnyxUtils.ts
```diff
@@ -1259,6 +1259,8 @@ function applyMerge<TValue extends OnyxInput<OnyxKey> | undefined, TChange exten
     existingValue: TValue,
     changes: TChange[],
     shouldRemoveNestedNulls: boolean,
+    isBatchingMergeChanges: boolean,
+    shouldReplaceMarkedObjects: boolean,
```
marcaaron, line 1263: Add docs for these ?

lib/storage/providers/SQLiteProvider.ts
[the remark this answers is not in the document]
marcaaron: +1 but I am less convinced that this is "fine for now". It feels like we are undoing work that we had a good reason to do at some point. I trust that we are moving in a good direction, but would rather let some benchmarks do the talking.

marcaaron (changes requested):
This is generally looking good. Thanks for the work on it @fabioh8010 @chrispader. My main feedback (that @neil-marcellini already called out) would be to make sure that `JSON_PATCH` in the native provider is something we can confidently get rid of. If so, then 👍.

lib/storage/providers/SQLiteProvider.ts
fabioh8010: Because we have another rule that forbids use of `for-of`

> iterators/generators require regenerator-runtime, which is too heavyweight for this guide to allow them. Separately, loops should be avoided in favor of array iterations. [no-restricted-syntax](https://eslint.org/docs/latest/rules/no-restricted-syntax)

We have other places in the codebase that are using simple for loops like this one, so I suggest we keep the same for now.

lib/storage/providers/SQLiteProvider.ts
fabioh8010: Answered above

lib/storage/providers/SQLiteProvider.ts
fabioh8010: Answered above

neil-marcellini:
@fabioh8010 please request our reviews again when it's ready

fabioh8010:
> @fabioh8010 please request our reviews again when it's ready

Sure, I started working through the comments today and preparing better explanation for you before requesting.

lib/Onyx.ts
fabioh8010: The comment still makes sense because our merge strategy stays the same regarding what is being described there.

lib/Onyx.ts
fabioh8010: I've decided to simplify Simplify and extract `applyMerge()` batching merges logic to a separate `batchMergeChanges()` function which will take care of this part specifically.

lib/Onyx.ts
fabioh8010: I rephrased the comments, please let me know if it's ok now!

lib/Onyx.ts
fabioh8010: I decided to change to `resultValue`, I think it conveys more the meaning of the variable, wdyt?

lib/OnyxUtils.ts
fabioh8010: Not necessary anymore as I simplified the function.
https://github.com/Expensify/react-native-onyx/pull/620#discussion_r2041183537

tests/unit/onyxTest.ts
fabioh8010: I've redesigned and simplified all the tests following your suggestion, please let me know later wdyt!

tests/unit/onyxTest.ts
fabioh8010: I've redesigned and simplified all the tests following your suggestion, please let me know later wdyt!

lib/Onyx.ts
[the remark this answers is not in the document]
fabioh8010: Agree, I just removed this comment and will add more explanations to the places where we actually have the merging logic

neil-marcellini:
@fabioh8010 what is the latest update? How is it coming along?

fabioh8010:
# Performance Analysis

For these analysis I tested on Android (the only provider that have substantial changes is the native one) and profiled `Storage.multiMerge` and `Storage.mergeItem` using the [Performance API](https://developer.mozilla.org/en-US/docs/Web/API/Performance).

## Baseline (main)

| Test | Method | Duration (ms) |
| - | - | - |
| Merging 1000 records with `mergeCollection()` | Storage.multiMerge | 83 |
| Merging 1 record with `mergeCollection()` | Storage.multiMerge | 4 |
| Merging the same record 10 times with `merge()` | Storage.mergeItem (uses Storage.multiMerge under the hood) | 3 |

## Delta (this PR)

| Test | Method | Duration (ms) |
| - | - | - |
| Merging 1000 records with `mergeCollection()` | Storage.multiMerge | 118 (+42,17%) |
| Merging 1 record with `mergeCollection()` | Storage.multiMerge | 37 (+89,19%) |
| Merging the same record 10 times with `merge()` | Storage.mergeItem (uses Storage.setItem under the hood) | 30 (+86,67%) |

---

As you can see the operations are quite more expensive now, probably due to having to use `this.multiGet()` and then `this.multiSet()` instead of JSON_PATCH.

I will try to look for another solution for the native provider, maybe I can keep the JSON_PATCH and only do additional SQL operations for the properties we need to fully replace.

fabioh8010:
**Updates:**

- Investigating and experimenting a way to keep JSON_PATCH operation and then execute the object replacements with separate SQL operations
- Also investigating why a simple `setItem` operation is taking whole 30 ms even on `main`, while `mergeItem` takes around 3-4 ms on `main`. I feel an SQL "SET" operation should have the same or even less execution time than a JSON_PATCH one.

fabioh8010:
Just FYI that I will be OOO next week and will return on May 5th 🌴

fabioh8010:
**Updates:**

- Resuming work on this one this week.

fabioh8010:
**Updates:**

- Designing/working on the `JSON_PATCH` + separate `JSON_REPLACE` operations solution.

fabioh8010:
**Updates:**

- I think I have a working solution for `JSON_PATCH` + separate `JSON_REPLACE` operations approach, will do a cleanup, more testing and finally the benchmarks.

chrispader:
> **Updates:**
> 
> * I think I have a working solution for `JSON_PATCH` + separate `JSON_REPLACE` operations approach, will do a cleanup, more testing and finally the benchmarks.

Cool! Lmk if you need any help with the Onyx storage layer or SQLite related stuff, since i'm the maintainer of `react-native-nitro-sqlite` i might be able to help there! 🙌🏼

fabioh8010:
> > **Updates:**
> > 
> > * I think I have a working solution for `JSON_PATCH` + separate `JSON_REPLACE` operations approach, will do a cleanup, more testing and finally the benchmarks.
> 
> Cool! Lmk if you need any help with the Onyx storage layer or SQLite related stuff, since i'm the maintainer of `react-native-nitro-sqlite` i might be able to help there! 🙌🏼

For sure man, thanks! I need to do some cleanup before, and for now I won't merge with `main` yet since I don't want to pollute the benchmarks.

fabioh8010:
**Updates:**

- Did the benchmarks and the logic I created to generate the JSON_REPLACE operations are causing some performance impact when merging big collections (like the `Merging 1000 records with mergeCollection()` case) because we need to traverse the objects again to search for the markers.
- To improve that I passed this logic to `mergeObject` because we can reuse the traverse object logic that already happen there to remove the performance impact.
- It seems to be working but needs more testing. Also started some discussion with @chrispader in DM.

fabioh8010:
**Updates:**

- Merged Chris' changes into my [JSON_PATCH solution branch](https://github.com/Expensify/react-native-onyx/compare/main...callstack-internal:react-native-onyx:bugfix/615-solution2-jsonPatchReplace).
- Fixing some TS errors and doing cleanup before proceeding with tests.

fabioh8010:
**Updates:**

- Merged [JSON_PATCH solution branch](https://github.com/Expensify/react-native-onyx/compare/main...callstack-internal:react-native-onyx:bugfix/615-solution2-jsonPatchReplace) into this PR.
- Did manual tests with Native and Web:
  - Unit tests are passing in E/App ✅
  - Solution works in Native (SQLiteProvider) ✅
  - Solution works in Web (IDBKeyValProvider) ✅
- Did a couple of minor TS, tests and logic fixes.

Seems that Reassure is now reporting some performance degradation, I will investigate it further.

```
🔴 OnyxCache merge one call merging 10k keys > Duration deviation of 40.12 ms (40.74%) exceeded the allowed range of 10.00 ms (20.00%).
🔴 Onyx merge 10k calls with heavy objects > Duration deviation of 152.49 ms (35.04%) exceeded the allowed range of 10.00 ms (20.00%).
🔴 OnyxUtils initializeWithDefaultKeyStates one call initializing 10k heavy objects > Duration deviation of 108.72 ms (33.77%) exceeded the allowed range of 10.00 ms (20.00%).
🔴 Onyx init one call with 10k records to init > Duration deviation of 77.55 ms (33.54%) exceeded the allowed range of 10.00 ms (20.00%).
🔴 Onyx mergeCollection one call with 10k heavy objects > Duration deviation of 75.94 ms (28.81%) exceeded the allowed range of 10.00 ms (20.00%).
🔴 Onyx set 10k calls with heavy objects > Duration deviation of 65.38 ms (26.47%) exceeded the allowed range of 10.00 ms (20.00%).
🔴 Onyx multiSet one call with 10k heavy objects > Duration deviation of 49.66 ms (26.11%) exceeded the allowed range of 10.00 ms (20.00%).
🔴 Onyx update one call with 5k sets and 5k merges updates > Duration deviation of 91.49 ms (23.49%) exceeded the allowed range of 10.00 ms (20.00%).
🟢 OnyxUtils subscribeToKey one call subscribing to a whole collection of 10k heavy objects > Duration deviation of 25.56 ms (18.27%) is within the allowed range of 10.00 ms (20.00%).
🟢 Onyx setCollection one call with 10k heavy objects > Duration deviation of 36.80 ms (17.42%) is within the allowed range of 10.00 ms (20.00%).
🟢 OnyxUtils multiGet one call getting 10k heavy objects from storage > Duration deviation of 18.30 ms (13.81%) is within the allowed range of 10.00 ms (20.00%).
```

neil-marcellini:
Those performance hits don't look too bad to me! Maybe we have done enough optimization? It might be good to see how this works in App on some large accounts.

fabioh8010:
**Updates:**

- Fixed `mergeCollection` not applying the solution in Web platform, and added a unit test to cover it.
- Add additional Storage checks in the unit tests covering this solution, to make sure both data in Storage and cache are the same.
- For next and final steps, add the remaining comments and open PR to review.

Note: I will be OOO tomorrow and Friday due to holidays, but I'll be back on Monday to focus on this.

jest.config.js
```diff
@@ -14,4 +14,8 @@ module.exports = {
     setupFilesAfterEnv: ['./jestSetup.js'],
     testTimeout: 60000,
     transformIgnorePatterns: ['node_modules/(?!((@)?react-native|@ngneat/falso|uuid)/)'],
+    moduleNameMapper: {
+        // Redirect all imports of OnyxMerge to its web version during unit tests.
+        '^(.*)/OnyxMerge$': '<rootDir>/lib/OnyxMerge/index.ts',
```
chrispader, line 20: Can we somehow generalize this so this applies to all files with `.ts` or `.web.ts` file ending, so we don't have to update this in the future when we add more platform-specific files?

lib/OnyxUtils.ts
```diff
@@ -1142,77 +1145,101 @@ function hasPendingMergeForKey(key: OnyxKey): boolean {
...
 
 /**
- * Merges an array of changes with an existing value
+ * Merges an array of changes with an existing value or creates a single change.
  *
- * @param changes Array of changes that should be applied to the existing value
+ * @param changes Array of changes that should be merged
+ * @param existingValue The existing value that should be merged with the changes
  */
 function applyMerge<TValue extends OnyxInput<OnyxKey> | undefined, TChange extends OnyxInput<OnyxKey> | undefined>(
```
chrispader: To avoid confusion, could we rename this (internal) function to sth different, since we also have `OnyxMerge.applyMerge`.

This was my mistake when i refactored this part but i think it would be good to have unique function names for further discussion. Maybe `mergeImplementation` or `mergeInternal` is appropriate, since we are not exposing this function anyway

lib/utils.ts
```diff
@@ -35,83 +94,112 @@ function mergeObject<TObject extends Record<string, unknown>>(target: TObject |
...
+
+        // If "shouldRemoveNestedNulls" is true, we want to remove (nested) null values from the merged object.
+        // If either the source value is null, we want to omit the key from the merged object.
+        const shouldOmitNullishProperty = options.shouldRemoveNestedNulls && sourceProperty === null;
+
+        if (sourceProperty === undefined || shouldOmitNullishProperty) {
+            return;
+        }
+
+        // If source value is not a mergable object, we need to set the source value it directly.
```
chrispader:
```suggestion
        // If the source value is not a mergable object, we need to set the key directly.
```

chrispader (dismissed):
Changes look good to me aside from a few minor requests 

Thanks for your great work @fabioh8010 ! 🙌🏼

ikevin127 (approved):

jest.config.js
fabioh8010: I tried some approaches but I wasn't able yet to come up with something generic enough to cover all possible cases (also tried a custom resolver). I think we can explore this in a follow-up if you don't mind.

ikevin127:
**♻️ Testing the PR on the E/App, will drop a report soon.**

ikevin127:
### 🧪 Manual Review Report

#### Report Actions (these directly use Onyx's merge/update )
- ✅ 
- - ✅ 
- - ✅ 
- - ✅ 
•  src/libs/actions/Report.ts
•  Test creation, updates, deletions, and interactions with report actions since these directly use Onyx's merge/update functionalities.
2. Expense Tracking and Submission:
•  src/pages/iou/request/step/IOURequestStepCategory.tsx
•  src/libs/actions/IOU.ts
•  Test tracking and submitting expenses following the steps in the issue, focusing on merging action updates related to expenses.
3. Pending Actions:
•  src/libs/actions/Transaction.ts
•  This area handles transaction actions, including nullifying pending actions, which are significant for the issue at hand.
4. User Interactions:
•  src/pages/home/ReportScreen.tsx
•  src/pages/home/report/ReportActionsView.tsx
•  Ensure user interactions such as marking actions as read or unread work correctly with updates to the Onyx cache.

Testing Focus

•  Interactions involving pendingAction: Ensure any functionality that marks actions as complete (nulls the pendingAction) operates correctly.
•  Race Conditions: Especially around clearing cache and restarting actions, which will heavily utilize the merge logic.
•  Consistency Across State Resets: Verify state consistency when cache is cleared, as this mimics real-world scenarios described in the issue.

ikevin127:
- ✅

ikevin127:
### 🧪 E/App Manual Review Report

> [!note]
> - performed the testing by building this PR's `react-native-onyx` and replacing the built `dist` in E/App
> - tests will be performed on all devices: **Android & iOS: HybridApp (Native), Android & iOS: mWeb, Web and Desktop**

#### SignUp / Login Flow
- ✅ on clear cache, sign up with a new account (public & private domain)
- ✅ while logged-in, logout and sign up with a new account (public & private domain)
- ✅ on clear cache, login with an existing account (public & private domain)
- ✅ while logged-in, login with another existing account (public & private domain)
- ✅ while anonymous in a public room, login / sign-up
  - once logged-in / signed-up -> navigate through the app to verify everything looks as expected
  
#### Troubleshoot: Clear Cache and Restart Functionality
- ✅ on a new account, start / create / add some messages in a: selfDM, 1-1 DM, workspace chat, group & room chats
  - verify that upon `Clear cache and restart` trigger, the LHN report previews look as expected for each of the 'start / create / add messages' report types

#### Report Actions (aka Messages / Comments: these directly use Onyx's merge/update)
- ✅ posting a message on a report / report thread (including offline -> online transition)
- ✅ updating a message on a report / report thread (including offline -> online transition)
- ✅ deleting a message on a report / report thread (including offline -> online transition)
- ✅ adding emojis on messages / nested thread messages (including offline -> online transition)

#### Expense Tracking and Submission
•  src/pages/iou/request/step/IOURequestStepCategory.tsx
•  src/libs/actions/IOU.ts
•  Test tracking and submitting expenses following the steps in the issue, focusing on merging action updates related to expenses.
- * [ ] TO BE CONTINUED...

#### Pending Actions
•  src/libs/actions/Transaction.ts
•  This area handles transaction actions, including nullifying pending actions, which are significant for the issue at hand.
- * [ ] TO BE CONTINUED...

#### User Interactions
•  src/pages/home/ReportScreen.tsx
•  src/pages/home/report/ReportActionsView.tsx
•  Ensure user interactions such as marking actions as read or unread work correctly with updates to the Onyx cache.
- * [ ] TO BE CONTINUED...

### ♻️ Testing focus

•  Interactions involving pendingAction: Ensure any functionality that marks actions as complete (nulls the pendingAction) operates correctly.
•  Race Conditions: Especially around clearing cache and restarting actions, which will heavily utilize the merge logic.
•  Consistency Across State Resets: Verify state consistency when cache is cleared, as this mimics real-world scenarios described in the issue.

ikevin127 (approved):
### 🧪 E/App Manual Review Report

> [!note]
> - performed the testing by building this PR's `react-native-onyx` and replacing the built `dist` in E/App
> - tests will be performed on all devices: **Android & iOS: HybridApp (Native), Android & iOS: mWeb, Web and Desktop**

#### SignUp / Login Flow
- ✅ on clear cache, sign up with a new account (public & private domain)
- ✅ while logged-in, logout and sign up with a new account (public & private domain)
- ✅ on clear cache, login with an existing account (public & private domain)
- ✅ while logged-in, login with another existing account (public & private domain)
- ✅ while anonymous in a public room, login / sign-up
- ✅ switch between ND / OD and back
  - once logged-in / signed-up -> navigate through the app to verify everything looks as expected
  
#### Troubleshoot: Clear Cache and Restart Functionality
- ✅ on a new account, start / create / add some messages in a: selfDM, 1-1 DM, workspace chat, group & room chats
  - verify that upon `Clear cache and restart` trigger, the LHN report previews look as expected for each of the 'start / create / add messages' report types

#### Report Actions (aka Messages / Comments: these directly use Onyx's merge/update)
- ✅ post / update / delete a message on a report / report thread (including offline -> online transition)
- ✅ add emojis on messages / nested thread messages (including offline -> online transition)
- ✅ search router functionality (`CMD + K`) outside of Reports page
- ✅ reports page functionality (filtering, sorting, manually typed queries)

#### Expense Tracking and Submission
- ✅ submit manual / scan / distance expense (including offline -> online transition)
- ✅ replace scan expense receipt after it was posted (including offline -> online transition)

#### Pending Actions (offline -> online transitions)
- ✅ while offline, post / update / delete a message on a report / report thread
- ✅ while offline, submit manual / scan / distance expense
  - then delete them, then return online

#### User Interactions
- ✅ mark reports read / unread (including offline -> online transition)
- ✅ pin / unpin reports (including offline -> online transition)
- ✅ expenses: approve / submit / mark as paid (including offline -> online transition)
- ✅ copy onyx / export onyx data functionality
- ✅ offline mode functionality

## Summary

🎉 No issues found, **E/App** behaving as expected with the update in all scenarios mentioned above.

lib/utils.ts
```diff
@@ -3,29 +3,88 @@ import type {ConnectOptions, OnyxInput, OnyxKey} from './types';
...
+    /** The list of tuples that will be used in SQLiteProvider to replace the nested objects using `JSON_REPLACE`. */
+    replaceNullPatches: FastMergeReplaceNullPatch[];
+};
+
+type FastMergeResult<TValue> = {
+    /** The result of the merge. */
+    result: TValue;
 
-// Mostly copied from https://medium.com/@lubaka.a/how-to-remove-lodash-performance-improvement-b306669ad0e1
+    /** The path to the object that contains the internal "ONYX_INTERNALS__REPLACE_OBJECT_MARK" flag. */
```
neil-marcellini: This explanation doesn't seem quite right given the replaceNullPatches comment above.

lib/utils.ts
```diff
@@ -35,83 +94,112 @@ function mergeObject<TObject extends Record<string, unknown>>(target: TObject |
...
-                destination[key] = fastMerge(targetValueWithFallback, sourceValue, shouldRemoveNestedNulls);
-            } else {
-                destination[key] = sourceValue;
-            }
+    Object.keys(source).forEach((key) => {
+        let targetProperty = targetObject?.[key];
+        const sourceProperty = source?.[key] as Record<string, unknown>;
+
+        // If "shouldRemoveNestedNulls" is true, we want to remove (nested) null values from the merged object.
+        // If either the source value is null, we want to omit the key from the merged object.
```
neil-marcellini: Remove the word "either" since it doesn't have a use in the sentence. I expect "either X or Y" but there's only an "X".

lib/utils.ts
```diff
@@ -35,83 +94,112 @@ function mergeObject<TObject extends Record<string, unknown>>(target: TObject |
...
+
+    if (typeof value !== 'object' || Array.isArray(value)) {
+        return value;
+    }
+
+    const result: Record<string, unknown> = {};
+
+    // eslint-disable-next-line no-restricted-syntax, guard-for-in
+    for (const key in value) {
+        const newValue = value[key];
```
neil-marcellini: NAB: The name `propertyValue` or `subValue` might be more clear.

lib/utils.ts
neil-marcellini, line 1: Reviewed ✅

lib/types.ts
```diff
@@ -493,11 +494,23 @@ type InitOptions = {
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 type GenericFunction = (...args: any[]) => any;
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type GenericDeepRecord = DeepRecord<string, any>;
```
neil-marcellini: Why do we need to use `any` here?

lib/types.ts
neil-marcellini, line 1: Reviewed ✅

neil-marcellini (commented):
Here's a very small start on reviewing this. I will tackle it tomorrow.

lib/OnyxMerge/index.native.ts
```diff
@@ -0,0 +1,46 @@
...
+    const updatePromise = OnyxUtils.broadcastUpdate(key, mergedValue as OnyxValue<TKey>, hasChanged);
+
+    // If the value has not changed, calling Storage.setItem() would be redundant and a waste of performance, so return early instead.
+    if (!hasChanged) {
+        return Promise.resolve({mergedValue, updatePromise});
+    }
+
+    // For native platforms we use `mergeItem` that will take advantage of JSON_PATCH and JSON_REPLACE SQL operations to
+    // merge the object in a performant way.
+    return Storage.mergeItem(key, batchedChanges as OnyxValue<TKey>, replaceNullPatches).then(() => ({
```
lakchote, line 38: NAB: should we catch eventual errors and log them?

Also: perf tests are failing @fabioh8010 

image

lib/types.ts
fabioh8010: If I use `unknown` TS will require me to specify a type every time I access a deep nested property of an object that uses this type. Since this type is used only for testing purposes and I don't want to pollute code with lots of assertions, I'm using `any`.

lib/OnyxMerge/index.native.ts
fabioh8010: > NAB: should we catch eventual errors and log them?

Yep I believe so, will check this 👍

lib/types.ts
neil-marcellini: Ok that sounds good. If the type is only going to be used for testing, can we please make that more clear by moving it into the tests directory? We could also add a comment at the top of the file explaining that these types should only be used for testing.

fabioh8010: Makes sense, done!

lib/OnyxUtils.ts
```diff
@@ -1532,8 +1561,6 @@ GlobalSettings.addGlobalSettingsChangeListener(({enablePerformanceMetrics}) => {
     // @ts-expect-error Reassign
     getCollectionKeys = decorateWithMetrics(getCollectionKeys, 'OnyxUtils.getCollectionKeys');
     // @ts-expect-error Reassign
-    addEvictableKeysToRecentlyAccessedList = decorateWithMetrics(cache.addEvictableKeysToRecentlyAccessedList, 'OnyxCache.addEvictableKeysToRecentlyAccessedList');
-    // @ts-expect-error Reassign
     keysChanged = decorateWithMetrics(keysChanged, 'OnyxUtils.keysChanged');
```
neil-marcellini, line 1734: Why was this removed?

lib/OnyxUtils.ts
neil-marcellini, line 1: Reviewed ✅

lib/OnyxMerge/types.ts
```diff
@@ -0,0 +1,10 @@
+import type {OnyxKey, OnyxValue} from '../types';
+
+type ApplyMergeResult = {
+    mergedValue: OnyxValue<OnyxKey>;
+    updatePromise: Promise<void>;
+};
+
+type ApplyMerge = <TKey extends OnyxKey>(key: TKey, existingValue: OnyxValue<TKey>, validChanges: unknown[]) => Promise<ApplyMergeResult>;
```
neil-marcellini: Why is validChanges an unknown array when it's passed into methods from OnyxUtils like mergeChanges that have a more precise type definition? Can't we reuse that same type here?

lib/OnyxMerge/index.native.ts
```diff
@@ -0,0 +1,46 @@
...
+import * as Logger from '../Logger';
+import OnyxUtils from '../OnyxUtils';
+import type {OnyxKey, OnyxValue} from '../types';
+import cache from '../OnyxCache';
+import Storage from '../storage';
+import type {ApplyMerge} from './types';
+
+const applyMerge: ApplyMerge = <TKey extends OnyxKey>(key: TKey, existingValue: OnyxValue<TKey>, validChanges: unknown[]) => {
+    // If any of the changes is null, we need to discard the existing value.
+    const baseValue = validChanges.includes(null) ? undefined : existingValue;
```
neil-marcellini: I'm not quite following this comment. Why do we need to discard the existing value if any changes are null? For example, if we had a list of changes where an object was set to null and then back to the same existing value, we wouldn't need to discard the existing value right? 

Would this process be handled by merge and mark changes as well as merge changes?

lib/Onyx.ts
```diff
@@ -171,38 +172,31 @@ function set<TKey extends OnyxKey>(key: TKey, value: OnyxSetInput<TKey>): Promis
...
         return Promise.resolve();
     }
 
-    const valueWithoutNullValues = valueAfterRemoving as OnyxValue<TKey>;
-    const hasChanged = cache.hasValueChanged(key, valueWithoutNullValues);
+    const valueWithoutNestedNullValues = utils.removeNestedNullValues(value) as OnyxValue<TKey>;
+    const hasChanged = cache.hasValueChanged(key, valueWithoutNestedNullValues);
 
-    logSetCall(hasChanged);
+    Logger.logInfo(`set called for key: ${key}${_.isObject(value) ? ` properties: ${_.keys(value).join(',')}` : ''} hasChanged: ${hasChanged}`);
```
neil-marcellini: I have seen this line several times now. If it is present three times, let's please extract it to a utility function.

lib/Onyx.ts
```diff
@@ -718,38 +686,48 @@ function update(data: OnyxUpdate[]): Promise<void> {
...
             promises.push(() => set(key, batchedChanges));
-        } else {
-            promises.push(() => merge(key, batchedChanges));
+            return;
         }
+
+        const mergePromises = operations.map((operation) => {
+            return merge(key, operation);
+        });
+        promises.push(() => mergePromises.at(0) ?? Promise.resolve());
```
neil-marcellini, line 616: Why do we only push the first merge promise when there could be multiple?

lib/storage/providers/SQLiteProvider.ts
```diff
@@ -82,57 +103,78 @@ const provider: StorageProvider = {
...
+                    replaceQueryArguments.push(...queries);
+                }
+            }
         }
 
-        return this.multiMerge([[key, deltaChanges]]) as Promise<BatchQueryResult>;
+        commands.push({query: patchQuery, params: patchQueryArguments});
+        if (replaceQueryArguments.length > 0) {
+            commands.push({query: replaceQuery, params: replaceQueryArguments});
+        }
```
neil-marcellini, line 157: NAB: Future performance optimization?: If we need to do several replacements on the same key, maybe we could pass a series of `key, value` pairs to the JSON_REPLACE SQLite function. Given that it [supports that](https://www.sqlite.org/json1.html#jrepl), I wonder if it's faster than this approach.

Not something we need to change since the performance of this PR seems to be good enough. Would be interesting to look into though.

API.md
```diff
@@ -157,7 +157,7 @@ Onyx.merge(ONYXKEYS.POLICY, {name: 'My Workspace'}); // -> {id: 1, name: 'My Wor
 ```
 <a name="mergeCollection"></a>
 
-## mergeCollection(collectionKey, collection)
+## mergeCollection(collectionKey, collection, mergeReplaceNullPatches)
```
neil-marcellini: Should we really be exposing the merge replacement patches parameter to the external facing API? It seems like something we are really only planning to use within Onyx. And I can't think of why Onyx users would need this.

I see that we do pass this value to merge collection within onyx. I think we should make a private and internal version of the function to keep the external facing API the same.

API-INTERNAL.md
```diff
@@ -88,20 +82,6 @@ or if the provided key is a collection member key (in case our configured key is
...
-<dd><p>Add a key to the list of recently accessed keys. The least
-recently accessed key should be at the head and the most
-recently accessed key at the tail.</p>
-</dd>
-<dt><a href="#addEvictableKeysToRecentlyAccessedList">addEvictableKeysToRecentlyAccessedList()</a></dt>
-<dd><p>Take all the keys that are safe to evict and add them to
-the recently accessed list when initializing the app. This
-enables keys that have not recently been accessed to be
-removed.</p>
-</dd>
```
neil-marcellini, line 104: Similar to another comment, why are we removing these?

tests/perf-test/OnyxUtils.perf-test.ts
```diff
@@ -634,9 +625,7 @@ describe('OnyxUtils', () => {
             const changedReportAction4 = createRandomReportAction(Number(reportAction.reportActionID));
             const changedReportAction5 = createRandomReportAction(Number(reportAction.reportActionID));
 
-            await measureFunction(() =>
-                OnyxUtils.applyMerge(reportAction, [changedReportAction1, changedReportAction2, changedReportAction3, changedReportAction4, changedReportAction5], false),
-            );
+            await measureFunction(() => OnyxUtils.mergeChanges([changedReportAction1, changedReportAction2, changedReportAction3, changedReportAction4, changedReportAction5], reportAction));
```
neil-marcellini, line 628: Please rename the test case to `mergeChanges` from `applyMerge`.

tests/unit/onyxUtilsTest.ts
neil-marcellini, line 1: I really like the way you wrote the tests in this file. I find them to be easy to read and track how the values are changing.

neil-marcellini (changes requested):
I have reviewed the whole thing before you extracted the removeNestNulls thing, so I can go and quickly review that separate PR. I will also copy over any relevant comments.

I'm requesting changes because I have a few important questions and some small tweaks I would like to see. Great job on this and especially persisting with it over time.

Also as a heads up, I'm going to be out of office from July 4-16. If we don't get this merged by then for any reason, please go ahead without my review.

lib/OnyxUtils.ts
fabioh8010: This `addEvictableKeysToRecentlyAccessedList ` method doesn't belong here anymore, it was a leftover mistake from another PR.

API-INTERNAL.md
fabioh8010: Leftovers from another PR, these methods are inside OnyxCache now.

tests/unit/onyxUtilsTest.ts
fabioh8010: Thanks!

lib/OnyxMerge/index.native.ts
fabioh8010: When one of the `validChanges` are `null`, it means that we actually want to fully replace that value in Storage (like a SET). If we are going to fully replace the existing value we don't need it anymore, so we just merge all the `validChanges` that, together, will become the new value.

lib/OnyxMerge/types.ts
fabioh8010: Makes sense, will adjust!

lib/Onyx.ts
fabioh8010: I asked @chrispader about this one

lib/storage/providers/SQLiteProvider.ts
fabioh8010: Yeah I think we can explore this in a follow up PR, I will also include [this](https://github.com/Expensify/react-native-onyx/pull/620#discussion_r2179286146) if you don't mind cc @lakchote

fabioh8010:
@neil-marcellini Addressed/asked all comments!

lib/OnyxMerge/index.native.ts
neil-marcellini: Ah I see, it makes sense that `null` is included in `validChanges` to signal that the value should be replaced.

lib/Onyx.ts
neil-marcellini: This can be NAB because the tests and app seem to work well. I will await Chris' answer because I'm still curious.

neil-marcellini (approved):
Thank you for the updates. It all looks good to me now. I think it's good to go. We can follow up on some NAB items later.

lib/Onyx.ts
fabioh8010: Actually I'm rolling back this because it's causing race conditions issues in E/App -> https://github.com/Expensify/react-native-onyx/pull/657

PR #623, opened 2025-03-19
useOnyx `canBeMissing` option

fabioh8010:
### Details

This PR adds a new `canBeMissing` configuration flag to `useOnyx`. This flag, when explicitly set to `false`, will log an alert when Onyx doesn't return data for that key.

### Related Issues

https://github.com/Expensify/App/issues/58499

### Automated Tests

Unit tests were added to cover this feature.

### Manual Tests

1. Checkout this branch and link it to your E/App repo by following [these instructions](https://github.com/Expensify/react-native-onyx/blob/main/README.md#development).
2. Change one of the `useOnyx()`s of [FreeTrial.tsx](https://github.com/Expensify/App/blob/main/src/pages/settings/Subscription/FreeTrial.tsx) component to set the `canBeMissing` flag.
   ```diff
   diff --git a/src/pages/settings/Subscription/FreeTrial.tsx b/src/pages/settings/Subscription/FreeTrial.tsx
   index 1d82614d66f..b3872873c3f 100644
   --- a/src/pages/settings/Subscription/FreeTrial.tsx
   +++ b/src/pages/settings/Subscription/FreeTrial.tsx
   @@ -23,7 +23,7 @@ type FreeTrialProps = {
   function FreeTrial({badgeStyles, pressable = false, addSpacing = false, success = true, inARow = false}: FreeTrialProps) {
       const styles = useThemeStyles();
       const [policies] = useOnyx(ONYXKEYS.COLLECTION.POLICY);
   -    const [firstDayFreeTrial] = useOnyx(ONYXKEYS.NVP_FIRST_DAY_FREE_TRIAL);
   +    const [firstDayFreeTrial] = useOnyx(ONYXKEYS.NVP_FIRST_DAY_FREE_TRIAL, {canBeMissing: false});
       const [lastDayFreeTrial] = useOnyx(ONYXKEYS.NVP_LAST_DAY_FREE_TRIAL);
       const [privateSubscription] = useOnyx(ONYXKEYS.NVP_PRIVATE_SUBSCRIPTION);
   
   
   ```
3. Change [Expensify.tsx](https://github.com/Expensify/App/blob/main/src/Expensify.tsx) to display the alert.
   ```diff
   diff --git a/src/Expensify.tsx b/src/Expensify.tsx
   index 13542380888..b01542872b1 100644
   --- a/src/Expensify.tsx
   +++ b/src/Expensify.tsx
   @@ -4,6 +4,7 @@ import type {NativeEventSubscription} from 'react-native';
    import {AppState, Linking, Platform} from 'react-native';
    import type {OnyxEntry} from 'react-native-onyx';
    import Onyx, {useOnyx} from 'react-native-onyx';
   +import alert from './components/Alert';
    import ConfirmModal from './components/ConfirmModal';
    import DeeplinkWrapper from './components/DeeplinkWrapper';
    import EmojiPicker from './components/EmojiPicker/EmojiPicker';
   @@ -45,10 +46,14 @@ import type {Route} from './ROUTES';
    import SplashScreenStateContext from './SplashScreenStateContext';
    import type {ScreenShareRequest} from './types/onyx';
    
   -Onyx.registerLogger(({level, message}) => {
   +Onyx.registerLogger(({level, message, extraData}) => {
        if (level === 'alert') {
            Log.alert(message);
            console.error(message);
   +
   +        if (extraData?.showAlert) {
   +            alert(message);
   +        }
        } else if (level === 'hmmm') {
            Log.hmmm(message);
        } else {
   
   ```
4. Create a new account and sign in.
5. Switch between Settings and Inbox in order to trigger the alert.
6. Assert an alert is displayed both visually and in the console.

### Author Checklist

    - [x] Android / native
    - [x] Android / Chrome
    - [x] iOS / native
    - [x] iOS / Safari
    - [x] MacOS / Chrome / Safari
    - [x] MacOS / Desktop

### Screenshots/Videos
<details>

https://github.com/user-attachments/assets/724123bf-0b9b-473a-b340-0c1f23f2fc36

</details>

<details>

I'm having problems with my emulators when opening the Chrome app (they crash instantly), so I couldn't record videos for this platform.

</details>

<details>

https://github.com/user-attachments/assets/f26df8d4-9325-49a7-89b7-f1f944c45559

</details>

<details>

https://github.com/user-attachments/assets/cbdd86f3-0cd9-4dec-99d6-bc29a74e3398

</details>

<details>

https://github.com/user-attachments/assets/164bb01f-8ba5-4944-89e8-f4e577b67819

https://github.com/user-attachments/assets/642755dd-736f-409c-8334-3e1998ab026c

</details>

<details>

https://github.com/user-attachments/assets/b4463cc2-51d6-48aa-9d4f-46d38287121c

</details>

### DIFF
--- a/lib/useOnyx.ts
+++ b/lib/useOnyx.ts
@@ -10,6 +10,7 @@ import type {CollectionKeyBase, KeyValueMapping, OnyxCollection, OnyxKey, OnyxVa
 import useLiveRef from './useLiveRef';
 import usePrevious from './usePrevious';
 import decorateWithMetrics from './metrics';
+import * as Logger from './Logger';
 
 type BaseUseOnyxOptions = {
     /**
@@ -37,6 +38,11 @@ type BaseUseOnyxOptions = {
      * If set to `true`, the key can be changed dynamically during the component lifecycle.
      */
     allowDynamicKey?: boolean;
+
+    /**
+     * If set to `false`, the hook will log an alert when Onyx doesn't return data for that key.
+     */
+    canBeMissing?: boolean;
 };
 
 type UseOnyxInitialValueOption<TInitialValue> = {
@@ -280,11 +286,17 @@ function useOnyx<TKey extends OnyxKey, TReturnValue = OnyxValue<TKey>>(
             previousValueRef.current = newValueRef.current;
 
             // If the new value is `null` we default it to `undefined` to ensure the consumer gets a consistent result from the hook.
-            resultRef.current = [previousValueRef.current ?? undefined, {status: newFetchStatus ?? 'loaded'}];
+            const newValue = previousValueRef.current ?? undefined;
+            const newStatus = newFetchStatus ?? 'loaded';
+            resultRef.current = [previousValueRef.current ?? undefined, {status: newStatus}];
+
+            if (options?.canBeMissing === false && newStatus === 'loaded' && newValue === undefined) {
+                Logger.logAlert(`useOnyx returned no data for key '${key}' while canBeMissing was set to false.`);
+            }
         }
 
         return resultRef.current;
-    }, [options?.initWithStoredValues, options?.allowStaleData, options?.initialValue, key, selectorRef]);
+    }, [options?.initWithStoredValues, options?.allowStaleData, options?.initialValue, options?.canBeMissing, key, selectorRef]);
 
     const subscribe = useCallback(
         (onStoreChange: () => void) => {
--- a/lib/useOnyx.ts
+++ b/lib/useOnyx.ts
@@ -289,7 +289,7 @@ function useOnyx<TKey extends OnyxKey, TReturnValue = OnyxValue<TKey>>(
             // If `canBeMissing` is set to `false` and the Onyx value of that key is not defined,
             // we log an alert so it can be acknowledged by the consumer.
             if (options?.canBeMissing === false && newStatus === 'loaded' && !isOnyxValueDefined) {
-                Logger.logAlert(`useOnyx returned no data for key '${key}' while canBeMissing was set to false.`, {showAlert: true});
+                Logger.logAlert(`useOnyx returned no data for key with canBeMissing set to false.`, {key, showAlert: true});
             }
         }
 
--- a/lib/useOnyx.ts
+++ b/lib/useOnyx.ts
@@ -104,17 +104,6 @@ function tryGetCachedValue<TKey extends OnyxKey>(key: TKey): OnyxValue<OnyxKey> 
     return values;
 }
 
-/**
- * Gets the value from cache and maps it with selector. It changes `null` to `undefined` for `useOnyx` compatibility.
- */
-function getCachedValue<TKey extends OnyxKey, TValue>(key: TKey, selector?: UseOnyxSelector<TKey, TValue>) {
-    const value = tryGetCachedValue(key) as OnyxValue<TKey>;
-
-    const selectedValue = selector ? selector(value) : (value as TValue);
-
-    return selectedValue ?? undefined;
-}
-
 function useOnyx<TKey extends OnyxKey, TReturnValue = OnyxValue<TKey>>(
     key: TKey,
     options?: BaseUseOnyxOptions & UseOnyxInitialValueOption<TReturnValue> & Required<UseOnyxSelectorOption<TKey, TReturnValue>>,
@@ -225,6 +214,8 @@ function useOnyx<TKey extends OnyxKey, TReturnValue = OnyxValue<TKey>>(
     }, [key, options?.canEvict]);
 
     const getSnapshot = useCallback(() => {
+        let isOnyxValueDefined = true;
+
         // We return the initial result right away during the first connection if `initWithStoredValues` is set to `false`.
         if (isFirstConnectionRef.current && options?.initWithStoredValues === false) {
             return resultRef.current;
@@ -234,11 +225,14 @@ function useOnyx<TKey extends OnyxKey, TReturnValue = OnyxValue<TKey>>(
         // so we can return any cached value right away. After the connection is made, we only
         // update `newValueRef` when `Onyx.connect()` callback is fired.
         if (isFirstConnectionRef.current || shouldGetCachedValueRef.current) {
-            // If `newValueRef.current` is `undefined` it means that the cache doesn't have a value for that key yet.
-            // If `newValueRef.current` is `null` or any other value it means that the cache does have a value for that key.
-            // This difference between `undefined` and other values is crucial and it's used to address the following
-            // conditions and use cases.
-            newValueRef.current = getCachedValue(key, selectorRef.current);
+            // Gets the value from cache and maps it with selector. It changes `null` to `undefined` for `useOnyx` compatibility.
+            const value = tryGetCachedValue(key) as OnyxValue<TKey>;
+            const selectedValue = selectorRef.current ? selectorRef.current(value) : value;
+            newValueRef.current = (selectedValue ?? undefined) as TReturnValue | undefined;
+
+            // This flag is `false` when the original Onyx value (without selector) is not defined yet.
+            // It will be used later to check if we need to log an alert that the value is missing.
+            isOnyxValueDefined = value !== null && value !== undefined;
 
             // We set this flag to `false` again since we don't want to get the newest cached value every time `getSnapshot()` is executed,
             // and only when `Onyx.connect()` callback is fired.
@@ -261,7 +255,7 @@ function useOnyx<TKey extends OnyxKey, TReturnValue = OnyxValue<TKey>>(
         // If data is not present in cache and `initialValue` is set during the first connection,
         // we set the new value to `initialValue` and fetch status to `loaded` since we already have some data to return to the consumer.
         if (isFirstConnectionRef.current && !hasCacheForKey && options?.initialValue !== undefined) {
-            newValueRef.current = (options?.initialValue ?? undefined) as TReturnValue;
+            newValueRef.current = options.initialValue;
             newFetchStatus = 'loaded';
         }
 
@@ -275,7 +269,7 @@ function useOnyx<TKey extends OnyxKey, TReturnValue = OnyxValue<TKey>>(
             areValuesEqual = shallowEqual(previousValueRef.current ?? undefined, newValueRef.current);
         }
 
-        // We updated the cached value and the result in the following conditions:
+        // We update the cached value and the result in the following conditions:
         // We will update the cached value and the result in any of the following situations:
         // - The previously cached value is different from the new value.
         // - The previously cached value is `null` (not set from cache yet) and we have cache for this key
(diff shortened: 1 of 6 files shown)


lib/useOnyx.ts
```diff
@@ -280,11 +286,17 @@ function useOnyx<TKey extends OnyxKey, TReturnValue = OnyxValue<TKey>>(
             previousValueRef.current = newValueRef.current;
 
             // If the new value is `null` we default it to `undefined` to ensure the consumer gets a consistent result from the hook.
-            resultRef.current = [previousValueRef.current ?? undefined, {status: newFetchStatus ?? 'loaded'}];
+            const newValue = previousValueRef.current ?? undefined;
+            const newStatus = newFetchStatus ?? 'loaded';
+            resultRef.current = [previousValueRef.current ?? undefined, {status: newStatus}];
+
+            if (options?.canBeMissing === false && newStatus === 'loaded' && newValue === undefined) {
+                Logger.logAlert(`useOnyx returned no data for key '${key}' while canBeMissing was set to false.`);
```
fabioh8010, line 294: @iwiznia Could you confirm if this is the kind of log (`Logger.logAlert`) we should use?

tests/unit/useOnyxTest.ts
```diff
@@ -749,6 +750,133 @@ describe('useOnyx', () => {
...
+            await act(async () => Onyx.set(ONYXKEYS.TEST_KEY, 'test'));
+
+            expect(result1.current[0]).toBe('test_changed');
+
+            await act(async () => Onyx.set(ONYXKEYS.TEST_KEY, null));
+
+            expect(result1.current[0]).toBe('undefined_changed');
+            expect(logAlertFn).not.toBeCalledWith(alert(ONYXKEYS.TEST_KEY));
+        });
+    });
```
fabioh8010, line 878: @iwiznia We have this following use case. Basically:

- We have a `useOnyx` with a selector that will always return some value to the consumer, even if the value from Onyx store doesn't exist

What should we do in this case? Is it valid to NOT log an alert in this case?

lib/useOnyx.ts
iwiznia: Yeah I think it is, since in App, we register logger [here](https://github.com/Expensify/App/blob/c15a8b2cfdfc34af8b76c1abdcaa6e06274071aa/src/Expensify.tsx#L48) which means that these calls will end up calling `Log.alert(message);` which should be sending the data to the backend, but let's ensure that's the case when we test this in App.

iwiznia: hmmmmm weird that the logging functions in this repo don't have a `params` object passed like the `Log` function in App... we _might_ need to change that later as it is possible that otherwise this log line would create an issue per key that is missing, which would mean for example 1 issue per reportID which is not what we want... but let's deal with that later.

tests/unit/useOnyxTest.ts
iwiznia: Not sure I follow, if there's no value in the onyx store, how is it returning a value? What value is it returning?

lib/useOnyx.ts
fabioh8010: Ok so I will do some testing before opening the PR to check if the logs are ok

tests/unit/useOnyxTest.ts
fabioh8010: @iwiznia Sure, let me explain!

Let's say we subscribe to a key using `useOnyx` and passing a `selector`:

```ts
const [value] = useOnyx(ONYXKEYS.TEST_KEY, {
    selector: (entry: OnyxEntry<string>) => `${entry}_changed`,
    canBeMissing: false,
}),
```

With this `selector` in particular, it will always return a string no matter if we have or not a value stored in Onyx for that key. So, let's say we don't have data in Onyx for `ONYXKEYS.TEST_KEY`. When running the hook, `selector` will be executed, `entry` will be `undefined` (because there is no data in Onyx) but the selector will return `undefined_changed` to the consumer. It means that this `useOnyx` will always return something because of this `selector`, regardless if we have data or not in Onyx.

fabioh8010: That's basically the use case described in this unit test, and at the moment it won't log an alert because consumer will always receive some value (because of this `selector`).

iwiznia: Ahhh thanks for the explanation! 
Ideally, we would detect that `entry` is `undefined` and log an alert if `canBeMissing` was false, can we do that?

fabioh8010:
@iwiznia About [your comment](https://github.com/Expensify/App/issues/58499#issuecomment-2743663946):

> Oh, I was referring about [this](https://github.com/Expensify/App/blob/4fe2f6e04ce475d2e344151e6bd3232c90a7d46d/src/libs/HttpUtils.ts#L144), but it seems it just uses an alert, since this is dev only I guess we can also use alert. I don't think you can test this, other than modifying that code and triggering it manually, since that data is never returned in the production/staging api, only in dev.

[I added support in Onyx Logger to pass extra data when logging stuff](https://github.com/Expensify/react-native-onyx/pull/623/commits/c5eb697d6b5c1a47e13cc14d67e49c1baf1ed98f), and in E/App `Expensify.tsx` I would have something like this:

```ts
Onyx.registerLogger(({level, message, extraData}) => {
    if (level === 'alert') {
        Log.alert(message);
        console.error(message);

        if (extraData?.showAlert) {
            alert(message);
        }
    } else if (level === 'hmmm') {
        Log.hmmm(message);
    } else {
        Log.info(message);
    }
});
```

But I still think it's too intrusive for devs and each alert stops the entire App's execution flow, please check out the video:

https://github.com/user-attachments/assets/8fbe9e99-d1aa-4002-9350-6ea2b9380e05

I think we should only show the errors in the console. According to the checklist C+ should always check for new errors when reviewing PRs. WDYT?

iwiznia:
Let's try with alert at first and see if it is annoying? I understand the alerts are a bit annoying, but it only depends on how many you see normally.
My fear of the console is that people will miss it.

lib/useOnyx.ts
```diff
@@ -280,11 +280,18 @@ function useOnyx<TKey extends OnyxKey, TReturnValue = OnyxValue<TKey>>(
...
 
             // If the new value is `null` we default it to `undefined` to ensure the consumer gets a consistent result from the hook.
-            resultRef.current = [previousValueRef.current ?? undefined, {status: newFetchStatus ?? 'loaded'}];
+            const newStatus = newFetchStatus ?? 'loaded';
+            resultRef.current = [previousValueRef.current ?? undefined, {status: newStatus}];
+
+            // If `canBeMissing` is set to `false` and the Onyx value of that key is not defined,
+            // we log an alert so it can be acknowledged by the consumer.
+            if (options?.canBeMissing === false && newStatus === 'loaded' && !isOnyxValueDefined) {
+                Logger.logAlert(`useOnyx returned no data for key '${key}' while canBeMissing was set to false.`, {showAlert: true});
```
iwiznia, line 289: We need to do the `alert()` call if we are in dev, no?

iwiznia: Also, can you move `key` to the params? I think otherwise this will create an issue for each key, for example one issue per reportID

fabioh8010: > Also, can you move key to the params? I think otherwise this will create an issue for each key, for example one issue per reportID

@iwiznia Could you help me this? I'm not sure how I should pass this data. I saw some example [here](https://github.com/Expensify/App/blob/main/src/libs/ReportUtils.ts#L1113), so I guess we need to pass this way?

```ts
Logger.logAlert(`useOnyx returned no data for key '${key}' while canBeMissing was set to false.`, {key, showAlert: true});
```

Also, I noticed that in E/App our `Log.alert()` has an additional `parameters` parameter that is used to pass this kind of data, right? In E/App we use `Logger` from `expensify-common` which has [this signature](https://github.com/callstack-internal/expensify-common/blob/main/lib/Logger.ts#L108). So I guess we must have same signature in our Onyx logger?

iwiznia: > We are going to handle this in E/App, with something like this in Expensify.tsx:

Ah ok, is that because `__DEV__` is not accessible here or what?

tests/unit/useOnyxTest.ts
```diff
@@ -749,6 +750,135 @@ describe('useOnyx', () => {
...
+            expect(result1.current[0]).toBe('test_changed');
+
+            await act(async () => Onyx.set(ONYXKEYS.TEST_KEY, null));
+
+            expect(result1.current[0]).toBeUndefined();
+            expect(logAlertFn).toHaveBeenCalledTimes(2);
+            expect(logAlertFn).toHaveBeenNthCalledWith(2, alert(ONYXKEYS.TEST_KEY), {showAlert: true});
+        });
+
+        it('should log an alert if Onyx doesn\'t return data but there is a selector that always return something and "canBeMissing" property is false', async () => {
```
amyevans, line 853: Interesting, I must say I didn't think of this case!

amyevans (commented):
I don't have a lot to add here but I am a fan of the idea!

lib/useOnyx.ts
fabioh8010: @iwiznia I adjusted Onyx Logger functions to use same parameters signature we have in `expensify-common`, so code in `Expensify.tsx` will look like this:

```diff
diff --git a/src/Expensify.tsx b/src/Expensify.tsx
index 13542380888..3424147387d 100644
--- a/src/Expensify.tsx
+++ b/src/Expensify.tsx
@@ -4,6 +4,7 @@ import type {NativeEventSubscription} from 'react-native';
 import {AppState, Linking, Platform} from 'react-native';
 import type {OnyxEntry} from 'react-native-onyx';
 import Onyx, {useOnyx} from 'react-native-onyx';
+import alert from './components/Alert';
 import ConfirmModal from './components/ConfirmModal';
 import DeeplinkWrapper from './components/DeeplinkWrapper';
 import EmojiPicker from './components/EmojiPicker/EmojiPicker';
@@ -45,14 +46,18 @@ import type {Route} from './ROUTES';
 import SplashScreenStateContext from './SplashScreenStateContext';
 import type {ScreenShareRequest} from './types/onyx';
 
-Onyx.registerLogger(({level, message}) => {
+Onyx.registerLogger(({level, message, parameters}) => {
     if (level === 'alert') {
-        Log.alert(message);
+        Log.alert(message, parameters);
         console.error(message);
+
+        if (typeof parameters === 'object' && !Array.isArray(parameters) && 'showAlert' in parameters) {
+            alert(message);
+        }
     } else if (level === 'hmmm') {
-        Log.hmmm(message);
+        Log.hmmm(message, parameters);
     } else {
-        Log.info(message);
+        Log.info(message, undefined, parameters);
     }
 });
 

```

tests/unit/useOnyxTest.ts
```diff
@@ -749,6 +750,136 @@ describe('useOnyx', () => {
         });
     });
 
+    describe('canBeMissing', () => {
+        let logAlertFn = jest.fn();
+        // const alert = (key: string) => `useOnyx returned no data for key '${key}' while canBeMissing was set to false.`;
```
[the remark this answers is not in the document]
fabioh8010: Oops 😄

amyevans (approved):

iwiznia (approved):

PR #640, opened 2025-06-05
Fix - added merge all snapshot config

FitseTLT:
This pr creates a mergeAllPropsSnapshotKeys config so that the BE response which is member of those collection keys will be directly merged to snapshot data without picking props that only exist in the old snapshot data.
### Related Issues

https://github.com/Expensify/App/issues/60116
https://github.com/Expensify/App/issues/63295

### Manual Tests

Use https://github.com/FitseTLT/App/tree/fix-bump-onyx-mergeAllPropsSnapshotKeys to test it.
Precondition: The user has sent a message with an attachment or a link

1. Navigate to the Reports
2. Click on Chats
3. Click on the Status drop-down menu
4. Select Attachments
5. Click on the message item (not over the attachment)
6. The filter is non reset, the user can close RHP with the conversation and see the page with filter results.

Precondition: a workspace has an employee with an expense without receipt

1. From the admin side click on Reports tab
2. From the employee side Click on Reports tab
3. Click on expense without receipt
4. Upload image for this expense
5. From the admin side
6. The uploaded image should be displayed

### Author Checklist

    - [x] Android / native
    - [x] Android / Chrome
    - [x] iOS / native
    - [x] iOS / Safari
    - [x] MacOS / Chrome / Safari
    - [x] MacOS / Desktop

### Screenshots/Videos
<details>

https://github.com/user-attachments/assets/15df8650-23ce-48b3-95cb-95a150d2d58f

</details>

<details>

https://github.com/user-attachments/assets/7754fe3c-1c06-4d66-8ab9-8dc8cd32a8fc

</details>

<details>

</details>

<details>

https://github.com/user-attachments/assets/38729c55-8e1a-4b5d-8292-d56e59885699

</details>

<details>

https://github.com/user-attachments/assets/a6be6f7a-84a2-4e2c-8599-335af23cded0

</details>

<details>

https://github.com/user-attachments/assets/654e1ddb-d7a5-4dec-81c2-990a16e6848b

</details>

### DIFF
--- a/lib/OnyxUtils.ts
+++ b/lib/OnyxUtils.ts
@@ -87,10 +87,6 @@ function getSnapshotKey(): OnyxKey | null {
     return snapshotKey;
 }
 
-function getFullyMergedSnapshotKeys(): Set<string> | undefined {
-    return fullyMergedSnapshotKeys;
-}
-
 /**
  * Getter - returns the merge queue.
  */
@@ -1421,7 +1417,6 @@ function updateSnapshots(data: OnyxUpdate[], mergeFn: typeof Onyx.merge): Array<
             }
 
             const oldValue = updatedData[key] || {};
-            const fullyMergedKeys = getFullyMergedSnapshotKeys();
             let collectionKey: string | undefined;
             try {
                 collectionKey = getCollectionKey(key);
@@ -1429,7 +1424,7 @@ function updateSnapshots(data: OnyxUpdate[], mergeFn: typeof Onyx.merge): Array<
                 // If getCollectionKey() throws an error it means the key is not a collection key.
                 collectionKey = undefined;
             }
-            const shouldFullyMerge = fullyMergedKeys?.has(collectionKey || key);
+            const shouldFullyMerge = fullyMergedSnapshotKeys?.has(collectionKey || key);
             const newValue = shouldFullyMerge ? value : lodashPick(value, Object.keys(snapshotData[key]));
 
             updatedData = {...updatedData, [key]: Object.assign(oldValue, newValue)};


lib/OnyxUtils.ts
```diff
@@ -84,6 +87,10 @@ function getSnapshotKey(): OnyxKey | null {
     return snapshotKey;
 }
 
+function getMergeAllPropsSnapshotKeys(): string[] | undefined {
```
iwiznia: Besides in this file, where is this going to be used? Is there an App PR that goes with this PR?

lib/OnyxUtils.ts
```diff
@@ -133,7 +140,7 @@ function setSkippableCollectionMemberIDs(ids: Set<string>): void {
  * @param initialKeyStates - initial data to set when `init()` and `clear()` are called
  * @param evictableKeys - This is an array of keys (individual or collection patterns) that when provided to Onyx are flagged as "safe" for removal.
  */
-function initStoreValues(keys: DeepRecord<string, OnyxKey>, initialKeyStates: Partial<KeyValueMapping>, evictableKeys: OnyxKey[]): void {
+function initStoreValues(keys: DeepRecord<string, OnyxKey>, initialKeyStates: Partial<KeyValueMapping>, evictableKeys: OnyxKey[], mergeAllSnapshotKeys?: string[]): void {
```
iwiznia: Please add a JSDoc comment explaining what this new param does

lib/OnyxUtils.ts
iwiznia: Also maybe it would be better if this (and mergeAllPropsSnapshotKeys) are defined as `string[]` and remove the undefined, no? `undefined` is not really used for anything (as in it does not signify anything different than `[]`), right?

lib/OnyxUtils.ts
iwiznia: And do we need this param? Which keys we don't want to merge?

luacmartins: `reportActions` comes to mind since this issue came from us merging all report action updates to snapshots and that breaks the chat search. Not sure if there are others

lib/OnyxUtils.ts
FitseTLT: No strong feelings, but something similar to snapshotKey here
https://github.com/Expensify/react-native-onyx/blob/e745f5401a463138657f21aaee783ec6e69c4081/lib/OnyxUtils.ts#L73 
When it is not initialized we will avoid iterating here
https://github.com/Expensify/react-native-onyx/blob/e745f5401a463138657f21aaee783ec6e69c4081/lib/OnyxUtils.ts#L1429

lib/OnyxUtils.ts
FitseTLT: BTW I have tried to refer from SearchResults type [here](https://github.com/FitseTLT/App/blob/aeffce7e48153b4ba33a7eabe04e4c81b698c026/src/types/onyx/SearchResults.ts#L462-L468) to identify the mergeAllPropsSnapshotKeys and here is my list
```
 mergeAllPropsSnapshotKeys: [
            ONYXKEYS.COLLECTION.TRANSACTION,
            ONYXKEYS.COLLECTION.REPORT,
            ONYXKEYS.COLLECTION.POLICY,
            ONYXKEYS.COLLECTION.REPORT_NAME_VALUE_PAIRS,
            ONYXKEYS.PERSONAL_DETAILS_LIST,
        ],
```

lib/Onyx.ts
```diff
@@ -44,6 +44,7 @@ function init({
     debugSetState = false,
     enablePerformanceMetrics = false,
     skippableCollectionMemberIDs = [],
+    mergeAllPropsSnapshotKeys = [],
```
parasharrajat: Can we have a better name for this? My suggestion is `unrestrictedSnapshotKeys`

FitseTLT: I think your suggestion is more confusing than the current one but if you guys have a better name we can update.

parasharrajat: I don't have any other suggestion. @luacmartins What do you think?

luacmartins: hmm maybe `fullyMergedSnapshotKeys` or `directMergeSnapshotKeys`. Not sure what's best tbh

FitseTLT: `fullyMergedSnapshotKeys` Looks good to me.

parasharrajat: Let's make that change.

FitseTLT: Done

lib/OnyxUtils.ts
```diff
@@ -132,8 +139,9 @@ function setSkippableCollectionMemberIDs(ids: Set<string>): void {
  * @param keys - `ONYXKEYS` constants object from Onyx.init()
  * @param initialKeyStates - initial data to set when `init()` and `clear()` are called
  * @param evictableKeys - This is an array of keys (individual or collection patterns) that when provided to Onyx are flagged as "safe" for removal.
+ * @param fullyMergedSnapshotKeys - Array of collection keys whose backend returned data can be directly merged into snapshot without selectively picking the existing keys in the existing snapshot data.
```
parasharrajat:
```suggestion
 * @param fullyMergedSnapshotKeys - Array of snapshot collection keys where full merge is supported and data structure can be changed after merge.
```

lib/types.ts
```diff
@@ -480,6 +480,12 @@ type InitOptions = {
      * Additionally, any subscribers from these keys to won't receive any data from Onyx.
      */
     skippableCollectionMemberIDs?: string[];
+
+    /**
+     * Array of collection keys whose backend returned data can be directly merged into
+     * snapshot without selectively picking the existing keys in the existing snapshot data.
```
parasharrajat:
```suggestion
    /**
     * Array of snapshot collection keys where full merge is supported and the data structure can be changed after the merge.
```

FitseTLT: @luacmartins WDYT about this suggestion ^ 😕

luacmartins: Sounds good, but I'd add an example to it to make it clearer.

lib/OnyxUtils.ts
iwiznia: I think in general having variables with less types is better than having them with more types, so unless undefined means something different logically than `[]` then let's change it

lib/types.ts
FitseTLT: Added

lib/OnyxUtils.ts
parasharrajat: +1

lib/types.ts
parasharrajat: Can we update the comment as well? There is no concept of backend in Onyx. The comment should be platform/project agnostic.

FitseTLT: Makes sense updated

lib/OnyxUtils.ts
```diff
@@ -71,6 +72,8 @@ const lastConnectionCallbackData = new Map<number, OnyxValue<OnyxKey>>();
 
 let snapshotKey: OnyxKey | null = null;
 
+let fullyMergedSnapshotKeys: string[] | undefined = [];
```
iwiznia: I think I commented this already, can we remove undefined from here given we have a default of `[]`?

lib/OnyxUtils.ts
```diff
@@ -84,6 +87,10 @@ function getSnapshotKey(): OnyxKey | null {
     return snapshotKey;
 }
 
+function getFullyMergedSnapshotKeys(): string[] | undefined {
```
iwiznia: And remove undefined from here too

lib/OnyxUtils.ts
```diff
@@ -1416,8 +1425,13 @@ function updateSnapshots(data: OnyxUpdate[], mergeFn: typeof Onyx.merge): Array<
             }
 
             const oldValue = updatedData[key] || {};
+            const fullyMergedKeys = getFullyMergedSnapshotKeys();
+            const newValue =
+                fullyMergedKeys && fullyMergedKeys.some((collectionKey) => isCollectionMemberKey(collectionKey, key, collectionKey.length))
```
iwiznia: And remove `fullyMergedKeys &&` from here too

iwiznia: Why is `isCollectionMemberKey` receiving as param `collectionKey.length` instead of the method computing it, given it also is receiving `collectionKey`? Seems like some kind of optimization? Not sure if I get it because getting the length of an array should be basically instant, no?
Anyway, let's either:
- Completely remove the parameter 
- Make it optional and get the length of the first param when not passed

FitseTLT:
Dealt with the last comments but regarding the performance tests, it doesn't show the specific test that is failing. How can I run it locally, so that I can spot the exact test that is failing? cc @luacmartins @iwiznia TIA

luacmartins:
I retried the tests and they failed with a different error now. Re-running again.

FitseTLT:
Still failed BTW have you tried runiing it locally? on main

luacmartins:
>BTW have you tried runiing it locally? on main

I haven't. I'm not sure what's causing this. Maybe you can start a discussion in Slack and ping some people you see in the blame for the failing tets

FitseTLT:
[Started](https://expensify.slack.com/archives/C01GTK53T8Q/p1750256781604239)

luacmartins (approved):

luacmartins:
@iwiznia do you wanna give this another review before we merge?

lib/OnyxUtils.ts
```diff
@@ -1416,8 +1421,10 @@ function updateSnapshots(data: OnyxUpdate[], mergeFn: typeof Onyx.merge): Array<
             }
 
             const oldValue = updatedData[key] || {};
+            const fullyMergedKeys = getFullyMergedSnapshotKeys();
+            const newValue = fullyMergedKeys.some((collectionKey) => isCollectionMemberKey(collectionKey, key)) ? value : lodashPick(value, Object.keys(snapshotData[key]));
```
iwiznia: I know `fullyMergedKeys` probably won't have that many entries in it, but shouldn't we make it a Set?

iwiznia (approved):
NAB

lib/OnyxUtils.ts
luacmartins: I agree we can optimize this. @FitseTLT let's update this to use a set for faster access instead

FitseTLT: Yep Done

lib/OnyxUtils.ts
```diff
@@ -1422,7 +1422,15 @@ function updateSnapshots(data: OnyxUpdate[], mergeFn: typeof Onyx.merge): Array<
 
             const oldValue = updatedData[key] || {};
             const fullyMergedKeys = getFullyMergedSnapshotKeys();
-            const newValue = fullyMergedKeys.some((collectionKey) => isCollectionMemberKey(collectionKey, key)) ? value : lodashPick(value, Object.keys(snapshotData[key]));
+            let collectionKey: string | undefined;
+            try {
+                collectionKey = getCollectionKey(key);
+            } catch (e) {
```
iwiznia, line 1423: Is it better to call `isCollectionKey` instead of `getCollectionKey` and catching the error?

lib/OnyxUtils.ts
```diff
@@ -84,6 +87,10 @@ function getSnapshotKey(): OnyxKey | null {
     return snapshotKey;
 }
 
+function getFullyMergedSnapshotKeys(): Set<string> | undefined {
+    return fullyMergedSnapshotKeys;
+}
```
iwiznia: Just realized this method is basically useless since we are not exporting it and anything in this file can just use `fullyMergedSnapshotKeys`

lib/OnyxUtils.ts
FitseTLT: @iwiznia to exploit set `has` function (which is basically the reason we are using Set) we need to find the collection key 👍

iwiznia: Yes I know, I am asking if instead of catching the error is better to do something like:
```
if (isCollectionKey(key)) {
    key = getCollectionKey(key);
}
const shouldFullyMerge = fullyMergedSnapshotKeys?.has(key);
```

FitseTLT: @iwiznia what we are trying to extract is the collection key of a member key and `isCollectionKey` is false for a member key.

luacmartins: I'm not sure that I follow. Can you give an example of what you mean by collection key and member key? I'm not sure how `isCollectionKey` would return false but `getCollectionKey` would work

FitseTLT: Example: `key` : `report_1`, fullyMergedSnapshotKeys would contain `report_` 
But here key is not a collection key but you can get the collecktion key from it `getCollectionKey`. If we are going to take advantage of `fullyMergedSnapshotKeys` Set we should know the exact value of the key we want to search in the set. Which means in this case we get the collection key from the `key` and search if the collection key is found in the Set.

luacmartins: I'm still confused. In your example, `report_1` is a collection key, no?

FitseTLT: obviously member key @luacmartins

luacmartins: Haha I'm even more confused now. What are you calling a member key? AFAIK `report_1` would be a collection key

FitseTLT: @luacmartins `getCollectionKey` is a helper that returns the collection key of a member key. So if you pass `report_1` it will return `report_` the collection key. And in this case `fullyMergedSnapshotKeys` might be holding collection keys like `report_` so if we use a Set for it then to use a Set `has` function you have to know the value you are searching for. But in our case  the `key` (onyx Update ) let say is `report_1` so `isCollectionKey` function doesn't help at all as it will be false for `report_1`.

https://github.com/Expensify/react-native-onyx/blob/3726a5469967457b787d07fbc40f445564240dae/lib/OnyxUtils.ts#L446-L452

luacmartins: Ah I see what you're saying now. That makes sense

luacmartins (approved):

luacmartins:
Perf tests are failing again 😞

iwiznia (approved):

PR #653, opened 2025-07-07
Implement storage eviction system with tracking and cleanup functiona…

kubabutkiewicz:
### Related Issues

https://github.com/Expensify/App/issues/65525

### Manual Tests

1. Link locally the react-native-onyx with the App
2. Checkout to this branch `add-storage-eviction`
3. in the app repo to [Onyx.init](https://github.com/Expensify/App/blob/feae7c58f60c7eba24eab56047c3ebb6a4747bb7/src/setup/index.ts#L29-L54)  add storage eviction configuration to do a cleanup more frequently for tests 
```
        storageManager: {
            //  2 hours
            maxAgeDays: 0.08333333333333333,
            //  5 minutes
            maxIdleDays: 0.003472222222222222,
            evictableKeys: [
                ONYXKEYS.COLLECTION.REPORT_ACTIONS,
                ONYXKEYS.COLLECTION.REPORT_ACTIONS_DRAFTS,
                ONYXKEYS.COLLECTION.REPORT_ACTIONS_PAGES,
                ONYXKEYS.COLLECTION.REPORT_ACTIONS_REACTIONS,
            ],
        },
```
4. Verify that evictable keys are remove after 5 mins of not using the key, and after 2h when its in the storage (used or not used) 

### Author Checklist

    - [x] Android / native
    - [x] Android / Chrome
    - [x] iOS / native
    - [x] iOS / Safari
    - [x] MacOS / Chrome / Safari
    - [x] MacOS / Desktop

### Screenshots/Videos
<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>


lib/Onyx.ts
```diff
@@ -72,9 +74,10 @@ function init({
     }
 
     OnyxUtils.initStoreValues(keys, initialKeyStates, evictableKeys, fullyMergedSnapshotKeys);
+    const manager = createStorageManager(storageManager);
 
     // Initialize all of our keys with data provided then give green light to any pending connections
-    Promise.all([cache.addEvictableKeysToRecentlyAccessedList(OnyxUtils.isCollectionKey, OnyxUtils.getAllKeys), OnyxUtils.initializeWithDefaultKeyStates()]).then(
+    Promise.all([cache.addEvictableKeysToRecentlyAccessedList(OnyxUtils.isCollectionKey, OnyxUtils.getAllKeys), OnyxUtils.initializeWithDefaultKeyStates(), manager?.initialize()]).then(
```
fabioh8010: I think would be better to follow how we done for `OnyxConnectionManager` and `OnyxCache`, that is having a single instance for the storage manager. I don't see how having the ability to creating multiple instances can benefit us, and it makes code more complicated (you have to do if-checks every time) than just importing the manager file.

I think you could instead combine `createStorageManager` and `initialize()` to be just `initialize()` and passing the config.

`Onyx.init()` is called only once in the application, so we don't need that concept of instance shutdown etc.

lib/Onyx.ts
```diff
@@ -225,6 +228,15 @@ function set<TKey extends OnyxKey>(key: TKey, value: OnyxSetInput<TKey>, options
...
         .catch((error) => OnyxUtils.evictStorageAndRetry(error, set, key, valueWithoutNestedNullValues))
         .then(() => {
+            try {
+                const storageManager = getStorageManager();
+                if (storageManager) {
+                    storageManager.trackKeySet(key);
+                }
+            } catch (error) {
+                Logger.logInfo(`Failed to track key set: ${error}`);
+            }
```
fabioh8010:
```suggestion
            storageManager.trackKeySet(key);
```

As mentioned previously, this kind of thing could be much simpler if we just import the storage manager singleton (like we do for connection manager and cache). Also the failing scenario can be handled inside `trackKeySet` itself, wdyt?

lib/storage-eviction/index.ts
fabioh8010, line 1: I think we could rename the folder to `storageManager` and the class to `OnyxStorageManager`, to be more aligned with the other files.

lib/storage-eviction/index.ts
```diff
@@ -0,0 +1,523 @@
...
+import type {OnyxKey} from '../types';
+import Storage from '../storage';
+import * as Logger from '../Logger';
+import type {StorageUsageConfig, StorageKeyInfo, StorageMetadata} from './types';
+import {DEFAULT_STORAGE_CONFIG} from './types';
+
+const MILLISECONDS_PER_DAY = 1000 * 60 * 60 * 24;
+
+type StorageManagerConfig = Partial<StorageUsageConfig>;
+type StorageCleanupResult = {cleanedKeys: OnyxKey[]; timeElapsed: number; errors: string[]};
```
fabioh8010: Could we add comments for these types?

lib/storage-eviction/index.ts
```diff
@@ -0,0 +1,523 @@
...
+
+    private evictableKeys: string[] = [];
+
+    private isInitialized = false;
+
+    private monitoringInterval: NodeJS.Timeout | null = null;
+
+    private isMonitoring = false;
+
+    private cleanupInProgress = false;
```
fabioh8010: Let's add meaningful comments for all these

lib/storage-eviction/index.ts
```diff
@@ -0,0 +1,523 @@
...
+
+type StorageManagerConfig = Partial<StorageUsageConfig>;
+type StorageCleanupResult = {cleanedKeys: OnyxKey[]; timeElapsed: number; errors: string[]};
+
+class StorageManager {
+    private keyInfoMap = new Map<OnyxKey, StorageKeyInfo>();
+
+    public config: StorageUsageConfig;
+
+    private evictableKeys: string[] = [];
```
fabioh8010: Couldn't the cache eviction keys be reused here? Or is it a problem?

lib/storage-eviction/index.ts
```diff
@@ -0,0 +1,523 @@
...
+            .then((allKeys) => {
+                Logger.logInfo(`Found ${allKeys.length} keys in storage`);
+                return this.trackExistingKeys(allKeys);
+            })
+            .catch((error) => {
+                Logger.logAlert(`Failed to initialize storage tracking: ${error}`);
+            });
+    }
+
+    private trackExistingKeys(keys: OnyxKey[]): Promise<Array<void | QueryResult<QueryResultRow>>> {
```
fabioh8010: Why would we return `Array<void | QueryResult<QueryResultRow>>`?

lib/storage-eviction/index.ts
```diff
@@ -0,0 +1,523 @@
...
+        const dataKeys = keys.filter((key) => !key.startsWith('__onyx_meta_'));
+        const evictableDataKeys = dataKeys.filter((key) => this.isKeyEvictable(key));
+        Logger.logInfo(`[StorageManager] Filtering keys: ${dataKeys.length} total, ${evictableDataKeys.length} evictable`);
+
+        const trackingPromises = evictableDataKeys.map((key) => this.initializeKeyTracking(key));
+
+        return Promise.all(trackingPromises);
+    }
+
+    private initializeKeyTracking(key: OnyxKey): Promise<void | QueryResult<QueryResultRow>> {
```
fabioh8010: Same about `QueryResult<QueryResultRow>`

lib/storage-eviction/index.ts
```diff
@@ -0,0 +1,523 @@
...
+                this.isInitialized = true;
+                Logger.logInfo('StorageManager initialized successfully');
+            })
+            .catch((error) => {
+                Logger.logAlert(`Failed to initialize StorageManager: ${error}`);
+                throw error;
+            });
+    }
+
+    private initializeTracking(): Promise<void | Array<void | QueryResult<QueryResultRow>>> {
```
fabioh8010: Same about `Array<void | QueryResult<QueryResultRow>>`

lib/storage-eviction/index.ts
```diff
@@ -0,0 +1,523 @@
...
+
+    private initializeKeyTracking(key: OnyxKey): Promise<void | QueryResult<QueryResultRow>> {
+        const now = Date.now();
+
+        return this.loadMetadata(key)
+            .then((metadata) => {
+                const keyInfo = {
+                    key,
+                    lastAccessed: metadata?.lastAccessed || now,
+                    createdAt: metadata?.createdAt || now,
```
fabioh8010:
```suggestion
                    lastAccessed: metadata?.lastAccessed ?? now,
                    createdAt: metadata?.createdAt ?? now,
```

lib/storage-eviction/index.ts
```diff
@@ -0,0 +1,523 @@
...
+            const metadataKey = this.getMetadataKey(key);
+            Storage.removeItem(metadataKey).catch((error) => {
+                Logger.logInfo(`Failed to remove metadata for ${key}: ${error}`);
+            });
+        }
+    }
+
+    /**
+     * Updates the list of keys that are eligible for eviction.
+     * @param keys - Array of key names or patterns (patterns ending with '_' match prefixes)
```
fabioh8010: "patterns (patterns ending with '_' match prefixes)" what does that mean?

lib/storage-eviction/index.ts
```diff
@@ -0,0 +1,523 @@
...
+     */
+    setEvictableKeys(keys: string[]): void {
+        this.evictableKeys = keys;
+    }
+
+    private isKeyEvictable(key: OnyxKey): boolean {
+        return this.evictableKeys.some((pattern) => {
+            return pattern.endsWith('_') ? key.startsWith(pattern) : key === pattern;
```
fabioh8010: If this is because of collection keys, you shouldn't use `.endsWith('_')`. You should use `splitCollectionMemberKey` or something related, we have some functions in OnyxUtils for checking collection/collection member keys.

lib/storage-eviction/index.ts
```diff
@@ -0,0 +1,523 @@
...
+        for (const keyInfo of this.keyInfoMap.values()) {
+            if (this.isKeyEvictable(keyInfo.key)) {
+                count++;
+            }
+        }
+        return count;
+    }
+
+    private getMetadataKey(key: OnyxKey): string {
+        return `__onyx_meta_${key}`;
```
fabioh8010: Could we create a constant from `__onyx_meta_`?

lib/storage-eviction/index.ts
```diff
@@ -0,0 +1,523 @@
...
+            Logger.logInfo(`All tracked keys: [${allKeys.join(', ')}]`);
+            Logger.logInfo(`Evictable patterns: [${this.evictableKeys.join(', ')}]`);
+            return false;
+        }
+
+        let expiredCount = 0;
+        for (const keyInfo of this.keyInfoMap.values()) {
+            if (this.isKeyEvictable(keyInfo.key)) {
+                const shouldEvict = this.shouldEvictKey(keyInfo);
+                const daysSinceAccess = (now - keyInfo.lastAccessed) / (1000 * 60 * 60 * 24);
```
fabioh8010: This `(1000 * 60 * 60 * 24)` should be in a variabble

lib/storage-eviction/index.ts
```diff
@@ -0,0 +1,523 @@
...
+        if (!this.config.enabled) return false;
+
+        const now = Date.now();
+        const daysSinceAccess = (now - keyInfo.lastAccessed) / MILLISECONDS_PER_DAY;
+        const ageInDays = (now - keyInfo.createdAt) / MILLISECONDS_PER_DAY;
+
+        return daysSinceAccess > this.config.maxIdleDays || ageInDays > this.config.maxAgeDays;
+    }
+
+    private executeCleanup(keysToEvict: StorageKeyInfo[]): Promise<{successfulKeys: OnyxKey[]; failedKeys: OnyxKey[]}> {
```
fabioh8010: Please extract `{successfulKeys: OnyxKey[]; failedKeys: OnyxKey[]}` to a separate type.

lib/storage-eviction/types.ts
```diff
@@ -0,0 +1,32 @@
...
+interface StorageKeyInfo {
+    key: OnyxKey;
+    lastAccessed: number;
+    createdAt: number;
+}
+
+interface StorageMetadata {
+    lastAccessed: number;
+    createdAt: number;
+}
```
fabioh8010: Let's add comments for all these types and properties

tests/unit/cacheEvictionTest.ts
```diff
@@ -20,6 +20,7 @@ test('Cache eviction', () => {
             Onyx.init({
                 keys: ONYX_KEYS,
                 evictableKeys: [ONYX_KEYS.COLLECTION.TEST_KEY],
+                storageManager: undefined, // Disable StorageManager for this test to use cache eviction
```
fabioh8010: In `lib/types.ts` is saying

>     /**
>      * Configuration for persistent storage eviction system.
>      * Set to `false` to disable storage eviction entirely.
>      * @default undefined (uses default storage eviction configuration)
>      */

So this doesn't make sense and it should be `false`, correct?

fabioh8010 (commented):
First round of reviews

lib/storage-eviction/index.ts
kubabutkiewicz: I remember testing with the same keys which we are using for eviction and there was some keys which were good for cache eviction but when I used them for storage eviction app was having a problems, dont remeber which keys was that but I remember that there was that case

lib/storage-eviction/index.ts
kubabutkiewicz: I could use the `isCollectionKey` method, but the problem is that when I am importing the OnyxUtils here its creates cycle dependecy as we are importing OnyxStorageManger also there :(

fabioh8010:
Will do a second review tomorrow!

lib/OnyxStorageManager/index.ts
```diff
@@ -0,0 +1,496 @@
...
+     * Current configuration for storage usage and eviction policies.
+     * Controls parameters like max idle days, max age, and which keys are evictable.
+     */
+    public config: StorageUsageConfig;
+
+    /**
+     * List of keys that are eligible for eviction based on the current config.
+     * Only these keys will be considered for automatic cleanup.
+     */
+    private evictableKeys: string[] = [];
```
fabioh8010:
```suggestion
    private evictableKeys: OnyxKey[] = [];
```

lib/OnyxStorageManager/index.ts
fabioh8010, line 1: Could you create Unit and Reassure tests for this class? You can take `OnyxConnectionManager` as inspiration

lib/OnyxStorageManager/index.ts
```diff
@@ -0,0 +1,496 @@
...
+                this.isInitialized = true;
+                Logger.logInfo('StorageManager initialized successfully');
+            })
+            .catch((error) => {
+                Logger.logAlert(`Failed to initialize StorageManager: ${error}`);
+                throw error;
+            });
+    }
+
+    private initializeTracking(): Promise<void | void[]> {
```
fabioh8010:
```suggestion
    private initializeTracking(): Promise<void> {
```

lib/OnyxStorageManager/index.ts
```diff
@@ -0,0 +1,496 @@
...
+
+    private trackExistingKeys(keys: OnyxKey[]): Promise<void[]> {
+        const dataKeys = keys.filter((key) => !key.startsWith('__onyx_meta_'));
+        const evictableDataKeys = dataKeys.filter((key) => this.isKeyEvictable(key));
+        Logger.logInfo(`[StorageManager] Filtering keys: ${dataKeys.length} total, ${evictableDataKeys.length} evictable`);
+
+        const trackingPromises = evictableDataKeys.map((key) => this.initializeKeyTracking(key));
+
+        return Promise.all(trackingPromises);
+    }
```
fabioh8010:
```suggestion
    private trackExistingKeys(keys: OnyxKey[]): Promise<void> {
        const dataKeys = keys.filter((key) => !key.startsWith('__onyx_meta_'));
        const evictableDataKeys = dataKeys.filter((key) => this.isKeyEvictable(key));
        Logger.logInfo(`[StorageManager] Filtering keys: ${dataKeys.length} total, ${evictableDataKeys.length} evictable`);

        const trackingPromises = evictableDataKeys.map((key) => this.initializeKeyTracking(key));

        return Promise.all(trackingPromises).then(() => undefined);
    }
```

lib/OnyxStorageManager/index.ts
```diff
@@ -0,0 +1,496 @@
...
+
+    private initializeKeyTracking(key: OnyxKey): Promise<void> {
+        const now = Date.now();
+
+        return this.loadMetadata(key)
+            .then((metadata) => {
+                const keyInfo = {
```
fabioh8010:
```suggestion
                const keyInfo: StorageKeyInfo = {
```

lib/OnyxStorageManager/index.ts
```diff
@@ -0,0 +1,496 @@
...
+        const isEvictable = this.isKeyEvictable(key);
+        const now = Date.now();
+        const existing = this.keyInfoMap.get(key);
+
+        this.updateKeyInfo(key, now, isEvictable, !existing);
+    }
+
+    private updateKeyInfo(key: OnyxKey, now: number, isEvictable: boolean, isNewKey = false): void {
+        const existing = this.keyInfoMap.get(key);
+        const keyInfo = {
```
fabioh8010:
```suggestion
        const keyInfo: StorageKeyInfo = {
```

lib/OnyxStorageManager/index.ts
```diff
@@ -0,0 +1,496 @@
...
+            key,
+            lastAccessed: now,
+            createdAt: isNewKey ? now : existing?.createdAt || now,
+        };
+
+        this.keyInfoMap.set(key, keyInfo);
+
+        if (isEvictable) {
+            const metadata = {
```
fabioh8010:
```suggestion
            const metadata: StorageMetadata = {
```

lib/OnyxStorageManager/index.ts
```diff
@@ -0,0 +1,496 @@
...
+                Logger.logInfo(`Failed to remove metadata for ${key}: ${error}`);
+            });
+        }
+    }
+
+    /**
+     * Updates the list of keys that are eligible for eviction.
+     * @param keys - Array of Onyx keys or collection keys
+     */
+    setEvictableKeys(keys: string[]): void {
```
fabioh8010:
```suggestion
    setEvictableKeys(keys: OnyxKey[]): void {
```

lib/storage-eviction/index.ts
fabioh8010: I think we could extract some function from OnyxUtils like isCollectionKey, isCollectionMemberKey, splitCollectionMemberKey, getCollectionKey, etc. in a separate OnyxKeyUtils file, that would solve the problem I think.

Anyway we need a good solution for this, `.endsWith('_')` is not the correct way for checking

lib/Onyx.ts
```diff
@@ -225,6 +229,12 @@ function set<TKey extends OnyxKey>(key: TKey, value: OnyxSetInput<TKey>, options
     return Storage.setItem(key, valueWithoutNestedNullValues)
         .catch((error) => OnyxUtils.evictStorageAndRetry(error, set, key, valueWithoutNestedNullValues))
         .then(() => {
+            try {
```
fabioh8010: I wonder if we really need these `trackKeySet`, I checked the function and we are not throwing anything

lib/OnyxStorageManager/index.ts
```diff
@@ -0,0 +1,496 @@
...
+const MILLISECONDS_PER_DAY = 1000 * 60 * 60 * 24;
+const METADATA_KEY_PREFIX = '__onyx_meta_';
+
+/**
+ * Represents the result of a storage cleanup operation.
+ * @property cleanedKeys - An array of Onyx keys that were successfully cleaned/evicted.
+ * @property timeElapsed - The time taken (in milliseconds) to perform the cleanup.
+ * @property errors - An array of error messages encountered during the cleanup process.
+ */
+type StorageCleanupResult = {
```
dukenv0307: we should move it to types.ts

lib/OnyxStorageManager/index.ts
```diff
@@ -0,0 +1,496 @@
...
+    cleanedKeys: OnyxKey[];
+    timeElapsed: number;
+    errors: string[];
+};
+
+type CleanupExecutionResult = {
```
dukenv0307: same as above

lib/OnyxStorageManager/index.ts
```diff
@@ -0,0 +1,478 @@
...
+
+    private updateKeyInfo(key: OnyxKey, now: number, isEvictable: boolean, isNewKey = false): void {
+        const existing = this.keyInfoMap.get(key);
+        const keyInfo: StorageKeyInfo = {
+            key,
+            lastAccessed: now,
+            createdAt: isNewKey ? now : existing?.createdAt || now,
+        };
+
+        this.keyInfoMap.set(key, keyInfo);
```
dukenv0307, line 178: Should we store the unevictable keys?

lib/OnyxStorageManager/index.ts
```diff
@@ -0,0 +1,478 @@
...
+
+        const isEvictable = this.isKeyEvictable(key);
+        const now = Date.now();
+        const existing = this.keyInfoMap.get(key);
+
+        this.updateKeyInfo(key, now, isEvictable, !existing);
+    }
+
+    private updateKeyInfo(key: OnyxKey, now: number, isEvictable: boolean, isNewKey = false): void {
+        const existing = this.keyInfoMap.get(key);
```
dukenv0307: Why don't we use isNewKey?

fabioh8010 (commented):
@kubabutkiewicz I think the only thing left here is update documentation here -> https://github.com/Expensify/react-native-onyx/blob/main/README.md#cache-eviction

Let's make sure it's very descriptive how the cache/storage eviction system works!

mountiny:
100% can we use LRU mechanism instead of some arbitrary time?

roryabraham:
Slack thread is here: https://expensify.slack.com/archives/C05LX9D6E07/p1755077335799619

PR #671, opened 2025-08-05
perf: cache useOnyx snapshots

TMisiukiewicz:
### Related Issues

$ https://github.com/Expensify/App/issues/68014

### Automated Tests

Added corresponding tests to `useOnyxTest.ts` file and created tests for snapshot cache. All tests in the E/App codebase are passing as well.
### Manual Tests

1. Create fresh account
2. Create new workspace
3. Verify the workspace was created
4. Create two expenses in a workspace with the same amount, merchant and category
5. Verify the RBR is displayed in LHN
6. Enter the expense chat
7. Resolve duplicates
9. Send message
10. Go back to LHN
11. Verify the RBR is not displayed in the LHN 
12. Long press on the Concierge chat
13. Pin the chat
14. Verify the chat was pinned
### Author Checklist

    - [x] Android / native
    - [x] Android / Chrome
    - [x] iOS / native
    - [x] iOS / Safari
    - [x] MacOS / Chrome / Safari
    - [x] MacOS / Desktop

### Screenshots/Videos
<details>

https://github.com/user-attachments/assets/2c307b2f-7441-4db2-9362-d20d67e35865

</details>

<details>

</details>

<details>

https://github.com/user-attachments/assets/81f6a25c-2704-449a-a746-0282830285f4

</details>

<details>

https://github.com/user-attachments/assets/473a59b2-de8f-4dab-b7c5-ed05b3c3cae2

</details>

<details>

https://github.com/user-attachments/assets/ed446cec-ab28-4734-a7ca-ab7cad84efe3

</details>

<details>

https://github.com/user-attachments/assets/53b652d5-18fe-4f91-92b9-0a6d629d295d

</details>

### DIFF
--- a/lib/OnyxConnectionManager.ts
+++ b/lib/OnyxConnectionManager.ts
@@ -263,6 +263,7 @@ class OnyxConnectionManager {
         });
 
         this.connectionsMap.clear();
+
         // Clear snapshot cache when all connections are disconnected
         onyxSnapshotCache.clear();
     }
@@ -272,6 +273,7 @@ class OnyxConnectionManager {
      */
     refreshSessionID(): void {
         this.sessionID = Str.guid();
+
         // Clear snapshot cache when session refreshes to avoid stale cache issues
         onyxSnapshotCache.clear();
     }
--- a/lib/OnyxConnectionManager.ts
+++ b/lib/OnyxConnectionManager.ts
@@ -6,6 +6,7 @@ import * as Str from './Str';
 import type {CollectionConnectCallback, DefaultConnectCallback, DefaultConnectOptions, OnyxKey, OnyxValue} from './types';
 import utils from './utils';
 import cache from './OnyxCache';
+import onyxSnapshotCache from './OnyxSnapshotCache';
 
 type ConnectCallback = DefaultConnectCallback<OnyxKey> | CollectionConnectCallback<OnyxKey>;
 
@@ -262,6 +263,8 @@ class OnyxConnectionManager {
         });
 
         this.connectionsMap.clear();
+        // Clear snapshot cache when all connections are disconnected
+        onyxSnapshotCache.clear();
     }
 
     /**
@@ -269,6 +272,8 @@ class OnyxConnectionManager {
      */
     refreshSessionID(): void {
         this.sessionID = Str.guid();
+        // Clear snapshot cache when session refreshes to avoid stale cache issues
+        onyxSnapshotCache.clear();
     }
 
     /**
--- a/lib/OnyxSnapshotCache.ts
+++ b/lib/OnyxSnapshotCache.ts
@@ -68,10 +68,22 @@ class OnyxSnapshotCache {
     }
 
     /**
-     * O(1) cache invalidation - just delete the entire cache for this key
+     * O(1) cache invalidation - delete cache for this key and related collection keys
      */
     invalidateForKey(keyToInvalidate: string): void {
+        // Always invalidate the exact key
         this.snapshotCache.delete(keyToInvalidate);
+        
+        // For collection member keys, also invalidate the parent collection
+        if (keyToInvalidate.includes('_')) {
+            const baseKey = keyToInvalidate.split('_')[0];
+            // Invalidate all related collection members
+            for (const [cacheKey] of this.snapshotCache) {
+                if (cacheKey.startsWith(baseKey + '_') || cacheKey === baseKey) {
+                    this.snapshotCache.delete(cacheKey);
+                }
+            }
+        }
     }
 
     /**
--- a/lib/OnyxSnapshotCache.ts
+++ b/lib/OnyxSnapshotCache.ts
@@ -39,9 +39,9 @@ class OnyxSnapshotCache {
     }
 
     /**
-     * Fast cache key generation for selector/options combination
+     * Fast cache key generation for useOnyx options combination
      */
-    generateSelectorKey<TKey extends OnyxKey, TReturnValue>(options?: UseOnyxOptions<TKey, TReturnValue>): string {
+    generateCacheKey<TKey extends OnyxKey, TReturnValue>(options?: UseOnyxOptions<TKey, TReturnValue>): string {
         const selectorId = options?.selector ? this.getSelectorId(options.selector) : 'no_selector';
         // Create options hash without expensive JSON.stringify
         const initWithStoredValues = options?.initWithStoredValues ?? true;
@@ -51,21 +51,21 @@ class OnyxSnapshotCache {
     }
 
     /**
-     * Get cached snapshot result for a key and selector combination
+     * Get cached snapshot result for a key and cache key combination
      */
-    getCachedResult(key: string, selectorKey: string): any {
+    getCachedResult(key: string, cacheKey: string): any {
         const keyCache = this.snapshotCache.get(key);
-        return keyCache?.get(selectorKey);
+        return keyCache?.get(cacheKey);
     }
 
     /**
-     * Set cached snapshot result for a key and selector combination
+     * Set cached snapshot result for a key and cache key combination
      */
-    setCachedResult(key: string, selectorKey: string, result: any): void {
+    setCachedResult(key: string, cacheKey: string, result: any): void {
         if (!this.snapshotCache.has(key)) {
             this.snapshotCache.set(key, new Map());
         }
-        this.snapshotCache.get(key)!.set(selectorKey, result);
+        this.snapshotCache.get(key)!.set(cacheKey, result);
     }
 
     /**
--- a/lib/OnyxSnapshotCache.ts
+++ b/lib/OnyxSnapshotCache.ts
@@ -17,15 +17,22 @@ class OnyxSnapshotCache {
      */
     private selectorIdMap: Map<UseOnyxSelector<any, any>, string>;
 
-    /**
-     * Counter for generating unique selector IDs
-     */
-    private selectorCounter: number;
-
     constructor() {
         this.snapshotCache = new Map();
         this.selectorIdMap = new Map();
-        this.selectorCounter = 0;
+    }
+
+    /**
+     * Hash string to generate deterministic ID
+     */
+    private hashString(str: string): string {
+        let hash = 0;
+        for (let i = 0; i < str.length; i++) {
+            const char = str.charCodeAt(i);
+            hash = (hash << 5) - hash + char;
+            hash = hash & hash; // Convert to 32-bit integer
+        }
+        return Math.abs(hash).toString(36);
     }
 
     /**
@@ -33,7 +40,11 @@ class OnyxSnapshotCache {
      */
     getSelectorId(selector: UseOnyxSelector<any, any>): string {
         if (!this.selectorIdMap.has(selector)) {
-            this.selectorIdMap.set(selector, `selector_${++this.selectorCounter}`);
+            const funcStr = selector.toString();
+            const funcName = selector.name || 'anonymous';
+            const hash = this.hashString(funcStr);
+            const id = `${funcName}_${hash}`;
+            this.selectorIdMap.set(selector, id);
         }
         return this.selectorIdMap.get(selector)!;
     }
@@ -99,7 +110,6 @@ class OnyxSnapshotCache {
      */
     clearSelectorIds(): void {
         this.selectorIdMap.clear();
-        this.selectorCounter = 0;
     }
 }
 
--- a/lib/OnyxSnapshotCache.ts
+++ b/lib/OnyxSnapshotCache.ts
@@ -54,7 +54,7 @@ class OnyxSnapshotCache {
      * Other options like `canEvict`, `reuseConnection`, and `allowDynamicKey` don't affect the data transformation
      * or timing behavior of getSnapshot, so they're excluded from the cache key for better cache hit rates.
      */
-    generateCacheKey<TKey extends OnyxKey, TReturnValue>(options?: UseOnyxOptions<TKey, TReturnValue>): string {
+    generateCacheKey<TKey extends OnyxKey, TReturnValue>(options: Pick<UseOnyxOptions<TKey, TReturnValue>, 'selector' | 'initWithStoredValues' | 'allowStaleData' | 'canBeMissing'>): string {
         const selectorId = options?.selector ? this.getSelectorId(options.selector) : 'no_selector';
         // Create options hash without expensive JSON.stringify
         const initWithStoredValues = options?.initWithStoredValues ?? true;
--- a/lib/OnyxSnapshotCache.ts
+++ b/lib/OnyxSnapshotCache.ts
@@ -9,7 +9,7 @@ import type {UseOnyxOptions, UseOnyxResult, UseOnyxSelector} from './useOnyx';
 class OnyxSnapshotCache {
     /**
      * Snapshot cache is a two-level map. The top-level keys are Onyx keys. The top-level values maps.
-     * The second-level keys are a custom composite string defined by this.generateCacheKey. These represent a unique useOnyx config, which is not fully represented by the Onyx key alone.
+     * The second-level keys are a custom composite string defined by this.registerConsumer. These represent a unique useOnyx config, which is not fully represented by the Onyx key alone.
      * The reason we have two levels is for performance: not to make cache access faster, but to make cache invalidation faster.
      * We can invalidate the snapshot cache for a given Onyx key with one map.delete operation on the top-level map, rather than having to loop through a large single-level map and delete any matching keys.
      */
@@ -25,10 +25,17 @@ class OnyxSnapshotCache {
      */
     private selectorIDCounter: number;
 
+    /**
+     * Reference counting for cache keys to enable automatic cleanup.
+     * Maps cache key (string) to number of consumers using it.
+     */
+    private cacheKeyRefCounts: Map<string, number>;
+
     constructor() {
         this.snapshotCache = new Map();
         this.selectorIDMap = new Map();
         this.selectorIDCounter = 0;
+        this.cacheKeyRefCounts = new Map();
     }
 
     /**
@@ -44,7 +51,8 @@ class OnyxSnapshotCache {
     }
 
     /**
-     * Fast cache key generation for useOnyx options combination.
+     * Register a consumer for a cache key and return the cache key.
+     * Generates cache key and increments reference counter.
      *
      * The properties used to generate the cache key are handpicked for performance reasons and
      * according to their purpose and effect they produce in the useOnyx hook behavior:
@@ -57,14 +65,46 @@ class OnyxSnapshotCache {
      * Other options like `canEvict`, `reuseConnection`, and `allowDynamicKey` don't affect the data transformation
      * or timing behavior of getSnapshot, so they're excluded from the cache key for better cache hit rates.
      */
-    generateCacheKey<TKey extends OnyxKey, TReturnValue>(options: Pick<UseOnyxOptions<TKey, TReturnValue>, 'selector' | 'initWithStoredValues' | 'allowStaleData' | 'canBeMissing'>): string {
+    registerConsumer<TKey extends OnyxKey, TReturnValue>(options: Pick<UseOnyxOptions<TKey, TReturnValue>, 'selector' | 'initWithStoredValues' | 'allowStaleData' | 'canBeMissing'>): string {
         const selectorID = options?.selector ? this.getSelectorID(options.selector) : 'no_selector';
 
         // Create options hash without expensive JSON.stringify
         const initWithStoredValues = options?.initWithStoredValues ?? true;
         const allowStaleData = options?.allowStaleData ?? false;
         const canBeMissing = options?.canBeMissing ?? true;
-        return `${selectorID}_${initWithStoredValues}_${allowStaleData}_${canBeMissing}`;
+        const cacheKey = `${selectorID}_${initWithStoredValues}_${allowStaleData}_${canBeMissing}`;
+
+        // Increment reference count for this cache key
+        const currentCount = this.cacheKeyRefCounts.get(cacheKey) || 0;
+        this.cacheKeyRefCounts.set(cacheKey, currentCount + 1);
+
+        return cacheKey;
+    }
+
+    /**
+     * Deregister a consumer for a cache key.
+     * Decrements reference counter and removes cache entry if no consumers remain.
+     */
+    deregisterConsumer(key: OnyxKey, cacheKey: string): void {
+        const currentCount = this.cacheKeyRefCounts.get(cacheKey) || 0;
+
+        if (currentCount <= 1) {
+            // Last consumer - remove from reference counter and cache
+            this.cacheKeyRefCounts.delete(cacheKey);
+
+            // Remove from snapshot cache
+            const keyCache = this.snapshotCache.get(key);
+            if (keyCache) {
+                keyCache.delete(cacheKey);
+                // If this was the last cache entry for this Onyx key, remove the key entirely
+                if (keyCache.size === 0) {
+                    this.snapshotCache.delete(key);
+                }
+            }
+        } else {
+            // Still has other consumers - just decrement count
+            this.cacheKeyRefCounts.set(cacheKey, currentCount - 1);
+        }
     }
 
     /**
--- a/lib/OnyxSnapshotCache.ts
+++ b/lib/OnyxSnapshotCache.ts
@@ -1,0 +1,97 @@
+import type {OnyxKey} from './types';
+import type {UseOnyxOptions, UseOnyxSelector} from './useOnyx';
+
+/**
+ * Manages snapshot caching for useOnyx hook performance optimization.
+ * Handles selector function tracking and memoized getSnapshot results.
+ */
+class OnyxSnapshotCache {
+    /**
+     * Snapshot cache for ultimate performance - separate cache per Onyx key
+     */
+    private snapshotCache: Map<string, Map<string, any>>;
+
+    /**
+     * Maps selector functions to unique IDs for cache key generation
+     */
+    private selectorIdMap: Map<UseOnyxSelector<any, any>, string>;
+
+    /**
+     * Counter for generating unique selector IDs
+     */
+    private selectorCounter: number;
+
+    constructor() {
+        this.snapshotCache = new Map();
+        this.selectorIdMap = new Map();
+        this.selectorCounter = 0;
+    }
+
+    /**
+     * Generate unique ID for selector functions
+     */
+    getSelectorId(selector: UseOnyxSelector<any, any>): string {
+        if (!this.selectorIdMap.has(selector)) {
+            this.selectorIdMap.set(selector, `selector_${++this.selectorCounter}`);
+        }
+        return this.selectorIdMap.get(selector)!;
+    }
+
+    /**
+     * Fast cache key generation for selector/options combination
+     */
+    generateSelectorKey<TKey extends OnyxKey, TReturnValue>(options?: UseOnyxOptions<TKey, TReturnValue>): string {
+        const selectorId = options?.selector ? this.getSelectorId(options.selector) : 'no_selector';
+        // Create options hash without expensive JSON.stringify
+        const initWithStoredValues = options?.initWithStoredValues ?? true;
+        const allowStaleData = options?.allowStaleData ?? false;
+        const canBeMissing = options?.canBeMissing ?? true;
+        return `${selectorId}_${initWithStoredValues}_${allowStaleData}_${canBeMissing}`;
+    }
+
+    /**
+     * Get cached snapshot result for a key and selector combination
+     */
+    getCachedResult(key: string, selectorKey: string): any {
+        const keyCache = this.snapshotCache.get(key);
+        return keyCache?.get(selectorKey);
+    }
+
+    /**
+     * Set cached snapshot result for a key and selector combination
+     */
+    setCachedResult(key: string, selectorKey: string, result: any): void {
+        if (!this.snapshotCache.has(key)) {
+            this.snapshotCache.set(key, new Map());
+        }
+        this.snapshotCache.get(key)!.set(selectorKey, result);
+    }
+
+    /**
+     * O(1) cache invalidation - just delete the entire cache for this key
+     */
+    invalidateForKey(keyToInvalidate: string): void {
+        this.snapshotCache.delete(keyToInvalidate);
+    }
+
+    /**
+     * Clear all snapshot cache
+     */
+    clear(): void {
+        this.snapshotCache.clear();
+    }
+
+    /**
+     * Clear selector ID mappings (useful for testing)
+     */
+    clearSelectorIds(): void {
+        this.selectorIdMap.clear();
+        this.selectorCounter = 0;
+    }
+}
+
+// Create and export a singleton instance
+const onyxSnapshotCache = new OnyxSnapshotCache();
+
+export default onyxSnapshotCache;
+export {OnyxSnapshotCache};
--- a/lib/OnyxSnapshotCache.ts
+++ b/lib/OnyxSnapshotCache.ts
@@ -17,33 +17,23 @@ class OnyxSnapshotCache {
      */
     private selectorIdMap: Map<UseOnyxSelector<any, any>, string>;
 
+    /**
+     * Counter for generating incremental selector IDs
+     */
+    private selectorIdCounter: number;
+
     constructor() {
         this.snapshotCache = new Map();
         this.selectorIdMap = new Map();
+        this.selectorIdCounter = 0;
     }
 
     /**
-     * Hash string to generate deterministic ID
-     */
-    private hashString(str: string): string {
-        let hash = 0;
-        for (let i = 0; i < str.length; i++) {
-            const char = str.charCodeAt(i);
-            hash = (hash << 5) - hash + char;
-            hash = hash & hash; // Convert to 32-bit integer
-        }
-        return Math.abs(hash).toString(36);
-    }
-
-    /**
-     * Generate unique ID for selector functions
+     * Generate unique ID for selector functions using incrementing numbers
      */
     getSelectorId(selector: UseOnyxSelector<any, any>): string {
         if (!this.selectorIdMap.has(selector)) {
-            const funcStr = selector.toString();
-            const funcName = selector.name || 'anonymous';
-            const hash = this.hashString(funcStr);
-            const id = `${funcName}_${hash}`;
+            const id = `selector_${this.selectorIdCounter++}`;
             this.selectorIdMap.set(selector, id);
         }
         return this.selectorIdMap.get(selector)!;
@@ -110,6 +100,7 @@ class OnyxSnapshotCache {
      */
     clearSelectorIds(): void {
         this.selectorIdMap.clear();
+        this.selectorIdCounter = 0;
     }
 }
 
--- a/lib/OnyxSnapshotCache.ts
+++ b/lib/OnyxSnapshotCache.ts
@@ -90,15 +90,12 @@ class OnyxSnapshotCache {
         // Always invalidate the exact key
         this.snapshotCache.delete(keyToInvalidate);
 
-        // Handle collection-related keys with selective invalidation
-        if (OnyxUtils.isCollectionKey(keyToInvalidate)) {
-            const baseKey = OnyxUtils.getCollectionKey(keyToInvalidate);
-
-            // If this is a collection member key (e.g., "reports_123")
-            // Only invalidate upward to the collection base key
-            if (baseKey !== keyToInvalidate) {
-                this.snapshotCache.delete(baseKey);
-            }
+        try {
+            // Check if the key is a collection member and invalidate the collection base key
+            const collectionBaseKey = OnyxUtils.getCollectionKey(keyToInvalidate);
+            this.snapshotCache.delete(collectionBaseKey);
+        } catch (e) {
+            // do nothing - this just means the key is not a collection member
         }
     }
 
--- a/lib/OnyxSnapshotCache.ts
+++ b/lib/OnyxSnapshotCache.ts
@@ -58,13 +58,13 @@ class OnyxSnapshotCache {
      * or timing behavior of getSnapshot, so they're excluded from the cache key for better cache hit rates.
      */
     generateCacheKey<TKey extends OnyxKey, TReturnValue>(options: Pick<UseOnyxOptions<TKey, TReturnValue>, 'selector' | 'initWithStoredValues' | 'allowStaleData' | 'canBeMissing'>): string {
-        const selectorId = options?.selector ? this.getSelectorID(options.selector) : 'no_selector';
+        const selectorID = options?.selector ? this.getSelectorID(options.selector) : 'no_selector';
 
         // Create options hash without expensive JSON.stringify
         const initWithStoredValues = options?.initWithStoredValues ?? true;
         const allowStaleData = options?.allowStaleData ?? false;
         const canBeMissing = options?.canBeMissing ?? true;
-        return `${selectorId}_${initWithStoredValues}_${allowStaleData}_${canBeMissing}`;
+        return `${selectorID}_${initWithStoredValues}_${allowStaleData}_${canBeMissing}`;
     }
 
     /**
--- a/lib/OnyxSnapshotCache.ts
+++ b/lib/OnyxSnapshotCache.ts
@@ -15,29 +15,29 @@ class OnyxSnapshotCache {
     /**
      * Maps selector functions to unique IDs for cache key generation
      */
-    private selectorIdMap: Map<UseOnyxSelector<OnyxKey, unknown>, number>;
+    private selectorIDMap: Map<UseOnyxSelector<OnyxKey, unknown>, number>;
 
     /**
      * Counter for generating incremental selector IDs
      */
-    private selectorIdCounter: number;
+    private selectorIDCounter: number;
 
     constructor() {
         this.snapshotCache = new Map();
-        this.selectorIdMap = new Map();
-        this.selectorIdCounter = 0;
+        this.selectorIDMap = new Map();
+        this.selectorIDCounter = 0;
     }
 
     /**
      * Generate unique ID for selector functions using incrementing numbers
      */
-    getSelectorId<TKey extends OnyxKey, TReturnValue>(selector: UseOnyxSelector<TKey, TReturnValue>): number {
+    getSelectorID<TKey extends OnyxKey, TReturnValue>(selector: UseOnyxSelector<TKey, TReturnValue>): number {
         const typedSelector = selector as unknown as UseOnyxSelector<OnyxKey, unknown>;
-        if (!this.selectorIdMap.has(typedSelector)) {
-            const id = this.selectorIdCounter++;
-            this.selectorIdMap.set(typedSelector, id);
+        if (!this.selectorIDMap.has(typedSelector)) {
+            const id = this.selectorIDCounter++;
+            this.selectorIDMap.set(typedSelector, id);
         }
-        return this.selectorIdMap.get(typedSelector)!;
+        return this.selectorIDMap.get(typedSelector)!;
     }
 
     /**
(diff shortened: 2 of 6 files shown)


lib/OnyxSnapshotCache.ts
```diff
@@ -0,0 +1,110 @@
...
+        this.selectorIdMap = new Map();
+        this.selectorCounter = 0;
+    }
+
+    /**
+     * Generate unique ID for selector functions
+     */
+    getSelectorId(selector: UseOnyxSelector<any, any>): string {
+        if (!this.selectorIdMap.has(selector)) {
+            this.selectorIdMap.set(selector, `selector_${++this.selectorCounter}`);
```
gedu: why using an string interpolation, can we just use numbers as keys? given that the string "selector_" is fix
I benchmark it and using just numbers will be faster

Screenshot 2025-08-05 at 17 41 15

lib/OnyxSnapshotCache.ts
```diff
@@ -0,0 +1,113 @@
...
+
+/**
+ * Manages snapshot caching for useOnyx hook performance optimization.
+ * Handles selector function tracking and memoized getSnapshot results.
+ */
+class OnyxSnapshotCache {
+    /**
+     * Snapshot cache for ultimate performance - separate cache per Onyx key
+     */
+    private snapshotCache: Map<string, Map<string, unknown>>;
```
fabioh8010:
```suggestion
    private snapshotCache: Map<OnyxKey, Map<string, UseOnyxResult<OnyxValue<OnyxKey>>>>;
```

lib/OnyxSnapshotCache.ts
```diff
@@ -0,0 +1,113 @@
...
+        const initWithStoredValues = options?.initWithStoredValues ?? true;
+        const allowStaleData = options?.allowStaleData ?? false;
+        const canBeMissing = options?.canBeMissing ?? true;
+        return `${selectorId}_${initWithStoredValues}_${allowStaleData}_${canBeMissing}`;
+    }
+
+    /**
+     * Get cached snapshot result for a key and cache key combination
+     */
+    getCachedResult<T>(key: string, cacheKey: string): T | undefined {
```
fabioh8010:
```suggestion
    getCachedResult<TResult extends UseOnyxResult<OnyxValue<OnyxKey>>>(key: OnyxKey, cacheKey: string): TResult | undefined {
```

lib/OnyxSnapshotCache.ts
```diff
@@ -0,0 +1,113 @@
...
+        const keyCache = this.snapshotCache.get(key);
+        return keyCache?.get(cacheKey) as T | undefined;
+    }
+
+    /**
+     * Set cached snapshot result for a key and cache key combination
+     */
+    setCachedResult<T>(key: string, cacheKey: string, result: T): void {
```
fabioh8010:
```suggestion
    setCachedResult<TResult extends UseOnyxResult<OnyxValue<OnyxKey>>>(key: OnyxKey, cacheKey: string, result: TResult): void {
```

lib/OnyxSnapshotCache.ts
```diff
@@ -0,0 +1,113 @@
...
+            this.snapshotCache.set(key, new Map());
+        }
+        this.snapshotCache.get(key)!.set(cacheKey, result);
+    }
+
+    /**
+     * Selective cache invalidation to prevent data unavailability
+     * Collection members invalidate upward, collections don't cascade downward
+     */
+    invalidateForKey(keyToInvalidate: string): void {
```
fabioh8010:
```suggestion
    invalidateForKey(keyToInvalidate: OnyxKey): void {
```

lib/useOnyx.ts
```diff
@@ -146,6 +148,15 @@ function useOnyx<TKey extends OnyxKey, TReturnValue = OnyxValue<TKey>>(
...
     const sourceValueRef = useRef<NonNullable<TReturnValue> | undefined>(undefined);
 
+    // Cache the options key to avoid regenerating it every getSnapshot call
+    const cacheKeyRef = useRef<string | null>(null);
+
+    // Generate cache key and update when options change
+    const currentCacheKey = onyxSnapshotCache.generateCacheKey(options);
+    if (cacheKeyRef.current !== currentCacheKey) {
+        cacheKeyRef.current = currentCacheKey;
+    }
```
fabioh8010: Is it ok for this piece of code running every time?

lib/useOnyx.ts
```diff
@@ -190,6 +201,8 @@ function useOnyx<TKey extends OnyxKey, TReturnValue = OnyxValue<TKey>>(
             return;
         }
 
+        // Invalidate cache when dependencies change so selector runs with new closure values
+        onyxSnapshotCache.invalidateForKey(key as string);
```
fabioh8010:
```suggestion
        onyxSnapshotCache.invalidateForKey(key);
```

lib/useOnyx.ts
```diff
@@ -213,11 +226,28 @@ function useOnyx<TKey extends OnyxKey, TReturnValue = OnyxValue<TKey>>(
     }, [key, options?.canEvict]);
 
     const getSnapshot = useCallback(() => {
+        const cacheKey = cacheKeyRef.current!;
+        const keyStr = key as string;
```
fabioh8010:
```suggestion
```

With my proposed changes we don't need this anymore, you can just use `key`.

lib/useOnyx.ts
```diff
@@ -213,11 +226,28 @@ function useOnyx<TKey extends OnyxKey, TReturnValue = OnyxValue<TKey>>(
...
+
+        // Check if we have any cache for this Onyx key
+        // Don't use cache for first connection with initWithStoredValues: false
+        // Also don't use cache during active data updates (when shouldGetCachedValueRef is true)
+        if (!(isFirstConnectionRef.current && options?.initWithStoredValues === false) && !shouldGetCachedValueRef.current) {
+            const cachedResult = onyxSnapshotCache.getCachedResult<UseOnyxResult<TReturnValue>>(keyStr, cacheKey);
```
fabioh8010:
```suggestion
            const cachedResult = onyxSnapshotCache.getCachedResult<UseOnyxResult<TReturnValue>>(key, cacheKey);
```

lib/useOnyx.ts
```diff
@@ -213,11 +226,28 @@ function useOnyx<TKey extends OnyxKey, TReturnValue = OnyxValue<TKey>>(
...
+        }
+
         let isOnyxValueDefined = true;
 
         // We return the initial result right away during the first connection if `initWithStoredValues` is set to `false`.
         if (isFirstConnectionRef.current && options?.initWithStoredValues === false) {
-            return resultRef.current;
+            const result = resultRef.current;
+            // Store result in snapshot cache
+            onyxSnapshotCache.setCachedResult<UseOnyxResult<TReturnValue>>(keyStr, cacheKey, result);
```
fabioh8010:
```suggestion
            onyxSnapshotCache.setCachedResult<UseOnyxResult<TReturnValue>>(key, cacheKey, result);
```

lib/useOnyx.ts
```diff
@@ -292,6 +322,8 @@ function useOnyx<TKey extends OnyxKey, TReturnValue = OnyxValue<TKey>>(
             }
         }
 
+        onyxSnapshotCache.setCachedResult<UseOnyxResult<TReturnValue>>(keyStr, cacheKey, resultRef.current);
```
fabioh8010:
```suggestion
        onyxSnapshotCache.setCachedResult<UseOnyxResult<TReturnValue>>(key, cacheKey, resultRef.current);
```

lib/useOnyx.ts
```diff
@@ -316,6 +348,9 @@ function useOnyx<TKey extends OnyxKey, TReturnValue = OnyxValue<TKey>>(
                     // sourceValue is unknown type, so we need to cast it to the correct type.
                     sourceValueRef.current = sourceValue as NonNullable<TReturnValue>;
 
+                    // Invalidate snapshot cache for this key when data changes
+                    onyxSnapshotCache.invalidateForKey(key as string);
```
fabioh8010:
```suggestion
                    onyxSnapshotCache.invalidateForKey(key);
```

lib/OnyxSnapshotCache.ts
```diff
@@ -0,0 +1,113 @@
...
+            const id = this.selectorIdCounter++;
+            this.selectorIdMap.set(typedSelector, id);
+        }
+        return this.selectorIdMap.get(typedSelector)!;
+    }
+
+    /**
+     * Fast cache key generation for useOnyx options combination
+     */
+    generateCacheKey<TKey extends OnyxKey, TReturnValue>(options?: UseOnyxOptions<TKey, TReturnValue>): string {
```
fabioh8010: Could you expand the description here a little bit, explaining why we are using these specific properties of options? You can use `OnyxConnectionManager.generateConnectionID` as example/inspiration.

lib/useOnyx.ts
```diff
@@ -251,15 +281,15 @@ function useOnyx<TKey extends OnyxKey, TReturnValue = OnyxValue<TKey>>(
...
+        // Optimized equality checking:
+        // - For memoized selectors, use deep equality since reference equality is too strict when cache is involved
         // - Non-selector cases use shallow equality for object reference checks
         // - Normalize null to undefined to ensure consistent comparison (both represent "no value")
         let areValuesEqual: boolean;
         if (memoizedSelector) {
             const normalizedPrevious = previousValueRef.current ?? undefined;
             const normalizedNew = newValueRef.current ?? undefined;
-            areValuesEqual = normalizedPrevious === normalizedNew;
+            areValuesEqual = deepEqual(normalizedPrevious, normalizedNew);
```
fabioh8010: Hmm not sure if I get this part, aren't we degrading performance here again? 😕

fabioh8010 (commented):
First round of reviews.

@TMisiukiewicz Could you also implement Reassure tests for `OnyxSnapshotCache`? They will be mostly for coverage reasons, you can use `tests/perf-test/OnyxConnectionManager.perf-test.ts` as example/reference.

lib/useOnyx.ts
TMisiukiewicz: yeah I think i over-vibecoded it 😆 Let me validate it

chuckdries: > reference equality is too strict when cache is involved

Can you expand on this comment? Sounds like a serious issue with our caching system. [React docs explicitly state](https://react.dev/reference/react/useSyncExternalStore#caveats) that useSyncExternalStore should return immutable snapshots. I strongly hesitate to introduce a deep equality check for busting memoization. If that is truly necessary, we need to change our caches to preserve referential equality.

TMisiukiewicz: Yeah, I can confirm this looks like over-optimization here. We already handle the deep equality check on selectors internally to avoid re-running expensive selectors, so if we do reference comparison, the worst that can happen is a re-render when the selector returns a new object reference. But the selector itself won't re execute thanks to internal memoization, so we end up with cheap re-renders using fast comparison with already-cached data.

Thanks for pointing this out - we already got rid of `deepEqual` when we introduced memoized selectors, so it wouldn't make sense to reintroduce it just to avoid these cheap re-renders

lib/useOnyx.ts
```diff
@@ -149,13 +149,7 @@ function useOnyx<TKey extends OnyxKey, TReturnValue = OnyxValue<TKey>>(
...
 
     // Cache the options key to avoid regenerating it every getSnapshot call
-    const cacheKeyRef = useRef<string | null>(null);
-
-    // Generate cache key and update when options change
-    const currentCacheKey = onyxSnapshotCache.generateCacheKey(options);
-    if (cacheKeyRef.current !== currentCacheKey) {
-        cacheKeyRef.current = currentCacheKey;
-    }
+    const cacheKey = useMemo(() => onyxSnapshotCache.generateCacheKey(options), [options]);
```
fabioh8010:
```suggestion
    const cacheKey = useMemo(
        () =>
            onyxSnapshotCache.generateCacheKey({
                selector: options?.selector,
                initWithStoredValues: options?.initWithStoredValues,
                allowStaleData: options?.allowStaleData,
                canBeMissing: options?.canBeMissing,
            }),
        [options?.selector, options?.initWithStoredValues, options?.allowStaleData, options?.canBeMissing],
    );
```

I think the cache key will be re-created every time if you just pass `options` to the dependencies array, because of the way we usually pass it when using `useOnyx` e.g:

```ts
const [data] = useOnyx(ONYXKEYS.DATA, {initWithStoredValues: false}); // new options object reference being passed every time.
```

To solve this, we can select only the options we really need in the dependencies array and pass them to `generateCacheKey`. WDYT?

lib/OnyxSnapshotCache.ts
```diff
@@ -41,7 +41,18 @@ class OnyxSnapshotCache {
...
+     *
+     * - `selector`: Different selectors produce different results, so each selector needs its own cache entry
+     * - `initWithStoredValues`: This flag changes the initial loading behavior and affects the returned fetch status
+     * - `allowStaleData`: Controls whether stale data can be returned during pending merges, affecting result timing
+     * - `canBeMissing`: Determines logging behavior for missing data, but doesn't affect the actual data returned
+     *
+     * Other options like `canEvict`, `reuseConnection`, and `allowDynamicKey` don't affect the data transformation
+     * or timing behavior of getSnapshot, so they're excluded from the cache key for better cache hit rates.
      */
     generateCacheKey<TKey extends OnyxKey, TReturnValue>(options?: UseOnyxOptions<TKey, TReturnValue>): string {
```
fabioh8010:
```suggestion
    generateCacheKey<TKey extends OnyxKey, TReturnValue>(options: Pick<UseOnyxOptions<TKey, TReturnValue>, 'selector' | 'initWithStoredValues' | 'allowStaleData' | 'canBeMissing'>): string {
```

Based on my comment above, we need to adjust this a little bit.

tests/unit/OnyxSnapshotCacheTest.ts
```diff
@@ -17,10 +17,7 @@ type TestData = {
     name?: string;
 };
 
-type TestResult = {
-    data: string;
-    status?: string;
-};
+type TestResult = UseOnyxResult<{data: string}>;
 
 type TestSelector = (data: unknown) => string;
```
fabioh8010:
```suggestion
type TestSelector = UseOnyxSelector<OnyxKey, string>;
```

tests/unit/OnyxSnapshotCacheTest.ts
```diff
@@ -0,0 +1,241 @@
...
+    let cache: OnyxSnapshotCache;
+
+    beforeEach(() => {
+        cache = new OnyxSnapshotCache();
+        jest.clearAllMocks();
+    });
+
+    describe('basic cache operations', () => {
+        it('should generate unique cache keys for different options', () => {
+            const selector: TestSelector = (data: unknown) => {
```
fabioh8010:
```suggestion
            const selector: TestSelector = (data) => {
```

tests/unit/OnyxSnapshotCacheTest.ts
```diff
@@ -0,0 +1,241 @@
...
+                selector,
+                initWithStoredValues: true,
+                allowStaleData: false,
+                canBeMissing: true,
+            };
+            const optionsWithoutSelector: UseOnyxOptions<string, string> = {
+                initWithStoredValues: false,
+                allowStaleData: true,
+                canBeMissing: false,
+            };
```
fabioh8010:
```suggestion
            const optionsWithSelector: UseOnyxOptions<OnyxKey, string> = {
                selector,
                initWithStoredValues: true,
                allowStaleData: false,
                canBeMissing: true,
            };
            const optionsWithoutSelector: UseOnyxOptions<OnyxKey, string> = {
                initWithStoredValues: false,
                allowStaleData: true,
                canBeMissing: false,
            };
```

tests/unit/OnyxSnapshotCacheTest.ts
```diff
@@ -0,0 +1,241 @@
...
+            const undefinedOptions: UseOnyxOptions<string, string> | undefined = undefined;
+
+            const keyWithSelector = cache.generateCacheKey(optionsWithSelector);
+            const keyWithoutSelector = cache.generateCacheKey(optionsWithoutSelector);
+            const keyWithUndefined = cache.generateCacheKey(undefinedOptions);
```
fabioh8010:
```suggestion
            const keyWithSelector = cache.generateCacheKey(optionsWithSelector);
            const keyWithoutSelector = cache.generateCacheKey(optionsWithoutSelector);
            const keyWithUndefined = cache.generateCacheKey({});
```

With the change I suggested to `generateCacheKey`, we now need to at least pass a empty object

tests/unit/OnyxSnapshotCacheTest.ts
```diff
@@ -0,0 +1,241 @@
...
+        });
+    });
+
+    describe('selector ID management', () => {
+        it('should generate unique IDs for different selectors', () => {
+            const nameSelector: TestSelector = (data: unknown) => {
+                const testData = data as TestData | undefined;
+                return testData?.name ?? '';
+            };
+            const idSelector: TestSelector = (data: unknown) => {
```
fabioh8010:
```suggestion
            const nameSelector: TestSelector = (data) => {
                const testData = data as TestData | undefined;
                return testData?.name ?? '';
            };
            const idSelector: TestSelector = (data) => {
```

tests/unit/OnyxSnapshotCacheTest.ts
```diff
@@ -0,0 +1,241 @@
...
+
+            const nameId = cache.getSelectorId(nameSelector);
+            const idSelectorId = cache.getSelectorId(idSelector);
+
+            // Different selectors should get different IDs
+            expect(nameId).not.toBe(idSelectorId);
+        });
+
+        it('should return the same ID for the same selector function', () => {
+            const selector: TestSelector = (data: unknown) => {
```
fabioh8010:
```suggestion
            const selector: TestSelector = (data) => {
```

tests/unit/OnyxSnapshotCacheTest.ts
```diff
@@ -0,0 +1,241 @@
...
+            expect(firstCall).toBe(secondCall);
+            expect(secondCall).toBe(thirdCall);
+        });
+
+        it('should clear selector IDs and reset counter', () => {
+            const selector1: TestSelector = (data: unknown) => {
+                const testData = data as TestData | undefined;
+                return testData?.name ?? '';
+            };
+            const selector2: TestSelector = (data: unknown) => {
```
fabioh8010:
```suggestion
            const selector1: TestSelector = (data) => {
                const testData = data as TestData | undefined;
                return testData?.name ?? '';
            };
            const selector2: TestSelector = (data) => {
```

tests/perf-test/OnyxSnapshotCache.perf-test.ts
```diff
@@ -0,0 +1,266 @@
...
+    TEST_KEY_2: 'test2',
+    COLLECTION: {
+        TEST_KEY: 'test_',
+        TEST_KEY_2: 'test2_',
+        REPORTS: 'reports_',
+    },
+};
+
+// Mock selector functions
+const simpleSelector = (data: unknown) => (data as MockData | undefined)?.value;
```
fabioh8010:
```suggestion
const simpleSelector: UseOnyxSelector<OnyxKey, number | undefined> = (data) => (data as MockData | undefined)?.value;
```

tests/perf-test/OnyxSnapshotCache.perf-test.ts
```diff
@@ -0,0 +1,266 @@
...
+
+const complexSelector = (data: unknown) => {
+    const mockData = data as MockData | undefined;
+    return {
+        id: mockData?.id,
+        name: mockData?.name,
+        computed: mockData?.value ? mockData.value * 2 : 0,
+        formatted: `${mockData?.name}: ${mockData?.value}`,
+    };
+};
```
fabioh8010:
```suggestion
type ComplexSelectorResult = {id?: number; name?: string; computed: number; formatted: string};
const complexSelector: UseOnyxSelector<OnyxKey, ComplexSelectorResult> = (data) => {
    const mockData = data as MockData | undefined;
    return {
        id: mockData?.id,
        name: mockData?.name,
        computed: mockData?.value ? mockData.value * 2 : 0,
        formatted: `${mockData?.name}: ${mockData?.value}`,
    };
};
```

tests/perf-test/OnyxSnapshotCache.perf-test.ts
```diff
@@ -0,0 +1,266 @@
...
+
+const selectorOptions: UseOnyxOptions<string, number | undefined> = {
+    selector: simpleSelector,
+    initWithStoredValues: true,
+    allowStaleData: false,
+    canBeMissing: true,
+};
+
+const complexSelectorOptions: UseOnyxOptions<string, {id?: number; name?: string; computed: number; formatted: string}> = {
```
fabioh8010:
```suggestion
const complexSelectorOptions: UseOnyxOptions<string, ComplexSelectorResult> = {
```

tests/perf-test/OnyxSnapshotCache.perf-test.ts
```diff
@@ -0,0 +1,266 @@
...
+    selector: complexSelector,
+    initWithStoredValues: true,
+    allowStaleData: false,
+    canBeMissing: true,
+};
+
+// Mock results
+const mockResult: UseOnyxResult<MockData> = [
+    {id: 1, name: 'Test', value: 42},
+    {status: 'loaded' as const, sourceValue: {id: 1, name: 'Test', value: 42}},
```
fabioh8010:
```suggestion
    {status: 'loaded', sourceValue: {id: 1, name: 'Test', value: 42}},
```

tests/perf-test/OnyxSnapshotCache.perf-test.ts
```diff
@@ -0,0 +1,266 @@
...
+];
+
+const mockResults = Array.from(
+    {length: 1000},
+    (_, i) =>
+        [
+            {id: i, name: `Test${i}`, value: i * 10},
+            {status: 'loaded' as const, sourceValue: {id: i, name: `Test${i}`, value: i * 10}},
+        ] as UseOnyxResult<MockData>,
+);
```
fabioh8010:
```suggestion
const mockResults = Array.from(
    {length: 1000},
    (_, i): UseOnyxResult<MockData> => [
        {id: i, name: `Test${i}`, value: i * 10},
        {status: 'loaded', sourceValue: {id: i, name: `Test${i}`, value: i * 10}},
    ],
);
```

tests/perf-test/OnyxSnapshotCache.perf-test.ts
```diff
@@ -0,0 +1,266 @@
...
+            await measureFunction(
+                () => {
+                    cache.getSelectorId(simpleSelector);
+                },
+                {
+                    beforeEach: () => {
+                        resetCacheBeforeEachMeasure();
+                        // Pre-populate with 1000 selectors
+                        for (let i = 0; i < 1000; i++) {
+                            const selector = (data: unknown) => ((data as MockData | undefined)?.field ?? '') + i;
```
fabioh8010:
```suggestion
                            const selector: UseOnyxSelector<OnyxKey, string> = (data) => ((data as MockData | undefined)?.field ?? '') + i;
```

tests/perf-test/OnyxSnapshotCache.perf-test.ts
```diff
@@ -0,0 +1,266 @@
...
+                    beforeEach: resetCacheBeforeEachMeasure,
+                },
+            );
+        });
+
+        test('generating 1000 cache keys with different selectors', async () => {
+            await measureFunction(
+                () => {
+                    for (let i = 0; i < 1000; i++) {
+                        const selector = (data: unknown) => ((data as MockData | undefined)?.field ?? '') + i;
```
fabioh8010:
```suggestion
                        const selector: UseOnyxSelector<OnyxKey, string> = (data) => ((data as MockData | undefined)?.field ?? '') + i;
```

tests/perf-test/OnyxSnapshotCache.perf-test.ts
```diff
@@ -0,0 +1,266 @@
...
+                },
+            );
+        });
+
+        test('getting cached result with complex selector (cache hit)', async () => {
+            const cacheKey = cache.generateCacheKey(complexSelectorOptions);
+            const complexResult: UseOnyxResult<{id?: number; name?: string; computed: number; formatted: string}> = [
+                {id: 1, name: 'Test', computed: 84, formatted: 'Test: 42'},
+                {status: 'loaded' as const, sourceValue: {id: 1, name: 'Test', computed: 84, formatted: 'Test: 42'}},
+            ];
```
fabioh8010:
```suggestion
            const complexResult: UseOnyxResult<ComplexSelectorResult> = [
                {id: 1, name: 'Test', computed: 84, formatted: 'Test: 42'},
                {status: 'loaded', sourceValue: {id: 1, name: 'Test', computed: 84, formatted: 'Test: 42'}},
            ];
```

lib/useOnyx.ts
TMisiukiewicz: yeah good point, thank you 🙏

lib/OnyxConnectionManager.ts
```diff
@@ -262,13 +263,17 @@ class OnyxConnectionManager {
         });
 
         this.connectionsMap.clear();
+        // Clear snapshot cache when all connections are disconnected
```
mountiny, line 267: NAB: Leave empty line before comments for better readability
```suggestion

        // Clear snapshot cache when all connections are disconnected
```

lib/OnyxConnectionManager.ts
```diff
@@ -262,13 +263,17 @@ class OnyxConnectionManager {
...
+        onyxSnapshotCache.clear();
     }
 
     /**
      * Refreshes the connection manager's session ID.
      */
     refreshSessionID(): void {
         this.sessionID = Str.guid();
+        // Clear snapshot cache when session refreshes to avoid stale cache issues
```
mountiny, line 277: NAB: Leave empty line before comments for better readability
```suggestion

        // Clear snapshot cache when session refreshes to avoid stale cache issues
```

lib/OnyxSnapshotCache.ts
```diff
@@ -0,0 +1,124 @@
...
+    constructor() {
+        this.snapshotCache = new Map();
+        this.selectorIdMap = new Map();
+        this.selectorIdCounter = 0;
+    }
+
+    /**
+     * Generate unique ID for selector functions using incrementing numbers
+     */
+    getSelectorId<TKey extends OnyxKey, TReturnValue>(selector: UseOnyxSelector<TKey, TReturnValue>): number {
```
mountiny: I believe our style guides say the naming should have uppercase shortcuts in the variable names
```suggestion
    getSelectorID<TKey extends OnyxKey, TReturnValue>(selector: UseOnyxSelector<TKey, TReturnValue>): number {
```

lib/OnyxSnapshotCache.ts
```diff
@@ -0,0 +1,124 @@
...
+    private selectorIdMap: Map<UseOnyxSelector<OnyxKey, unknown>, number>;
+
+    /**
+     * Counter for generating incremental selector IDs
+     */
+    private selectorIdCounter: number;
+
+    constructor() {
+        this.snapshotCache = new Map();
+        this.selectorIdMap = new Map();
```
mountiny:
```suggestion
        this.selectorIDMap = new Map();
```

lib/OnyxSnapshotCache.ts
```diff
@@ -0,0 +1,124 @@
...
+        this.selectorIdCounter = 0;
```
mountiny:
```suggestion
        this.selectorIDCounter = 0;
```

lib/OnyxSnapshotCache.ts
```diff
@@ -0,0 +1,124 @@
...
+     * - `initWithStoredValues`: This flag changes the initial loading behavior and affects the returned fetch status
+     * - `allowStaleData`: Controls whether stale data can be returned during pending merges, affecting result timing
+     * - `canBeMissing`: Determines logging behavior for missing data, but doesn't affect the actual data returned
+     *
+     * Other options like `canEvict`, `reuseConnection`, and `allowDynamicKey` don't affect the data transformation
+     * or timing behavior of getSnapshot, so they're excluded from the cache key for better cache hit rates.
+     */
+    generateCacheKey<TKey extends OnyxKey, TReturnValue>(options: Pick<UseOnyxOptions<TKey, TReturnValue>, 'selector' | 'initWithStoredValues' | 'allowStaleData' | 'canBeMissing'>): string {
+        const selectorId = options?.selector ? this.getSelectorId(options.selector) : 'no_selector';
+        // Create options hash without expensive JSON.stringify
```
mountiny, line 71:
```suggestion

        // Create options hash without expensive JSON.stringify
```

lib/useOnyx.ts
```diff
@@ -213,11 +229,25 @@ function useOnyx<TKey extends OnyxKey, TReturnValue = OnyxValue<TKey>>(
...
+            }
+        }
+
         let isOnyxValueDefined = true;
 
         // We return the initial result right away during the first connection if `initWithStoredValues` is set to `false`.
         if (isFirstConnectionRef.current && options?.initWithStoredValues === false) {
-            return resultRef.current;
+            const result = resultRef.current;
+            // Store result in snapshot cache
```
mountiny, line 250:
```suggestion

            // Store result in snapshot cache
```

TMisiukiewicz:
@chuckdries can I ask you for re-approval, it was dismissed by the commit that was just changing a couple of NABs from Vit 🙄

mountiny (approved):
@roryabraham Would you be able to review please?

lib/OnyxSnapshotCache.ts
```diff
@@ -0,0 +1,125 @@
...
+class OnyxSnapshotCache {
+    /**
+     * Snapshot cache for ultimate performance - separate cache per Onyx key
+     */
+    private snapshotCache: Map<OnyxKey, Map<string, UseOnyxResult<OnyxValue<OnyxKey>>>>;
+
+    /**
+     * Maps selector functions to unique IDs for cache key generation
+     */
+    private selectorIDMap: Map<UseOnyxSelector<OnyxKey, unknown>, number>;
```
roryabraham: I think a `WeakMap` might be more appropriate here - that way when a selector is no longer used anywhere and is garbage-collected, its entry in the `selectorIDMap` will be garbage-collected as well:

```suggestion
    private selectorIDMap: WeakMap<UseOnyxSelector<OnyxKey, unknown>, number>;
```

lib/OnyxSnapshotCache.ts
```diff
@@ -0,0 +1,125 @@
...
+
+/**
+ * Manages snapshot caching for useOnyx hook performance optimization.
+ * Handles selector function tracking and memoized getSnapshot results.
+ */
+class OnyxSnapshotCache {
+    /**
+     * Snapshot cache for ultimate performance - separate cache per Onyx key
+     */
+    private snapshotCache: Map<OnyxKey, Map<string, UseOnyxResult<OnyxValue<OnyxKey>>>>;
```
roryabraham, line 16: Why do you have a separate snapshot cache per key? It's meaningfully contributing to complexity - is it meaningfully improving performance?

I think this would be simpler and more maintainable with a 1-level cache, most likely without a meaningful performance cost.

lib/OnyxSnapshotCache.ts
```diff
@@ -0,0 +1,125 @@
...
+    constructor() {
+        this.snapshotCache = new Map();
+        this.selectorIDMap = new Map();
+        this.selectorIDCounter = 0;
+    }
+
+    /**
+     * Generate unique ID for selector functions using incrementing numbers
+     */
+    getSelectorID<TKey extends OnyxKey, TReturnValue>(selector: UseOnyxSelector<TKey, TReturnValue>): number {
```
roryabraham, line 44:
```suggestion
    private getSelectorID<TKey extends OnyxKey, TReturnValue>(selector: UseOnyxSelector<TKey, TReturnValue>): number {
```

lib/OnyxSnapshotCache.ts
roryabraham: This would also mean that you can get rid of `clearSelectorMap`

lib/OnyxSnapshotCache.ts
roryabraham: Hmmmm, okay actually I see why you did this. I think we should add more context to the comments to explain why we have a two-level cache (rather than the "ultimate performance" comment 😂). Something like this:

```ts
/**
 * Snapshot cache is a two-level map. The top-level keys are Onyx keys. The top-level values maps.
 * The second-level keys are a custom composite string defined by this.generateCacheKey. These represent a unique useOnyx config, which is not fully represented by the Onyx key alone.
 * The reason we have two levels is for performance: not to make cache access faster, but to make cache invalidation faster.
 * We can invalidate the snapshot cache for a given Onyx key with one map.delete operation on the top-level map, rather than having to loop through a large single-level map and delete any matching keys.
 */
```

roryabraham: However, I do see a potential optimization opportunity here.

The top-level map is great for cache invalidation when keys change. However, right now there's nothing to cleanup the second-level map when data in there is no longer needed anywhere.

We should consider using a `WeakMap` with a `Symbol` wrapper around the string keys, such that entries in the second-level map can be garbage-collected when there's no longer any component holding a reference to the same `useOnyx` config:

```diff
diff --git a/lib/OnyxSnapshotCache.ts b/lib/OnyxSnapshotCache.ts
index 7918acf..fb47748 100644
--- a/lib/OnyxSnapshotCache.ts
+++ b/lib/OnyxSnapshotCache.ts
@@ -8,14 +8,24 @@ import type {UseOnyxOptions, UseOnyxResult, UseOnyxSelector} from './useOnyx';
  */
 class OnyxSnapshotCache {
     /**
-     * Snapshot cache for ultimate performance - separate cache per Onyx key
+     * Snapshot cache is a two-level map. The top-level keys are Onyx keys. The top-level values are WeakMaps.
+     * The second-level keys are Symbols that represent unique useOnyx configs. Using WeakMap with Symbols
+     * provides automatic garbage collection when a config is no longer used by any component.
+     * The reason we have two levels is for performance: not to make cache access faster, but to make cache invalidation faster.
+     * We can invalidate the snapshot cache for a given Onyx key with one map.delete operation on the top-level map, rather than having to loop through a large single-level map and delete any matching keys.
      */
-    private snapshotCache: Map<OnyxKey, Map<string, UseOnyxResult<OnyxValue<OnyxKey>>>>;
+    private snapshotCache: Map<OnyxKey, WeakMap<symbol, UseOnyxResult<OnyxValue<OnyxKey>>>>;
+
+    /**
+     * Registry to share Symbols across components with the same cache key combination.
+     * Maps Onyx key → cache key string → WeakRef<symbol>
+     */
+    private symbolRegistry: Map<OnyxKey, Map<string, WeakRef<symbol>>>;
 
     /**
      * Maps selector functions to unique IDs for cache key generation
      */
-    private selectorIDMap: Map<UseOnyxSelector<OnyxKey, unknown>, number>;
+    private selectorIDMap: WeakMap<UseOnyxSelector<OnyxKey, unknown>, number>;
 
     /**
      * Counter for generating incremental selector IDs
@@ -24,7 +34,8 @@ class OnyxSnapshotCache {
 
     constructor() {
         this.snapshotCache = new Map();
-        this.selectorIDMap = new Map();
+        this.symbolRegistry = new Map();
+        this.selectorIDMap = new WeakMap();
         this.selectorIDCounter = 0;
     }
 
@@ -64,22 +75,47 @@ class OnyxSnapshotCache {
         return `${selectorId}_${initWithStoredValues}_${allowStaleData}_${canBeMissing}`;
     }
 
+    /**
+     * Get or create a Symbol for a cache key combination.
+     * Shares Symbols across components with the same key + cacheKey combination.
+     */
+    private getSymbolForCacheKey(key: OnyxKey, cacheKey: string): symbol {
+        if (!this.symbolRegistry.has(key)) {
+            this.symbolRegistry.set(key, new Map());
+        }
+
+        const keyRegistry = this.symbolRegistry.get(key)!;
+        const existingRef = keyRegistry.get(cacheKey);
+        const existingSymbol = existingRef?.deref();
+
+        if (existingSymbol) {
+            return existingSymbol;
+        }
+
+        // Create new Symbol for this key + cacheKey combination
+        const newSymbol = Symbol(`${key}_${cacheKey}`);
+        keyRegistry.set(cacheKey, new WeakRef(newSymbol));
+        return newSymbol;
+    }
+
     /**
      * Get cached snapshot result for a key and cache key combination
      */
     getCachedResult<TResult extends UseOnyxResult<OnyxValue<OnyxKey>>>(key: OnyxKey, cacheKey: string): TResult | undefined {
+        const symbol = this.getSymbolForCacheKey(key, cacheKey);
         const keyCache = this.snapshotCache.get(key);
-        return keyCache?.get(cacheKey) as TResult | undefined;
+        return keyCache?.get(symbol) as TResult | undefined;
     }
 
     /**
      * Set cached snapshot result for a key and cache key combination
      */
     setCachedResult<TResult extends UseOnyxResult<OnyxValue<OnyxKey>>>(key: OnyxKey, cacheKey: string, result: TResult): void {
+        const symbol = this.getSymbolForCacheKey(key, cacheKey);
         if (!this.snapshotCache.has(key)) {
-            this.snapshotCache.set(key, new Map());
+            this.snapshotCache.set(key, new WeakMap());
         }
-        this.snapshotCache.get(key)!.set(cacheKey, result);
+        this.snapshotCache.get(key)!.set(symbol, result);
     }
 
     /**
@@ -89,6 +125,7 @@ class OnyxSnapshotCache {
     invalidateForKey(keyToInvalidate: OnyxKey): void {
         // Always invalidate the exact key
         this.snapshotCache.delete(keyToInvalidate);
+        this.symbolRegistry.delete(keyToInvalidate);
 
         // Handle collection-related keys with selective invalidation
         if (OnyxUtils.isCollectionKey(keyToInvalidate)) {
@@ -98,6 +135,7 @@ class OnyxSnapshotCache {
             // Only invalidate upward to the collection base key
             if (baseKey !== keyToInvalidate) {
                 this.snapshotCache.delete(baseKey);
+                this.symbolRegistry.delete(baseKey);
             }
         }
     }
@@ -107,14 +145,7 @@ class OnyxSnapshotCache {
      */
     clear(): void {
         this.snapshotCache.clear();
-    }
-
-    /**
-     * Clear selector ID mappings (useful for testing)
-     */
-    clearSelectorIds(): void {
-        this.selectorIDMap.clear();
-        this.selectorIDCounter = 0;
+        this.symbolRegistry.clear();
     }
 }
```

This would minimize the memory footprint of `OnyxSnapshotCache` and help keep it lightweight. ~~but has a trade-off where if a snapshot ref is garbage-collected in the cache because nothing's holding a reference to it anymore, and then a component mounts that needs that, they end up recomputing the snapshot. So it's a memory-vs-time tradeoff.~~

*Edit* There's actually no tradeoff there without centralized/reused selectors. Because if a component remounts it will create a new selector, which will have a new selectorID, which will result in a cache recomputation anyways. So I think we _should_ implement this optimization or some version thereof.

lib/OnyxSnapshotCache.ts
```diff
@@ -0,0 +1,125 @@
...
+            this.snapshotCache.set(key, new Map());
+        }
+        this.snapshotCache.get(key)!.set(cacheKey, result);
+    }
+
+    /**
+     * Selective cache invalidation to prevent data unavailability
+     * Collection members invalidate upward, collections don't cascade downward
+     */
+    invalidateForKey(keyToInvalidate: OnyxKey): void {
```
roryabraham, line 132: NAB I found this implementation a bit confusing. I personally think this is a touch clearer:

```diff
diff --git a/lib/OnyxSnapshotCache.ts b/lib/OnyxSnapshotCache.ts
index 7918acf..16acb2d 100644
--- a/lib/OnyxSnapshotCache.ts
+++ b/lib/OnyxSnapshotCache.ts
@@ -90,15 +90,12 @@ class OnyxSnapshotCache {
         // Always invalidate the exact key
         this.snapshotCache.delete(keyToInvalidate);
 
-        // Handle collection-related keys with selective invalidation
-        if (OnyxUtils.isCollectionKey(keyToInvalidate)) {
-            const baseKey = OnyxUtils.getCollectionKey(keyToInvalidate);
-
-            // If this is a collection member key (e.g., "reports_123")
-            // Only invalidate upward to the collection base key
-            if (baseKey !== keyToInvalidate) {
-                this.snapshotCache.delete(baseKey);
-            }
+        try {
+            // Check if the key is a collection member and invalidate the collection base key
+            const collectionBaseKey = OnyxUtils.getCollectionKey(keyToInvalidate);
+            this.snapshotCache.delete(collectionBaseKey);
+        } catch (e) {
+            // do nothing - this just means the key is not a collection member
         }
     }
```

lib/OnyxSnapshotCache.ts
TMisiukiewicz: > (rather than the "ultimate performance" comment 😂)

LOL I completely missed that comment before, AI autocomplete really went wild on this one 😆 

Yeah I think the proposed diff makes sense 👏 However when I applied it, the app broke on iOS. I was able to narrow down the issue to `WeakRef` usage 🤔 When I tried using `WeakRef` directly in the E/App it worked, might be some kind of an issue directly in Onyx? Do you have any idea why it happens?

image

TMisiukiewicz:
> Also, can we please make sure that useOnyx invocations that don’t provide a selector (i.e: selector is undefined) don’t produce separate selectorID, which would in turn lead to duplicate cache keys?

this one is covered, if there is no selector provided, it adds `no_selector` to the snapshot key

lib/OnyxSnapshotCache.ts
TMisiukiewicz: Unit and perf tests are using this method, so unfortunately I couldn't make it private 😞

fabioh8010: @TMisiukiewicz you can do this trick -> https://github.com/Expensify/react-native-onyx/blob/27f8a11ac35ebb2e9b715bf5edc24f4769142d72/tests/unit/OnyxConnectionManagerTest.ts#L13-L20

lib/useOnyx.ts
```diff
@@ -98,9 +99,10 @@ function useOnyx<TKey extends OnyxKey, TReturnValue = OnyxValue<TKey>>(
                 // Only proceed if we have a valid selector
                 if (currentSelector) {
                     const newOutput = currentSelector(input);
+                    const areEqual = deepEqual(lastOutput, newOutput);
```
roryabraham: I'm not sure this local `areEqual` variable is helpful. In fact, it might be hurting readability and, to a minor degree performance. With the `deepEqual` call inlined where it was before, the comparison might be skipped if `hasComputed` is true.

lib/OnyxSnapshotCache.ts
```diff
@@ -0,0 +1,124 @@
...
+     * - `selector`: Different selectors produce different results, so each selector needs its own cache entry
+     * - `initWithStoredValues`: This flag changes the initial loading behavior and affects the returned fetch status
+     * - `allowStaleData`: Controls whether stale data can be returned during pending merges, affecting result timing
+     * - `canBeMissing`: Determines logging behavior for missing data, but doesn't affect the actual data returned
+     *
+     * Other options like `canEvict`, `reuseConnection`, and `allowDynamicKey` don't affect the data transformation
+     * or timing behavior of getSnapshot, so they're excluded from the cache key for better cache hit rates.
+     */
+    generateCacheKey<TKey extends OnyxKey, TReturnValue>(options: Pick<UseOnyxOptions<TKey, TReturnValue>, 'selector' | 'initWithStoredValues' | 'allowStaleData' | 'canBeMissing'>): string {
+        const selectorId = options?.selector ? this.getSelectorID(options.selector) : 'no_selector';
```
roryabraham:
```suggestion
        const selectorID = options?.selector ? this.getSelectorID(options.selector) : 'no_selector';
```

lib/OnyxSnapshotCache.ts
roryabraham: oh, no worries then I don't know that it's necessary. `OnyxSnapshotCache` is essentially private to the Onyx package anyways.

lib/OnyxSnapshotCache.ts
roryabraham: hmmmm so I think that the idea with using a WeakMap for the second-level map doesn't work as well as I hoped because the symbol needs to be unregistered (i.e: global symbol registry with `Symbol.for` and `Symbol.keyFor` doesn't work).

PR #684, opened 2025-09-29
Add IDBKeyValProvider with createStore function for manual store crea…

blazejkustra:


fabioh8010:
Hold for my review

lib/storage/providers/IDBKeyValProvider/createStore.ts
```diff
@@ -0,0 +1,60 @@
+import {promisifyRequest} from 'idb-keyval';
+import type {UseStore} from 'idb-keyval';
+import {logInfo} from '../../../Logger';
+
+// This is a copy of the createStore function from idb-keyval, but we need to use it to create the store manually.
+// We do this because we need to create the database manually in order to ensure that the store exists before we use it.
+// If the store does not exist, idb-keyval will throw an error
+// source: https://github.com/jakearchibald/idb-keyval/blob/9d19315b4a83897df1e0193dccdc29f78466a0f3/src/index.ts#L12
+function createStore(dbName: string, storeName: string): UseStore {
+    let database: Promise<IDBDatabase> | undefined;
```
fabioh8010:
```suggestion
    let dbp: Promise<IDBDatabase> | undefined;

```

Just to match better with original source code and make diff simpler to compare.

lib/storage/providers/IDBKeyValProvider/createStore.ts
```diff
@@ -0,0 +1,60 @@
...
+    const getDB = () => {
+        if (database) return database;
+        const request = indexedDB.open(dbName);
+        request.onupgradeneeded = () => request.result.createObjectStore(storeName);
+        database = promisifyRequest(request);
+        database.then(
```
fabioh8010:
```suggestion

        database.then(
```

Same

lib/storage/providers/IDBKeyValProvider/createStore.ts
```diff
@@ -0,0 +1,60 @@
...
+                // It seems like Safari sometimes likes to just close the connection.
+                // It's supposed to fire this event when that happens. Let's hope it does!
+                // eslint-disable-next-line no-param-reassign
+                db.onclose = () => (database = undefined);
+            },
+            // eslint-disable-next-line @typescript-eslint/no-empty-function
+            () => {},
+        );
+
+        return database;
```
fabioh8010:
```suggestion
        return database;
```

Same

lib/storage/providers/IDBKeyValProvider/createStore.ts
```diff
@@ -0,0 +1,60 @@
...
+    };
+
+    const fixStore = (db: IDBDatabase) => {
```
fabioh8010: Could you add a comment here explaining what this function is doing exactly?

lib/storage/providers/IDBKeyValProvider/createStore.ts
```diff
@@ -0,0 +1,60 @@
...
+        if (db.objectStoreNames.contains(storeName)) {
+            return db;
+        }
+
+        logInfo(`Store ${storeName} does not exist in database ${dbName}.`);
+        const nextVersion = db.version + 1;
```
fabioh8010, line 38: Why do we need to do this?

lib/storage/providers/IDBKeyValProvider/createStore.ts
```diff
@@ -0,0 +1,60 @@
+import {promisifyRequest} from 'idb-keyval';
+import type {UseStore} from 'idb-keyval';
+import {logInfo} from '../../../Logger';
+
+// This is a copy of the createStore function from idb-keyval, but we need to use it to create the store manually.
+// We do this because we need to create the database manually in order to ensure that the store exists before we use it.
+// If the store does not exist, idb-keyval will throw an error
+// source: https://github.com/jakearchibald/idb-keyval/blob/9d19315b4a83897df1e0193dccdc29f78466a0f3/src/index.ts#L12
```
fabioh8010:
```suggestion
// This is a copy of the createStore function from idb-keyval, we need a custom implementation
// because we need to create the database manually in order to ensure that the store exists before we use it.
// If the store does not exist, idb-keyval will throw an error
// source: https://github.com/jakearchibald/idb-keyval/blob/9d19315b4a83897df1e0193dccdc29f78466a0f3/src/index.ts#L12
```

lib/storage/providers/IDBKeyValProvider/createStore.ts
blazejkustra: IndexedDB only allows creating object stores during a version upgrade. If the store is missing (e.g., first run, cleared storage, or a previous version didn’t create it)

fabioh8010 (approved):

jnowakow (approved):
Looks good for me 🚀

staszekscp (approved):

PR #787, opened 2026-05-19
Make mergeCollectionWithPatches cache-first

elirangoshen:
### Details

`mergeCollectionWithPatches` was the only Onyx write method that did **not** follow the cache-first / storage-second invariant. Every other write method (`setWithRetry`, `applyMerge`, `setCollectionWithRetry`, `partialSetCollection`, `clear`) updates the in-memory cache synchronously before issuing a storage write — so when storage fails (e.g., the IDB corruption population analyzed in Expensify/App#87862), the cache is already correct and subscribers still see the right data.

`mergeCollectionWithPatches` instead pushed `Storage.multiMerge` / `Storage.multiSet` first and bundled the `cache.merge` + `keysChanged` update into a `previousCollectionPromise.then(...)` chain that ran in parallel with storage. In the warm-cache common case the cache update happened to win the race, but if `previousCollectionPromise` lost the race (cache miss falling back to a slow storage read), subscribers could observe a state inconsistent with what the merge logically applied.

This change restores the invariant:

1. `Promise.all(existingKeys.map((key) => get(key)))` runs first to pre-warm the cache from storage for any cache-miss existing keys. `get()` is a sync-resolved no-op for cache hits, so this is microtask-cheap for the warm-cache common case. The pre-warm is required because `cache.merge`'s `fastMerge` needs the previous storage value as the merge base — without it, cold-cache merges would start from `undefined` and silently drop existing data (the behavior the old `previousCollectionPromise` happened to provide as a side effect via `get()`'s cache-population path).
2. `cache.merge(finalMergedCollection)` and `keysChanged(...)` then run synchronously on the now-warm cache, **before** the storage promises are issued.
3. `Storage.multiMerge` / `Storage.multiSet` are pushed afterwards. The `retryOperation` / DevTools chain is unchanged.

Result: subscribers receive the merged data **before** any storage write is attempted, and storage failure no longer races the cache update — matching the behavior of every other Onyx write method.

### Related Issues

https://github.com/Expensify/App/issues/90634

### Companion PR

Expensify/App#91160 — bumps the App's react-native-onyx pin to this branch's head for end-to-end verification.

### Automated Tests

Added a new regression test in \`tests/unit/onyxUtilsTest.ts\`:

> \`mergeCollection cache-first ordering > updates cache and notifies subscribers even when Storage.multiMerge rejects\`

[...]

PR #792, opened 2026-05-25
[Do Not Merge] Avoid subscriber re-fire when storage writes retry

elirangoshen:
### Details

Follow-up to #787. After the cache-first/storage-second refactor of `mergeCollectionWithPatches`, the storage-retry path was identified as re-notifying subscribers once per retry attempt — `waitForCollectionCallback` subscribers fire on every `keysChanged` call by contract, so a transient storage error could surface 2-6 callback invocations for a single logical write. Same shape applies to `multiSetWithRetry`, `setCollectionWithRetry`, and `partialSetCollection` — all four flow through `retryOperation`, which re-enters the orchestrator on retry.

`setWithRetry` was already safe because `broadcastUpdate` short-circuits on `hasChanged === false`.

#### Fix shape

Two small pieces:

1. **`!retryAttempt` guard around the subscriber notification** in each of the four functions. `cache.set/merge` stays unguarded — it's idempotent, runs every attempt, and is what restores cache state after `retryOperation`'s eviction branch drops a key. The notification (`keysChanged`/`keyChanged`) is the one piece that's not safe to re-fire.

2. **`skipNotify` flag on `remove()`** + passing `true` from `retryOperation`'s eviction branch. Without this, eviction-during-retry would fire `keyChanged(undefined)` and strand subscribers in the "removed" state, since the orchestrator no longer re-fires `keysChanged` on retry. `cache.drop` and `Storage.removeItem` still run — only the spurious `keyChanged(undefined)` is skipped, and the imminent retry's `cache.set` covers cache restoration.

That's the entire diff: +157 / -9 across `lib/OnyxUtils.ts` and `tests/unit/onyxUtilsTest.ts`.

### Related Issues

No separate GH issue — this PR addresses [#787 review comment](https://github.com/Expensify/react-native-onyx/pull/787#discussion_r3281758212) directly.

### Linked E/App PR

https://github.com/Expensify/App/pull/91626

### Automated Tests

Added a new `retry side-effect idempotency` describe block in `tests/unit/onyxUtilsTest.ts` with one test per affected function. Each seeds a `waitForCollectionCallback: true` subscriber, mocks `Storage.multiSet` (or `Storage.multiMerge` for `mergeCollection`) to reject once then resolve, runs the write, and asserts the subscriber was called exactly **1** time across the retry chain:

1. **`mergeCollection — waitForCollectionCallback subscriber fires once across retries`**
2. **`Onyx.multiSet — collection subscriber fires once across retries`**
3. **`Onyx.setCollection — collection subscriber fires once across retries`**
4. **`OnyxUtils.partialSetCollection — collection subscriber fires once across retries`**

All 4 fail with `Received number of calls: 2` if the `!retryAttempt` guard is reverted in any of the four orchestrators — confirmed locally via `git stash` of `lib/OnyxUtils.ts`.

All 6 existing **storage eviction** tests still pass without modification, validating that the `skipNotify` change to `remove()` preserves the eviction recovery path: the imminent retry's `cache.set` re-populates the cache for any evicted in-flight key, and subscribers never see a transient `undefined` because `keyChanged(undefined)` no longer fires from the eviction branch.

Local results:
- `npx jest` — **455 / 455** pass across 17 suites.
- `npx tsc --noEmit` — clean.

### Manual Tests

The four refactored paths (`mergeCollectionWithPatches`, `multiSetWithRetry`, `setCollectionWithRetry`, `partialSetCollection`) are reached from the App via Pusher events, the `OpenApp` response, every `Onyx.update` batch that contains a `MERGE_COLLECTION` / `MULTI_SET` / `SET_COLLECTION` op, LHN refresh, Search filters, chat sends, mark-all-as-read, hold/unhold, workspace switching, etc. The companion App PR [Expensify/App#91626](https://github.com/Expensify/App/pull/91626) pins `react-native-onyx` to this branch's head for end-to-end exercise; the steps below are the same protocol used for #787.

**Setup**

1. In the App repo, check out [Expensify/App#91626](https://github.com/Expensify/App/pull/91626) (`elirangoshen/fix/retry-side-effects-idempotency` on `callstack-internal/Expensify-App`), which bumps `package.json`'s `react-native-onyx` to this branch's head.
2. `npm install` under Node 20.20.0, then `npm run web`.
3. Open https://dev.new.expensify.com:8082/ in Chrome with DevTools open.
4. Sign in to a test account.

**Functional smoke — no regression in the refactored paths**

The refactor preserves the cache-first invariant from #787, so all of these should look indistinguishable from `main` to the user:

1. **Initial hydration** — after login, LHN reports list and workspace switcher populate within a few seconds. No missing rows vs. a baseline session on `main`. _(exercises `mergeCollectionWithPatches` via Pusher / `OpenApp`)_
2. **Chat message** — open a chat, send a message. Appears immediately (optimistic merge into `reportActions_`), confirms via Pusher, persists after reload. _(`mergeCollectionWithPatches`)_
3. **Mark all as read** — click the mark-all-as-read action. All unread badges clear immediately and stay cleared after reload. _(`mergeCollectionWithPatches`)_
4. **Search & filter** — open Search, apply a filter. Results populate within a few seconds; live updates as filters change; no duplicates. _(`mergeCollectionWithPatches` + `setCollectionWithRetry`)_
5. **Hold / unhold an expense** — toggle hold on an expense in a report. Badge appears/clears immediately; persists after reload. _(`mergeCollectionWithPatches`)_
6. **Submit expense** — FAB → Submit expense. Expense appears in the report immediately, no duplicate, persists after reload. _(`mergeCollectionWithPatches`)_
7. **Switch workspaces** — switch via the workspace switcher. LHN reports filter to the new workspace within a few seconds. _(`mergeCollectionWithPatches`)_

**Storage-failure simulation — core regression guard for this PR**

This is the test that exercises the retry path the fix targets. With the fix in place, retries on storage failure should produce **one** subscriber notification per logical operation (not one per retry attempt), and brand-new keys should stay routed through `multiSet` even when an earlier `multiMerge` retry kicked in.

1. With the App running and authenticated, open Chrome DevTools → **Application** tab → **IndexedDB**.
2. Find the database used by Onyx (typically named `OnyxDB`).
3. **Right-click → Delete database** while the App is still running and connected (do **not** reload).
4. Immediately trigger a `mergeCollection`-driven action — switch workspaces, send a chat message, or apply a search filter.
5. **Expected with this fix:** the UI updates correctly; the new state is visible to subscribers even though the underlying IDB write fails and `retryOperation` kicks in. Console shows storage error logs and retry attempts (look for `Failed to save to storage. Error: ... retryAttempt: N/5` in `Logger.logInfo`), but **no white screen, no stale UI, no data loss within the session**, and **no duplicate `waitForCollectionCallback` subscriber invocations** (verify via React DevTools "Highlight updates when components render" or via a `console.count` instrumented in a temporary `useOnyx` callback if you want to be precise).
6. Try the same with a `multiSet`-driven action (least common — typically a deep-link prep) and a `setCollection`-driven action (Search filter result population) to cover the other refactored paths.
7. (Optional regression check) Revert this PR's commits locally and re-run step 4. Without the fix, you'll either see (a) `Onyx.connect({waitForCollectionCallback: true})` callbacks firing twice in DevTools traces, or (b) on the routing bug specifically, no observable user-facing difference on IDB but the issue is fixed under the hood for non-IDB backends.

**Cold-cache merge — preserves the pre-warm fast path from #787**

1. Sign out and clear site data to force a cold cache.
2. Sign back in. LHN should populate within the same window as a baseline `main` session — the cache pre-warm path (existing in `mergeCollectionWithPatches`) is untouched by this PR.

### Author Checklist

- [x] I linked the corresponding Expensify/App PR in the `### Linked E/App PR` section above, and verified this change against it (E/App CI passed and manual testing completed) _(companion PR https://github.com/Expensify/App/pull/91626 pins this branch's HEAD for end-to-end exercise)_
    - [x] I tested this PR with a [High Traffic account](https://github.com/Expensify/App/blob/main/contributingGuides/CONTRIBUTING.md#high-traffic-accounts) against the staging or production API to ensure there are no regressions (e.g. long loading states that impact usability). _(pending companion App PR #91626 manual run)_
- [x] I included screenshots or videos for tests on [all platforms](https://github.com/Expensify/App/blob/main/contributingGuides/CONTRIBUTING.md#make-sure-you-can-test-on-all-platforms) _(pending companion App PR #91626; this PR has no UI surface)_
    - [x] Android / native
    - [x] Android / Chrome
    - [x] iOS / native
    - [x] iOS / Safari
    - [x] MacOS / Chrome / Safari
- [x] If a new code pattern is added I verified it was agreed to be used by multiple Expensify engineers _(orchestrator/write-helper split was suggested in #787 review)_
- [x] I tested other components that can be impacted by my changes (i.e. if the PR modifies a shared library or component like `Avatar`, I verified the components using `Avatar` are working as expected) _(459/459 unit tests green)_
- [x] I verified all code is DRY (the PR doesn't include any logic written more than once, with the exception of tests) _(`persistCollectionWrite` is shared between `setCollectionWithRetry` and `partialSetCollection` since their storage tails are byte-identical)_
- [x] If a new component is created I verified that: _(N/A — no new components)_
    - [x] A similar component doesn't exist in the codebase _(N/A)_
    - [x] All props are defined accurately and each prop has a `/** comment above it */` _(N/A)_
    - [x] The file is named correctly _(N/A)_
    - [x] The component has a clear name that is non-ambiguous and the purpose of the component can be inferred from the name alone _(N/A)_
    - [x] The only data being stored in the state is data necessary for rendering and nothing else _(N/A)_
    - [x] If we are not using the full Onyx data that we loaded, I've added the proper selector in order to ensure the component only re-renders when the data it is using changes _(N/A)_
    - [x] For Class Components, any internal methods passed to components event handlers are bound to `this` properly so there are no scoping issues (i.e. for `onClick={this.submit}` the method `this.submit` should be bound to `this` in the constructor) _(N/A)_
    - [x] Any internal methods bound to `this` are necessary to be bound (i.e. avoid `this.submit = this.submit.bind(this);` if `this.submit` is never passed to a component event handler like `onClick`) _(N/A)_
    - [x] All JSX used for rendering exists in the render method _(N/A)_
    - [x] The component has the minimum amount of code necessary for its purpose, and it is broken down into smaller components in order to separate concerns and functions _(N/A)_
- [x] If any new file was added I verified that: _(N/A — no new files)_
    - [x] The file has a description of what it does and/or why is needed at the top of the file if the code is not self explanatory _(N/A)_
- [x] If the PR modifies a generic component, I tested and verified that those changes do not break usages of that component in the rest of the App (i.e. if a shared library or component like `Avatar` is modified, I verified that `Avatar` is working as expected in all cases) _(pending companion App PR #91626)_
- [x] If the `main` branch was merged into this PR after a review, I tested again and verified the outcome was still expected according to the `Test` steps. _(N/A — no merge from main yet)_

### Screenshots/Videos

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

https://github.com/user-attachments/assets/58849db2-2a78-4cea-a73e-cdb4bed29777

https://github.com/user-attachments/assets/2d2889f1-5b8c-4514-afa5-a93396eeeeee

https://github.com/user-attachments/assets/4de8005a-623f-41f0-80e6-b2f588387656

https://github.com/user-attachments/assets/ef0089c1-1c71-4ed8-bf5e-c688dbc9e80b

https://github.com/user-attachments/assets/c8793e5c-0b46-4a5b-8d05-14f931db0350

https://github.com/user-attachments/assets/0dfb2aec-2127-41a6-9c9d-34e66b2f6904

</details>

### DIFF
--- a/lib/OnyxUtils.ts
+++ b/lib/OnyxUtils.ts
@@ -862,8 +862,6 @@ function retryOperation<TMethod extends RetriableOnyxOperation>(
     Logger.logInfo(`Out of storage. Evicting least recently accessed key (${keyForRemoval}) and retrying. Error: ${error}`);
     reportStorageQuota(error);
 
-    // The evicted key is never in-flight (excluded above), so this is a genuine removal —
-    // subscribers must see the removed state.
     // @ts-expect-error No overload matches this call.
     return remove(keyForRemoval).then(() => onyxMethod(defaultParams, nextRetryAttempt));
 }
--- a/lib/OnyxUtils.ts
+++ b/lib/OnyxUtils.ts
@@ -769,9 +769,14 @@ function getCollectionDataAndSendAsObject<TKey extends OnyxKey>(matchingKeys: Co
 /**
  * Remove a key from Onyx and update the subscribers
  */
-function remove<TKey extends OnyxKey>(key: TKey, isProcessingCollectionUpdate?: boolean): Promise<void> {
+function remove<TKey extends OnyxKey>(key: TKey, isProcessingCollectionUpdate?: boolean, skipNotify?: boolean): Promise<void> {
     cache.drop(key);
-    keyChanged(key, undefined as OnyxValue<TKey>, undefined, isProcessingCollectionUpdate);
+    // skipNotify is used by retryOperation's eviction branch — the imminent retry's cache.set
+    // will re-populate cache, so firing keyChanged(undefined) here would only strand subscribers
+    // in the "removed" state across the retry.
+    if (!skipNotify) {
+        keyChanged(key, undefined as OnyxValue<TKey>, undefined, isProcessingCollectionUpdate);
+    }
 
     if (OnyxKeys.isRamOnlyKey(key)) {
         return Promise.resolve();
@@ -842,8 +847,10 @@ function retryOperation<TMethod extends RetriableOnyxOperation>(error: Error, on
     Logger.logInfo(`Out of storage. Evicting least recently accessed key (${keyForRemoval}) and retrying. Error: ${error}`);
     reportStorageQuota(error);
 
+    // skipNotify=true: retry's orchestrator skips keysChanged on retryAttempt > 0, so we
+    // must not let remove() fire keyChanged(undefined) — cache.set on retry restores the value.
     // @ts-expect-error No overload matches this call.
-    return remove(keyForRemoval).then(() => onyxMethod(defaultParams, nextRetryAttempt));
+    return remove(keyForRemoval, undefined, true).then(() => onyxMethod(defaultParams, nextRetryAttempt));
 }
 
 /**
@@ -1395,14 +1402,21 @@ function multiSetWithRetry(data: OnyxMultiSetInput, retryAttempt?: number): Prom
             // so re-entrant callbacks (e.g. Onyx.set inside a callback) see consistent cache
             // and subscriber state, matching the original per-key notification semantics.
             cache.set(key, value);
-            keyChanged(key, value);
+            // Skip subscriber notification on retry — already notified on attempt 0.
+            // waitForCollectionCallback subscribers re-fire on every keyChanged by contract.
+            if (!retryAttempt) {
+                keyChanged(key, value);
+            }
         }
     }
 
     // One keysChanged() per collection — fires each collection-level subscriber once and lets
     // keysChanged() internally decide which individual member subscribers need notification.
-    for (const [collectionKey, batch] of collectionBatches) {
-        keysChanged(collectionKey as CollectionKeyBase, batch.partial, batch.previous);
+    // Skip on retry — already notified on attempt 0 (see same-reason comment above).
+    if (!retryAttempt) {
+        for (const [collectionKey, batch] of collectionBatches) {
+            keysChanged(collectionKey as CollectionKeyBase, batch.partial, batch.previous);
+        }
     }
 
     const keyValuePairsToStore = keyValuePairsToSet.filter((keyValuePair) => {
@@ -1476,7 +1490,11 @@ function setCollectionWithRetry<TKey extends CollectionKeyBase>({collectionKey, 
 
         for (const [key, value] of keyValuePairs) cache.set(key, value);
 
-        keysChanged(collectionKey, mutableCollection, previousCollection);
+        // Skip subscriber notification on retry — already notified on attempt 0.
+        // waitForCollectionCallback subscribers re-fire on every keysChanged by contract.
+        if (!retryAttempt) {
+            keysChanged(collectionKey, mutableCollection, previousCollection);
+        }
 
         // RAM-only keys are not supposed to be saved to storage
         if (OnyxKeys.isRamOnlyKey(collectionKey)) {
@@ -1611,7 +1629,11 @@ function mergeCollectionWithPatches<TKey extends CollectionKeyBase>(
                 // write fails.
                 const previousCollection = getCachedCollection(collectionKey, existingKeys);
                 cache.merge(finalMergedCollection);
-                keysChanged(collectionKey, finalMergedCollection, previousCollection);
+                // Skip subscriber notification on retry — already notified on attempt 0.
+                // waitForCollectionCallback subscribers re-fire on every keysChanged by contract.
+                if (!retryAttempt) {
+                    keysChanged(collectionKey, finalMergedCollection, previousCollection);
+                }
 
                 const promises = [];
 
@@ -1690,7 +1712,11 @@ function partialSetCollection<TKey extends CollectionKeyBase>({collectionKey, co
 
         for (const [key, value] of keyValuePairs) cache.set(key, value);
 
-        keysChanged(collectionKey, mutableCollection, previousCollection);
+        // Skip subscriber notification on retry — already notified on attempt 0.
+        // waitForCollectionCallback subscribers re-fire on every keysChanged by contract.
+        if (!retryAttempt) {
+            keysChanged(collectionKey, mutableCollection, previousCollection);
+        }
 
         if (OnyxKeys.isRamOnlyKey(collectionKey)) {
             sendActionToDevTools(METHOD.SET_COLLECTION, undefined, mutableCollection);
--- a/lib/OnyxUtils.ts
+++ b/lib/OnyxUtils.ts
@@ -777,14 +777,9 @@ function getCollectionDataAndSendAsObject<TKey extends OnyxKey>(matchingKeys: Co
 /**
  * Remove a key from Onyx and update the subscribers
  */
-function remove<TKey extends OnyxKey>(key: TKey, isProcessingCollectionUpdate?: boolean, skipNotify?: boolean): Promise<void> {
+function remove<TKey extends OnyxKey>(key: TKey, isProcessingCollectionUpdate?: boolean): Promise<void> {
     cache.drop(key);
-    // skipNotify is used by retryOperation's eviction branch — the imminent retry's cache.set
-    // will re-populate cache, so firing keyChanged(undefined) here would only strand subscribers
-    // in the "removed" state across the retry.
-    if (!skipNotify) {
-        keyChanged(key, undefined as OnyxValue<TKey>, undefined, isProcessingCollectionUpdate);
-    }
+    keyChanged(key, undefined as OnyxValue<TKey>, undefined, isProcessingCollectionUpdate);
 
     if (OnyxKeys.isRamOnlyKey(key)) {
         return Promise.resolve();
@@ -851,8 +846,10 @@ function retryOperation<TMethod extends RetriableOnyxOperation>(
         return onyxMethod(defaultParams, nextRetryAttempt);
     }
 
-    // Find the least recently accessed evictable key that we can remove
-    const keyForRemoval = cache.getKeyForEviction();
+    // Find the least recently accessed evictable key that we can remove. Never evict an in-flight
+    // key — its cache value is the merge base this retry depends on, so dropping it would truncate
+    // the write to just the delta and diverge cache from storage.
+    const keyForRemoval = cache.getKeyForEviction(inFlightKeys);
     if (!keyForRemoval) {
         // If we have no acceptable keys to remove then we are possibly trying to save mission critical data. If this is the case,
         // then we should stop retrying as there is not much the user can do to fix this. Instead of getting them stuck in an infinite loop we
@@ -865,12 +862,10 @@ function retryOperation<TMethod extends RetriableOnyxOperation>(
     Logger.logInfo(`Out of storage. Evicting least recently accessed key (${keyForRemoval}) and retrying. Error: ${error}`);
     reportStorageQuota(error);
 
-    // Only suppress keyChanged(undefined) when the evicted key is part of the in-flight
-    // write — then cache.set on retry will restore it. For unrelated keys, eviction is a
-    // genuine loss and subscribers must see the removed state.
-    const willBeRestored = inFlightKeys?.has(keyForRemoval) ?? false;
+    // The evicted key is never in-flight (excluded above), so this is a genuine removal —
+    // subscribers must see the removed state.
     // @ts-expect-error No overload matches this call.
-    return remove(keyForRemoval, undefined, willBeRestored).then(() => onyxMethod(defaultParams, nextRetryAttempt));
+    return remove(keyForRemoval).then(() => onyxMethod(defaultParams, nextRetryAttempt));
 }
 
 /**
@@ -1666,8 +1661,9 @@ function mergeCollectionWithPatches<TKey extends CollectionKeyBase>(
 
                 const promises = [];
 
-                // New keys will be added via multiSet while existing keys will be updated using multiMerge
-                // This is because setting a key that doesn't exist yet with multiMerge will throw errors
+                // New keys go through multiSet and existing keys through multiMerge. multiMerge on a
+                // missing key stores the value just like multiSet across all backends; splitting them lets
+                // multiSet strip nested nulls (the merge layer keeps them to delete nested storage keys).
                 // We can skip this step for RAM-only keys as they should never be saved to storage
                 if (!OnyxKeys.isRamOnlyKey(collectionKey) && keyValuePairsForExistingCollection.length > 0) {
                     promises.push(Storage.multiMerge(keyValuePairsForExistingCollection));
--- a/lib/OnyxUtils.ts
+++ b/lib/OnyxUtils.ts
@@ -802,7 +802,13 @@ function reportStorageQuota(error?: Error): Promise<void> {
  * - Non-retriable errors: logs an alert and resolves without retrying
  * - Other errors: retries the operation
  */
-function retryOperation<TMethod extends RetriableOnyxOperation>(error: Error, onyxMethod: TMethod, defaultParams: Parameters<TMethod>[0], retryAttempt: number | undefined): Promise<void> {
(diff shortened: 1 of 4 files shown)


lib/OnyxUtils.ts
```diff
@@ -1333,6 +1333,21 @@ function setWithRetry<TKey extends OnyxKey>({key, value, options}: SetParams<TKe
...
+/**
+ * Storage-write tail of multiSetWithRetry, isolated so that retryOperation re-enters only the
+ * storage step. Cache and subscriber notifications already happened in the orchestrator, so
+ * retries no longer re-fire `waitForCollectionCallback` subscribers with the same payload.
+ */
+function persistMultiSetWrite(params: {keyValuePairsToStore: StorageKeyValuePair[]; newData: OnyxMultiSetInput}, retryAttempt?: number): Promise<void> {
+    const {keyValuePairsToStore, newData} = params;
+
+    return Storage.multiSet(keyValuePairsToStore)
+        .catch((error) => OnyxUtils.retryOperation(error, persistMultiSetWrite, params, retryAttempt))
```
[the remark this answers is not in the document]
elirangoshen: Addressed in 11bc0a62.

**Fix:** each orchestrator (`multiSetWithRetry`, `setCollectionWithRetry`, `partialSetCollection`, `mergeCollectionWithPatches`) now constructs a `restoreEvictedKey(key)` closure that re-applies its cache write + subscriber notification for a single in-flight key. The closure is passed via the write helper's params; `retryOperation` invokes it after `remove(keyForRemoval)` and before re-entering the write helper. No-op when the evicted key isn't part of this write (the typical case).

**`mergeCollection` nuance:** the closure snapshots the cache's *post-merge* state (not just the delta) so previously-merged-in fields survive when the eviction's `cache.drop` forces us to re-populate cache from scratch — otherwise we'd lose fields like `{id: 1}` when restoring after a `mergeCollection({key: {value: 'new'}})` retry.

**Regression coverage:** added 3 tests under `storage eviction` — one per write helper — that mock `Storage.multiSet`/`Storage.multiMerge` to fail with `database or disk is full` once on a setup where the LRU evictable key is the same key being written, and assert both cache + subscriber reflect the restored value (not `undefined`). All 3 fail without the lib changes (`Received: undefined` / missing fields) and pass with them.

459/459 tests pass; tsc and prettier clean.

fabioh8010:
i will review it soon

elirangoshen:
@codex review

lib/OnyxUtils.ts
```diff
@@ -842,8 +847,10 @@ function retryOperation<TMethod extends RetriableOnyxOperation>(error: Error, on
     Logger.logInfo(`Out of storage. Evicting least recently accessed key (${keyForRemoval}) and retrying. Error: ${error}`);
     reportStorageQuota(error);
 
+    // skipNotify=true: retry's orchestrator skips keysChanged on retryAttempt > 0, so we
+    // must not let remove() fire keyChanged(undefined) — cache.set on retry restores the value.
     // @ts-expect-error No overload matches this call.
-    return remove(keyForRemoval).then(() => onyxMethod(defaultParams, nextRetryAttempt));
+    return remove(keyForRemoval, undefined, true).then(() => onyxMethod(defaultParams, nextRetryAttempt));
```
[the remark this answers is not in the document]
elirangoshen: i add a fix in last commit

fabioh8010:
@elirangoshen

### Routing claim in PR description (general PR comment)

The PR description lists bug #1 as "new vs existing routing recomputed against an already-mutated cache on retry," but I don't think this is actually a bug worth fixing. `multiMerge` on a key that doesn't exist yet behaves identically to `multiSet` across all our storage backends — the inline comment in `mergeCollectionWithPatches` saying it "will throw errors" is stale. So even when attempt 0's `cache.merge` causes the retry to reclassify a brand-new key as existing and route it through `multiMerge`, there's no observable difference in behavior.

Could we first:
1. Add one test to `tests/unit/storage/providers/IDBKeyvalProviderTest.ts` and `tests/unit/storage/providers/SQLiteProviderTest.ts` that assert a `multiMerge` of a non-existing key will just store the value as expected? You need to merge with `main`
2. Update PR description and the comment in `mergeCollectionWithPatches` about `will throw errors` which is not true?

### Eviction-restore (inline on `mergeCollectionWithPatches`, line ~1643 `cache.merge(finalMergedCollection)`)

I think this loses pre-existing fields on eviction-driven retry of a `mergeCollection`. Trace:

1. Seed `{id: 1, value: 'orig'}` for `memberKey`.
3. `mergeCollection({memberKey: {value: 'merged'}})`. On attempt 0, `finalMergedCollection = {memberKey: {value: 'merged'}}` (the delta — `existingKeyCollection[key] = resultCollection[key]` at line ~1606 stores the delta, not the merged value). `cache.merge` against the cache's `{id:1, value:'orig'}` produces `{id:1, value:'merged'}` ✅.
4. Storage rejects with disk-full. `retryOperation` evicts `memberKey` via `remove()`. `cache.drop` removes it from cache *and* from `storageKeys`.
5. Retry attempt 1: `getAllKeys()` no longer includes `memberKey` → classified as **new**, not existing. `cache.merge({memberKey: {value: 'merged'}})` against an empty cache produces `{value: 'merged'}` — `{id: 1}` is lost. `keysChanged` is skipped by `if (!retryAttempt)`, so subscribers stay on the attempt-0 value, but storage now persists the truncated `{value: 'merged'}`.

The previous iteration of this PR specifically called this out and handled it with a post-merge snapshot in `restoreEvictedKey`. The new approach relies on `cache.merge(finalMergedCollection)` re-running on retry, but `finalMergedCollection` is the delta, not the post-merge value, so it doesn't reconstruct what attempt 0 produced.

A repro test would be roughly:
```ts
await Onyx.set(memberKey, {id: 1, value: 'orig'});
// make memberKey the LRU evictable, mock Storage.multiMerge to reject with disk-full once
await Onyx.mergeCollection(collectionKey, {[memberKey]: {value: 'merged'}});
expect(OnyxCache.get(memberKey)).toEqual({id: 1, value: 'merged'});
// currently resolves to {value: 'merged'} — {id: 1} is gone
```

Easiest fix is probably the same shape as before — snapshot the post-merge cache values for in-flight keys and re-apply them in the eviction branch. Or skip the `cache.drop` for in-flight keys entirely since the retry's `cache.merge` is the source of truth (would also let you drop the `skipNotify` plumbing).

elirangoshen:
> @elirangoshen
> 
> ### Routing claim in PR description (general PR comment)
> The PR description lists bug #1 as "new vs existing routing recomputed against an already-mutated cache on retry," but I don't think this is actually a bug worth fixing. `multiMerge` on a key that doesn't exist yet behaves identically to `multiSet` across all our storage backends — the inline comment in `mergeCollectionWithPatches` saying it "will throw errors" is stale. So even when attempt 0's `cache.merge` causes the retry to reclassify a brand-new key as existing and route it through `multiMerge`, there's no observable difference in behavior.
> 
> Could we first:
> 
> 1. Add one test to `tests/unit/storage/providers/IDBKeyvalProviderTest.ts` and `tests/unit/storage/providers/SQLiteProviderTest.ts` that assert a `multiMerge` of a non-existing key will just store the value as expected? You need to merge with `main`
> 2. Update PR description and the comment in `mergeCollectionWithPatches` about `will throw errors` which is not true?
> 
> ### Eviction-restore (inline on `mergeCollectionWithPatches`, line ~1643 `cache.merge(finalMergedCollection)`)
> I think this loses pre-existing fields on eviction-driven retry of a `mergeCollection`. Trace:
> 
> 1. Seed `{id: 1, value: 'orig'}` for `memberKey`.
> 2. `mergeCollection({memberKey: {value: 'merged'}})`. On attempt 0, `finalMergedCollection = {memberKey: {value: 'merged'}}` (the delta — `existingKeyCollection[key] = resultCollection[key]` at line ~1606 stores the delta, not the merged value). `cache.merge` against the cache's `{id:1, value:'orig'}` produces `{id:1, value:'merged'}` ✅.
> 3. Storage rejects with disk-full. `retryOperation` evicts `memberKey` via `remove()`. `cache.drop` removes it from cache _and_ from `storageKeys`.
> 4. Retry attempt 1: `getAllKeys()` no longer includes `memberKey` → classified as **new**, not existing. `cache.merge({memberKey: {value: 'merged'}})` against an empty cache produces `{value: 'merged'}` — `{id: 1}` is lost. `keysChanged` is skipped by `if (!retryAttempt)`, so subscribers stay on the attempt-0 value, but storage now persists the truncated `{value: 'merged'}`.
> 
> The previous iteration of this PR specifically called this out and handled it with a post-merge snapshot in `restoreEvictedKey`. The new approach relies on `cache.merge(finalMergedCollection)` re-running on retry, but `finalMergedCollection` is the delta, not the post-merge value, so it doesn't reconstruct what attempt 0 produced.
> 
> A repro test would be roughly:
> 
> ```ts
> await Onyx.set(memberKey, {id: 1, value: 'orig'});
> // make memberKey the LRU evictable, mock Storage.multiMerge to reject with disk-full once
> await Onyx.mergeCollection(collectionKey, {[memberKey]: {value: 'merged'}});
> expect(OnyxCache.get(memberKey)).toEqual({id: 1, value: 'merged'});
> // currently resolves to {value: 'merged'} — {id: 1} is gone
> ```
> 
> Easiest fix is probably the same shape as before — snapshot the post-merge cache values for in-flight keys and re-apply them in the eviction branch. Or skip the `cache.drop` for in-flight keys entirely since the retry's `cache.merge` is the source of truth (would also let you drop the `skipNotify` plumbing).

Hi, I pushed a fix , can you check ?

lib/OnyxUtils.ts
```diff
@@ -865,12 +862,10 @@ function retryOperation<TMethod extends RetriableOnyxOperation>(
     Logger.logInfo(`Out of storage. Evicting least recently accessed key (${keyForRemoval}) and retrying. Error: ${error}`);
     reportStorageQuota(error);
 
-    // Only suppress keyChanged(undefined) when the evicted key is part of the in-flight
-    // write — then cache.set on retry will restore it. For unrelated keys, eviction is a
-    // genuine loss and subscribers must see the removed state.
-    const willBeRestored = inFlightKeys?.has(keyForRemoval) ?? false;
+    // The evicted key is never in-flight (excluded above), so this is a genuine removal —
+    // subscribers must see the removed state.
```
fabioh8010: Looks like unecessary comment

tests/unit/storage/providers/IDBKeyvalProviderTest.ts
```diff
@@ -172,6 +172,11 @@ describe('IDBKeyValProvider', () => {
                 ),
             ).toEqual(expectedEntries.map((e) => (e[1] === null ? undefined : e[1])));
         });
+
+        it('should insert a new record when key does not exist', async () => {
```
fabioh8010, line 176: Could you add the same test for SQLiteProviderTest too?

elirangoshen: it seems it already added in main in the pr #794 
```
  it('should insert a new record when key does not exist', async () => {
      await SQLiteProvider.multiMerge([[ONYXKEYS.TEST_KEY_2, {fresh: true}]]);
      expect(await SQLiteProvider.getItem(ONYXKEYS.TEST_KEY_2)).toEqual({fresh: true});
  });

```

lib/OnyxUtils.ts
elirangoshen: removed

fabioh8010 (approved):
@Julesssss could you have a look?

Julesssss (approved):
PR description slightly outdated, but no problem ..

PR #793, opened 2026-05-26
perf: batch cold-cache pre-warm in mergeCollectionWithPatches via multiGet

elirangoshen:
### Details

Hybrid pre-warm strategy for `mergeCollectionWithPatches` that replaces the unconditional `Promise.all(existingKeys.map((key) => get(key)))`:

- **Fast path** (every `existingKey` is already warm in cache): use a sync-resolved `Promise.resolve()`. No extra microtask hops, preserving the original promise-chain depth and the subscriber-callback timing that dependent tests rely on (`Onyx.update` batch tests broadcast a single merged callback rather than `undefined` followed by the merged value).

- **Slow path** (at least one `existingKey` is a cache miss): use `multiGet`, which batches the missing keys into a single `Storage.multiGet` round-trip instead of N parallel `get()` invocations and writes the storage values back to cache before resolving.

Net result: same correctness as before, fewer storage operations on cold-cache merges, identical broadcast timing for warm-cache merges.

Addresses follow-up from Expensify/react-native-onyx#787 review.

### Fixed Issues

$ https://github.com/Expensify/App/issues/90634

### Linked E/App PR

https://github.com/Expensify/App/pull/91585

### Automated Tests

Added a new `describe('mergeCollection pre-warm', ...)` block in `tests/unit/onyxUtilsTest.ts` with 5 tests:

1. **`fast path: skips storage reads entirely when every existing key is warm in cache`** — Seeds two members via `Onyx.set`, spies on `StorageMock.multiGet` and `StorageMock.getItem`, then runs `Onyx.mergeCollection`. Asserts both spies are called **0 times**. Confirms the diff's `Promise.resolve()` shortcut.

2. **`slow path: batches cold existing keys into a single Storage.multiGet, with no individual getItem calls`** — Seeds three members, evicts two from cache (cold), leaves one warm. Asserts `Storage.multiGet` is called **exactly once** with only the two cold keys (the warm key is filtered out by `OnyxUtils.multiGet`), and `Storage.getItem` is never called during pre-warm.

3. **`slow path: cold-cache merge layers the new delta on top of existing storage data (no field drops)`** — Seeds a key with `{a:1, b:2}` then evicts from cache. Merges `{c:3}` and asserts the cache holds `{a:1, b:2, c:3}`. Without the pre-warm reading from storage, `cache.merge` would start from `undefined` and drop `{a:1, b:2}` — this guards the correctness invariant the in-code comment specifically calls out.

4. **`warm cache: subscriber receives a single merged broadcast for an Onyx.update batch (no transient undefined)`** — Subscribes to a warm collection key, fires an `Onyx.update` with a `MERGE_COLLECTION` op. Asserts subscriber was called exactly once with the final merged value (NOT `undefined` then merged on a later microtask). Guards the promise-chain-depth invariant.

5. **`equivalence: warm-path and cold-path produce the same final cache state for the same merge`** — Runs the same effective merge against a warm cache and against a cold cache; asserts the post-merge collection state is identical in both runs.

**Suite-pollution fix:** the `retryOperation` describe block earlier in the file mutates `StorageMock.setItem` (and other methods) without restoring them. This block captures pristine references at file-load time and restores them in `beforeEach` so seeding via `Onyx.set` actually persists.

**Helper note:** `OnyxCache.drop` removes the key from `storageKeys`, which makes `getAllKeys()` miss it when other keys remain in cache. The `evictFromCache` helper calls `OnyxCache.addKey(key)` after `drop` so the key stays "tracked but unloaded" — exactly the cold-but-persisted state the slow path is meant to handle.

Local results:

- `npx jest` — **450/450** pass across 16 suites.
- `npx tsc --noEmit` — clean.

### Manual Tests

End-to-end verification against `Expensify/App` via the companion PR **Expensify/App#91585**, which pins `react-native-onyx`'s `package.json` to this branch's head SHA.

**Setup**

1. In the App repo, check out the companion branch that pins `react-native-onyx` to this PR's head SHA.
2. `npm install` under Node 20.20.0, then `npm run web`.
3. Open https://dev.new.expensify.com:8082/ in Chrome with DevTools open.
4. Sign in.

**Functional smoke (same flows as #787, expect no regression)**

For each: open the screen, perform the action, verify UI updates immediately, persist after reload.

1. Initial hydration after login — LHN populates correctly.
2. Send a chat message — appears immediately, confirms via Pusher, persists after reload.
3. Mark-all-as-read — badges clear and stay cleared.
4. Search filter — results populate and update live.
5. Hold / unhold an expense — badge toggles and persists.
6. Submit expense via FAB — appears in report immediately, persists.
7. Switch workspaces — LHN filters to new workspace.

**Cold-cache merge correctness**

1. Sign in and perform an action that writes collection data.
2. Reload (clears cache, keeps storage).
3. Trigger a `MERGE_COLLECTION` against one of those keys **before** the LHN hydrates it.
4. **Expected**: merged value retains all fields from storage; the new delta is layered on top — no silent drops. (Slow-path correctness.)

**Storage-failure regression (carry-over from #787)**

1. With App running and authenticated, Application → IndexedDB → right-click → delete `OnyxDB`. **Do not reload.**
2. Immediately trigger a `MERGE_COLLECTION` action.
3. **Expected**: UI updates correctly; new state is visible to subscribers even though the IDB write fails. Console shows storage errors, but no white screen / no stale UI / no data loss within the session.

### Author Checklist

  - [x] Android / native
  - [x] Android / Chrome
  - [x] iOS / native
  - [x] iOS / Safari
  - [x] MacOS / Chrome / Safari

### Screenshots/Videos

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

https://github.com/user-attachments/assets/07611d7d-4c0b-488d-a440-25cbd39d8cae

https://github.com/user-attachments/assets/28948810-6417-4da1-8182-be21f1b2b863

https://github.com/user-attachments/assets/49d5e79f-e7f8-4476-9c38-4669d42f2383

https://github.com/user-attachments/assets/9c743510-d5a7-412e-9986-6fa577a36532

Screenshot 2026-05-25 at 12 25 01

</details>

### DIFF
--- a/lib/OnyxUtils.ts
+++ b/lib/OnyxUtils.ts
@@ -1597,12 +1597,20 @@ function mergeCollectionWithPatches<TKey extends CollectionKeyBase>(
             // finalMergedCollection contains all the keys that were merged, without the keys of incompatible updates
             const finalMergedCollection = {...existingKeyCollection, ...newCollection};
 
-            // Pre-warm cache for any existing storage keys that aren't yet in cache. get() is a no-op
-            // (sync-resolved) for cache hits, and on a cache miss it reads from storage and writes the
-            // value back to cache. This is required so the subsequent cache.merge() merges the new delta
-            // into the real previous storage value (rather than starting from `undefined` and dropping
-            // the existing keys).
-            return Promise.all(existingKeys.map((key) => get(key))).then(() => {
+            // Pre-warm cache for any existing storage keys that aren't yet in cache so the subsequent
+            // cache.merge() merges the new delta into the real previous storage value (rather than
+            // starting from `undefined` and dropping the existing keys).
+            //
+            // Fast path: when every existingKey is already in cache, skip the pre-warm entirely. This
+            // preserves the original promise-chain depth and the subscriber-callback timing that
+            // dependent tests rely on.
+            //
+            // Slow path: when at least one existingKey is a cache miss, use multiGet — it batches the
+            // missing keys into a single Storage.multiGet call (vs. N parallel get() invocations) and
+            // writes the storage values back to cache before resolving.
+            const hasColdExistingKey = existingKeys.some((key) => !cache.hasCacheForKey(key));
+            const prewarmPromise = hasColdExistingKey ? multiGet(existingKeys) : Promise.resolve();
+            return prewarmPromise.then(() => {
                 // Snapshot previous values from the (now-warm) cache for keysChanged's diff, then update
                 // cache and notify subscribers synchronously BEFORE issuing storage writes. This matches
                 // the cache-first / storage-second invariant followed by every other Onyx write method
--- a/lib/OnyxUtils.ts
+++ b/lib/OnyxUtils.ts
@@ -347,9 +347,11 @@ function multiGet<TKey extends OnyxKey>(keys: CollectionKeyBase[]): Promise<Map<
             continue;
         }
 
-        const cacheValue = cache.get(key) as OnyxValue<TKey>;
-        if (cacheValue) {
-            dataMap.set(key, cacheValue);
+        // Use hasCacheForKey, not a truthy check on the value — otherwise cached falsy values
+        // (0, '', false, null) get treated as cache misses and re-fetched from storage, which
+        // can overwrite the warm cached value with a stale storage value via cache.merge().
+        if (cache.hasCacheForKey(key)) {
+            dataMap.set(key, cache.get(key) as OnyxValue<TKey>);
             continue;
         }
 
--- a/lib/OnyxUtils.ts
+++ b/lib/OnyxUtils.ts
@@ -402,6 +402,13 @@ function multiGet<TKey extends OnyxKey>(keys: CollectionKeyBase[]): Promise<Map<
                         }
                     }
 
+                    // Prefer cache over stale storage if a concurrent write populated it during
+                    // the read — otherwise cache.merge(temp) below would resurrect dropped fields.
+                    if (cache.hasCacheForKey(key)) {
+                        dataMap.set(key, cache.get(key) as OnyxValue<TKey>);
+                        continue;
+                    }
+
                     dataMap.set(key, value as OnyxValue<TKey>);
                     temp[key] = value as OnyxValue<TKey>;
                 }
--- a/lib/OnyxUtils.ts
+++ b/lib/OnyxUtils.ts
@@ -1611,7 +1611,14 @@ function mergeCollectionWithPatches<TKey extends CollectionKeyBase>(
             // missing keys into a single Storage.multiGet call (vs. N parallel get() invocations) and
             // writes the storage values back to cache before resolving.
             const hasColdExistingKey = existingKeys.some((key) => !cache.hasCacheForKey(key));
-            const prewarmPromise = hasColdExistingKey ? multiGet(existingKeys) : Promise.resolve();
+            // Swallow pre-warm read failures the same way the previous get()-based pre-warm did
+            // (see get() catch at the bottom of its definition). Without this, a transient
+            // Storage.multiGet rejection would skip cache.merge() + keysChanged() below and
+            // regress the cache-first invariant established in #787 — subscribers would miss
+            // the merge and Onyx.mergeCollection would reject up to the caller.
+            const prewarmPromise = hasColdExistingKey
+                ? multiGet(existingKeys).catch((err) => Logger.logInfo(`mergeCollectionWithPatches pre-warm failed; proceeding with cache-only merge. Error: ${err}`))
+                : Promise.resolve();
             return prewarmPromise.then(() => {
                 // Snapshot previous values from the (now-warm) cache for keysChanged's diff, then update
                 // cache and notify subscribers synchronously BEFORE issuing storage writes. This matches
--- a/lib/OnyxUtils.ts
+++ b/lib/OnyxUtils.ts
@@ -347,9 +347,8 @@ function multiGet<TKey extends OnyxKey>(keys: CollectionKeyBase[]): Promise<Map<
             continue;
         }
 
-        // Use hasCacheForKey, not a truthy check on the value — otherwise cached falsy values
-        // (0, '', false, null) get treated as cache misses and re-fetched from storage, which
-        // can overwrite the warm cached value with a stale storage value via cache.merge().
+        // hasCacheForKey catches cached falsy values (0, '', false, null) as cache hits, which
+        // a truthy check on the value would miss.
         if (cache.hasCacheForKey(key)) {
             dataMap.set(key, cache.get(key) as OnyxValue<TKey>);
             continue;
@@ -1611,11 +1610,9 @@ function mergeCollectionWithPatches<TKey extends CollectionKeyBase>(
             // missing keys into a single Storage.multiGet call (vs. N parallel get() invocations) and
             // writes the storage values back to cache before resolving.
             const hasColdExistingKey = existingKeys.some((key) => !cache.hasCacheForKey(key));
-            // Swallow pre-warm read failures the same way the previous get()-based pre-warm did
-            // (see get() catch at the bottom of its definition). Without this, a transient
-            // Storage.multiGet rejection would skip cache.merge() + keysChanged() below and
-            // regress the cache-first invariant established in #787 — subscribers would miss
-            // the merge and Onyx.mergeCollection would reject up to the caller.
+            // Swallow pre-warm read failures so a transient Storage.multiGet rejection doesn't
+            // skip the cache.merge() + keysChanged() below. Subscribers still see the merge even
+            // when storage reads fail.
             const prewarmPromise = hasColdExistingKey
                 ? multiGet(existingKeys).catch((err) => Logger.logInfo(`mergeCollectionWithPatches pre-warm failed; proceeding with cache-only merge. Error: ${err}`))
                 : Promise.resolve();
--- a/lib/OnyxUtils.ts
+++ b/lib/OnyxUtils.ts
@@ -1598,17 +1598,9 @@ function mergeCollectionWithPatches<TKey extends CollectionKeyBase>(
             // finalMergedCollection contains all the keys that were merged, without the keys of incompatible updates
             const finalMergedCollection = {...existingKeyCollection, ...newCollection};
 
-            // Pre-warm cache for any existing storage keys that aren't yet in cache so the subsequent
-            // cache.merge() merges the new delta into the real previous storage value (rather than
-            // starting from `undefined` and dropping the existing keys).
-            //
-            // Fast path: when every existingKey is already in cache, skip the pre-warm entirely. This
-            // preserves the original promise-chain depth and the subscriber-callback timing that
-            // dependent tests rely on.
-            //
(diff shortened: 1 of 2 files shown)


elirangoshen:
Addressing the [review on the predecessor PR (callstack-internal/react-native-onyx#5)](https://github.com/callstack-internal/react-native-onyx/pull/5#pullrequestreview-4361502804) — both points handled in this branch:

**1. `multiGet` falsy-value cache hit bug — fixed.** Aligned the inner check (`if (cacheValue)`) with the outer (`!cache.hasCacheForKey(key)`) so cached falsy values (`0`, `''`, `false`, `null`) are no longer re-fetched from storage and clobbered via `cache.merge(temp)`. New commit: `d66152a3 fix: align multiGet cache hit check with cache.hasCacheForKey`.

Added a regression test under `describe('multiGet cache hit consistency', ...)` that seeds cache with `0` and asserts `Storage.multiGet` / `Storage.getItem` are never called. Confirmed it fails on the pre-fix code (`multiGetSpy` is called once with `["test"]`).

**2. Unrelated reformat hunks in `tests/unit/onyxUtilsTest.ts` — dropped.** All 8 prettier-drift hunks above the new `mergeCollection pre-warm` describe block were noise from a different prettier version. Reset the file to `origin/main` and re-applied only the new describe block. Final diff vs main is now a single +171-line hunk for the new tests (the existing tests in the file are byte-identical to main).

Resulting 3-commit history vs `origin/main`:

```
d66152a3 fix: align multiGet cache hit check with cache.hasCacheForKey
d02382fd test: cover mergeCollectionWithPatches multiGet pre-warm fast/slow paths
32956378 perf: batch cold-cache pre-warm in mergeCollectionWithPatches via multiGet
```

Total: lib +28/−6, tests +209/−0. 457/457 unit tests green, typecheck clean, prettier clean.

lib/OnyxUtils.ts
```diff
@@ -1597,12 +1599,20 @@ function mergeCollectionWithPatches<TKey extends CollectionKeyBase>(
...
+            // Fast path: when every existingKey is already in cache, skip the pre-warm entirely. This
+            // preserves the original promise-chain depth and the subscriber-callback timing that
+            // dependent tests rely on.
+            //
+            // Slow path: when at least one existingKey is a cache miss, use multiGet — it batches the
+            // missing keys into a single Storage.multiGet call (vs. N parallel get() invocations) and
+            // writes the storage values back to cache before resolving.
+            const hasColdExistingKey = existingKeys.some((key) => !cache.hasCacheForKey(key));
+            const prewarmPromise = hasColdExistingKey ? multiGet(existingKeys) : Promise.resolve();
+            return prewarmPromise.then(() => {
```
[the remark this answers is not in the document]
elirangoshen: Good catch — applied Option A in `7fb447dc fix: preserve cache-first merge when multiGet pre-warm rejects`.

**Fix:** wrapped the slow-path call site in `.catch()` so a `Storage.multiGet` rejection logs via `Logger.logInfo` and resolves to undefined — `cache.merge()` + `keysChanged()` then run as if the pre-warm produced no data, matching the previous `get()`-based behavior where each `get()` swallows its own read errors.

```ts
const prewarmPromise = hasColdExistingKey
    ? multiGet(existingKeys).catch((err) => Logger.logInfo(`mergeCollectionWithPatches pre-warm failed; proceeding with cache-only merge. Error: ${err}`))
    : Promise.resolve();
```

Didn't change `multiGet` itself — other callers (`getCollectionDataAndSendAsObject`, `subscribeToKey`, perf tests) may legitimately rely on rejection visibility, and the failure semantics for the pre-warm specifically are well-defined: `cache.merge` starts from whatever's in cache, which is the same fallback the existing missing-key path already handles correctly.

**Regression test** in `tests/unit/onyxUtilsTest.ts` under `describe('mergeCollection pre-warm', ...)`:
- Seeds an existing collection member, evicts from cache.
- Connects a `waitForCollectionCallback: true` subscriber and flushes its initial data load before installing the mock (otherwise the connect's own no-catch multiGet call would consume the rejection and leak unrelated).
- Re-evicts to force the merge into the cold-key slow path.
- Mocks `Storage.multiGet` with `.mockRejectedValueOnce(transientError).mockImplementation(pristineMultiGet)`.
- Asserts `Onyx.mergeCollection` resolves (no outer rejection) AND subscriber receives the merged delta.

Confirmed it fails on pre-fix code with `Received: [Error: Transient IndexedDB read error]` on `expect(outerRejected).toBeNull()`.

fabioh8010:
i will review it soon

lib/OnyxUtils.ts
```diff
@@ -347,9 +347,11 @@ function multiGet<TKey extends OnyxKey>(keys: CollectionKeyBase[]): Promise<Map<
             continue;
         }
 
-        const cacheValue = cache.get(key) as OnyxValue<TKey>;
-        if (cacheValue) {
-            dataMap.set(key, cacheValue);
+        // Use hasCacheForKey, not a truthy check on the value — otherwise cached falsy values
+        // (0, '', false, null) get treated as cache misses and re-fetched from storage, which
+        // can overwrite the warm cached value with a stale storage value via cache.merge().
```
fabioh8010: Can we simplify this comment?

lib/OnyxUtils.ts
```diff
@@ -1597,12 +1599,27 @@ function mergeCollectionWithPatches<TKey extends CollectionKeyBase>(
...
+            //
+            // Slow path: when at least one existingKey is a cache miss, use multiGet — it batches the
+            // missing keys into a single Storage.multiGet call (vs. N parallel get() invocations) and
+            // writes the storage values back to cache before resolving.
+            const hasColdExistingKey = existingKeys.some((key) => !cache.hasCacheForKey(key));
+            // Swallow pre-warm read failures the same way the previous get()-based pre-warm did
+            // (see get() catch at the bottom of its definition). Without this, a transient
+            // Storage.multiGet rejection would skip cache.merge() + keysChanged() below and
+            // regress the cache-first invariant established in #787 — subscribers would miss
+            // the merge and Onyx.mergeCollection would reject up to the caller.
```
fabioh8010: Remove references to another PR, simplify or remove this comment

tests/unit/onyxUtilsTest.ts
```diff
@@ -920,6 +920,271 @@ describe('OnyxUtils', () => {
...
+            evictFromCache(memberKey);
+            await Onyx.mergeCollection(collectionKey, {
+                [memberKey]: delta,
+            } as GenericCollection);
+            const coldResult = OnyxCache.getCollectionData(collectionKey)?.[memberKey];
+
+            expect(warmResult).toEqual(coldResult);
+            expect(coldResult).toEqual({value: 'after', extra: 'kept'});
+        });
+        it('preserves cache-first invariant when Storage.multiGet rejects on the slow path', async () => {
```
fabioh8010, line 1090:
```suggestion

        it('preserves cache-first invariant when Storage.multiGet rejects on the slow path', async () => {
```

tests/unit/onyxUtilsTest.ts
```diff
@@ -920,6 +920,271 @@ describe('OnyxUtils', () => {
...
+            // Regression for codex review #3302454987 / PR #793. Before the .catch() at the
+            // pre-warm call site, a Storage.multiGet rejection would propagate up and skip
+            // cache.merge() + keysChanged() entirely — subscribers would miss the merge and
+            // Onyx.mergeCollection would reject. The previous get()-based pre-warm swallowed
+            // these errors per-key inside get()'s own .catch(), so the cache-first invariant
+            // from #787 held even on a flaky read.
```
fabioh8010: Simplify this by removing any mentions to other PRs or comments

fabioh8010:
@codex review

lib/OnyxUtils.ts
elirangoshen: Trimmed to a single line in 1764d2cb.

lib/OnyxUtils.ts
elirangoshen: done

tests/unit/onyxUtilsTest.ts
elirangoshen: done

tests/unit/onyxUtilsTest.ts
elirangoshen: done

lib/OnyxUtils.ts
```diff
@@ -1597,12 +1599,27 @@ function mergeCollectionWithPatches<TKey extends CollectionKeyBase>(
...
+            const prewarmPromise = hasColdExistingKey
+                ? multiGet(existingKeys).catch((err) => Logger.logInfo(`mergeCollectionWithPatches pre-warm failed; proceeding with cache-only merge. Error: ${err}`))
```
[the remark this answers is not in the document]
elirangoshen: https://github.com/Expensify/react-native-onyx/pull/793/changes#r3310600265

lib/OnyxUtils.ts
```diff
@@ -401,6 +402,13 @@ function multiGet<TKey extends OnyxKey>(keys: CollectionKeyBase[]): Promise<Map<
                         }
                     }
 
+                    // Prefer cache over stale storage if a concurrent write populated it during
+                    // the read — otherwise cache.merge(temp) below would resurrect dropped fields.
+                    if (cache.hasCacheForKey(key)) {
+                        dataMap.set(key, cache.get(key) as OnyxValue<TKey>);
+                        continue;
+                    }
```
elirangoshen, line 410: Hi @fabioh8010 here I addressed codex comment, this change might be better to do in a new pr, because it affect all multiGet calls, so I pushed it here to show the change I want to do, but I want to ask what you think if its ok to try it on this pr to open new one for that ?

fabioh8010 (approved):

lib/OnyxUtils.ts
fabioh8010: The change makes sense and I think it's ok to stay here

fabioh8010:
@elirangoshen could you test your recent changes in E/App again? Also, CI is failing

elirangoshen:
> @elirangoshen could you test your recent changes in E/App again? Also, CI is failing

I tested it seems ok. also the failing seems to be flake. maybe worth to run CI again 

Screenshot 2026-05-28 at 11 10 45

cristipaval:
Screenshot 2026-05-@elirangoshen at 10 53 41

@elirangoshen, could you please always link the fixed issue with the $ symbol? Otherwise, the automation breaks

cristipaval (approved):

cristipaval:
All checks are green now. @Julesssss, this is on you now

elirangoshen:
> Screenshot 2026-05-@elirangoshen at 10 53 41
> @elirangoshen, could you please always link the fixed issue with the $ symbol? Otherwise, the automation breaks

Sure, thank for the clarification I see that the PR template doesn't show that, maybe worth to update it there as well ?

cristipaval:
> I see that the PR template doesn't show that

oh, my bad, I just checked, and it looks like this repo is not wired to Melvin to add the assignees from the linked issue as reviewers. It randomly assigns and we are supposed to manually add specific reviewers

Julesssss (approved):

PR #816, opened 2026-07-29
Classify SQLite disk-pressure errors and stop retrying them

elirangoshen:
### Details

Fixes the app-side symptom of https://github.com/Expensify/App/issues/87869 — iOS bursts of `NitroSQLiteError: [NativeNitroSQLiteException][SqlExecutionError] disk I/O error` (Sentry APP-19Q).

**Root cause** (reproduced locally on an iOS simulator): when the device disk is full at database (re)open, SQLite fails to ftruncate the `OnyxDB-shm` WAL-index file to 32KB (errno 28 `ENOSPC` → `SQLITE_IOERR_SHMSIZE`), after which **every** read and write returns `disk I/O error` until space frees — then it recovers on its own. Two related shapes of the same condition:

- `unable to open database file` (`SQLITE_CANTOPEN`) — the `-shm` file cannot be created at all
- `cannot rollback - no transaction is active` — a batch-write failure whose original error was masked by the ROLLBACK itself failing (this one is a *mask*, not a cause — any error that already aborted the transaction produces it, so it deliberately stays `UNKNOWN`; nitro-sqlite stops the masking in 9.7.0)

**Before this PR** these errors fell in the `UNKNOWN` storage error class: 5 blind retries per operation (futile — the disk is still full) plus a per-operation alert, amplifying the very log storm being reported (~70x amplification seen in prod logs).

**This PR:**

- Adds a new `DISK_PRESSURE` storage error class (`lib/storage/errors.ts`)
- `classifySQLiteError` routes the two disk-pressure messages above to it (`lib/storage/providers/classifySQLiteError.ts`); the rollback message stays `UNKNOWN` (bounded retry + shape logging) since the real cause is unknowable
- `retryOperation` drops the write (the in-memory cache stays authoritative), skips retries and eviction (neither can free OS-level disk space), and logs a single throttled alert (60s window) plus one storage quota snapshot per burst (`lib/OnyxUtils.ts`)
- On native, the quota snapshot includes `getFreeDiskStorage()` bytes, giving prod telemetry the free-disk data to confirm disk pressure as the root cause

### Related Issues

https://github.com/Expensify/App/issues/97908

### Linked E/App PR

https://github.com/Expensify/App/pull/97944

### Automated Tests

Added to `tests/unit/onyxUtilsTest.ts`:

- `it.each` over the two disk-pressure error strings (`disk I/O error`, `unable to open database file`) asserting the operation is **not** retried
- A burst test asserting exactly one throttled alert + one storage quota snapshot across a burst of failing operations, and no "retries exhausted" alert

Full unit suite passes (569 tests).

### Manual Tests

Regression smoke (no user-facing change in this PR — it only alters error classification/logging):

1. Sign in and use the app normally: open a report, create an expense, send a message.
2. Go offline, create an expense, kill and relaunch the app — verify the offline change is still there and syncs when back online.
3. Verify no crashes or console errors during normal usage.

### Author Checklist

    - [x] Android / native
    - [x] Android / Chrome
    - [x] iOS / native
    - [x] iOS / Safari
    - [x] MacOS / Chrome / Safari

### Screenshots/Videos

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

https://github.com/user-attachments/assets/5766bfb1-ee87-4144-9f91-145308c9a8f5

</details>

### DIFF
--- a/lib/OnyxUtils.ts
+++ b/lib/OnyxUtils.ts
@@ -52,6 +52,16 @@ const METHOD = {
 // Max number of retries for failed storage operations
 const MAX_STORAGE_OPERATION_RETRY_ATTEMPTS = 5;
 
+/** Minimum interval between disk-pressure alerts. One disk-pressure burst fails every queued operation
+ * with the identical error, so per-operation logging would amplify the very storm it reports. */
+const DISK_PRESSURE_LOG_INTERVAL_MS = 60000;
+let lastDiskPressureLogTime = 0;
+
+/** Test-only: clears the disk-pressure log throttle so each test observes its own alert. */
+function resetDiskPressureLogThrottle(): void {
+    lastDiskPressureLogTime = 0;
+}
+
 type OnyxMethod = ValueOf<typeof METHOD>;
 
 // Key/value store of Onyx key and arrays of values to merge
@@ -783,6 +793,9 @@ function reportStorageQuota(error?: Error): Promise<void> {
  * - CAPACITY: evicts the least recently accessed evictable key and retries, under a session-level
  *   circuit breaker (see lib/StorageCircuitBreaker.ts) that halts the loop once eviction stops making
  *   progress or failures storm — the per-operation budget alone cannot stop a session-wide storm.
+ * - DISK_PRESSURE: the device disk itself is full (or the database files are unreadable), so neither
+ *   retries nor in-DB eviction can free space — the write is dropped (cache stays authoritative) with
+ *   a single throttled alert + quota snapshot per burst.
  * - UNKNOWN: the provider couldn't classify it — log the full error shape (name + message +
  *   provider) once so it's visible, then bounded retry without eviction.
  */
@@ -804,6 +817,21 @@ function retryOperation<TMethod extends RetriableOnyxOperation>(
     if (errorClass === StorageErrorClass.CAPACITY && !StorageCircuitBreaker.isAllowed()) {
         StorageCircuitBreaker.recordProbeFailure();
         return Promise.resolve();
+    }
+
+    // DISK_PRESSURE: the filesystem around the database is out of space or its files are unreadable, so
+    // neither retries nor in-DB eviction can succeed — every operation fails identically until the OS
+    // frees space, and the burst recovers on its own once it does. Drop the write (the cache stays
+    // authoritative) and log one alert + quota snapshot per interval instead of one per operation; the
+    // quota snapshot carries the free-disk bytes that confirm (or rule out) disk pressure in telemetry.
+    if (errorClass === StorageErrorClass.DISK_PRESSURE) {
+        const now = Date.now();
+        if (now - lastDiskPressureLogTime < DISK_PRESSURE_LOG_INTERVAL_MS) {
+            return Promise.resolve();
+        }
+        lastDiskPressureLogTime = now;
+        Logger.logAlert(`Disk-pressure storage error; skipping retries. provider: ${Storage.getStorageProvider().name}. message: ${error?.message}. onyxMethod: ${onyxMethod.name}.`);
+        return reportStorageQuota(error);
     }
 
     Logger.logInfo(
@@ -1816,6 +1844,7 @@ const OnyxUtils = {
     getCollectionDataAndSendAsObject,
     remove,
     reportStorageQuota,
+    resetDiskPressureLogThrottle,
     retryOperation,
     broadcastUpdate,
     hasPendingMergeForKey,
--- a/lib/OnyxUtils.ts
+++ b/lib/OnyxUtils.ts
@@ -819,11 +819,9 @@ function retryOperation<TMethod extends RetriableOnyxOperation>(
         return Promise.resolve();
     }
 
-    // DISK_PRESSURE: the filesystem around the database is out of space or its files are unreadable, so
-    // neither retries nor in-DB eviction can succeed — every operation fails identically until the OS
-    // frees space, and the burst recovers on its own once it does. Drop the write (the cache stays
-    // authoritative) and log one alert + quota snapshot per interval instead of one per operation; the
-    // quota snapshot carries the free-disk bytes that confirm (or rule out) disk pressure in telemetry.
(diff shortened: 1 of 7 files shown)


lib/storage/providers/classifySQLiteError.ts
```diff
@@ -18,6 +18,13 @@ function classifySQLiteError(error: unknown): ValueOf<typeof StorageErrorClass>
         return StorageErrorClass.CAPACITY;
     }
 
+    // Full-disk failures around the database files: SQLITE_IOERR (cannot size the -shm file on reopen,
+    // fails reads too), SQLITE_CANTOPEN (cannot create it), and a failed ROLLBACK masking the original
+    // write error. None can succeed until the OS frees space.
+    if (message.includes('disk i/o error') || message.includes('unable to open database file') || message.includes('cannot rollback - no transaction is active')) {
```
sosek108: Are you sure that `cannot rollback - no transaction is active` is 100% related to disk pressure?

lib/storage/errors.ts
```diff
@@ -15,6 +15,10 @@ const StorageErrorClass = {
     TRANSIENT: 'transient',
     /** Quota exceeded / disk full. Owner: operation layer — evict and retry. */
     CAPACITY: 'capacity',
+    /** Filesystem-level failure around the database files (device disk full, or the files cannot be
+     *  created/read). Owner: operation layer — skip retries and eviction (neither can free OS-level
+     *  space) and log one throttled alert + quota snapshot per burst. */
+    DISK_PRESSURE: 'diskPressure',
```
sosek108, line 21: How this is different from `CAPACITY`?

lib/OnyxUtils.ts
```diff
@@ -806,6 +819,19 @@ function retryOperation<TMethod extends RetriableOnyxOperation>(
...
+    // interval — the snapshot's free-disk bytes let telemetry confirm (or rule out) disk pressure.
+    if (errorClass === StorageErrorClass.DISK_PRESSURE) {
+        const now = Date.now();
+        if (now - lastDiskPressureLogTime < DISK_PRESSURE_LOG_INTERVAL_MS) {
+            return Promise.resolve();
+        }
+        lastDiskPressureLogTime = now;
+        Logger.logAlert(`Disk-pressure storage error; skipping retries. provider: ${Storage.getStorageProvider().name}. message: ${error?.message}. onyxMethod: ${onyxMethod.name}.`);
+        return reportStorageQuota(error);
+    }
```
sosek108, line 833: The goal is not to send so many logs to VL, right? 

This is why we introduced `CircuitBreaker`.

sosek108 (commented):
Got some comments,

lib/storage/providers/classifySQLiteError.ts
elirangoshen: its not 100 percent so I edit the comments to reflect it better

lib/storage/errors.ts
elirangoshen: their both related to disk full but `CAPACITY` usually means the onyx keys is full  and evict some keys can fix while in `DISK_PRESSURE` its when os file system is full so there is no way to fix it unless free space.

lib/OnyxUtils.ts
elirangoshen: They solve different problems. The circuit breaker stops the evict → retry loop for `CAPACITY` errors. `DISK_PRESSURE` has no loop to stop — we never retry or evict, the write is just dropped. The only thing left to limit is log volume, and the 60s throttle does that. Putting it through the breaker wouldn't work anyway: the breaker's recovery probe is an eviction, which can't fix a full disk.

sosek108 (approved):

lib/storage/providers/classifySQLiteError.ts
fabioh8010: Should we remove it then? Looks more like a generic error than disk-specific

elirangoshen: removed

fabioh8010 (approved):

lib/storage/providers/classifySQLiteError.ts
```diff
@@ -18,6 +18,15 @@ function classifySQLiteError(error: unknown): ValueOf<typeof StorageErrorClass>
...
 
+    // Full-disk failures around the database files: SQLITE_IOERR (cannot size the -shm file on reopen,
+    // fails reads too) and SQLITE_CANTOPEN (cannot create it). Neither can succeed until the OS frees space.
+    if (message.includes('disk i/o error') || message.includes('unable to open database file')) {
+        return StorageErrorClass.DISK_PRESSURE;
+    }
+
+    // "cannot rollback - no transaction is active" lands here deliberately: it is a mask, not a cause —
+    // nitro-sqlite's rollback-on-error replaced the original failure until 9.7.0. UNKNOWN gives it
+    // bounded retry + shape logging instead of a disk-pressure guess.
```
fabioh8010: Can we remove this?

lib/storage/providers/SQLiteProvider.ts
```diff
@@ -325,16 +325,21 @@ const provider: StorageProvider<NitroSQLiteConnection | undefined> = {
             throw new Error('Store is not initialized!');
         }
 
-        return Promise.all([provider.store.executeAsync<PageSizeResult>('PRAGMA page_size;'), provider.store.executeAsync<PageCountResult>('PRAGMA page_count;'), getFreeDiskStorage()]).then(
-            ([pageSizeResult, pageCountResult, bytesRemaining]) => {
+        // The PRAGMAs run on the SQLite connection — the very thing that's broken during disk pressure —
+        // while getFreeDiskStorage() is filesystem-level and still works. Degrade bytesUsed to -1 instead
+        // of failing the snapshot, so the free-disk bytes (the signal that matters) always get logged.
```
fabioh8010: Another comment that could be simplified e.g. "the very thing that's broken during disk pressure" is useless

elirangoshen: I simlified it

inimaga (approved):

PR #818, opened 2026-08-05
Skip keyless update entries in updateSnapshots instead of crashing

elirangoshen:
### Details

Came from [this slack thread](https://callstack-hq.slack.com/archives/C08CZDJFJ77/p1785928288177769)

Fixes a production crash in `OnyxUtils.updateSnapshots`: `TypeError: Cannot read properties of undefined (reading 'startsWith')` ([Sentry APP-6NT](https://expensify.sentry.io/issues/APP-6NT), ~94 events / ~35 users since 2026-03-01, plus duplicate groupings APP-JT0, APP-961, APP-H8X, APP-ED8 — all on web, mostly 2FA pages).

**Root cause:** `Onyx.update` validates its entries and skips any with a non-string `key` (and `clear`/`multiset` entries legitimately carry no key at all), but then passes the **unfiltered** update array to `OnyxUtils.updateSnapshots`. There, `OnyxKeys.isCollectionMemberKey(snapshotCollectionKey, key)` calls `key.startsWith(collectionKey)` with no guard, so any keyless entry throws and rejects the whole `Onyx.update` promise. It only reproduces when the snapshot collection is non-empty (i.e. the user has cached search snapshots), which is why the volume is low.

In Expensify/App this surfaced as an unhandled rejection from `QueuedOnyxUpdates.flushQueue` (the deferred WRITE flush), and Sentry mis-attributed the newest grouping (APP-JT0) to Expensify/App PR [#97098](https://github.com/Expensify/App/pull/97098), which merely touched those lines.

**Fix:** in `updateSnapshots`, skip entries where `typeof key !== 'string'` before calling `isCollectionMemberKey` — mirroring the per-entry validation `Onyx.update` already performs.

Fixes APP-6NT

https://github.com/Expensify/App/pull/97857 (TESTING ONLY pin of this PR's HEAD, so the full E/App CI suite runs against it). The fix itself is self-contained in Onyx and requires no App-side change; for context, https://github.com/Expensify/App/pull/97098 is the App PR that Sentry falsely attributed the newest crash grouping to.

### Related Issues

https://github.com/Expensify/App/issues/98438

### Linked E/App PR

https://github.com/Expensify/App/pull/97857

### Automated Tests

Added a regression test in `tests/unit/onyxTest.ts` — "should skip update entries without a key when updating Snapshots instead of rejecting": dispatches an `Onyx.update` containing a keyless `merge` entry alongside a valid one while a snapshot is cached. Without the fix it reproduces the exact production crash (`TypeError: Cannot read properties of undefined (reading 'startsWith')` rejecting `Onyx.update`); with the fix, `Onyx.update` resolves and the valid entry still lands in the snapshot.

Verification run: 173 unit tests pass (`onyxTest` + `onyxUtilsTest`), `tsc --noEmit` clean, ESLint clean, Prettier clean. Also verified the new test fails with the exact production stack trace when the guard is stashed.

### Manual Tests

On Android / iOS (native or mWeb) against staging:

1. Sign in with an account that has chats and expenses.
2. Go to the **Reports/Search** tab and run a search (e.g. All expenses). Scroll through the results and open one or two of them. *(This caches search snapshot data — the precondition for the code path this PR touches.)*
3. Go back to **Inbox** and, in a report that appeared in those search results: send a message, edit an expense amount, and mark the report as unread.
4. Return to the **Search** tab and re-open the search. Verify the results reflect the changes (new amount, latest message) and the app does not crash or freeze at any point.
5. Enable airplane mode. Make 2–3 more changes (add a comment, rename a report, edit an expense). Disable airplane mode and wait for sync to finish. Verify the app stays responsive, nothing crashes, and search results reflect the offline changes.
6. Sign out and sign back in. Verify no crash during sign-out/sign-in.

Expected: no crash and no stuck/undelivered updates in any step. (Before this fix, a malformed keyless entry in any server update batch would crash the app with a `startsWith` TypeError whenever a search had previously been run — steps 2–5 keep that code path continuously active.)

<details>
<summary>Deterministic developer repro (web console)</summary>

1. In Expensify/App (web), sign in and run a Search so at least one `snapshot_` collection entry is cached.
2. In the dev console, dispatch an update containing a keyless entry alongside a valid one, e.g. `Onyx.update([{onyxMethod: 'merge', key: 'session', value: {}}, {onyxMethod: 'merge', value: {}}])`.
3. Without this fix: the promise rejects with `TypeError: Cannot read properties of undefined (reading 'startsWith')` from `updateSnapshots`.
4. With this fix: the promise resolves, the keyless entry is skipped, and the valid entry is applied to the cached snapshot.

</details>

### QA Steps

Same as Manual Tests.

### Author Checklist

    - [x] Android / native
    - [x] Android / Chrome
    - [x] iOS / native
    - [x] iOS / Safari
    - [x] MacOS / Chrome / Safari

### Screenshots/Videos

<details>

N/A — non-UI library fix, covered by the unit regression test.

</details>

<details>

N/A — non-UI library fix, covered by the unit regression test.

</details>

<details>

https://github.com/user-attachments/assets/fc78d56f-bcbd-492e-8033-d8f9d116f595

</details>

<details>

N/A — non-UI library fix, covered by the unit regression test.

</details>

<details>

N/A — non-UI library fix, covered by the unit regression test.

</details>

### DIFF
--- a/lib/OnyxUtils.ts
+++ b/lib/OnyxUtils.ts
@@ -1211,6 +1211,12 @@ function updateSnapshots<TKey extends OnyxKey>(data: Array<OnyxUpdate<TKey>>, me
         let updatedData: Record<string, unknown> = {};
 
         for (const {key, value} of data) {
+            // `clear`/`multiset` updates legitimately carry no key (and `Onyx.update` skips other keyless
+            // entries), so ignore them here too instead of crashing on `key.startsWith` below.
+            if (typeof key !== 'string') {
+                continue;
+            }
+
             // snapshots are normal keys so we want to skip update if they are written to Onyx
             if (OnyxKeys.isCollectionMemberKey(snapshotCollectionKey, key)) {
                 continue;
--- a/lib/OnyxUtils.ts
+++ b/lib/OnyxUtils.ts
@@ -1202,6 +1202,15 @@ function updateSnapshots<TKey extends OnyxKey>(data: Array<OnyxUpdate<TKey>>, me
 
     const snapshotCollection = getCachedCollection(snapshotCollectionKey);
 
+    // Multiset entries are keyless but update every key of their payload, so expand them into
+    // per-key entries to keep cached snapshots in sync with the real Onyx data.
+    const flattenedData = data.flatMap<{onyxMethod: OnyxMethod; key: unknown; value?: unknown}>((entry) => {
+        if (entry.onyxMethod === METHOD.MULTI_SET && typeof entry.key !== 'string' && entry.value && typeof entry.value === 'object' && !Array.isArray(entry.value)) {
+            return Object.entries(entry.value).map(([key, value]) => ({onyxMethod: METHOD.SET, key, value}));
+        }
+        return entry;
+    });
+
     for (const [snapshotEntryKey, snapshotEntryValue] of Object.entries(snapshotCollection)) {
         // Snapshots may not be present in cache. We don't know how to update them so we skip.
         if (!snapshotEntryValue) {
@@ -1210,11 +1219,11 @@ function updateSnapshots<TKey extends OnyxKey>(data: Array<OnyxUpdate<TKey>>, me
 
         let updatedData: Record<string, unknown> = {};
 
-        for (const {key, value, onyxMethod} of data) {
+        for (const {key, value, onyxMethod} of flattenedData) {
             if (typeof key !== 'string') {
-                // clear/multiset entries legitimately carry no key; snapshots have nothing to update for them
+                // clear entries legitimately carry no key, and malformed multiset payloads are already logged by update() itself
                 if (onyxMethod !== METHOD.CLEAR && onyxMethod !== METHOD.MULTI_SET) {
-                    Logger.logInfo(`Invalid ${typeof key} key (method: ${onyxMethod}, key: ${String(key).slice(0, 50)}) provided in Onyx update. Skipping snapshot update for this entry.`);
+                    Logger.logHmmm(`Invalid ${typeof key} key (method: ${onyxMethod}, key: ${String(key).slice(0, 50)}) provided in Onyx update. Skipping snapshot update for this entry.`);
                 }
                 continue;
             }
--- a/lib/OnyxUtils.ts
+++ b/lib/OnyxUtils.ts
@@ -1210,9 +1210,12 @@ function updateSnapshots<TKey extends OnyxKey>(data: Array<OnyxUpdate<TKey>>, me
 
         let updatedData: Record<string, unknown> = {};
 
-        for (const {key, value} of data) {
+        for (const {key, value, onyxMethod} of data) {
             if (typeof key !== 'string') {
-                Logger.logInfo(`Invalid ${typeof key} key provided in Onyx update. Key must be of type string. Skipping snapshot update for this entry.`);
+                // clear/multiset entries legitimately carry no key; snapshots have nothing to update for them
+                if (onyxMethod !== METHOD.CLEAR && onyxMethod !== METHOD.MULTI_SET) {
+                    Logger.logInfo(`Invalid ${typeof key} key (method: ${onyxMethod}, key: ${String(key).slice(0, 50)}) provided in Onyx update. Skipping snapshot update for this entry.`);
+                }
                 continue;
             }
 
--- a/lib/OnyxUtils.ts
+++ b/lib/OnyxUtils.ts
@@ -1212,6 +1212,7 @@ function updateSnapshots<TKey extends OnyxKey>(data: Array<OnyxUpdate<TKey>>, me
 
         for (const {key, value} of data) {
             if (typeof key !== 'string') {
+                Logger.logInfo(`Invalid ${typeof key} key provided in Onyx update. Key must be of type string. Skipping snapshot update for this entry.`);
                 continue;
             }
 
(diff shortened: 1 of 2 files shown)


lib/OnyxUtils.ts
```diff
@@ -1211,6 +1211,12 @@ function updateSnapshots<TKey extends OnyxKey>(data: Array<OnyxUpdate<TKey>>, me
         let updatedData: Record<string, unknown> = {};
 
         for (const {key, value} of data) {
+            // `clear`/`multiset` updates legitimately carry no key (and `Onyx.update` skips other keyless
+            // entries), so ignore them here too instead of crashing on `key.startsWith` below.
```
fabioh8010: Unnecessary comment imo

tests/unit/onyxTest.ts
```diff
@@ -1535,6 +1535,35 @@ describe('Onyx', () => {
...
+
+        Onyx.connect({
+            key: ONYX_KEYS.COLLECTION.SNAPSHOT,
+            callback,
+        });
+
+        await waitForPromisesToResolve();
+
+        // A keyless entry (e.g. a malformed server update) used to crash updateSnapshots with
+        // "can't access property 'startsWith', key is undefined" and reject the whole update.
```
fabioh8010: Same

lib/OnyxUtils.ts
elirangoshen: removed

tests/unit/onyxTest.ts
elirangoshen: removed

lib/OnyxUtils.ts
```diff
@@ -1211,6 +1211,10 @@ function updateSnapshots<TKey extends OnyxKey>(data: Array<OnyxUpdate<TKey>>, me
         let updatedData: Record<string, unknown> = {};
 
         for (const {key, value} of data) {
+            if (typeof key !== 'string') {
+                continue;
```
VickyStash, line 1215: Please, add Log message, same as it's done in Onyx.update

lib/OnyxUtils.ts
```diff
@@ -1211,6 +1211,11 @@ function updateSnapshots<TKey extends OnyxKey>(data: Array<OnyxUpdate<TKey>>, me
         let updatedData: Record<string, unknown> = {};
 
         for (const {key, value} of data) {
+            if (typeof key !== 'string') {
+                Logger.logInfo(`Invalid ${typeof key} key provided in Onyx update. Key must be of type string. Skipping snapshot update for this entry.`);
```
LukasMod: NAB: Maybe its worth to log something extra here instead of only typeof key? `onyxMethod`? Something to help with debugging later? Trimmed String(key) maybe?

lib/OnyxUtils.ts
elirangoshen: Done

lib/OnyxUtils.ts
elirangoshen: good idea, done

Valforte:
@roryabraham could you review this PR? I'm not familiar with this code or what it's trying to achieve

lib/OnyxUtils.ts
```diff
@@ -1210,7 +1210,15 @@ function updateSnapshots<TKey extends OnyxKey>(data: Array<OnyxUpdate<TKey>>, me
 
         let updatedData: Record<string, unknown> = {};
 
-        for (const {key, value} of data) {
+        for (const {key, value, onyxMethod} of data) {
+            if (typeof key !== 'string') {
+                // clear/multiset entries legitimately carry no key; snapshots have nothing to update for them
+                if (onyxMethod !== METHOD.CLEAR && onyxMethod !== METHOD.MULTI_SET) {
+                    Logger.logInfo(`Invalid ${typeof key} key (method: ${onyxMethod}, key: ${String(key).slice(0, 50)}) provided in Onyx update. Skipping snapshot update for this entry.`);
```
roryabraham:
```suggestion
                    Logger.logWarn(`Invalid ${typeof key} key (method: ${onyxMethod}, key: ${String(key).slice(0, 50)}) provided in Onyx update. Skipping snapshot update for this entry.`);
```

lib/OnyxUtils.ts
```diff
@@ -1210,7 +1210,15 @@ function updateSnapshots<TKey extends OnyxKey>(data: Array<OnyxUpdate<TKey>>, me
...
+                }
+                continue;
```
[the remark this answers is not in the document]
roryabraham: ok, so this is not a _regression_ with this code, so we could choose to handle it in a follow-up. What we have here is strictly better than what we have on main, which errors in this case.

That said, we probably _should_ update `updateSnapshots` to work with `multiSet` and `clear`, which would be strictly better than just avoiding a crash. Right @elirangoshen ?

roryabraham (changes requested):

lib/OnyxUtils.ts
elirangoshen: there's no `logWarn` in onyx so I added `logHmmm` which should do the same thing

elirangoshen:
@roryabraham  Done for multiSet — the payload keys are now expanded into per-key entries, so snapshots get updated the same way the real data does. Added a test for it.

For `clear` I left the skip as is, on purpose: `Onyx.clear()` deletes the snapshot keys themselves (they're regular Onyx keys with no default), so there's nothing left to update. And since the snapshot merges run *after* the clear promise resolves, "syncing" snapshots on clear would just recreate the snapshot key with pre-clear data — resurrecting exactly the data clear is supposed to delete (e.g. user data lingering after sign-out).

What do you think ?

lib/OnyxUtils.ts
roryabraham: NAB, but for future reference `logWarn` does more than `logHmmm`. If we see a certain threshold of warnings logged in our back-end, we'll automatically create a GitHub issue.

roryabraham (approved):

PR #819, opened 2026-08-05
fix: add getSnapshot memoizedShallowEqual to skip rerenders

LukasMod:
### Details

`useOnyx` no longer changes its result identity when the shared snapshot cache holds a content-equal result from another subscriber

The snapshot cache slot is shared by every subscriber of the same `(key, selector)` pair, while each subscriber's memoized selector owns a distinct output object. A subscriber that mounts later (its per-hook memoization cache is empty) computes a fresh object, content-equal but referentially new, and publishes it into the slot. Existing subscribers then adopted it blindly in the `getSnapshot()` fast path, which changed their result identity and re-rendered their memoized subtrees for no reason. The fast path now keeps the hook's own result when the cached one is content-equal (`memoizedShallowEqual` on the value plus a status match).

### Why

We hit this in E/App. In a Concierge chat, every sent message and every Concierge reply re-rendered the whole `ReportActionsList` content. React DevTools Profiler blamed the `report` prop of `ReportActionsListItemRenderer`, and prop-diff instrumentation showed the tell-tale signature on each message: `report prop (stableReport <id>): NEW REFERENCE, but every field is shallow-equal`. A pure object-identity change with zero data change.

Root cause: every report row (`ActionContentRouter`) and the report list itself subscribe to the same `report_<id>` key with the same shared `getStableReportSelector`. Every incoming message mounts a new row, which republished a fresh identity into the shared slot. That flipped the `report` prop of every `ReportActionsListItemRenderer` and re-rendered the entire chat list on each message, even though the selected data never changed. The stable-report projection exists precisely to prevent this class of re-render, and the shared slot was silently defeating it.

### Where the fresh object comes from

A projection selector builds a fresh object literal on every call (`return {reportID: report.reportID, ...}`), so two calls with the same input produce two content-equal objects with different identities. Identity stability is never provided by the selector itself. It comes from the per-hook memoized wrapper (`createMemoizedSelector`), which deep-compares each recompute against its `lastOutput` and returns the old reference when equal. A newly mounted subscriber has no `lastOutput` yet, so the fresh literal becomes its output and gets published into the shared slot, where the other subscribers' fast paths adopted it.

### Why a per-call-site selector wrapper also "fixes" it (and why it's not the fix)

The slot key is `${key}_${selectorID}`, where `selectorID` is assigned per function identity. Wrapping the shared selector in a module-level function (`const stableReportSelectorForList = (report) => getStableReportSelector(report)`) mints a new identity, giving that call site a private slot with a single writer and reader, which is the always-working sole-subscriber case. But that only shields one call site. All remaining subscribers of the shared selector keep flipping each other in their shared slot. The fast-path guard fixes the adoption itself, making sharing safe for every consumer without per-call-site workarounds. (The wrapper must be module-level. Defined inside a component it would mint a new identity every render, causing cache-slot churn and a selector recompute per render.)

### Why this implementation

- Guard only the fast path. The recompute path already preserves identity via the `memoizedShallowEqual` check on `previousValueRef`. The fast path was the single place adopting a result without any comparison.
- Near-zero cost. Converged and single-subscriber hits short-circuit on the first `!==` pointer compare. Distinct-but-equal objects pay one top-level walk, memoized by identity pair in the `memoizedShallowEqual` WeakMap, so N hooks comparing the same two objects pay for one walk total. Real content changes arrive via slot invalidation (recompute path), so the guard rarely sees them.
- Status must match too. A cached `loaded` result never masks this hook's `loading` state (and vice versa). Value equality alone isn't enough to skip adoption.
- Regression test reproduces the exact failure: an existing subscriber's result identity must survive a new subscriber mounting with the same key and selector. It fails on the previous code at the identity assertion and passes with the guard. The full `useOnyx` suite is unaffected.

Before:

concierge baseline

After:

concierge fixed

| Scenario | Before | After | Change |
| --- | ---: | ---: | ---: |
| ManualSendMessage - Concierge chat (narrow web) | 301 ms | 237 ms | -21.1% |
| ManualSendMessage - DM test | 465 ms | 336 ms | -27.7% |

### Related Issues

https://github.com/Expensify/App/issues/95584

### Linked E/App PR
 
https://github.com/Expensify/App/pull/97952

### Manual Tests

No visible changes. Smoke tests:

#### Test 1: Send messages in open report
1. As account A, open Report A
2. Send 5 messages in quick succession
3. Confirm each message appears
4. Confirm the LHN row for Report A shows the last sent message as its preview

#### Test 2: Rapid report switching
1. As account A, switch between Report A and Report B 5 times quickly
2. Confirm each report shows its own content immediately
3. Confirm no report gets stuck in a loading skeleton
4. Send a message mid-switching and confirm it lands in the right report

#### Test 3: New subscriber mounts on an already-open report
1. As account A, open Report A and note its content
2. Open the report details page (header click), then a thread on any message
3. Close both and return to the report
4. As account B, send a message to Report A
5. Confirm the message appears in the open report and the LHN preview updates

#### Test 4: Report data change reaches all subscribers
1. As account A, open a group chat or workspace room
2. Rename the room (or update its description)
3. Confirm the header, the report screen, and the LHN row all show the new name
4. Pin and unpin the report and confirm the LHN reorders both times

#### Test 5: Receive messages while report is open
1. As account A, keep Report A open
2. As account B, send 3 messages to Report A, one with an edit right after sending
3. Confirm all messages and the edit appear for account A in real time
4. As account B, add a reaction to the last message and confirm it appears for account A

#### Test 6: Fresh key and loading transitions
1. As account A, sign out and sign back in
2. Open a chat that was never o
3. Confirm the skeleton is replaced by content and never gets stuck on loading
4. Start a brand-new DM with a new user and send a message
5. Confirm the new chat renders and appears in the LHN

### Author Checklist

    - [x] Android / native
    - [x] Android / Chrome
    - [x] iOS / native
    - [x] iOS / Safari
    - [x] MacOS / Chrome / Safari

### Screenshots/Videos

Android: Native

Android: mWeb 

ChromeiOS: Native

iOS: mWeb Safari

MacOS: Chrome / Safari

Test 1-2:

https://github.com/user-attachments/assets/667cf6e3-b92c-42c6-956d-8c723d56a5a2

Test 3:

https://github.com/user-attachments/assets/cae34314-a866-4f26-8c29-9db5e3fdc8b3

Test 4:

https://github.com/user-attachments/assets/56efff9d-890b-4b6b-bcdf-500ba7c2266d

Test 5:

https://github.com/user-attachments/assets/15f5b04e-d287-408c-af46-54232d2055e0

Test 6:

https://github.com/user-attachments/assets/67020400-6798-4813-8ca7-fa7fee34935c

### DIFF
--- a/lib/useOnyx.ts
+++ b/lib/useOnyx.ts
@@ -108,6 +108,12 @@ function useOnyx<TKey extends OnyxKey, TReturnValue = OnyxValue<TKey>>(key: TKey
         if (!shouldGetCachedValueRef.current) {
             const cachedResult = onyxSnapshotCache.getCachedResult<UseOnyxResult<TReturnValue>>(key, cacheKey);
             if (cachedResult !== undefined) {
+                // The slot is shared by all subscribers of the same (key, selector) pair, so it can hold a content-equal
+                // result computed by another subscriber. Keep our own result then, otherwise we would needlessly change
+                // this hook's result identity and re-render its consumer.
+                if (cachedResult !== resultRef.current && memoizedShallowEqual(cachedResult[0], resultRef.current[0]) && cachedResult[1].status === resultRef.current[1].status) {
+                    return resultRef.current;
+                }
                 resultRef.current = cachedResult;
                 return cachedResult;
             }
--- a/tests/unit/useOnyxTest.ts
+++ b/tests/unit/useOnyxTest.ts
@@ -574,6 +574,28 @@ describe('useOnyx', () => {
             expect(oldResult).toBe(result.current);
         });
 
+        it('should keep result identity when a new subscriber with the same key and selector mounts', async () => {
+            Onyx.set(ONYXKEYS.TEST_KEY, {id: 'test_id', name: 'test_name'});
+
+            const selector = ((entry: OnyxEntry<{id: string; name: string}>) => ({id: entry?.id})) as UseOnyxSelector<OnyxKey, {id?: string}>;
+
+            const {result, rerender} = renderHook(() => useOnyx(ONYXKEYS.TEST_KEY, {selector}));
+
+            await act(async () => waitForPromisesToResolve());
+
+            const oldResult = result.current;
+
+            // A subscriber that mounts later computes its own content-equal selector output and publishes it into the shared snapshot cache slot.
+            renderHook(() => useOnyx(ONYXKEYS.TEST_KEY, {selector}));
+
+            await act(async () => waitForPromisesToResolve());
+
+            rerender(undefined);
+
+            // must be the same reference — the new subscriber's content-equal result must not replace it
+            expect(result.current).toBe(oldResult);
+        });
+
         it('should always use the current selector reference to return new data', async () => {
             Onyx.set(ONYXKEYS.TEST_KEY, {id: 'test_id', name: 'test_name'});
 


LukasMod:
@codex review

arosiclair (approved):

Julesssss (approved):
Should we run wider tests for this change?

LukasMod:
@Julesssss Do you mean I should add more tests, or some general smoke QA tests? If there’s an option for wider test coverage, that would be great. 

The issue with this could be that we might rely on those unnecessary rerenders without realizing it, but that should be fixed right there instead

Julesssss:
Some additional tests here would be great, wider testing can wait for the full QA regression suite -it'll take a couple of releases due to the merge queue

LukasMod:
@Julesssss  I added 4 new tests that could could potentially show some stale content when multiple subscribers use same keys. PR is rebased also

cristipaval (approved):

PR #820, opened 2026-08-10
Batch collection member removals and coalesce cross-tab sync events

elirangoshen:
### Details

Fixes the two tabs with a heavy account crash both tabs

**Root cause.** A collection write (`mergeCollection`/`multiSet`/`setCollection`/`partialSetCollection`) persists its non-null members in one batched `multiMerge`/`multiSet` call (one cross-tab `SYNC_ONYX` event), but every `null` member went through `remove(key)` individually — one storage call and one cross-tab event **per removed key**, with no check that the key even exists. On a heavy account, OpenApp's `transactionViolations_` mergeCollection contains ~2,000 `null` members on **every boot** (keys that were never stored, so the removals are pure no-ops). Each refresh therefore flooded the other tab with ~2,000 single-key events; each event is its own macrotask, so the receiving tab ran one `keysChanged` → one derived-value recompute (`reportAttributes`, `reportTransactionsAndViolations`) → one `Onyx.set` of an MB-sized derived value **per event**, saturating the main thread for ~1 minute and broadcasting ~4,000 derived-key events back to the first tab, which then performed ~4,000 IndexedDB reads of those MB-sized values. On Applause-sized accounts both tabs run out of memory and crash.

**Fix (two layers):**

1. **Sender — batch and skip removals.** `prepareKeyValuePairsForStorage` no longer removes null keys as a side effect; it returns them as `keysToRemove`. All four collection-write paths now:
   - skip removals that are complete no-ops (key neither cached nor persisted);
   - fold real removals into the existing single `keysChanged()` batch (as `undefined`, with previous values snapshotted for the diff);
   - delete them from storage with **one** `Storage.removeItems()` call, which raises **one** cross-tab event for the whole batch.
2. **Receiver — coalesce event floods.** `InstanceSync` (web) now buffers incoming `SYNC_ONYX` storage events and flushes once per tick (single `multiGet` + single dispatch). A tab still running an older bundle during a deploy emits one event per key; the flush timer is scheduled on the first event of the burst, so every event already in the task queue joins one batch instead of re-running the notification pipeline per key.

**Measured on a heavy (Applause) account, two tabs, per boot:** cross-tab events 1,971 → 41 (zero per-key `transactionViolations_` events); derived-value writes in the other tab 3,891 → 8; no crash; pin/draft/read-unread/message sync, a 19.5 MB state import, and Clear cache and restart all work across tabs without refresh.

### Related Issues

https://github.com/Expensify/App/issues/94839

### Linked E/App PR

https://github.com/Expensify/App/pull/98121

### Automated Tests

- `tests/unit/onyxTest.ts` — new `describe('batched collection member removals')`:
  - `mergeCollection` deletes null members from cache and storage via one batched `removeItems` call (and never per-key `removeItem`);
  - member subscribers are notified about batched removals (`undefined` delivered);
  - removals of members that are neither cached nor persisted are skipped entirely;
  - `setCollection` deletes missing members via one batched `removeItems` call;
  - `multiSet` deletes null keys via one batched `removeItems` call.
- `tests/unit/storage/instanceSyncWebTest.ts` — new coalescing tests:
  - a burst of storage events collapses into one `multiGet` and one dispatch (with key dedup);
  - separate bursts still dispatch separately.

### Manual Tests

With this branch pinned in E/App (web dev build), on a heavy account:

1. Open two tabs signed in to the same heavy account (e.g. Applause).
2. Refresh both tabs and stay in Tab A. Wait ~30 seconds.
3. Assert neither tab crashes and both stay responsive (before this fix: ~2,000 cross-tab events per boot, derived-value recompute storm, eventual crash of both tabs).
4. In Tab A: pin/unpin a report, draft a message in a mid-list report, mark a report unread/read, and send a message. Assert each action syncs to Tab B without refreshing.
5. In Tab A, import a really heavy Onyx state (Account > Troubleshoot > Import Onyx state). Assert Tab B adopts the imported state.
6. Go to Account > Troubleshoot > Clear cache and restart. Assert Tab B is restored to the normal state as well.

### Author Checklist

    - [x] Android / native
    - [x] Android / Chrome
    - [x] iOS / native
    - [x] iOS / Safari
    - [x] MacOS / Chrome / Safari

### Screenshots/Videos
<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

https://github.com/user-attachments/assets/5429f69d-ce1a-421a-bf8f-b2a46f0c7cce

https://github.com/user-attachments/assets/06989785-3c6b-493c-9e10-689313c943d9

</details>

### DIFF
--- a/lib/OnyxUtils.ts
+++ b/lib/OnyxUtils.ts
@@ -629,12 +629,7 @@ function keysChanged<TKey extends CollectionKeyBase>(
 /**
  * When a key change happens, search for any callbacks matching the key or collection key and trigger those callbacks
  */
-function keyChanged<TKey extends OnyxKey>(
-    key: TKey,
-    value: OnyxValue<TKey>,
-    canUpdateSubscriber: (subscriber?: CallbackToStateMapping<OnyxKey>) => boolean = () => true,
-    isProcessingCollectionUpdate = false,
-): void {
+function keyChanged<TKey extends OnyxKey>(key: TKey, value: OnyxValue<TKey>, canUpdateSubscriber: (subscriber?: CallbackToStateMapping<OnyxKey>) => boolean = () => true): void {
     // Add or remove this key from the recentlyAccessedKeys list
     if (value !== null && value !== undefined) {
         cache.addLastAccessedKey(key, OnyxKeys.isCollectionKey(key));
@@ -678,11 +673,6 @@ function keyChanged<TKey extends OnyxKey>(
                 }
 
                 if (OnyxKeys.isCollectionKey(subscriber.key)) {
-                    // Skip individual key changes during collection updates to prevent duplicate
-                    // callbacks - the collection update will handle this properly.
-                    if (isProcessingCollectionUpdate) {
-                        continue;
-                    }
                     // Cache once per dispatch to ensure all subscribers see a consistent snapshot
                     // even if a previous callback synchronously wrote to the same collection.
                     let cachedCollection = cachedCollections[subscriber.key];
@@ -761,9 +751,9 @@ function getCollectionDataAndSendAsObject<TKey extends OnyxKey>(matchingKeys: Co
 /**
  * Remove a key from Onyx and update the subscribers
  */
-function remove<TKey extends OnyxKey>(key: TKey, isProcessingCollectionUpdate?: boolean): Promise<void> {
+function remove<TKey extends OnyxKey>(key: TKey): Promise<void> {
     cache.drop(key);
-    keyChanged(key, undefined as OnyxValue<TKey>, undefined, isProcessingCollectionUpdate);
+    keyChanged(key, undefined as OnyxValue<TKey>);
 
     if (OnyxKeys.isRamOnlyKey(key)) {
         return Promise.resolve();
@@ -1413,7 +1403,12 @@ function multiSetWithRetry(data: OnyxMultiSetInput, retryAttempt?: number): Prom
         }, {});
     }
 
-    const {pairs: keyValuePairsToSet, keysToRemove} = OnyxUtils.prepareKeyValuePairsForStorage(newData, true);
+    const {pairs: keyValuePairsToSet, keysToRemove: removalCandidates} = OnyxUtils.prepareKeyValuePairsForStorage(newData, true);
+
+    // Removals of keys that are neither cached nor persisted are no-ops and skipped. When the key
+    // index has not been loaded yet (empty set), keep the removal to be safe.
+    const persistedKeys = cache.getAllKeys();
+    const keysToRemove = removalCandidates.filter((key) => cache.get(key) !== undefined || persistedKeys.size === 0 || persistedKeys.has(key));
 
     // Group collection members by their parent collection key so each collection can be notified
     // via a single batched keysChanged() call instead of one keyChanged() per member. For each
@@ -1500,9 +1495,10 @@ function multiSetWithRetry(data: OnyxMultiSetInput, retryAttempt?: number): Prom
 
     const inFlightKeys = new Set<OnyxKey>(keyValuePairsToSet.map(([key]) => key));
 
+    // A failed removal is logged, not retried — keysToRemove cannot be re-derived after the cache update.
     const storagePromises = [Storage.multiSet(keyValuePairsToStore)];
     if (keysToRemoveFromStorage.length > 0) {
-        storagePromises.push(Storage.removeItems(keysToRemoveFromStorage));
+        storagePromises.push(Storage.removeItems(keysToRemoveFromStorage).catch((error) => Logger.logAlert(`multiSet failed to remove keys from storage. Error: ${error}`)));
     }
 
     return Promise.all(storagePromises)
@@ -1578,7 +1574,9 @@ function setCollectionWithRetry<TKey extends CollectionKeyBase>({collectionKey, 
         // Skip subscriber notification on retry — already notified on attempt 0.
         // Collection-root subscribers re-fire on every keysChanged by contract.
         if (!retryAttempt) {
-            keysChanged(collectionKey, mutableCollection, previousCollection);
+            // Removed members are notified as undefined, matching mergeCollection/multiSet.
+            const partialForNotify = Object.fromEntries(Object.entries(mutableCollection).map(([key, value]) => [key, value ?? undefined]));
+            keysChanged(collectionKey, partialForNotify, previousCollection);
         }
 
         // RAM-only keys are not supposed to be saved to storage
@@ -1589,10 +1587,11 @@ function setCollectionWithRetry<TKey extends CollectionKeyBase>({collectionKey, 
 
         const inFlightKeys = new Set<OnyxKey>(keyValuePairs.map(([key]) => key));
 
-        // One batched removal = one cross-tab sync event instead of one per key.
+        // One batched removal = one cross-tab sync event instead of one per key. A failed removal is
+        // logged, not retried — keysToRemove cannot be re-derived after the cache update.
         const storagePromises = [Storage.multiSet(keyValuePairs)];
         if (keysToRemove.length > 0) {
-            storagePromises.push(Storage.removeItems(keysToRemove));
+            storagePromises.push(Storage.removeItems(keysToRemove).catch((error) => Logger.logAlert(`setCollection failed to remove keys from storage. Error: ${error}`)));
         }
 
         return Promise.all(storagePromises)
@@ -1614,11 +1613,10 @@ function setCollectionWithRetry<TKey extends CollectionKeyBase>({collectionKey, 
  * @param params.collection Object collection keyed by individual collection member keys and values
  * @param params.mergeReplaceNullPatches Record where the key is a collection member key and the value is a list of
  * tuples that we'll use to replace the nested objects of that collection member record with something else.
- * @param params.isProcessingCollectionUpdate whether this is part of a collection update operation.
  * @param retryAttempt retry attempt
  */
 function mergeCollectionWithPatches<TKey extends CollectionKeyBase>(
-    {collectionKey, collection, mergeReplaceNullPatches, isProcessingCollectionUpdate = false}: MergeCollectionWithPatchesParams<TKey>,
+    {collectionKey, collection, mergeReplaceNullPatches}: MergeCollectionWithPatchesParams<TKey>,
     retryAttempt?: number,
 ): Promise<void> {
     if (!isValidNonEmptyCollectionForMerge(collection)) {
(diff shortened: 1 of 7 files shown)


lib/OnyxUtils.ts
```diff
@@ -1777,10 +1841,14 @@ function partialSetCollection<TKey extends CollectionKeyBase>({collectionKey, co
     return getAllKeys().then((persistedKeys) => {
         const mutableCollection: OnyxInputKeyValueMapping = {...resultCollection};
         const existingKeys = resultCollectionKeys.filter((key) => persistedKeys.has(key));
+        const {pairs: keyValuePairs, keysToRemove: removalCandidates} = prepareKeyValuePairsForStorage(mutableCollection, true);
+        // Removals of keys that are neither cached nor persisted are no-ops and skipped.
+        const keysToRemove = removalCandidates.filter((key) => cache.get(key) !== undefined || persistedKeys.has(key));
+        // Snapshot before cache mutations so keysChanged() can diff removed members.
         const previousCollection = getCachedCollection(collectionKey, existingKeys);
```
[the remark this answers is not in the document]
elirangoshen: `getAllKeys()` returns the cache-augmented key set (`cache.getAllKeys()`), so any member with a cached value — RAM-only included — is in existingKeys, and its previous value is in the diff snapshot. Verified empirically: probed `keysChanged()` inputs during a RAM-only member removal through the exact batched-set path; previous contained the member's old value with unmodified code, and the subscriber received undefined correctly.

mountiny:
@codex review
@MelvinBot review

lib/OnyxUtils.ts
```diff
@@ -1688,15 +1734,33 @@ function mergeCollectionWithPatches<TKey extends CollectionKeyBase>(
...
                 // write fails.
                 const previousCollection = getCachedCollection(collectionKey, existingKeys);
+
+                // Removed members join the same keysChanged() batch (as undefined); previous values
+                // are snapshotted before the cache drop so keysChanged() can diff them.
+                const removedPreviousValues: OnyxInputKeyValueMapping = {};
+                for (const key of keysToRemove) {
+                    removedPreviousValues[key] = cache.get(key);
+                    cache.drop(key);
+                }
```
mountiny: **A key that gets deleted here can wipe out a write that came after it.**

On `main`, the deletion happened right away, early in the function. In this PR it moved later, into the block that waits for `Storage.multiGet` to finish. That wait is real whenever any member of the collection is not in the cache yet.

So if something writes to one of the deleted keys during that wait, the deletion runs afterwards and throws that write away.

Here is the repro. Put `test_A` and `test_B` in storage, clear them from the cache so the slow path runs, then fire both calls in the same tick:

```js
const removal = Onyx.mergeCollection('test_', {test_A: null, test_B: {b: 2}});
const concurrent = Onyx.merge('test_A', {y: 2});   // this one is called second
await Promise.all([removal, concurrent]);
```

- On `main`, `test_A` ends up as `{a: 1, y: 2}` in the cache and in storage.
- On this branch, `test_A` ends up gone from both.

The second call should win, since it was made last. Could the `cache.drop()` move back up into the first `.then()`, and only the `keysChanged()` batch stay down here?

lib/OnyxUtils.ts
```diff
@@ -1553,7 +1589,13 @@ function setCollectionWithRetry<TKey extends CollectionKeyBase>({collectionKey,
 
         const inFlightKeys = new Set<OnyxKey>(keyValuePairs.map(([key]) => key));
 
-        return Storage.multiSet(keyValuePairs)
+        // One batched removal = one cross-tab sync event instead of one per key.
+        const storagePromises = [Storage.multiSet(keyValuePairs)];
+        if (keysToRemove.length > 0) {
+            storagePromises.push(Storage.removeItems(keysToRemove));
+        }
```
mountiny: **When `removeItems` fails, the retry redoes the write but never retries the deletion.**

`removeItems` is now part of the same promise as `multiSet`, so if the delete fails, the whole `setCollection` is retried. But by then the keys are already gone from the cache, so the filter on line 1571 comes back empty and the delete is never attempted again. Only the `multiSet` runs a second time, and it had already succeeded.

I saw this with a storage error that Onyx classifies as `UNKNOWN`: `multiSet` ran twice, `removeItems` ran once, and the row was left behind in storage even though `getAllKeys()` says it is gone.

The leftover row is not new. On `main` a failed `removeItem` was dropped too, and this PR at least stops it from becoming an unhandled rejection. The pointless second write is the new part. `partialSetCollection` and `multiSet` do the same thing.

Would it be safer to give the removal its own `.catch()` so a failed delete does not re-run the whole write?

mountiny (commented):
Reviewed this locally on the branch. The core fix looks right and the win is real, so nothing here blocks it.

What I checked and found clean:

- `npx jest tests/unit` passes (567 tests). The only failing suite is `SQLiteProviderTest`, which is a `better-sqlite3` native binding problem in my environment and has nothing to do with this diff.
- `eslint` is clean on all four changed files.
- RAM-only keys are guarded on all four write paths.
- `SQLiteProvider.removeItems` chunks by the SQLite variable limit, and the IndexedDB provider does one transaction, so batching a few thousand keys into one call is safe on every platform.
- `keysChanged()` does deliver `undefined` to member subscribers for a batched removal.
- The coalescing in `InstanceSync` uses a fixed one-tick window rather than a debounce, so a steady stream of events cannot delay a dispatch indefinitely.
- `Storage.removeItems` is wired to `raiseStorageSyncManyKeysEvent`, so the batch really is one cross-tab event.

Two things I could not check locally: the two-tab measurements on a heavy account, and the native platforms.

I left two comments below. The first is a behaviour change I was able to reproduce and I think it is worth fixing. The second is smaller.

A few nits that need no action, just flagging:

- `multiSet` did not get the same "skip removals that are no-ops" filter the other three paths got.
- `isProcessingCollectionUpdate` is now dead. Nothing passes the second argument to `remove()` any more, so the `if (isProcessingCollectionUpdate)` branch in `keyChanged()` can no longer be reached.
- `setCollection` and `partialSetCollection` notify with `null` for removed members, while `mergeCollection` and `multiSet` notify with `undefined`. Same result either way, but it would read better if all four matched.
- A `mergeCollection` made up entirely of nulls that are all no-ops now notifies collection subscribers zero times, where `main` notified them once. That is where most of the win comes from, so this is intended, but it is worth calling out to the App reviewers in case anything treats a collection callback as a signal that data arrived.

lib/OnyxUtils.ts
elirangoshen: Hi I modified it

lib/OnyxUtils.ts
elirangoshen: I fixed it too

lib/OnyxUtils.ts
elirangoshen: > **A key that gets deleted here can wipe out a write that came after it.**
> 
> On `main`, the deletion happened right away, early in the function. In this PR it moved later, into the block that waits for `Storage.multiGet` to finish. That wait is real whenever any member of the collection is not in the cache yet.
> 
> So if something writes to one of the deleted keys during that wait, the deletion runs afterwards and throws that write away.
> 
> Here is the repro. Put `test_A` and `test_B` in storage, clear them from the cache so the slow path runs, then fire both calls in the same tick:
> 
> ```js
> const removal = Onyx.mergeCollection('test_', {test_A: null, test_B: {b: 2}});
> const concurrent = Onyx.merge('test_A', {y: 2});   // this one is called second
> await Promise.all([removal, concurrent]);
> ```
> 
> * On `main`, `test_A` ends up as `{a: 1, y: 2}` in the cache and in storage.
> * On this branch, `test_A` ends up gone from both.
> 
> The second call should win, since it was made last. Could the `cache.drop()` move back up into the first `.then()`, and only the `keysChanged()` batch stay down here?

I fixed the issues that you mentioned. and also tested on native platforms. its all looks good.

lib/OnyxUtils.ts
```diff
@@ -1688,15 +1734,33 @@ function mergeCollectionWithPatches<TKey extends CollectionKeyBase>(
                 // ensuring subscribers still reflect the merged data even if the subsequent storage
                 // write fails.
                 const previousCollection = getCachedCollection(collectionKey, existingKeys);
+
+                // Removed members join the same keysChanged() batch (as undefined); previous values
+                // are snapshotted before the cache drop so keysChanged() can diff them.
+                const removedPreviousValues: OnyxInputKeyValueMapping = {};
+                for (const key of keysToRemove) {
+                    removedPreviousValues[key] = cache.get(key);
+                    cache.drop(key);
```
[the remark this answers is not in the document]
elirangoshen: resolved

mountiny:
@FitseTLT please complete checklist and the testing. Thank you very much

FitseTLT:
## Reviewer Checklist

    - [x] Android: HybridApp
    - [x] Android: mWeb Chrome
    - [x] iOS: HybridApp
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari
    - [x] I verified that any callback methods that were added or modified are named for what the method does and never what callback they handle (i.e. `toggleReport` and not `onIconClick`).

### Screenshots/Videos
<details>


https://github.com/user-attachments/assets/1fb8dd21-b952-4492-bb8a-a56c4ad0c890

https://github.com/user-attachments/assets/21ae131f-694b-4d45-b405-4b3bd377a6b3


</details>

<details>




https://github.com/user-attachments/assets/a49ccfb2-b3b0-494c-832d-696ec552ccdb




</details>

<details>





</details>

<details>



https://github.com/user-attachments/assets/be4efab6-6b60-4926-9b11-10da25465391


https://github.com/user-attachments/assets/79ff5e9a-90eb-44ff-ad33-c647cb5db612


</details>

<details>



https://github.com/user-attachments/assets/60f1cb8d-a774-4cf5-96ad-3bb114b227cc

https://github.com/user-attachments/assets/e888a210-ce6e-40d5-b513-7a0b55c98d5e



</details>

FitseTLT:
Test is mostly looking good. But
I receive uninconsistent bug on clear cache which I am not sure it is from this pr
Android web - On the end of the vid below there was a lack of correct report name until I refreshed the pages 

https://github.com/user-attachments/assets/f2619116-e6d1-497e-ad43-a3cdc3beb285


Also regarding the heavy onyx state can you provide me with one @elirangoshen which you used to test it. Thx

elirangoshen:
> Test is mostly looking good. But I receive uninconsistent bug on clear cache which I am not sure it is from this pr Android web - On the end of the vid below there was a lack of correct report name until I refreshed the pages
> 
>  2026-08-12.02-28-05.mp4 
> Also regarding the heavy onyx state can you provide me with one @elirangoshen which you used to test it. Thx

I couldn't reproduce the report names issue also on main , but I did notice stale data  - some reports that needs attention right after clear cache  - happens on main as well so its not related to this pr. 


this is onyx state you can use  [onyx-state (6).txt](https://github.com/user-attachments/files/30973961/onyx-state.6.txt)

FitseTLT (approved):
LGTM

mountiny (approved):
Thanks for the unit tests added to cover this change

PR #821, opened 2026-08-11
perf: skip per-key clone when Onyx.init() hydrates the cache

WojtekBoman:
### Details

`Onyx.init()` loaded the persisted database into `OnyxCache` through `merge()`, which runs `fastMerge(existing, value)` per key. At init time `existing` is always `undefined`, so `fastMerge` never actually merges anything - it just walks and deep-clones every nested object of every row for no reason. On a heavy account this clone-heavy cache-write phase was measured at ~29% of total `Onyx.init()` time.

Adds `OnyxCache.hydrate()`, a bulk-load variant used only by `Onyx.init()`. Values are stored by reference when safe (`utils.needsNormalization()` is a read-only, non-allocating check for nested null/undefined and the internal `ONYX_INTERNALS__REPLACE_OBJECT_MARK`). Falls back to the existing `fastMerge` path in two cases: the key already has a value in cache (a cross-tab sync listener can race ahead of init and write a key before `hydrate()` reaches it), or the value itself needs normalizing.

`OnyxUtils.ts`'s `initializeWithDefaultKeyStates()` now calls `cache.hydrate(allDataFromStorage)` instead of `cache.merge(allDataFromStorage)`.

`merge()`'s inline `fastMerge` options were also extracted into a shared `CACHE_MERGE_OPTIONS` constant used by both `merge()` and `hydrate()`'s fallback, so the cache write paths can't drift apart.

### Related Issues

https://github.com/Expensify/App/issues/98553

### Linked E/App PR

https://github.com/Expensify/App/pull/98649

### Automated Tests

`tests/unit/onyxCacheTest.tsx` covers `hydrate()` directly and through `Onyx.init()`:

- Equivalence with `merge()`: identical cache state on an empty cache, and when keys already hold values, across fixtures with nested nulls, nested `undefined`, the replace-object mark, and plain values.
- Every key is registered, including keys whose stored value is nullish.
- Values are stored by reference when no normalization is needed, and normalized when they carry nested nullish properties.
- Reference stability: a hydrated value that changes nothing keeps the existing reference.
- Init-level: collection members loaded from storage are readable through collection reads, keys written while storage was still being read are merged rather than overwritten, and the resulting cache matches a merge-based init.

### Manual Tests

1. Sign in to an account with a reasonably large Onyx database (many reports/reportActions), force-quit and cold-start the app.
   Verify that the app boots normally, the LHN and reports render with the same data as before the change, and no console errors appear.
2. Cold-start the app, open a report with actions containing deleted/optimistic-cleared fields (values that were `null` on disk).
   Verify that the report renders correctly with those fields absent, matching pre-change behavior.
3. On web, open the app in two browser tabs signed into the same account. In tab A, trigger a write (e.g. send a message) right as tab B is cold-starting/reloading.
   Verify that tab B's cache ends up consistent with tab A's write - i.e. the value isn't clobbered by `hydrate()`'s bulk load racing the cross-tab sync listener.

### Offline tests

1. Put the device in airplane mode, force-quit and cold-start the app.
   Verify that the app boots with the last-persisted cache (same as before this change) and no data is lost or duplicated.
2. While offline, open a report and confirm previously-persisted reportActions/transactions render correctly, then go back online and verify no reconciliation errors occur.

### QA Steps

Same as tests.

### Author Checklist

    - [x] Android / native
    - [x] Android / Chrome
    - [x] iOS / native
    - [x] iOS / Safari
    - [x] MacOS / Chrome / Safari

### Screenshots/Videos

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

https://github.com/user-attachments/assets/ec06c564-79ef-4570-8304-0529bb19a1d2

</details>

### DIFF
--- a/lib/OnyxCache.ts
+++ b/lib/OnyxCache.ts
@@ -77,6 +77,7 @@ class OnyxCache {
             'set',
             'drop',
             'merge',
+            'hydrate',
             'hasPendingTask',
             'getTaskPromise',
             'captureTask',
@@ -197,6 +198,73 @@ class OnyxCache {
 
         this.storageKeys.delete(key);
         OnyxKeys.deregisterMemberKey(key);
+    }
+
+    /**
+     * Bulk-loads values into a cache that's expected to be empty, skipping merge()'s per-key clone when
+     * safe. Falls back to a real merge for any key that already has a value, in case the cache wasn't
+     * empty after all. Used only by `Onyx.init()`.
+     * @param data - a map of (cache) key - values
+     */
+    hydrate(data: Record<OnyxKey, OnyxValue<OnyxKey>>): void {
+        if (typeof data !== 'object' || Array.isArray(data)) {
+            throw new Error('data passed to cache.hydrate() must be an Object of onyx key/value pairs');
+        }
+
+        const affectedCollections = new Set<OnyxKey>();
+
+        // eslint-disable-next-line no-restricted-syntax, guard-for-in
+        for (const key in data) {
+            if (!Object.hasOwn(data, key)) {
+                continue;
+            }
+
+            const value = data[key];
+            this.addKey(key);
+
+            if (value === undefined) {
+                this.addNullishStorageKey(key);
+                continue;
+            }
+
+            const collectionKey = OnyxKeys.getCollectionKey(key);
+
+            if (value === null) {
+                this.addNullishStorageKey(key);
+                delete this.storageMap[key];
+
+                if (collectionKey) {
+                    affectedCollections.add(collectionKey);
+                }
+            } else {
+                this.nullishStorageKeys.delete(key);
+
+                const existing = this.storageMap[key];
+
+                if (existing !== undefined) {
+                    // Key already has a value, so the empty-cache assumption doesn't hold here - merge instead of clobbering.
+                    this.storageMap[key] = utils.fastMerge(existing, value, {
+                        shouldRemoveNestedNulls: true,
+                        objectRemovalMode: 'replace',
+                    }).result;
+                } else if (utils.needsNormalization(value)) {
+                    this.storageMap[key] = utils.fastMerge(undefined, value, {
+                        shouldRemoveNestedNulls: true,
+                        objectRemovalMode: 'replace',
+                    }).result;
+                } else {
+                    this.storageMap[key] = value;
+                }
+
+                if (collectionKey) {
+                    affectedCollections.add(collectionKey);
+                }
+            }
+        }
+
+        for (const collectionKey of affectedCollections) {
+            this.dirtyCollections.add(collectionKey);
+        }
     }
 
     /**
--- a/lib/OnyxCache.ts
+++ b/lib/OnyxCache.ts
@@ -225,17 +225,14 @@ class OnyxCache {
         const affectedCollections = new Set<OnyxKey>();
 
         // Use for-in loop to avoid an unnecessary array allocation from Object.keys()
-        // eslint-disable-next-line no-restricted-syntax
+        // eslint-disable-next-line no-restricted-syntax, guard-for-in
         for (const key in data) {
-            if (!Object.hasOwn(data, key)) {
-                continue;
-            }
-
             const value = data[key];
             this.addKey(key);
 
             if (value === undefined) {
                 this.addNullishStorageKey(key);
+                // undefined means "no change" — skip storageMap modification
                 continue;
             }
 
@@ -262,8 +259,8 @@ class OnyxCache {
                     // clears the nullish marker too, so a deletion racing init is still undone - same as before.
                     const merged = utils.fastMerge(existing, value, CACHE_MERGE_OPTIONS).result;
 
-                    // fastMerge is reference-stable: returns the original target when nothing changed, so a
-                    // simple === check detects no-ops and avoids dirtying the collection for nothing.
+                    // fastMerge is reference-stable: returns the original target when
+                    // nothing changed, so a simple === check detects no-ops.
                     if (merged === existing) {
                         continue;
                     }
@@ -281,6 +278,7 @@ class OnyxCache {
             }
         }
 
+        // Mark affected collections as dirty — snapshots will be lazily rebuilt on next read
         for (const collectionKey of affectedCollections) {
             this.dirtyCollections.add(collectionKey);
         }
@@ -297,7 +295,10 @@ class OnyxCache {
 
         const affectedCollections = new Set<OnyxKey>();
 
-        for (const [key, value] of Object.entries(data)) {
+        // Use for-in loop to avoid an unnecessary array allocation from Object.entries()
+        // eslint-disable-next-line no-restricted-syntax, guard-for-in
+        for (const key in data) {
+            const value = data[key];
             this.addKey(key);
 
             const collectionKey = OnyxKeys.getCollectionKey(key);
--- a/lib/OnyxCache.ts
+++ b/lib/OnyxCache.ts
@@ -2,6 +2,7 @@ import {deepEqual} from 'fast-equals';
 import bindAll from 'lodash/bindAll';
 import type {ValueOf} from 'type-fest';
 import utils from './utils';
+import type {FastMergeOptions} from './utils';
 import type {CollectionKeyBase, KeyValueMapping, NonUndefined, OnyxCollection, OnyxKey, OnyxValue} from './types';
 import OnyxKeys from './OnyxKeys';
 
@@ -14,6 +15,16 @@ type CollectionSnapshot = Readonly<NonUndefined<OnyxCollection<KeyValueMapping[O
  * which relies on === equality to detect changes.
  */
 const FROZEN_EMPTY_COLLECTION: Readonly<NonUndefined<OnyxCollection<KeyValueMapping[OnyxKey]>>> = Object.freeze({});
+
+/**
+ * Merge options shared by every cache write path (`merge()` and `hydrate()`'s fallback), so the
+ * three call sites can't drift apart. Cached values must never hold nested nulls, and a source
+ * object carrying the replace mark must replace the target object rather than merge into it.
+ */
+const CACHE_MERGE_OPTIONS: FastMergeOptions = {
+    shouldRemoveNestedNulls: true,
+    objectRemovalMode: 'replace',
+};
 
 // Task constants
 const TASK = {
@@ -242,16 +253,21 @@ class OnyxCache {
                 const existing = this.storageMap[key];
 
                 if (existing !== undefined) {
-                    // Key already has a value, so the empty-cache assumption doesn't hold here - merge instead of clobbering.
-                    this.storageMap[key] = utils.fastMerge(existing, value, {
-                        shouldRemoveNestedNulls: true,
-                        objectRemovalMode: 'replace',
-                    }).result;
+                    // Key already has a value, so the empty-cache assumption doesn't hold here (e.g. a write
+                    // landed while storage was still being read). Fall back to a real merge, which has exactly
+                    // the same semantics as the old `cache.merge(allDataFromStorage)` init path: the value
+                    // loaded from disk is the merge source, so it wins on any overlapping leaf key.
+                    const merged = utils.fastMerge(existing, value, CACHE_MERGE_OPTIONS).result;
+
+                    // fastMerge is reference-stable: returns the original target when nothing changed, so a
+                    // simple === check detects no-ops and avoids dirtying the collection for nothing.
+                    if (merged === existing) {
+                        continue;
+                    }
+
+                    this.storageMap[key] = merged;
                 } else if (utils.needsNormalization(value)) {
-                    this.storageMap[key] = utils.fastMerge(undefined, value, {
-                        shouldRemoveNestedNulls: true,
-                        objectRemovalMode: 'replace',
-                    }).result;
+                    this.storageMap[key] = utils.fastMerge(undefined, value, CACHE_MERGE_OPTIONS).result;
                 } else {
                     this.storageMap[key] = value;
                 }
@@ -301,10 +317,7 @@ class OnyxCache {
 
                 // Per-key merge instead of spreading the entire storageMap
                 const existing = this.storageMap[key];
-                const merged = utils.fastMerge(existing, value, {
-                    shouldRemoveNestedNulls: true,
-                    objectRemovalMode: 'replace',
-                }).result;
+                const merged = utils.fastMerge(existing, value, CACHE_MERGE_OPTIONS).result;
 
                 // fastMerge is reference-stable: returns the original target when
                 // nothing changed, so a simple === check detects no-ops.
--- a/lib/OnyxCache.ts
+++ b/lib/OnyxCache.ts
@@ -224,6 +224,7 @@ class OnyxCache {
 
         const affectedCollections = new Set<OnyxKey>();
 
+        // Use for-in loop to avoid an unnecessary array allocation from Object.keys()
         // eslint-disable-next-line no-restricted-syntax
         for (const key in data) {
             if (!Object.hasOwn(data, key)) {
--- a/lib/OnyxCache.ts
+++ b/lib/OnyxCache.ts
@@ -224,7 +224,7 @@ class OnyxCache {
 
         const affectedCollections = new Set<OnyxKey>();
 
-        // eslint-disable-next-line no-restricted-syntax, guard-for-in
+        // eslint-disable-next-line no-restricted-syntax
         for (const key in data) {
             if (!Object.hasOwn(data, key)) {
                 continue;
--- a/lib/OnyxCache.ts
+++ b/lib/OnyxCache.ts
@@ -218,7 +218,7 @@ class OnyxCache {
      * @param data - a map of (cache) key - values
      */
     hydrate(data: Record<OnyxKey, OnyxValue<OnyxKey>>): void {
-        if (typeof data !== 'object' || Array.isArray(data)) {
+        if (data === null || typeof data !== 'object' || Array.isArray(data)) {
             throw new Error('data passed to cache.hydrate() must be an Object of onyx key/value pairs');
         }
 
--- a/lib/OnyxCache.ts
+++ b/lib/OnyxCache.ts
@@ -258,6 +258,8 @@ class OnyxCache {
                     // landed while storage was still being read). Fall back to a real merge, which has exactly
                     // the same semantics as the old `cache.merge(allDataFromStorage)` init path: the value
                     // loaded from disk is the merge source, so it wins on any overlapping leaf key.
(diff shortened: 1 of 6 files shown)


WojtekBoman:

lib/OnyxCache.ts
```diff
@@ -199,6 +211,78 @@ class OnyxCache {
...
+     * @param data - a map of (cache) key - values
+     */
+    hydrate(data: Record<OnyxKey, OnyxValue<OnyxKey>>): void {
+        if (typeof data !== 'object' || Array.isArray(data)) {
+            throw new Error('data passed to cache.hydrate() must be an Object of onyx key/value pairs');
+        }
+
+        const affectedCollections = new Set<OnyxKey>();
+
+        // eslint-disable-next-line no-restricted-syntax
```
mateuuszzzzz: Can we add an explanation why we have to use restricted syntax here?

lib/utils.ts
```diff
@@ -196,6 +196,35 @@ function isMergeableObject<TObject extends Record<string, unknown>>(value: unkno
...
+/**
+ * Reports whether a value needs cleaning (nested null/undefined, or the replace-object mark) before it's
+ * safe to store by reference. Read-only, non-allocating.
+ */
+function needsNormalization(value: unknown): boolean {
+    if (value === null || value === undefined || typeof value !== 'object' || Array.isArray(value)) {
+        return false;
+    }
+
+    // eslint-disable-next-line no-restricted-syntax, guard-for-in
```
mateuuszzzzz, line 209: Same as https://github.com/Expensify/react-native-onyx/pull/821/changes#r3775004474

lib/OnyxCache.ts
```diff
@@ -199,6 +211,79 @@ class OnyxCache {
...
     }
 
+    /**
+     * Bulk-loads values into a cache that's expected to be empty, skipping merge()'s per-key clone when
+     * safe. Falls back to a real merge for any key that already has a value, in case the cache wasn't
+     * empty after all. Used only by `Onyx.init()`.
+     * @param data - a map of (cache) key - values
+     */
+    hydrate(data: Record<OnyxKey, OnyxValue<OnyxKey>>): void {
+        if (typeof data !== 'object' || Array.isArray(data)) {
```
collectioneur: Let's also add a `data === null` check here. The old `merge()` path had the same guard gap (`typeof null === 'object'`), but `Object.entries(data)` right after it would still throw on `null`. Here we iterate with `for-in`, which silently does nothing for `null`, so the guard is the only line of defense. NAB since data can't currently be `null` (the only caller builds it as an object literal), but worth adding for safety

lib/OnyxUtils.ts
```diff
@@ -1044,9 +1044,11 @@ function initializeWithDefaultKeyStates(): Promise<void> {
                 allDataFromStorage[key] = value;
             }
 
-            // Load all storage data into cache silently (no subscriber notifications)
+            // Load all storage data into cache silently (no subscriber notifications).
+            // hydrate() rather than merge(): the cache is empty at this point, so a per-key fastMerge
+            // would only deep-clone every row it was handed.
             cache.setAllKeys(Object.keys(allDataFromStorage));
```
collectioneur: It looks like `hydrate()` already covers what `setAllKeys()` does here, it calls `addKey()` for every key in data, which both populates `storageKeys` and calls `registerMemberKey()`. Since we pass the same keys to both, we end up registering every key twice on `init`. What do you think about dropping the `setAllKeys()` call and leaving a short comment that `hydrate()` handles key registration itself?

collectioneur (approved):
LGTM! 🚀

lib/OnyxCache.ts
```diff
@@ -199,6 +211,79 @@ class OnyxCache {
...
+            } else {
+                this.nullishStorageKeys.delete(key);
+
+                const existing = this.storageMap[key];
+
+                if (existing !== undefined) {
+                    // Key already has a value, so the empty-cache assumption doesn't hold here (e.g. a write
+                    // landed while storage was still being read). Fall back to a real merge, which has exactly
+                    // the same semantics as the old `cache.merge(allDataFromStorage)` init path: the value
+                    // loaded from disk is the merge source, so it wins on any overlapping leaf key.
```
mkhutornyi, line 257: This race fallback cannot detect a cross-tab **deletion** that landed before hydrate, so a stale row is resurrected.
`cache.set(key, null)` erases both `storageMap[key]` and the `nullishStorageKeys` marker, leaving hydrate no way to tell "**never written**" from "**just deleted**".
Same as the old `merge()` behavior (not a regression), but this new comment and the PR body claim the fallback makes the init race safe — it only does for non-null writes.

Suggested additional comment:
```
// Note: this only covers non-null writes. A `cache.set(key, null)` that lands before hydrate
// clears the nullish marker too, so a deletion racing init is still undone — same as before.

```

tests/unit/onyxCacheTest.tsx
```diff
@@ -362,6 +387,176 @@ describe('Onyx', () => {
...
+
+                // Then the collection snapshot is rebuilt and reports the same members
+                expect(cache.getCollectionData(collectionKey)).toEqual({
+                    [`${collectionKey}1`]: {id: 1},
+                    [`${collectionKey}2`]: {id: 2},
+                });
+                expect(cache.getCollectionData(collectionKey)).toEqual(mergeCache.getCollectionData(collectionKey));
+            });
+
+            it('Should throw if called with anything that is not an object', () => {
```
mkhutornyi: `hydrate(null)` is guarded but not covered by this test.
Worth one more line given the guard was added in 51d14dd specifically for it.

mkhutornyi (commented):
NAB comments:

mkhutornyi:
## Regression test cases

### 1. Cold start with existing data (the main path)

1. Sign in and let the app fully load (open a few chats, a workspace, and Search).
2. Force-quit the app (swipe away on mobile, close the tab/browser on web).
3. Relaunch.
4. Verify the LHN renders the same report list as before the restart, in the same order.
5. Open a report that was open before — verify all messages, avatars, and unread indicators are identical.
6. Open Settings → Workspaces — verify categories, tags, and members are present.
7. Verify no "empty state" or skeleton flashes and then fills in for data that was already on disk.

_Why:_ `cache.hydrate()` replaces `setAllKeys()` + `cache.merge()` for the entire eager load. A key registration failure shows up here as missing LHN rows or an empty workspace list.

### 2. Collections render fully

1. From the cold-start state above, verify the LHN report list, the workspace member list, and a report's transaction list all show every row.
2. Pull to refresh / navigate away and back — verify nothing appears or disappears.

_Why:_ `hydrate()` now registers collection member keys via `addKey()` and marks collections dirty itself. A miss here means a collection snapshot is built from an incomplete index — the symptom is _some_ members rendering, not zero.

### 3. Values with removed properties survive the restart

1. Set a workspace description, then clear it. Do the same for an optional profile field (e.g. pronouns).
2. Force-quit and relaunch.
3. Verify the cleared fields are still empty and no `null` leaks into the UI as the literal string "null" or a blank row that shouldn't exist.

_Why:_ `hydrate()` skips normalization entirely for values that pass `needsNormalization()`, and only deep-cleans the ones that don't. This is the scenario where a value carrying nested nulls is stored raw.

### 4. Offline data after restart

1. Create an expense, then go offline before it finishes syncing.
2. Force-quit and relaunch while still offline.
3. Verify the pending expense is still visible with its offline styling.
4. Go back online — verify it syncs and the pending styling clears.

_Why:_ exercises hydrate against optimistic data written by the previous session, including `pendingAction` fields.

### 5. Multi-tab race (web only — highest-risk scenario)

1. Open the app in two browser tabs and let both load.
2. In Tab A, send a message in a chat.
3. **Immediately** hard-refresh Tab B (Cmd+Shift+R) so its startup storage read overlaps the cross-tab broadcast.
4. Verify Tab B shows the new message and the chat is in the right LHN position.
5. Repeat 3–4 times — the timing window is narrow.
6. Now the inverse: in Tab A **delete** a message, and immediately hard-refresh Tab B.
7. Verify the deleted message does not reappear in Tab B.

_Why:_ the `existing !== undefined` fallback inside `hydrate()` exists specifically for this race. Steps 6–7 cover a known pre-existing gap — if the deleted message does briefly reappear, that reproduces on `main` too and is not caused by this PR. Worth capturing on video either way.

### 6. Fresh install / empty storage

1. Uninstall (or clear site data) and reinstall.
2. Launch before signing in — verify the sign-in page renders normally, no crash, no hang on the splash screen.
3. Sign in and verify normal loading.

_Why:_ `hydrate({})` on an empty payload no longer resets the key index. Confirms nothing depends on that reset.

### 7. Sign out / sign in / switch accounts

1. Sign out, then sign in as a **different** account.
2. Verify none of the previous account's chats, workspaces, or avatars appear.
3. Force-quit and relaunch — verify still clean.

_Why:_ `Onyx.clear()` interacts with the key index, and the index is no longer replaced wholesale at init.

### 8. Performance verification (this is the PR's actual goal)

1. On a **release** build with a large account (many reports/transactions), measure time from launch to first LHN paint.
2. Compare against a build on `main` with the same account and the same storage state — run each 5 times and take the median.
3. Verify the branch is not slower; a measurable improvement on large accounts is the expected result.

A Sentry span or a `console.time` around `Onyx.init()`'s resolution both work; use whichever the perf setup already tracks.

### Platforms

Run 1–4 and 6–7 on **iOS, Android, and web** (different storage providers: SQLite vs IndexedDB, and only SQLite batches the whole table through one `JSON.parse`). Scenario 5 is **web only**. Scenario 8 on **iOS and Android release builds**.

roryabraham (dismissed):
Code LGTM but we should link an App PR to help the Onyx merge queue go smoother.

mkhutornyi:
## Reviewer Checklist

    - [x] Android: HybridApp
    - [x] Android: mWeb Chrome
    - [x] iOS: HybridApp
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari
    - [x] I verified that any callback methods that were added or modified are named for what the method does and never what callback they handle (i.e. `toggleReport` and not `onIconClick`).

### Screenshots/Videos
<details>



</details>

<details>



</details>

<details>


https://github.com/user-attachments/assets/255d0c02-832e-430b-80d8-38f0268a8554



</details>

<details>



</details>



<details>


https://github.com/user-attachments/assets/ea468762-9969-436c-8c61-0082c352020b


</details>

mkhutornyi:
> Code LGTM but we should link an App PR to help the Onyx merge queue go smoother.

App PR is here - https://github.com/Expensify/App/pull/98649

mkhutornyi (approved):

WojtekBoman:
@roryabraham @mkhutornyi 
Comments have been addressed and I’ve added a link to the App PR in the description :)

mountiny (dismissed):
code looks good, waiting for @fabioh8010 review too

lib/OnyxCache.ts
```diff
@@ -199,6 +211,81 @@ class OnyxCache {
...
+        }
+
+        const affectedCollections = new Set<OnyxKey>();
+
+        // Use for-in loop to avoid an unnecessary array allocation from Object.keys()
+        // eslint-disable-next-line no-restricted-syntax
+        for (const key in data) {
+            if (!Object.hasOwn(data, key)) {
+                continue;
+            }
```
fabioh8010: Is this condition really necessary?

mountiny:
Checking internally the CLA action is failing due to new org wide rule

lib/OnyxCache.ts
```diff
@@ -199,6 +211,81 @@ class OnyxCache {
...
+
+            const value = data[key];
+            this.addKey(key);
+
+            if (value === undefined) {
+                this.addNullishStorageKey(key);
+                continue;
```
fabioh8010, line 236:
```suggestion
                // undefined means "no change" — skip storageMap modification
                continue;
```

Same as we have in merge()

lib/OnyxCache.ts
```diff
@@ -199,6 +211,81 @@ class OnyxCache {
...
+                    // loaded from disk is the merge source, so it wins on any overlapping leaf key.
+                    // Note: this only covers non-null writes. A `cache.set(key, null)` that lands before hydrate
+                    // clears the nullish marker too, so a deletion racing init is still undone - same as before.
+                    const merged = utils.fastMerge(existing, value, CACHE_MERGE_OPTIONS).result;
+
+                    // fastMerge is reference-stable: returns the original target when nothing changed, so a
+                    // simple === check detects no-ops and avoids dirtying the collection for nothing.
+                    if (merged === existing) {
+                        continue;
+                    }
```
fabioh8010:
```suggestion
                    // fastMerge is reference-stable: returns the original target when
                    // nothing changed, so a simple === check detects no-ops.
                    if (merged === existing) {
                        continue;
                    }
```

Same as we have in merge()

lib/OnyxCache.ts
```diff
@@ -199,6 +211,81 @@ class OnyxCache {
...
+                    this.storageMap[key] = value;
+                }
+
+                if (collectionKey) {
+                    affectedCollections.add(collectionKey);
+                }
+            }
+        }
+
+        for (const collectionKey of affectedCollections) {
```
fabioh8010, line 282:
```suggestion
        // Mark affected collections as dirty — snapshots will be lazily rebuilt on next read
        for (const collectionKey of affectedCollections) {
```

Same as we have in merge()

lib/OnyxCache.ts
```diff
@@ -199,6 +211,81 @@ class OnyxCache {
...
+    hydrate(data: Record<OnyxKey, OnyxValue<OnyxKey>>): void {
+        if (data === null || typeof data !== 'object' || Array.isArray(data)) {
+            throw new Error('data passed to cache.hydrate() must be an Object of onyx key/value pairs');
+        }
+
+        const affectedCollections = new Set<OnyxKey>();
+
+        // Use for-in loop to avoid an unnecessary array allocation from Object.keys()
+        // eslint-disable-next-line no-restricted-syntax
+        for (const key in data) {
```
fabioh8010, line 229: Should we do the same in merge()?

lib/utils.ts
```diff
@@ -196,6 +196,36 @@ function isMergeableObject<TObject extends Record<string, unknown>>(value: unkno
     return isNonNullObject && !(value instanceof RegExp) && !(value instanceof Date) && !Array.isArray(value);
 }
 
+/**
+ * Reports whether a value needs cleaning (nested null/undefined, or the replace-object mark) before it's
+ * safe to store by reference. Read-only, non-allocating.
+ */
+function needsNormalization(value: unknown): boolean {
```
fabioh8010, line 203: Please create unit tests for this function in `tests/unit/utilsTest.ts`

tests/unit/onyxCacheTest.tsx
```diff
@@ -362,6 +387,178 @@ describe('Onyx', () => {
...
+                nestedNullsKey: {keep: 1, drop: null, nested: {keep: 2, drop: null}},
+                arrayKey: [{ID: 1}, {ID: 2}],
+                objectWithArrayKey: {ID: [1, 2]},
+                nullKey: null,
+                undefinedKey: undefined,
+            });
+
+            const ALL_KEYS = Object.keys(buildStorageData());
+
+            it('Should produce the same cache state as merge() for an empty cache', () => {
```
fabioh8010:
```suggestion
            it('should produce the same cache state as merge() for an empty cache', () => {
```

tests/unit/onyxCacheTest.tsx
```diff
@@ -362,6 +387,178 @@ describe('Onyx', () => {
...
+
+                // When one is hydrated and the other merged with the same data
+                cache.hydrate(buildStorageData());
+                mergeCache.merge(buildStorageData());
+
+                // Then both caches are indistinguishable through their public read surface
+                expectSameCacheState(cache, mergeCache, ALL_KEYS);
+            });
+
+            it('Should produce the same cache state as merge() when keys already have values', () => {
```
fabioh8010:
```suggestion
            it('should produce the same cache state as merge() when keys already have values', () => {
```

tests/unit/onyxCacheTest.tsx
```diff
@@ -362,6 +387,178 @@ describe('Onyx', () => {
...
+                // Then the fallback merge inside hydrate() matches merge() exactly
+                expectSameCacheState(cache, mergeCache, ALL_KEYS);
+
+                // And the value loaded from storage wins on every overlapping leaf, as merge() does
+                expect(cache.get('plainObjectKey')).toEqual({value: 'mockValue', fromWrite: true});
+                expect(cache.get('nestedObjectKey')).toEqual({a: {b: {c: 1, fromWrite: true}}});
+                expect(cache.get('stringKey')).toBe('mockValue');
+            });
+
+            it('Should register every key, including keys with nullish values', () => {
```
fabioh8010:
```suggestion
            it('should register every key, including keys with nullish values', () => {
```

@WojtekBoman Could you apply the same suggestion for the other tests too?

WojtekBoman: Done!

lib/utils.ts
WojtekBoman: Added in this commit https://github.com/Expensify/react-native-onyx/pull/821/commits/2baf94693d7f59963e4b3d5ac1f57930ee6014fe

lib/OnyxCache.ts
WojtekBoman: I think we can adjust it :) Done in this commit https://github.com/Expensify/react-native-onyx/pull/821/changes/03d1f6ff5cb6328f386ac62d3ce424edb3583648

fabioh8010 (approved):

fabioh8010 (approved):

lib/OnyxCache.ts
```diff
@@ -199,6 +211,79 @@ class OnyxCache {
...
+    hydrate(data: Record<OnyxKey, OnyxValue<OnyxKey>>): void {
+        if (data === null || typeof data !== 'object' || Array.isArray(data)) {
+            throw new Error('data passed to cache.hydrate() must be an Object of onyx key/value pairs');
+        }
+
+        const affectedCollections = new Set<OnyxKey>();
+
+        // Use for-in loop to avoid an unnecessary array allocation from Object.keys()
+        // eslint-disable-next-line no-restricted-syntax, guard-for-in
+        for (const key in data) {
```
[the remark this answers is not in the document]
WojtekBoman: The mechanism is correct, but I don't think the guard is necessary here, and it was intentionally removed in review https://github.com/Expensify/react-native-onyx/pull/821#discussion_r3795288130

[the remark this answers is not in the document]
fabioh8010: Yeah I dont think this kind of situation would ever happen

mountiny (approved):
Thanks for the thorough review, I think we can move this ahead now

PR #833, opened 2026-08-31
Add OnyxStore as a standalone, inert module

fabioh8010:
### Details

Coming from https://expensify.slack.com/archives/C05LX9D6E07/p1788270903471539.

Adds `OnyxStore`, a standalone listener registry that will back Onyx subscriptions in the store-based redesign. Nothing calls it yet, so this PR is inert and changes no behavior.

`OnyxStore` keeps a single `Map<key, Set<listener>>`. Subscribing is a `Set.add`, unsubscribing a `Set.delete`, and a write looks up only the affected key's listeners:

- `subscribe(key, listener)` registers an exact-key listener (a single key, a collection root, or a collection member) and returns an unsubscribe function.
- `getState(key)` reads the eager cache: the frozen collection object for collection keys, the cached value otherwise.
- `notifyKey(key, value)` dispatches a single-key write and routes to the parent collection's listeners when the key is a collection member.
- `notifyCollection(key, partial, previous)` dispatches a batch update, firing collection listeners once and member listeners only for members whose value reference actually changed.

A throwing listener is caught and logged so it can't stop the others. The module depends only on `OnyxCache` and `OnyxKeys`.

The next PR wires the subscription layer onto `OnyxStore` and deletes the old per-subscription bookkeeping. Full context is in the proposal thread.

### Related Issues

https://github.com/Expensify/App/issues/100271

### Linked E/App PR

https://github.com/Expensify/App/pull/100316

### Automated Tests

Tests were added for the new module.

### Manual Tests

There is nothing to test right now as we are just adding a new module without using it.

### Author Checklist

    - [x] Android / native
    - [x] Android / Chrome
    - [x] iOS / native
    - [x] iOS / Safari
    - [x] MacOS / Chrome / Safari

### Screenshots/Videos
<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>

<details>

</details>


fabioh8010:
@codex review

lib/OnyxStore.ts
```diff
@@ -0,0 +1,204 @@
...
+     *   1. keyListeners.get(key): exact-key subscribers (always fires).
+     *   2. If key is a collection member, keyListeners.get(collectionKey): collection
+     *      listeners for the parent collection (unless `options.suppressCollectionNotify`).
+     */
+    notifyKey<TKey extends OnyxKey>(key: TKey, value: OnyxValue<TKey>, options?: NotifyKeyOptions): void {
+        // 1. Exact-key listeners
+        const exact = this.keyListeners.get(key);
+        if (exact && exact.size > 0) {
+            for (const listener of exact) {
+                this.safeInvoke(() => listener(value, key), key);
```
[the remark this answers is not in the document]
fabioh8010: Addressed in https://github.com/Expensify/react-native-onyx/pull/833/commits/844571b1f7592975cb06c4ba21d8f184b307d11e, although this is very unlikely to happen.

lib/OnyxSubscriptionManager.ts
roryabraham, line 1: Maybe `OnyxSubscriptionManager` or `OnyxListenerRegistry` is a better name. I know it's quite similar to `OnyxConnectionManager`, but that's what this is replacing. I worry `OnyxStore` and `OnyxCache` are confusingly similar, and the distinction between them isn't clear enough from just the names.

lib/OnyxStore.ts
```diff
@@ -0,0 +1,204 @@
...
+
+/**
+ * Listener fired when an exact key's value changes.
+ */
+type KeyListener<TKey extends OnyxKey = OnyxKey> = (value: OnyxValue<TKey>, key: TKey) => void;
+
+/**
+ * Storage form of a listener, value erased so one Map can hold listeners for every key type.
+ */
+type StoredListener = (value: unknown, key: OnyxKey) => void;
```
roryabraham: **suggestion:**

```ts
/**
 * Generic listener, without specific types for the keys or values.
 * This was a single Map can hold listeners for every key type.
 */
type GenericListener = (value: unknown, key: OnyxKey) => void;
```

lib/OnyxStore.ts
```diff
@@ -0,0 +1,204 @@
...
+
+type NotifyKeyOptions = {
+    /**
+     * Skips collection-level routing. Collection-batch write paths set it so each member write
+     * doesn't re-trigger the collection-level listeners; the outer `notifyCollection()` fires those once.
+     */
+    suppressCollectionNotify?: boolean;
```
roryabraham: **suggestion:**

```ts
type KeyNotificationOptions = {
    /**
     * Should subscribers of the root collection key be notified of writes to individual collection members?
     */
   shouldNotifyCollectionForIndividualMember?: boolean;
};
```

"should suppress" is effectively a negated variable ("should I NOT do X?"), which makes it harder to reason about than "should I do X?"

Also, it's not clear why this is needed at all. Don't collection batch writes use `notifyCollection` instead of `notifyKey`? We could guarantee at compile time that `notifyKey` is not called with collection root keys by doing something like this (note `Exclude<OnyxKey, CollectionKeyBase>`):

```ts
notifyKey<TKey extends Exclude<OnyxKey, CollectionKeyBase>>(key: TKey, value: OnyxValue<TKey>): void { ... }
```

could that allow us to remove this entirely along with the "Collection-level routing" in `notifyKey`?

lib/OnyxStore.ts
```diff
@@ -0,0 +1,204 @@
+import type {CollectionKeyBase, KeyValueMapping, OnyxCollection, OnyxKey, OnyxValue} from './types';
+
+import * as Logger from './Logger';
+import cache from './OnyxCache';
+import OnyxKeys from './OnyxKeys';
+
+/**
+ * Listener fired when an exact key's value changes.
+ */
+type KeyListener<TKey extends OnyxKey = OnyxKey> = (value: OnyxValue<TKey>, key: TKey) => void;
```
roryabraham: **suggestion:**

```suggestion
type Listener<TKey extends OnyxKey = OnyxKey> = (value: OnyxValue<TKey>, key: TKey) => void;
```

lib/OnyxSubscriptionManager.ts
```diff
@@ -0,0 +1,204 @@
...
+
+type NotifyKeyOptions = {
+    /**
+     * Skips collection-level routing. Collection-batch write paths set it so each member write
+     * doesn't re-trigger the collection-level listeners; the outer `notifyCollection()` fires those once.
+     */
+    suppressCollectionNotify?: boolean;
+};
+
+/**
```
roryabraham, line 18: **suggestion:**

```ts
/**
 * OnyxSubscriptionManager is a registry for Onyx subscriptions.
 * Subscriptions are stored in `keyListeners`, a flat map keyed by OnyxKey.
 * Subscribers are notified on a per-key basis:
 *    - `notifyKey` for individual keys or collection members
 *    - `notifyCollection` for batch updates to collections
 */
```

lib/OnyxStore.ts
```diff
@@ -0,0 +1,204 @@
...
+    private safeInvoke(fn: () => void, contextKey: OnyxKey): void {
+        try {
+            fn();
+        } catch (error) {
+            Logger.logAlert(`[OnyxStore] Listener threw an error for key '${contextKey}': ${error}`);
+        }
+    }
+}
+
+const onyxStore = new OnyxStore();
```
roryabraham: NAB: initialize explicitly in `Onyx.init` rather than on module import. I have found that things that initialize on import can be harder to mock in tests, if we ever want to do that.

roryabraham (changes requested):

lib/OnyxSubscriptionManager.ts
```diff
@@ -0,0 +1,204 @@
...
+
+    constructor() {
+        this.keyListeners = new Map();
+    }
+
+    /**
+     * Returns the frozen collection object for collection
+     * keys, the cached value for single keys, or `undefined` if not in cache.
+     */
+    getState<TKey extends OnyxKey>(key: TKey): OnyxValue<TKey> {
```
huult, line 36: @fabioh8010 Could we add a test for `getState()` to cover both the `collection-key` and `single-key paths`?

lib/OnyxStore.ts
```diff
@@ -0,0 +1,204 @@
...
+            for (const listener of [...collectionListeners]) {
+                this.safeInvoke(() => listener(collectionData, collectionKey), collectionKey);
+            }
+        }
+
+        // 2. Exact-member subscribers fire per changed key (skip if ref unchanged vs previous).
+        for (const memberKey of changedKeys) {
+            const value = collectionData?.[memberKey];
+            const prev = previous[memberKey];
+            if (value === prev) {
```
huult:
```suggestion
           if (partialPreviousCollection && Object.prototype.hasOwnProperty.call(partialPreviousCollection, memberKey) && value === partialPreviousCollection[memberKey]) {
```

@fabioh8010 Do you think this case can actually happen?

```diff
       it('should notify the exact-member listener when the member is deleted without a previous snapshot', () => {
            // member1 has already been removed from the cache by the time notifyCollection runs.
            jest.spyOn(cache, 'getCollectionData').mockReturnValue({[MEMBER_2]: {id: 2}});

            const member1Callback = jest.fn();
            onyxStore.subscribe(MEMBER_1, member1Callback);

            // No partialPreviousCollection passed — it's optional.
            onyxStore.notifyCollection(COLLECTION, {[MEMBER_1]: null});

            expect(member1Callback).toHaveBeenCalledTimes(1);
        });
```

When a member is deleted (`null`) and the caller omits `partialPreviousCollection`,
`previous` defaults to `{}`, so `previous[memberKey]` is `undefined`, which is the same value
that `collectionData?.[memberKey]` resolves to once the member is removed from the cache
snapshot (see `OnyxCache.rebuildCollectionSnapshot`). Since `undefined === undefined` is `true`,
the ref-equality check treats the deletion as "unchanged" and skips the listener,
even though `memberKey` is present in `changedKeys`.

huult:
@fabioh8010 Could you take a look at the comments above?

lib/OnyxStore.ts
fabioh8010: It's very unlikely but for correctness I applied the change

lib/OnyxStore.ts
fabioh8010: `suppressCollectionNotify` was a leftover we don't need anymore because of recent changes in Onyx, so I removed it 👍

lib/OnyxStore.ts
fabioh8010: @roryabraham The file follow the pattern of other modules like connection manager and cache, and they are mockable with `jest.mock('./OnyxSubscriptionManager')` for example so I don't think we need to do this now, if you don't mind I would leave as it is.

lib/OnyxSubscriptionManager.ts
```diff
@@ -0,0 +1,196 @@
...
+    notifyKey<TKey extends OnyxKey>(key: TKey, value: OnyxValue<TKey>): void {
+        // 1. Exact-key listeners
+        const exact = this.keyListeners.get(key);
+        if (exact && exact.size > 0) {
+            for (const listener of [...exact]) {
+                this.safeInvoke(() => listener(value, key), key);
+            }
+        }
+
+        // 2. Collection-level routing. Only fires when the write is to a member key.
```
roryabraham, line 90: Not sure if you missed this from my last review. if direct writes to a collection root are an unsupported anti-pattern, why not prevent them at compile time?

```diff
diff --git a/lib/OnyxSubscriptionManager.ts b/lib/OnyxSubscriptionManager.ts
index 7e61ec7e..8bc43472 100644
--- a/lib/OnyxSubscriptionManager.ts
+++ b/lib/OnyxSubscriptionManager.ts
@@ -78,7 +78,7 @@ class OnyxSubscriptionManager {
      *   2. If key is a collection member, keyListeners.get(collectionKey): collection
      *      listeners for the parent collection.
      */
-    notifyKey<TKey extends OnyxKey>(key: TKey, value: OnyxValue<TKey>): void {
+    notifyKey<TKey extends Exclude<OnyxKey, CollectionKeyBase>>(key: TKey, value: OnyxValue<TKey>): void {
         // 1. Exact-key listeners
         const exact = this.keyListeners.get(key);
         if (exact && exact.size > 0) {

```

fabioh8010: Sorry, forgot to answer about this one. While it makes sense for external consumers like E/App (if we exported this module), it breaks completely internally because `OnyxKey` and `CollectionKeyBase` are both `string` from library's own perspective – external consumers will augment Onyx types with the correct keys but internally we set them to `string`, see [TypeOptions type](https://github.com/Expensify/react-native-onyx/blob/956be8f239668e976a2278879f97d34f0870eb9c/lib/types.ts#L36-L43). Since they are both `string`, `Exclude<string, string>` will resolve to `never` and any call of this function inside the repo will break.

We shouldn't worry about this though, this module isn't supposed to be exported or used externally by consumers.

roryabraham (approved):

roryabraham:
@huult @Gonals back to you (though there's nothing to test here yet)

huult:
## Reviewer Checklist

    - [x] Android: HybridApp
    - [x] Android: mWeb Chrome
    - [x] iOS: HybridApp
    - [x] iOS: mWeb Safari
    - [x] MacOS: Chrome / Safari
    - [x] All props are defined accurately

### Screenshots/Videos
<details>



</details>

<details>



</details>

<details>



</details>

<details>



</details>

<details>



</details>

huult (approved):
LGTM.

PR #834, opened 2026-08-31
Flip the subscription layer onto OnyxStore (atomic swap)

fabioh8010:
### Details

Replaces Onyx's subscription layer with a store-based engine. Subscriptions are now held in a single registry (`OnyxSubscriptionManager`) keyed by Onyx key; `useOnyx` is rewritten on top of `useSyncExternalStoreWithSelector`, and `Onyx.connect`/`Onyx.connectWithoutView` deliver through the same store. `OnyxConnectionManager` and `OnyxSnapshotCache` are removed.

The public API is unchanged — `useOnyx`, `connect`, `connectWithoutView`, `disconnect`, and all write methods keep their existing signatures and semantics — so this is an internal swap with no intended user-facing behavior change.

Key points:
- A subscriber's initial callback is delivered after any in-flight writes that affect its own key have applied, so it reads post-write data and does not double-deliver. This wait is scoped to the connecting key (and, for a collection root, its members), so an unrelated or slow write elsewhere can never block or delay a subscription's first delivery.
- Collection-root subscribers receive the frozen collection object; individual member subscribers receive their own value, with reference-equality dedup so unchanged members do not re-render.
- Fewer subscription bookkeeping operations per render, reducing churn on high-traffic surfaces like the chat list.

### Related Issues

https://github.com/Expensify/App/issues/100271

### Linked E/App PR

https://github.com/Expensify/App/pull/98956

### Automated Tests

- `OnyxSubscriptionManagerTest`: subscribe/unsubscribe, per-key and collection routing, `notifyKey`/`notifyCollection` dedup, subscription mutation during dispatch, and listener error isolation.
- `onyxTest`: initial-fire ordering (a subscriber connected alongside a same-key write receives a single, post-write delivery) and a regression test that a cached key still delivers its initial callback while an unrelated write is stuck pending.
- `useOnyxTest`: loading→loaded transitions, selector output dedup, and value delivery across keys and collections.
- Perf tests (`useOnyx.perf-test`): render-count regression guards for single keys, collections, selector dedup, subscription isolation, and a large-list member-update scenario.

### Manual Tests

**Fresh start**

1. First load shows loading, then content:
    1. Log in fresh on a High Traffic account (or Troubleshoot → Clear cache and restart).
    2. Watch the chat list and a chat load.
    3. Assert the data loads correctly after reset.

**Online, single account (stay logged in, no reload)**

1. Profile changes show everywhere:
    1. Open Settings → Profile.
    2. Change your display name or status.
    3. Assert each change shows immediately everywhere it appears (header, chat list, profile) without a refresh.
2. Opening chats shows the right messages:
    1. Open a chat and send a few messages.
    2. Open a different chat, then come back.
    3. Assert each chat always shows its own current messages — never another chat's messages or an outdated view.
3. Search:
    1. Go to Spend tab, run a search and scroll the results.
    2. Change a filter or re-run it, then open a result.
    3. Assert results appear and update correctly, and the opened item shows the right data.

**Live updates from elsewhere**

1. Two tabs stay in sync (web):
    1. Open the app in two browser tabs, same account, side by side.
    2. In one tab, draft a message, mark one chat as unread.
    3. Assert the other tab reflects each change on its own, without a reload.
2. Chat list stays live:
    1. Stay on the chat list.
    2. From another account, send a message in an existing chat and start a brand-new chat with you.
    3. Assert the existing chat list shows the new message, and the new chat appears.

**General testing over the app**

1. Go to Inbox tab, scroll the chat list a bit.
2. Select a chat and send a message.
3. React to that message and create a thread.
4. Create a task and complete it.
5. Create an expense, go to the expense details, assert the data is correct.
6. Split a new expense, assert it works.
7. Delete one of the split expenses.
8. Go to Spend tab and check if expense is there. Go to its details.
9. Go to Workspaces tab, scroll the list a bit.

**Offline**

1. Works while offline:
    1. Turn off the network.
    2. Send a message to a chat.
    3. From the other user, send a message to you.
    4. Go back to online.
    5. Assert your message was sent and you received their message.

**Session teardown**

1. Session / account switching:
    1. Switch between accounts.
    2. Assert there is no crash or error; you end up signed in as the right person with the right data.
2. Log out and back in is clean:
    1. Log out.
    2. Assert you land on sign-in with no flashes of the previous account's data.
    3. Log back in and verify everything reloads correctly.

### Author Checklist

    - [x] Android / native
    - [x] Android / Chrome
    - [x] iOS / native
    - [x] iOS / Safari
    - [x] MacOS / Chrome / Safari

### Screenshots/Videos
<details>

</details>

<details>

</details>

<details>

https://github.com/user-attachments/assets/ba140831-a7ea-4c2d-a6c2-33819e4fb0dc

</details>

<details>

</details>

<details>

https://github.com/user-attachments/assets/4e839f73-27c8-4c48-a607-674b32cb5d0e

</details>


fabioh8010:
@codex review

lib/useOnyx.ts
```diff
@@ -35,209 +31,63 @@ type ResultMetadata = {
...
+    const value = useSyncExternalStoreWithSelector<OnyxValue<TKey> | undefined, TReturnValue | undefined>(subscribe, getSnapshot, undefined, select, isEqual);
+
+    // `loading` only on a key's first render (mount or key change) when a merge is still in flight for it.
+    // `connectedKeyRef` differs from `key` only on that first render; the effect below catches it up, so a
+    // later merge on an already-connected key never surfaces loading.
+    // Reading the ref during render is safe: it's written only in the effect below and re-renders are driven
+    // by `useSyncExternalStore` and the `key` prop, so it can't cause a missed update; it gates a one-shot signal.
+    // eslint-disable-next-line react-hooks/refs
+    const isLoading = connectedKeyRef.current !== key && OnyxUtils.hasPendingMergeForKey(key);
+    const loadingStatus: FetchStatus = isLoading ? 'loading' : 'loaded';
```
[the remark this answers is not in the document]
fabioh8010: Addressed in https://github.com/Expensify/react-native-onyx/pull/834/commits/3fd8b9142f844b8718a281a767fccc9898a5b143

fabioh8010:
@codex review

lib/OnyxUtils.ts
```diff
@@ -78,26 +71,63 @@ type PreparedKeyValuePairs = {
...
+ * Resolves once no write operations are in flight. Re-checks after each drain because a
+ * settling write can apply cache changes that spawn further writes (e.g. `Onyx.update`
+ * fans out into per-item merges), and those must be awaited too. Write failures are
+ * swallowed here: this only cares that writes have settled, not that they succeeded.
+ */
+function whenWritesSettled(): Promise<void> {
+    if (pendingWrites.size === 0) {
+        return Promise.resolve();
+    }
+    return Promise.all([...pendingWrites].map((promise) => promise.catch(() => undefined))).then(whenWritesSettled);
```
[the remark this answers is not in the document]
fabioh8010: https://github.com/Expensify/react-native-onyx/pull/834/commits/91b66b514fd0d92983930f35142bcb65d029c101

huult:
@fabioh8010 Is this PR ready for review?

fabioh8010:
Yes, go ahead

huult:
## Reviewer Checklist

- [ ] I have verified the author checklist is complete (all boxes are checked off).
- [ ] I verified the correct issue is linked in the `### Fixed Issues` section above
- [ ] I verified testing steps are clear and they cover the changes made in this PR
    - [ ] I verified the steps for local testing are in the `Tests` section
    - [ ] I verified the steps for Staging and/or Production testing are in the `QA steps` section
    - [ ] I verified the steps cover any possible failure scenarios (i.e. verify an input displays the correct error message if the entered data is not correct)
    - [ ] I turned off my network connection and tested it while offline to ensure it matches the expected behavior (i.e. verify the default avatar icon is displayed if app is offline)
- [ ] I checked that screenshots or videos are included for tests on [all platforms](https://github.com/Expensify/App/blob/main/contributingGuides/CONTRIBUTING.md#make-sure-you-can-test-on-all-platforms)
- [ ] I verified that the composer does not automatically focus or open the keyboard on mobile unless explicitly intended. This includes checking that returning the app from the background does not unexpectedly open the keyboard.
- [ ] I verified tests pass on **all platforms** & I tested again on:
    - [ ] Android: HybridApp
    - [ ] Android: mWeb Chrome
    - [ ] iOS: HybridApp
    - [ ] iOS: mWeb Safari
    - [ ] MacOS: Chrome / Safari
- [ ] If there are any errors in the console that are unrelated to this PR, I either fixed them (preferred) or linked to where I reported them in Slack
- [ ] I verified proper code patterns were followed (see [Reviewing the code](https://github.com/Expensify/App/blob/main/contributingGuides/PR_REVIEW_GUIDELINES.md#reviewing-the-code))
    - [ ] I verified any copy / text that was added to the app is grammatically correct in English. It adheres to proper capitalization guidelines (note: only the first word of header/labels should be capitalized), and is either coming verbatim from figma or has been approved by marketing (in order to get marketing approval, ask the Bug Zero team member to add the Waiting for copy label to the issue)
- [ ] I verified that this PR follows the guidelines as stated in the [Review Guidelines](https://github.com/Expensify/App/blob/main/contributingGuides/PR_REVIEW_GUIDELINES.md)
- [ ] I verified other components that can be impacted by these changes have been tested, and I retested again (i.e. if the PR modifies a shared library or component like `Avatar`, I verified the components using `Avatar` have been tested & I retested again)
    - [ ] All props are defined accurately
- [ ] If a new CSS style is added I verified that:
    - [ ] A similar style doesn't already exist
    - [ ] The style can't be created with an existing [StyleUtils](https://github.com/Expensify/App/blob/main/src/utils/index.ts) function (i.e. `StyleUtils.getBackgroundAndBorderStyle(theme.componentBG`)
- [ ] If the PR modifies code that runs when editing or sending messages, I tested and verified there is no unexpected behavior for all supported markdown - URLs, single line code, code blocks, quotes, headings, bold, strikethrough, and italic.
- [ ] If the PR modifies a component related to any of the existing Storybook stories, I tested and verified all stories for that component are still working as expected.
- [ ] If the PR modifies a component or page that can be accessed by a direct deeplink, I verified that the code functions as expected when the deeplink is used - from a logged in and logged out account.
- [ ] If the PR modifies the UI (e.g. new buttons, new UI components, changing the padding/spacing/sizing, moving components, etc) or modifies the form input styles:
    - [ ] I verified that all the inputs inside a form are aligned with each other.
    - [ ] I added `Design` label and/or tagged `@Expensify/design` so the design team can review the changes.
- [ ] For any bug fix or new feature in this PR, I verified that sufficient [unit tests](https://github.com/Expensify/App/blob/main/tests/README.md) are included to prevent regressions in this flow.
- [ ] I have checked off every checkbox in the PR reviewer checklist, including those that don't apply to this PR.

### Screenshots/Videos
<details>



</details>

<details>



</details>

<details>



</details>

<details>



</details>

<details>



</details>

huult:
@MelvinBot review

lib/useOnyx.ts
```diff
@@ -35,209 +30,43 @@ type ResultMetadata = {
...
+    const isEqual = selector ? deepEqual : undefined;
+
+    const value = useSyncExternalStoreWithSelector<OnyxValue<TKey> | undefined, TReturnValue | undefined>(subscribe, getSnapshot, undefined, select, isEqual);
+
+    // Reactive cache presence, so the first value landing re-renders even when the selector output is unchanged.
+    const isCached = useSyncExternalStore(subscribe, () => cache.hasCacheForKey(key));
+
+    // Loading only on a key's first render when a merge is in flight and nothing is cached yet.
+    // eslint-disable-next-line react-hooks/refs
+    const isLoading = connectedKeyRef.current !== key && !isCached && OnyxUtils.hasPendingMergeForKey(key);
```
huult, line 59: @fabioh8010 Do we need to check for any pending writes from set, multiSet, mergeCollection, or setCollection?

lib/OnyxUtils.ts
```diff
@@ -78,26 +71,110 @@ type PreparedKeyValuePairs = {
...
+ */
+function pendingWritesForKey(key: OnyxKey): Array<Promise<unknown>> {
+    const promises = [...pendingGlobalWrites];
+    const own = pendingWritesByKey.get(key);
+    if (own) {
+        promises.push(...own);
+    }
+    if (OnyxKeys.isCollectionKey(key)) {
+        for (const [writeKey, set] of pendingWritesByKey) {
+            if (writeKey !== key && OnyxKeys.isCollectionMemberKey(key, writeKey)) {
```
huult, line 153: @fabioh8010  `pendingWritesForKey` uses a raw prefix check (`isCollectionMemberKey`) to decide whether a pending write belongs to `key`'s collection. For nested/overlapping collection prefixes (e.g. `test_level_` and `test_level_last_`), this incorrectly matches a write meant for the more specific collection, so `scheduleInitialFire` can block a subscriber on an unrelated write — the opposite of the "unrelated write can never delay a subscription" guarantee stated above.

Suggest resolving membership the same way `notifyKey`/`notifyCollection` do, via longest-match:

```diff
- if (writeKey !== key && OnyxKeys.isCollectionMemberKey(key, writeKey)) {
+ if (writeKey !== key && OnyxKeys.getCollectionKey(writeKey) === key) {

File: lib/useOnyx.ts [hunk 5]
```diff
@@ -35,209 +30,43 @@ type ResultMetadata = {
...
+    const loadingStatus: FetchStatus = isLoading ? 'loading' : 'loaded';
+
+    useEffect(() => {
+        connectedKeyRef.current = key;
+    }, [key]);
```
huult, line 64:
```suggestion
    useEffect(() => {
        if (isCached) {
              connectedKeyRef.current = key;
        }
    }, [key]);
```

@fabioh8010 I noticed that `connectedKeyRef.current` is set unconditionally. If an unrelated re-render happens before the pending write for this key resolves, `isLoading` can flip to `false` prematurely and report `loaded`/`undefined` too early. In this case, should we only set it after the data has been cached?

lib/OnyxCache.ts
```diff
@@ -554,28 +554,22 @@ class OnyxCache {
...
         if (snapshot === undefined) {
             return undefined;
         }
 
-        // The collection is empty (it holds our shared empty object). But "empty" is ambiguous
-        // during startup: we can't tell an actually-empty collection apart from one whose data
-        // hasn't loaded yet. Once any key exists, we know setAllKeys has run and loaded everything,
-        // so an empty collection really is empty. Before that, return undefined so subscribers
-        // don't briefly see a collection as empty when it just hasn't loaded.
-        if (snapshot === FROZEN_EMPTY_COLLECTION) {
```
huult, line 575: @fabioh8010 Could you help explain why we need to remove this condition?

huult:
https://github.com/user-attachments/assets/f760da9e-0d34-4d36-a2ed-09369d88f697

BUG: Product marketing appears briefly and then disappears after re-login

huult:
https://github.com/user-attachments/assets/ad4b8c44-8b98-417a-a00b-3f779d9e8a5f


https://github.com/user-attachments/assets/9afa94c2-4f30-4fec-a621-9d5f2001c520



BUG: Login succeeds and redirects to Home, but reloading the page redirects back to the login page

fabioh8010:
Thanks for the review and bug reports @huult, here's the investigation notes I did with Claude today about Bug 1 (BUG: Product marketing appears briefly and then disappears after re-login):

In the OpenApp batch, `Onyx.update` writes a bunch of keys together: account, `isLoadingApp`, the policy collection, and three NVPs including `nvp_lastDismissedMarketingWindow`. On main every single-key write lands in the first tick (+3ms), then the loading flag and the collection. On this branch account lands at +3, the loading flag at +18, the collection at +19, but the three NVPs only at +80, +92 and +96. React renders in that gap, sees "app done loading" and "nothing dismissed", and paints the window. It disappears once the dismissal value finally lands.

The 60ms comes from `Onyx.merge` starting with `OnyxUtils.get(key)`. On main, subscribing to a key also went through `get()` (`subscribeToKey` does `get(key).then(() => sendDataToConnection(...))`), so a key with nothing in storage got marked as known-absent in the cache, and any later merge to it resolved from cache immediately. The new `useOnyx` reads the cache synchronously and never calls `get()`, so those keys are never marked, and the first write to one has to do a real IndexedDB read while the rest of the batch goes ahead without it.

The proposed fix: in `merge`, skip the storage round trip when the key isn't in cache. `merge` already runs after init, and it already prefers the cache over the fetched value, so with eager loading a cache miss at that point means the key genuinely has nothing stored. That puts every key in a batch back into the same tick, which is what main does today.

Later we could also follow up separately on making `Onyx.update` atomic, so all of a batch's cache writes happen before any subscriber is notified.

---

@huult About your review comments, I will take a look at them once I apply the fix for Bug 1, as from my investigations they're not what caused the flash anyway.

About Bug 2 (BUG: Login succeeds and redirects to Home, but reloading the page redirects back to the login page), I couldn't reproduce it yet.

roryabraham:
> Later we could also follow up separately on making Onyx.update atomic, so all of a batch's cache writes happen before any subscriber is notified.

@MelvinBot create a dedicated issue in the Expensify/App repo for this. Label it Weekly. Assign @fabioh8010, @roryabraham, and @huult. Make it a child issue of https://github.com/Expensify/App/issues/80355

roryabraham:
Created the requested child issue: https://github.com/Expensify/App/issues/101801

huult:
Thanks for checking. Please tag me when it’s ready for another review.

lib/useOnyx.ts
fabioh8010: I don't think we need to. `set`, `multiSet` and `setCollection` all write to the cache before they hit storage, so by the time their promise is pending the value is already there, and checking them would just hide a value that's already correct.

`merge` is different because it queues the delta and applies it a tick later, that's why the condition uses `hasPendingMergeForKey`, and it works the same on `main`.

`mergeCollection` does leave a small window since it applies after resolving the collection keys and never goes through the merge queue, but I don't want `loading` to end up meaning "a sibling member is being written" when today it means "this key has no value yet". That window is really about batch visibility, which is what https://github.com/Expensify/App/issues/101801 is for.

lib/OnyxCache.ts
fabioh8010: @huult It's redundant now. `Onyx.init` set a frozen empty entry (`{}`) for every collection we know about (`setCollectionKeys`), so if there's an entry we're already past init. And `storageKeys.size > 0` was actually wrong after `Onyx.clear()`, since clear wipes the key index and every empty collection would start reading `undefined` again even though we're initialised.

Dropping it also saves checking whether the object is empty on every read, we just compare against the shared frozen one.

The only thing that changes: a collection that's seeded but hasn't loaded yet reads `{}` instead of `undefined`. They match again once the keys load, and that's why some E/App tests needed updating.

lib/Onyx.ts
```diff
@@ -122,23 +124,95 @@ function init({
...
+ * @returns The `Connection` handle to use when calling `Onyx.disconnect()`.
  */
 function connect<TKey extends OnyxKey>(connectOptions: ConnectOptions<TKey>): Connection {
-    return connectionManager.connect(connectOptions);
+    const {key, callback} = connectOptions;
+
+    let active = true;
+    let unsubscribeFn: (() => void) | null = null;
+
+    const wireUp = () => {
```
roryabraham, line 150: Can we perhaps create a more specific/descriptive name than `wireUp`?

lib/Onyx.ts
```diff
@@ -122,23 +124,95 @@ function init({
...
+                if (Object.is(lastDeliveredCollection, rawCollection)) {
+                    return;
+                }
+                lastDeliveredCollection = rawCollection;
+                (callback as CollectionConnectCallback<TKey> | undefined)?.(rawCollection as NonNullable<OnyxCollection<KeyValueMapping[TKey]>>, k);
+            };
+            unsubscribeFn = onyxSubscriptionManager.subscribe(key, (value, k) => {
+                deliverCollection(value as unknown as OnyxValue<TKey>, k as TKey);
+            });
+            OnyxUtils.scheduleInitialFire(key, () => {
```
roryabraham, line 170: Again, perhaps a more specific name. `scheduleInitialSubscriberNotification`?

lib/Onyx.ts
```diff
@@ -181,7 +257,10 @@ function connectWithoutView<TKey extends OnyxKey>(connectOptions: ConnectOptions
  * @param connection Connection object returned by calling `Onyx.connect()` or `Onyx.connectWithoutView()`.
  */
 function disconnect(connection: Connection): void {
```
roryabraham, line 259: NABish: this function is kind of useless now that `unsubscribe` lives on the `Connection`.

lib/Onyx.ts
```diff
@@ -203,7 +285,10 @@ function set<TKey extends OnyxKey>(key: TKey, value: OnyxSetInput<TKey>, options
  * @param data object keyed by ONYXKEYS and the values to set
  */
 function multiSet(data: OnyxMultiSetInput): Promise<void> {
-    return OnyxUtils.afterInit(() => OnyxUtils.multiSetWithRetry(data));
+    return OnyxUtils.trackPendingWrite(
+        Object.keys(data),
+        OnyxUtils.afterInit(() => OnyxUtils.multiSetWithRetry(data)),
```
roryabraham, line 290: *question:* Should `OnyxUtils.afterInit` wrap the whole thing or just the write? Same question everywhere this pattern is used

lib/OnyxUtils.ts
```diff
@@ -78,26 +71,110 @@ type PreparedKeyValuePairs = {
...
+/**
+ * Registers an in-flight write under each key it can change, so `scheduleInitialFire` waits only for
+ * the writes relevant to a connecting key. Returns the same promise so callers can wrap a write's
+ * return value inline. The write is deregistered once it settles (success or failure).
+ */
+function trackPendingWrite<T>(keys: OnyxKey | OnyxKey[], promise: Promise<T>): Promise<T> {
+    // Drop nullish keys (e.g. a keyless `clear` item) so they never reach `pendingWritesForKey`'s scan.
+    const keyList = (Array.isArray(keys) ? keys : [keys]).filter((key) => typeof key === 'string');
+    for (const key of keyList) {
+        let set = pendingWritesByKey.get(key);
```
roryabraham, line 108:
```suggestion
        let pendingWritesForKey = pendingWritesByKey.get(key);
```

lib/OnyxUtils.ts
```diff
@@ -78,26 +71,110 @@ type PreparedKeyValuePairs = {
...
+ */
+// eslint-disable-next-line rulesdir/no-negated-variables
+const NOT_DELIVERED = Symbol('NOT_DELIVERED');
+
+/**
+ * Registers an in-flight write under each key it can change, so `scheduleInitialFire` waits only for
+ * the writes relevant to a connecting key. Returns the same promise so callers can wrap a write's
+ * return value inline. The write is deregistered once it settles (success or failure).
+ */
+function trackPendingWrite<T>(keys: OnyxKey | OnyxKey[], promise: Promise<T>): Promise<T> {
```
roryabraham, line 104:
```suggestion
function trackPendingWrite<T>(keys: OnyxKey | OnyxKey[], pendingWrite: Promise<T>): Promise<T> {
```

lib/OnyxUtils.ts
```diff
@@ -78,26 +71,110 @@ type PreparedKeyValuePairs = {
...
+        }
+    };
+    promise.then(deregister, deregister);
+    return promise;
+}
+
+/**
+ * Registers an in-flight write that affects every key (Onyx.clear). Deregistered once it settles.
+ */
+function trackPendingGlobalWrite<T>(promise: Promise<T>): Promise<T> {
```
roryabraham, line 134:
```suggestion
function trackPendingGlobalWrite<T>(pendingGlobalWrite: Promise<T>): Promise<T> {
```

lib/OnyxUtils.ts
roryabraham: Also, *question:* why do we not use `pendingWritesForKey` here?

lib/OnyxUtils.ts
```diff
@@ -78,26 +71,110 @@ type PreparedKeyValuePairs = {
...
+ * can't collide with any real Onyx value, so the first `Object.is` check never matches and
+ * the initial fire runs even when a key's genuine first value is `undefined`. It only needs
+ * to be distinct from real values, not unique per subscription, so one module-level instance
+ * is reused by every connection.
+ */
+// eslint-disable-next-line rulesdir/no-negated-variables
+const NOT_DELIVERED = Symbol('NOT_DELIVERED');
+
+/**
+ * Registers an in-flight write under each key it can change, so `scheduleInitialFire` waits only for
```
roryabraham, line 100: Can we put the pending-write-tracking logic in its own file rather than in OnyxUtils?

roryabraham (changes requested):
Not a comprehensive review, but here's some quick feedback

fabioh8010:
I will address remaining of the comments today

PR #848, opened 2026-09-18
feat: export persisted Onyx state through active provider

chrispader:
Expensify's Troubleshoot export opens Onyx's SQLite database separately, which NitroSQLite 9.8.1 rejects as a duplicate connection. Onyx has no public API for exporting its persisted store through the active provider, so the export can fail. This PR adds `Onyx.exportState()` for native and web consumers.

@NicolasBonet @ikevin127

### Details

`exportState()` waits for Onyx initialization, reads the active provider's existing `getAll()` operation, and returns a plain key/value object. It excludes stale persisted rows for keys now configured as RAM-only by default. Callers can pass `{includeStaleRamOnlyKeys: true}` to include those rows, as the linked App export does to match its previous raw-storage query. Neither setting adds current RAM-only values from memory or writes that have not reached storage.

The method does not mask data. Callers must redact sensitive values before sharing the result. The API reference and README document the option and these boundaries.

### Related Issues

https://github.com/Expensify/App/issues/101448

### Linked E/App PR

https://github.com/Expensify/App/pull/101568

### Automated Tests

Tests cover persisted values, default exclusion and explicit inclusion of stale RAM-only rows, exclusion of current RAM-only values, and storage-read failures. All 666 unit tests, typecheck, and lint pass locally.

### Manual Tests

1. In the linked App PR, open Settings → Troubleshoot and export Onyx state on native and web. Verify the exported file contains local data and masked sensitive fields.
2. Repeat offline; exporting local state should still work.
3. With the NitroSQLite 9.8.1 upgrade in https://github.com/Expensify/App/pull/101437, verify native export does not produce a duplicate-connection error.

These manual checks have not yet been completed on devices or browsers.

### Author Checklist

- [x] I linked the related issue and corresponding App PR.
- [x] I documented the new API and added automated tests.
- [ ] I verified the linked App PR's full CI and manual tests on all platforms.
- [x] I considered screenshots and videos. The export has no normal UI change, so none are attached.

### Screenshots/Videos

None attached; this adds a storage API without changing the normal UI.


chrispader:
@MelvinBot request review from @NicolasBonet and @ikevin127

lib/Onyx.ts
```diff
@@ -611,6 +611,21 @@ function setCollection<TKey extends CollectionKeyBase>(collectionKey: TKey, coll
     return OnyxUtils.afterInit(() => OnyxUtils.setCollectionWithRetry({collectionKey, collection}));
 }
 
+type ExportStateOptions = {
```
ikevin127, line 614: 🟠 **`lib/Onyx.ts:614`**

`ExportStateOptions` breaks the option-type convention this file already follows. Every other option type lives in `lib/types.ts` and is imported:

```ts
// lib/types.ts
type ConnectOptions<TKey extends OnyxKey> = ...   // :237
type SetOptions = { ... }                          // :325
type InitOptions = { ... }                         // :357
```

```ts
// lib/Onyx.ts, bottom of file
export type {OnyxUpdate, ConnectOptions, SetOptions};
```

`ExportStateOptions` is declared inline halfway down `Onyx.ts` and appears in neither that export list nor `lib/index.ts`. So a consumer can call `Onyx.exportState({includeStaleRamOnlyKeys: true})` but cannot name the type.

That matters concretely for App#101568. Anything wrapping this call, for example

```ts
function exportOnyxStateForTroubleshoot(options: ExportStateOptions) { ... }
```

has no type to import, so App will either re-declare the shape locally or inline `{includeStaleRamOnlyKeys?: boolean}`. Duplicating a published type is exactly what we avoid, and it silently drifts the first time a second option is added here.

Move it next to `SetOptions` in `lib/types.ts`, then:

```ts
// lib/Onyx.ts
export type {OnyxUpdate, ConnectOptions, SetOptions, ExportStateOptions};
```

```ts
// lib/index.ts
export type {
    ...
    ExportStateOptions,
};
```

lib/Onyx.ts
```diff
@@ -611,6 +611,21 @@ function setCollection<TKey extends CollectionKeyBase>(collectionKey: TKey, coll
...
+
+/**
+ * Returns persisted Onyx key-value pairs as a plain object.
+ * Live RAM-only values and writes that have not reached storage are not included.
+ * @param [options] Export options.
+ * @param [options.includeStaleRamOnlyKeys=false] Include persisted rows for keys that are now RAM-only.
+ */
+function exportState({includeStaleRamOnlyKeys = false}: ExportStateOptions = {}): Promise<Record<string, unknown>> {
+    return OnyxUtils.afterInit(() => Storage.getAll().then((entries) => Object.fromEntries(includeStaleRamOnlyKeys ? entries : entries.filter(([key]) => !OnyxKeys.isRamOnlyKey(key)))));
+}
```
ikevin127, line 627: 🟡 **`lib/Onyx.ts:625-626`**

Three things wrong with this line, all fixed by the same rewrite.

**1. The declared return type is unchecked.** `Storage.getAll()` returns `StorageKeyValuePair[]`, which is a three-element tuple, not a pair:

```ts
// lib/storage/providers/types.ts:6
type StorageKeyValuePair = [key: OnyxKey, value: OnyxValue<OnyxKey>, replaceNullPatches?: FastMergeReplaceNullPatch[]];
```

Because of that optional third slot, the array does not match `Object.fromEntries`'s typed overload (`Iterable<readonly [PropertyKey, T]>`) and falls through to `Iterable<readonly any[]> => any`. The result is `any`, so `Promise<Record<string, unknown>>` is an assertion nobody checks.

I confirmed this rather than guessing. Two snippets, same tsc invocation:

```ts
// A: the real StorageKeyValuePair shape
type StorageKeyValuePair = [key: string, value: unknown, replaceNullPatches?: Patch[]];
declare const entries: StorageKeyValuePair[];
const result = Object.fromEntries(entries);
const wrong: number = result;   // NO ERROR
```

```ts
// B: plain two-tuple, control
type Pair = [key: string, value: unknown];
declare const entries2: Pair[];
const result2 = Object.fromEntries(entries2);
const wrong2: number = result2; // error TS2322: Type '{ [k: string]: unknown; }' is not assignable to type 'number'
```

Case A reports nothing. So if `getAll()`'s shape ever moves, a provider starting to return `replaceNullPatches`, or the key/value order flipping, `npm run typecheck` stays green and the export silently ships the wrong thing. For a function whose entire job is "hand this data to a human debugging a customer report", silently wrong is the worst outcome.

**2. The types are not the repo's own.** `Record<string, unknown>` uses `string` where this codebase uses `OnyxKey`, and `unknown` forces every consumer to cast. `OnyxKey` and `OnyxValue` are already exported from `lib/index.ts`, so there is no reason to invent a looser shape.

**3. It walks the whole database twice.** `entries.filter(...)` allocates a second array before `fromEntries` walks it again. On a High Traffic account that is tens of thousands of rows copied for nothing, on the JS thread, while the user waits on the Troubleshoot screen.

All three go away with an explicit single pass:

```ts
function exportState({includeStaleRamOnlyKeys = false}: ExportStateOptions = {}): Promise<Record<OnyxKey, OnyxValue<OnyxKey>>> {
    return OnyxUtils.afterInit(() =>
        Storage.getAll().then((entries) => {
            const state: Record<OnyxKey, OnyxValue<OnyxKey>> = {};

            for (const [key, value] of entries) {
                if (!includeStaleRamOnlyKeys && OnyxKeys.isRamOnlyKey(key)) {
                    continue;
                }
                state[key] = value;
            }

            return state;
        }),
    );
}
```

This is also just readable. The current version is a 190-character line doing four things.

tests/unit/onyxTest.ts
```diff
@@ -66,6 +66,46 @@ describe('Onyx', () => {
         return Onyx.clear();
     });
 
+    describe('exportState', () => {
```
ikevin127, line 69: 🟠 **`tests/unit/onyxTest.ts:69`**

The `afterInit` gate is the one part of this function with a real ordering contract, and nothing tests it.

`exportState` is reached from a settings screen. If someone later "simplifies" it to call `Storage.getAll()` directly, which is tempting because it passes every test in this suite (init has already run by then), a cold-start export would hit the provider's own guard and reject:

```ts
// lib/storage/providers/SQLiteProvider.ts
getAll() {
    if (!provider.store) {
        throw new Error('Store is not initialized!');
    }
```

Add a test that calls `exportState()` before init resolves and asserts it still resolves with the persisted data. Cheap, and it pins the only behaviour in this function that is not obvious from reading it.

ikevin127, line 69: 🟡 **`tests/unit/onyxTest.ts:69`**

No test covers a normal collection member, which is the bulk of real Onyx data.

The suite has `TEST_KEY` (a single key) and `ramOnlyCollection_1` (RAM-only, mocked). Nothing proves that `report_1`-shaped keys survive the export. Since the filter uses `OnyxKeys.isRamOnlyKey`, which matches collection prefixes:

```
isRamOnlyKey("ramOnlyCollection_1") → true
```

a bug in prefix matching would drop every member of a normal collection and no test would fail. One line to cover:

```ts
it('exports persisted collection members', async () => {
    await Onyx.set(`${ONYX_KEYS.COLLECTION.TEST_KEY}1`, {nested: 'value'});

    const state = await Onyx.exportState();

    expect(state[`${ONYX_KEYS.COLLECTION.TEST_KEY}1`]).toEqual({nested: 'value'});
});
```

lib/Onyx.ts
```diff
@@ -611,6 +611,21 @@ function setCollection<TKey extends CollectionKeyBase>(collectionKey: TKey, coll
...
+    includeStaleRamOnlyKeys?: boolean;
+};
+
+/**
+ * Returns persisted Onyx key-value pairs as a plain object.
+ * Live RAM-only values and writes that have not reached storage are not included.
+ * @param [options] Export options.
+ * @param [options.includeStaleRamOnlyKeys=false] Include persisted rows for keys that are now RAM-only.
+ */
+function exportState({includeStaleRamOnlyKeys = false}: ExportStateOptions = {}): Promise<Record<string, unknown>> {
```
ikevin127, line 625: 🟢 **`lib/Onyx.ts:626` and `README.md`**

The mutability contract is undefined and provider-dependent. `MemoryOnlyProvider` hands back live references into its store:

```ts
// lib/storage/providers/MemoryOnlyProvider.ts:147
getAll() {
    return Promise.resolve(Object.entries(provider.store) as StorageKeyValuePair[]);
},
```

whereas SQLite returns freshly `JSON.parse`d values and IDB returns structured clones. So whether a caller may mutate the returned object depends on which provider is active, and `MemoryOnlyProvider` becomes active after `degradePerformance()` swaps it in.

A consumer that edits the export in place would corrupt Onyx's live store, but only on a device that had already degraded, which is about the least reproducible bug shape there is.

App's `maskOnyxState` does `let onyxState = {...data}` and replaces top-level entries, so it is fine today. That is luck, not contract. Add a line to the README saying the returned object must be treated as read-only, or clone on the way out.

README.md
```diff
@@ -288,6 +288,10 @@ function signOut() {
 }
 ```
```
ikevin127, line 290: 🟢 **`README.md`, design question**

**Why does the option exist at all?**

`includeStaleRamOnlyKeys` is there so App#101568 can pass `true` and match what its old raw `SELECT *` returned. But those rows are dead data: keys the app has since declared RAM-only, so nothing writes them any more and nothing reads them. "Match the previous behaviour" is a weak reason to put a permanent branch in a public Onyx API.

I checked the privacy side and it is covered. The sensitive RAM-only keys (`RAM_ONLY_WALLET_ONFIDO`, `RAM_ONLY_PLAID_LINK_TOKEN`, `RAM_ONLY_MERGE_HR_LINK_TOKEN`, `RAM_ONLY_IS_AUTHENTICATING_WITH_SHORT_LIVED_TOKEN`) are all handled in App's `src/libs/ExportOnyxState/common.ts`, so a stale row would be removed or masked before sharing.

So this is not a leak, it is scope. Worth asking whether App actually needs those stale rows in the dump, because if not, dropping the option leaves a one-argument-free API that is easier to reason about.

ikevin127:
### Verified ✅ 

- `OnyxKeys.isRamOnlyKey` (`lib/OnyxKeys.ts:212`) exists and already handles collection prefixes, and `Storage`, `OnyxKeys`, `OnyxUtils` are all imported in `lib/Onyx.ts` already, so no new imports were needed.
- The README claim "It works on native and web" holds. Both platform providers implement `getAll`: `lib/storage/providers/IDBKeyValProvider/index.ts:136` and `lib/storage/providers/SQLiteProvider.ts:264`, plus `MemoryOnlyProvider` and `NoopProvider`. No provider is missing it, so there is no platform where this throws.
- Exporting will not accidentally trip the storage fallback. `degradePerformance` only fires on `'IDBKeyVal store could not be created'`, an init-time message, so a failing `getAll()` rejects and propagates rather than silently swapping the app onto `MemoryOnlyProvider`. The `propagates storage read failures` test pins that.
- `API.md` ordering is consistent with source order, so a regenerate via `npm run build:docs` will not churn.
- `exportState` correctly does not need an entry in `OnyxUtils.METHOD`, that map is for update operations.

---

🟢 Small, focused diff, nothing here blocks on correctness. The comments I dropped above are type-consistency and test-contract issues.

README.md
chrispader: Good point, i'm going to remove the option

lib/Onyx.ts
chrispader: Thanks!

lib/Onyx.ts
chrispader: Moved!

chrispader:
Addressed all of @ikevin127's review comments!

ikevin127 (approved):
**🟢 LGTM** - Thanks for addressing the comments

NicolasBonet (approved):

